PR #8, opened 2019-06-06
Update `cabal.project` & `stack.yaml`

Jimbo4350:


cardano-node.cabal
```diff
@@ -11,43 +11,70 @@ cabal-version:  >= 1.10
 extra-source-files: README.md, ChangeLog.md
 
 library
+  hs-source-dirs:      src
+
   exposed-modules:
-      Lib
+                       Lib
```
erikd: Surely we can namespace this now rather than later? Like `Cardano.Node` or something?

cardano-node.cabal
```diff
@@ -11,43 +11,70 @@ cabal-version:  >= 1.10
...
+  default-extensions:  NoImplicitPrelude
+
+  ghc-options:         -Weverything
+                       -fno-warn-safe
+                       -fno-warn-unsafe
+                       -fno-warn-missing-import-lists
 
 executable cardano-node-exe
-  main-is: Main.hs
+  main-is:              Main.hs
```
erikd: Since this is the main Haskell file for an executable, it does not need to be camel case. All lower case would work.

Also, why not name it `cardano-node.hs` since that is the name of executable its going to create.

Finally, can we drop the "-exe" from the executable's name?

cardano-node.cabal
```diff
@@ -11,43 +11,70 @@ cabal-version:  >= 1.10
...
+
   build-depends:
-      base >=4.7 && <5
-    , cardano-ledger
-    , cardano-prelude
-    , cardano-shell
-    , ouroboros-consensus
-    , ouroboros-network
-  default-language: Haskell2010
+                       base >=4.7 && <5
```
erikd: I am almost certain this will not build with `base == 4.7.*`. Since ghc 8.6 uses `base` version `4.12`, maybe that should be `base >=4.12 && <5`.

cabal.project
```diff
@@ -22,18 +58,62 @@ source-repository-package
...
+source-repository-package
+  type: git
+  location: https://github.com/joelwilliamson/bimap
+  tag: 997fbb38b08dec14d225d064dac05b0a85f4ceae
 
+constraints:
+  ip < 1.5,
+  ekg-prometheus-adapter >= 0.1.0.4,
+  hedgehog >= 1.0,
+  libsystemd-journal >= 1.4.4
```
erikd: Someone's text editor should be configured to always make sure the file ends in a newline.

PR #13, opened 2019-06-18
Created node feature.

365andreas:
Created feature for node.

### DIFF
--- a/cardano-node.cabal
+++ b/cardano-node.cabal
@@ -47,16 +47,15 @@ executable cardano-node
                        Run
                        Topology
   build-depends:       base
-                     , contra-tracer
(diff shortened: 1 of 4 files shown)


cardano-node.cabal
```diff
@@ -55,6 +55,8 @@ executable cardano-node
                      , cardano-ledger
                      , cardano-ledger-test
                      , cardano-prelude
+                     , cardano-shell
+                    --  , cardano-prelude
```
ksaric: Maybe remove this from the cabal file?

ksaric:
BTW, this is probably missing since it's not portable - https://hydra.iohk.io/build/882894/nixlog/2

PR #14, opened 2019-06-19
Fix double start of node.

365andreas:


ksaric:
We can merge this for the presentation, but in the future, we need to make this more portable - the failure now is that this POSIX stuff is not going to build on Windows.

365andreas:
bors retry

PR #16, opened 2019-06-19
WIP:Restructure

Jimbo4350:
Restructuring the node based on `validate-mainnet`


Jimbo4350:
Outdated, will restructure after node demo.

PR #19, opened 2019-06-20
temporarily ignore failing windows builds on CI

cleverca22:


CodiePP:
I guess this needs to be merged by hand
@cleverca22 @disassembler

PR #22, opened 2019-06-21
Bump the dependency on ouroboros-network and port recent changes

mrBliss:
As ouroboros-network bumps its dependencies on some packages as well as
GHC (8.6.4 -> 8.6.5), we have to follow suit.


mrBliss:
Pushed (no changes) again to trigger CI.

Jimbo4350:
We should probably check that the demo scripts are still working

scripts/submit-tx.sh
```diff
@@ -5,5 +5,6 @@ now=`date "+%Y-%m-%d 00:00:00"`
 set -x
 cabal new-run cardano-node -- \
     --system-start "$now" --slot-duration 2 \
+    --log-config configuration/log-configuration.yaml \
```
mrBliss: Actually, this should not really be needed for submitting a transaction.

PR #26, opened 2019-06-21
Bump the dependency on ouroboros-network and port recent changes

mrBliss:


mrBliss:
I'll rebase after #17 is merged.

PR #28, opened 2019-06-25
genesis-tool: Byron genesis generator CLI

deepfire:
Issue: https://github.com/input-output-hk/cardano-node/issues/30

# TODO
- [x] Figure out secret key size discrepancy: 130 bytes in this implementation vs. 203 for actual secret keys in `cardano-sl` (**SHOWSTOPPER**).
  - this appears to be due to VSS keys also being there: https://github.com/input-output-hk/cardano-sl/blob/master/tools/src/keygen/Dump.hs#L48
- [x] Key pretty-printing.
- [x] Documentation.
- [x] Key material migration.
- [x] Factor.
- [?] Optional random generation of things meaningfully available to random generation (seed, protocol magic, etc.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,9 @@
 dist-newstyle/
 dist/
 *~
+\#*
+*.swp
 result*
 launch-*
 .stack-to-nix.cache
+/genesis.*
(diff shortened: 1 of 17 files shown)


deepfire:
Thank you, @CodiePP!

PR #35, opened 2019-06-28
Wallet client template

coot:
This is a dummy client, it consists of two parts:

* chain sync client which downloads blocks from a local core node and logs them to the console
* runs tx submission protocol which just blocks

Both can be extended to run a wallet application:
* update the wallet state when a new block is received (or rollback command is received)
* send transactions via tx-submission protocol to the core node

 The PR includes a `README.md` which documents how to run it.  It works with `--bft` (when using `--real-pbft`  the chain of core node is not growing, so nothing happens in the client).

@KtorZ


wallet-client/CLI.hs
```diff
@@ -0,0 +1,111 @@
+{-# LANGUAGE GADTs #-}
+
+module CLI
```
mrBliss: Hmm, it would be nicer to reuse `app/CLI.hs` (move it to `cardano-node`, the library folder?) instead of duplicating parts here.

coot: +1, this will also hide some of the `ouroboros-consensus` dependencies.

cardano-node/Cardano/Node/CLI.hs
```diff
@@ -0,0 +1,133 @@
...
+    )
+
+parseNumCoreNodes :: Parser NumCoreNodes
+parseNumCoreNodes =
+    option (fmap NumCoreNodes auto) (
+            long "num-core-nodes"
+         <> short 'm'
+         <> metavar "NUM-CORE-NODES"
+         <> help "The number of core nodes"
+    )
```
avieth: Ignoring this file because it looks like it's a copy/paste move.

wallet-client/Main.hs
```diff
@@ -0,0 +1,126 @@
...
+
+
+createNodeFeature :: LoggingLayer -> CLI -> CardanoEnvironment -> CardanoConfiguration -> IO (NodeLayer, CardanoFeature)
+createNodeFeature loggingLayer cli cardanoEnvironment cardanoConfiguration = do
+    -- we parse any additional configuration if there is any
+    -- We don't know where the user wants to fetch the additional configuration from, it could be from
+    -- the filesystem, so we give him the most flexible/powerful context, @IO@.
+
+    -- we construct the layer
+    nodeLayer <- (featureInit nodeCardanoFeatureInit) cardanoEnvironment loggingLayer cardanoConfiguration cli
```
avieth: What's with the parens? Isn't it the same without them?

wallet-client/Run.hs
```diff
@@ -0,0 +1,25 @@
...
+
+import           CLI
+import           WalletClient
+
+runClient :: CLI -> Trace IO Text -> IO ()
+runClient CLI{..} tracer = do
+    let CoreNodeId nid = cliCoreNodeId
+    tracer'
+      <- contramap pack . toLogObject <$>
+          appendName ("Wallet " <> pack (show nid)) tracer
```
avieth: Not an issue with this PR but I can't help but point it out: `appendName` has a bad type. It should be `LoggerName -> Trace m a -> Trace m a`, but they put it into `m` instead, for no apparent reason.

wallet-client/WalletClient.hs
```diff
@@ -0,0 +1,223 @@
...
+      Nothing
+      addr
+
+muxLocalInitiatorNetworkApplication
+  :: forall blk m.
+     (RunDemo blk, MonadST m, MonadThrow m, MonadTimer m)
+  -- TODO: the need of a 'Proxy' is an evidence that blk type is not really
+  -- needed here.  The wallet client should use some concrete type of block
+  -- from 'cardano-chain'.  This should remove the dependency of this module
+  -- from 'ouroboros-consensus'.
```
avieth: In this type signature, `blk` only appears in `BlockProtocol`, which is a non-injective type family. I think you could remove the proxy and flip on `AllowAmbiguousTypes` instead. It did this in https://github.com/input-output-hk/ouroboros-network/pull/635 but @dcoutts didn't like it.

avieth: Really though I think you should just leave the proxy. It's one of the fee options that we must pay for using type families so liberally, as ouroboros-consensus does.

wallet-client/WalletClient.hs
```diff
@@ -0,0 +1,223 @@
...
+  :: forall blk m. MonadTimer m
+  => Tracer m blk
+  -> ChainSyncClient blk (Point blk) m Void
+chainSyncClient tracer = ChainSyncClient $ pure $
+    -- Notify the core node about the our latest points at which we are
+    -- synchronised.  This client is not persistent and thus it just
+    -- synchronises from the genesis block.  A real implementation should send
+    -- a list of points up to a point which is k blocks deep.
+    SendMsgFindIntersect
+      [genesisPoint]
```
avieth: This is a point of confusion that I have begun to clarify in https://github.com/input-output-hk/ouroboros-network/pull/635

If this `SendMsgFindIntersect` were _not_ sent, what would be the read pointer? Wouldn't it still be `genesisPoint`? What if the server does not know `genesisPoint`? Can the protocol continue in that case?

wallet-client/WalletClient.hs
```diff
@@ -0,0 +1,223 @@
...
+    clientStNext :: ClientStNext blk (Point blk) m Void
+    clientStNext = ClientStNext {
+        recvMsgRollForward = \blk _tip -> ChainSyncClient $ do
+          -- trace the next block, the core note tip point is '_tip'.
+          traceWith tracer blk
+          pure clientStIdle
+      , recvMsgRollBackward = \_point _tip -> ChainSyncClient $ do
+          -- we are requested to roll backward to point '_point', the core
+          -- node's chain's tip is '_tip'.
+          pure clientStIdle
```
avieth: Should trace here too.

wallet-client/Main.hs
coot: +1, I'll remove them. The sin of copy pasting :D

wallet-client/WalletClient.hs
coot: If it was not given it would be `genesisPoint` I believe, @mrBliss will know the definite answer.  But the point here is to give the wallet team the starting point.  This is just a hook where they need to provide their state to the local node.

wallet-client/WalletClient.hs
coot: I will actually remove the trace.  I'll add a comment how to trace all network messages.

wallet-client/WalletClient.hs
mrBliss: Indeed, Genesis is the implicit starting point of the ChainSync protocol (I did not write the protocol). So `SendMsgFindIntersect` is not necessary.

wallet-client/Run.hs
coot: ^ @CodiePP

CodiePP: thanks for pointing this out. We will have to check this, as this has been annoying at other places too. Probably, it is just legacy we carry around..

erikd: Anything that can be a pure function *should* be a pure function.

365andreas: This is fixed now.

PR #37, opened 2019-07-03
Added Trace acceptors.

365andreas:
Nodes spawn TraceForwarders and an acceptor gathers all the objects that are being traced.


CodiePP:
bors try

PR #40, opened 2019-07-05
LiveViewState. TUI draw simplification. Node's data.

denisshevchenko:
`LiveViewState`. TUI draw simplification. Node's data.

### DIFF
--- a/app/LiveView.hs
+++ b/app/LiveView.hs
@@ -2,6 +2,7 @@
 {-# LANGUAGE GADTs               #-}
 {-# LANGUAGE NamedFieldPuns      #-}
 {-# LANGUAGE NumericUnderscores  #-}
+{-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE RecordWildCards     #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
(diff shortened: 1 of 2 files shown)


app/LiveView.hs
```diff
@@ -12,149 +13,171 @@ module LiveView (
...
-          deriving (Show, Eq)
+initLiveViewState
+    :: TopologyInfo
+    -> LiveViewState
+initLiveViewState (TopologyInfo nodeId _) = LiveViewState
+    { lvsQuit            = False
+    , lvsRelease         = "Shelley"   -- Should be taken from ..?
+    , lvsNodeId          = nodeIdNum
+    , lvsVersion         = showVersion version
+    , lvsCommit          = "bc39f7124" -- Should be taken from ..?
```
CodiePP: do the autogenerated data by Cabal contain such information?
https://www.haskell.org/cabal/users-guide/developing-packages.html#pkg-field-custom-setup-autogen-modules

CodiePP: I see we already have access to `Paths_cardano_node`

denisshevchenko: I see two autogenerated modules: `cabal_macros.h` and `Paths_cardano_node.hs`. But these modules contains different versions and system paths, it doesn't contain Git hash nor node's release..

app/LiveView.hs
```diff
@@ -12,149 +13,171 @@ module LiveView (
...
+    , lvsMaxNetDelay     = 0
+    , lvsMempool         = 0
+    , lvsMempoolPerc     = 0
+    , lvsCPUUsagePerc    = 0
+    , lvsMemoryUsageCurr = 0
+    , lvsMemoryUsageMax  = 0
+    }
+  where
+    nodeIdNum = case nodeId of
+        CoreId num  -> num
```
CodiePP: could we make this a `nodeId :: Text`? and display for example `C3` or `R1` for core, relay, respectively.

PR #41, opened 2019-07-05
LiveViewState: show current Git commit.

denisshevchenko:
Now `LiveViewState` contains the real Git commit.


denisshevchenko:
bors+

PR #53, opened 2019-07-12
Additional progress bars: disk IO and network IO.

denisshevchenko:


CodiePP:
I am not sure we will see 500 MB/s :-)

PR #54, opened 2019-07-12
Track minted blocks.

365andreas:
Live view now tracks how many blocks were minted by the node.

### DIFF
--- a/app/LiveView.hs
+++ b/app/LiveView.hs
@@ -121,9 +121,12 @@ instance IsEffectuator LiveViewBackend Text where
                                          }
                     _ -> return ()
             LogObject _ _ (LogMessage msg) ->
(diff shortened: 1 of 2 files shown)


app/LiveView.hs
```diff
@@ -121,6 +120,11 @@ instance IsEffectuator LiveViewBackend a where
                                          , lvsUpTime       = diffUTCTime (tstamp meta) (lvsStartTime lvs)
                                          }
                     _ -> return ()
+            LogObject _ _ (LogMessage msg) ->
+                when ("As leader of slot" `isPrefixOf` msg) $
+                    modifyMVar_ (getbe lvbe) $ \lvs ->
```
CodiePP: the block number could go into the `lvsBlockHeight`

CodiePP: here "slot" number...

PR #56, opened 2019-07-15
All The Jobs!

angerman:
Build
- [x] windows cross compiled libraries
- [x] re-enable the darwin builds for inspection.


angerman:
I'm reasonably confident this commit is correct, however the underlying libraries (cardano-node) just don't build.

365andreas:
> Looks hacky, but it builds.
> Is this just disabling the LiveView on Windows or?

Now using `--live-view` on Windows will just do nothing (`return ())

I will change it so that using `--live-view` will just spawn a simple node without the TUI.

PR #57, opened 2019-07-15
Track tip in trace prefix

mrBliss:
This was part of my branch while doing last week's demo: print the first four characters of the hash of the tip of the node's chain in the prefix of each traced message.


mrBliss:
I'm not sure if we want to include it in the node.

PR #60, opened 2019-07-15
Configurability of `GenesisConfig`

deepfire:
0. Restore compatibility with latest `cardano-shell`
0. Add dummy genesis dumping mode to `genesis-tool`
0. Extend genesis dumping to also dump delegation certificates
0. Wire up:
   - _Partial Options Monoid_-based `CardanoConfiguration`-sourced specification of
      - genesis
      - static key material (`PbftLeaderCredentials`)
      ..into `cardano-node` and `wallet-client`
   - unto `cardano-ledger`'s `mkConfigFromFile`
   - and the resulting `GenesisConfig` unto `DemoProtocol`
0. Commit dummy genesis & secrets as files
0. Wire up `demo2.sh` to work


app/Main.hs
```diff
@@ -66,11 +67,16 @@ main = do
...
     finalConfig <- either (throwIO . ConfigurationError) pure $
           finaliseCardanoConfiguration $
-          -- Here we perform merging of layers of configuration, but for now, only in a trivial way.
+          -- Here we perform merging of layers of configuration, but for now, only in a trivial way,
+          -- just for Cardano.Shell.Constants.Types.Genesis.
           -- We expect this process to become generic at some point.
-          partialConfig { pccNode = pccNode partialConfig }
+          partialConfig { pccCore = pccCore' <> (pure $ mempty { pcoGenesis = (join $ pcoGenesis <$> pccCore')
+                                                                 <> genesisSpec nodeCli
+                                                               })}
```
dcoutts:
```
pccCore' <> pure (...)
```
This is wrong, or not what you meant. Think about it. You've said that `pccCore' :: Last ...`, and now you're using the monoid operation with a pure (i.e. `Last`) as the right hand arg.

For `Last`:
```
forall x.  x <> pure y = pure y
```

Also, this expression is getting unwieldy. You've tried to make it smaller using a `let` but put the `let` far away. You can use a local let and make this all fit in 80 cols. Note also that we can probably simplify the merging of the config if the top level thing is not a `Last` monoid but a product monoid. e.g. something like:
```
finaliseCardanoConfiguration $
-- the comments
let mergeCLI :: PartialCore -> PartialCore
    mergeCLI core = core <> mempty { pcoGenesis = genesisSpec nodeCli }
in partialConfig { pccCore = fmap mergeCLI (pccCore partialConfig) }
```

deepfire: > This is wrong, or not what you meant. Think about it. You've said that `pccCore' :: Last ...`, and now you're using the monoid operation with a pure (i.e. `Last`) as the right hand arg.

Indeed.. I wanted a component-wise monoidal append, but got a whole-sale override by the right-side in its stead.  Thank you for catching it!

deepfire: Would you discourage a point-free variant I submitted?

deepfire:
> The last four patches keep chopping and changing. I'd be tempted to squash the penultimate 3 patches.

Done!

> So it looks like with the CLI and partial config that we can still run the existing simple demo with the hard coded demo genesis.

Yes, that was the intention.

> Any chance you'll be able to make an external JSON file equivalent of the current hard-coded demo genesis? That would be ideal really.

I would have loved it, actually.

> We don't really need a hard coded one (except in consensus tests), and it might simplify things.

Will do.

app/Main.hs
deepfire: The rabbit hole was embarassingly uniform.  The nested children need the same treatment.

cardano-node/Cardano/Node/CLI.hs
```diff
@@ -1,5 +1,6 @@
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE RecordWildCards     #-}
```
ruhatch: Spacing here is a bit odd

app/Main.hs
```diff
@@ -61,11 +64,19 @@ main = do
...
-initializeAllFeatures :: CLIArguments -> CardanoConfiguration -> CardanoEnvironment -> IO ([CardanoFeature], NodeLayer)
-initializeAllFeatures (CLIArguments logCli nodeCli) cardanoConfiguration cardanoEnvironment = do
+initializeAllFeatures :: CLIArguments -> PartialCardanoConfiguration -> CardanoEnvironment -> IO ([CardanoFeature], NodeLayer)
+initializeAllFeatures (CLIArguments logCli nodeCli) partialConfig cardanoEnvironment = do
+    finalConfig <- either (throwIO . ConfigurationError) pure $
+          finaliseCardanoConfiguration $
+          -- Here we perform merging of layers of configuration, but for now, only in a trivial way,
+          -- just for Cardano.Shell.Constants.Types.Genesis.
+          -- We expect this process to become generic at some point.
+          partialConfig { pccCore =
```
ruhatch: I think this is still a bit confusing - I think it would be worth pulling this out into a function with more comments

cardano-node/Cardano/Node/CLI.hs
erikd: Please avoid using `RecordWIldCards`. In almost all circumstances it makes code less clear, readable and maintainable.

deepfire: Dropped RWC & fixed spacing.

app/Main.hs
deepfire: Done!

cardano-node/Cardano/Node/CLI.hs
```diff
@@ -93,34 +118,67 @@ data SomeProtocol where
...
+    let plc = PbftLeaderCredentials sk dlg
+        p   = ProtocolRealPBFT defaultDemoPBftParams geneConfig (pure plc)
+
     case Consensus.runProtocol p of
-      Dict -> return $ SomeProtocol p
-  where
-    p = ProtocolRealPBFT defaultDemoPBftParams genesisConfig
-    genesisConfig = Dummy.dummyConfig
+      Dict -> do
+        return . SomeProtocol $ p
```
ruhatch: This can probably revert to `Dict -> return $ SomeProtocol p`

cardano-node/Cardano/Node/CLI.hs
```diff
@@ -93,34 +118,67 @@ data SomeProtocol where
...
   | SimpleView  -- Simple mode, just output text.
 
+{-------------------------------------------------------------------------------
+  Configuration merging
+-------------------------------------------------------------------------------}
+
+-- | Perform merging of layers of configuration, but for now, only in a trivial way,
+--   just for Cardano.Shell.Constants.Types.{Genesis,StaticKeyMaterial}.
+--   We expect this process to become generic at some point.
+mergeConfiguration :: PartialCardanoConfiguration -> Last PartialGenesis -> Last PartialStaticKeyMaterial -> PartialCardanoConfiguration
```
ruhatch: You can indent this type signature for readability :)
```suggestion
mergeConfiguration 
  :: PartialCardanoConfiguration 
  -> Last PartialGenesis 
  -> Last PartialStaticKeyMaterial 
  -> PartialCardanoConfiguration
```

ruhatch:
Some squashing of the commits would be good before merging :)

deepfire:
> Some squashing of the commits would be good before merging :)

Done!

cardano-node/Cardano/Node/CLI.hs
deepfire: Done!

cardano-node/Cardano/Node/CLI.hs
deepfire: Done!

app/Main.hs
coot: Yes point-free is discouraged, (which includes over using of `Control.Arrow`), often it's much easier to read non point-free, with explicit lambdas.  But I haven't seen your point-free term, so take it as guide-line, which at times might be good to break.

app/CLI.hs
```diff
@@ -37,6 +42,8 @@ import           Cardano.Node.CLI
 data NodeCLIArguments = NodeCLIArguments {
     systemStart  :: !SystemStart
   , slotDuration :: !SlotLength
+  , genesisSpec  :: !(Last PartialGenesis)
+  , keyMaterialSpec :: !(Last PartialStaticKeyMaterial)
```
coot: All `::` used to be aligned; Doing that pollutes the git diff, but keeps things tidy when reading a module.

deepfire: Diff pollution was the reason I didn't do it, yeah.. Done.

app/Main.hs
```diff
@@ -62,11 +64,20 @@ main = do
...
+                          flip fmap (pccCore partialConfig) $
+                          \x -> x { pcoGenesis            = (<>) <$> pcoGenesis            x <*> genesisSpec      nodeCli
```
coot: I think that's a sign that we should not define everything inside a `Last`.  You'd have here just `<>`, it should be just `<>`. With this design there are just too many monoids available, do we need them all?

app/Run.hs
```diff
@@ -116,8 +117,8 @@ instance Condense Peer where
     condense (Peer localAddr remoteAddr) = (show localAddr) ++ (show remoteAddr)
 
 
-runNode :: NodeCLIArguments -> LoggingLayer -> IO ()
-runNode nodeCli@NodeCLIArguments{..} loggingLayer = do
+runNode :: NodeCLIArguments -> LoggingLayer -> CardanoConfiguration -> IO ()
+runNode nodeCli@NodeCLIArguments{..} loggingLayer cc = do
     let !tr = (llAppendName loggingLayer) "node" (llBasicTrace loggingLayer)
```
coot: I know that this is outside of the scope of this PR, but let me just note that the first pair of brackets is redundant.

cardano-node/Cardano/Node/CLI.hs
```diff
@@ -93,28 +110,43 @@ data SomeProtocol where
...
-fromProtocol MockPBFT =
+fromProtocol _ MockPBFT =
     case Consensus.runProtocol p of
       Dict -> return $ SomeProtocol p
   where
     p = ProtocolMockPBFT defaultDemoPBftParams
-fromProtocol RealPBFT =
+fromProtocol cc@CardanoConfiguration{..} RealPBFT = do
+    let Genesis{..} = coGenesis ccCore
+        genHash = either (throw . ConfigurationError) id $ -- XXX: need a proper error tag
```
coot: Is this left for later, or did you intended to do this in this PR?

cardano-node/Cardano/Node/CLI.hs
```diff
@@ -58,8 +63,20 @@ import           Cardano.Chain.Common
 import           Cardano.Chain.Genesis
 import           Cardano.Crypto.ProtocolMagic
 
-import qualified Test.Cardano.Chain.Genesis.Dummy as Dummy
-
+import           Cardano.Shell.Lib (GeneralException (..))
+import qualified Cardano.Chain.Genesis as Genesis
```
coot: This module is already imported unqualified.  Maybe add a todo to remove the unqualified import.

cardano-node/Cardano/Node/CLI.hs
```diff
@@ -1,6 +1,7 @@
 {-# LANGUAGE ConstraintKinds  #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GADTs            #-}
+{-# LANGUAGE RecordWildCards  #-}
```
coot: Can we avoid `RecordWildCards`?

cardano-node/Cardano/Node/CLI.hs
```diff
@@ -93,28 +110,43 @@ data SomeProtocol where
...
+    let Genesis{..} = coGenesis ccCore
+        genHash = either (throw . ConfigurationError) id $ -- XXX: need a proper error tag
+                  decodeAbstractHash geGenesisHash
+        cvtRNM :: RequireNetworkMagic -> RequiresNetworkMagic
+        cvtRNM NoRequireNetworkMagic = RequiresNoMagic
+        cvtRNM RequireNetworkMagic   = RequiresMagic
+        StaticKeyMaterial{..} = coStaticKeyMaterial ccCore
+        kmoDeserialiseDelegateKey = Signing.SigningKey . snd . either (error . show) id . deserialiseFromBytes Signing.fromCBORXPrv
+    genesisConfig <- either (error . show) id <$> -- XXX: need a proper error tag
+      runExceptT (Genesis.mkConfigFromFile (cvtRNM $ coRequiresNetworkMagic ccCore) geSrc genHash)
```
coot: sometimes it is more readable to just use explicit case analysis:
```
cfg <- runExceptT ...
case cfg of
   Left err -> error (show err)
  Right res -> ...
```

cardano-node/Cardano/Node/CLI.hs
```diff
@@ -93,28 +110,43 @@ data SomeProtocol where
...
+    let skm = StaticKeyMaterialByronPbft sk dlg
+        p   = ProtocolRealPBFT defaultDemoPBftParams genesisConfig skm
+
     case Consensus.runProtocol p of
-      Dict -> return $ SomeProtocol p
-  where
-    p = ProtocolRealPBFT defaultDemoPBftParams genesisConfig
-    genesisConfig = Dummy.dummyConfig
+      Dict -> do
+        return . SomeProtocol $ p
```
coot: If you write it this way `return $ SomeProtocol p` it will be one character less.  :thinking: I just wrote a whole sentence to reduce a single character :D.

wallet-client/CLI.hs
```diff
@@ -8,21 +8,27 @@ module CLI
...
 import           Cardano.Node.CLI
+import           Cardano.Shell.Constants.CLI
+import           Cardano.Shell.Constants.PartialTypes (PartialGenesis (..))
 
 data CLI = CLI {
     cliCoreNodeId   :: CoreNodeId,
     cliNumCoreNodes :: NumCoreNodes,
-    cliProtocol     :: Protocol
+    cliProtocol     :: Protocol,
+    cliGenesis      :: (Last PartialGenesis)
```
coot: These parens are redundant.

wallet-client/CLI.hs
```diff
@@ -8,21 +8,27 @@ module CLI
...
   }
 
 parseCLI :: Parser CLI
 parseCLI = CLI
     <$> parseCoreNodeId
     <*> parseNumCoreNodes
     <*> parseProtocol
+    <*> (pure <$> configGenesisCLIParser)
```
coot: It's better to use `Last` than `pure`.

app/Main.hs
deepfire: Yes, the point-free variant did not survive in the end -- the whole thing was isolated into a separate function: https://github.com/input-output-hk/cardano-node/pull/60/files#diff-ba9f0f06843a3a67a21c656d60e82bc8R175

app/CLI.hs
```diff
@@ -49,6 +56,8 @@ nodeParser :: Parser NodeCLIArguments
 nodeParser = NodeCLIArguments
     <$> parseSystemStart
     <*> parseSlotDuration
+    <*> (pure <$> configGenesisCLIParser)
```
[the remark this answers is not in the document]
deepfire: @coot, the `pure` has to be replaced with `Last . Just ` -- do you still think that would be an improvement?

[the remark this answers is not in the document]
coot: In the first case it's isomorphic to `3`, in the latter case it might be :rocket: as well ;)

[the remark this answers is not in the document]
deepfire: Hmm, I'm not sure I get it -- isn't the context guiding inference here (as always) -- and so type of `NodeCLIArguments` makes it pretty unambiguous that it's not something wrong?

wallet-client/CLI.hs
deepfire: `Just . Last`, done.

cardano-node/Cardano/Node/CLI.hs
deepfire: Done!

app/CLI.hs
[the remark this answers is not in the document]
deepfire: Done.

app/Main.hs
deepfire: Hmm, this particular comment pertains to the code that no longer exists -- that piece of code was removed in `42f4e3b78d1eaf30ff97d37d5cc00f6de5ba05bb`.

Something odd with github?

app/Run.hs
deepfire: Done!

cardano-node/Cardano/Node/CLI.hs
deepfire: Since it'd involve changing `cardano-shell`, I think `ConfigurationError` might be sufficient for now.  I removed the TODO.

cardano-node/Cardano/Node/CLI.hs
deepfire: Removed the unqualified import.

cardano-node/Cardano/Node/CLI.hs
deepfire: Interesting, this again refers to an outdated commit -- that instance of RWC was removed in `42f4e3b78d1eaf30ff97d37d5cc00f6de5ba05bb`.

cardano-node/Cardano/Node/CLI.hs
deepfire: Done.

cardano-node/Cardano/Node/CLI.hs
deepfire: Interesting -- this again refers to an outdated commit -- this was recently shortened.

wallet-client/CLI.hs
deepfire: And again, that was already done in a recent commit.

app/CLI.hs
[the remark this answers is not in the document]
deepfire: I guess I see -- if we change the type of `NodeCLIArguments`, the `pure` users will be changed inadvertently.

wallet-client/Main.hs
```diff
@@ -54,26 +58,26 @@ instance Exception PartialConfigError
...
     runCardanoApplicationWithFeatures Development cardanoFeatures (cardanoApplication nodeLayer)
 
-initializeAllFeatures :: ArgParser -> CardanoConfiguration -> CardanoEnvironment -> IO ([CardanoFeature], NodeLayer)
-initializeAllFeatures (ArgParser logCli cli) cardanoConfiguration cardanoEnvironment = do
+initializeAllFeatures :: ArgParser -> PartialCardanoConfiguration -> CardanoEnvironment -> IO ([CardanoFeature], NodeLayer)
+initializeAllFeatures (ArgParser logCli cli) partialConfig cardanoEnvironment = do
+
+    finalConfig <- either (throwIO . ConfigurationError) pure $
+                   finaliseCardanoConfiguration $
+                   mergeConfiguration partialConfig (cliGenesis cli) (cliKeyMaterial cli)
```
coot: I think it's better to put this way:
```
case finaliseCardanoConfiguration ... of
  Left err -> throwIO (ConfigurationError err)
  Right finalConfig -> ...
```

wallet-client/CLI.hs
```diff
@@ -8,21 +8,30 @@ module CLI
...
+    cliKeyMaterial  :: Last PartialStaticKeyMaterial
   }
 
 parseCLI :: Parser CLI
 parseCLI = CLI
     <$> parseCoreNodeId
     <*> parseNumCoreNodes
     <*> parseProtocol
+    <*> (Last . Just <$> configGenesisCLIParser)
+    <*> (Last . Just <$> configStaticKeyMaterialCLIParser)
```
coot: ok, this looks much better now :)

wallet-client/Main.hs
deepfire: Done, as well as in `cardano-node`'s case.

wallet-client/CLI.hs
deepfire: : -)

deepfire:
Extended by https://github.com/input-output-hk/cardano-node/pull/79 for a more complete demo.

app/CLI.hs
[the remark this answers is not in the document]
coot: My point is that it's easy for the compiler to infer the right type, but it's not as easy for the reader.

cardano-node/Cardano/Node/CLI.hs
coot: It's still throwing in a pure code. If you'd like to be 100% kosher, then using `throwIO` is a nicer alternative.

coot:
What's left here? @deepfire if you address @dcoutts comments by now, consider dismissing the review and re-requesting a new one.

app/genesis-tool/Run.hs
```diff
@@ -144,6 +130,70 @@ runCommand
...
+  (configGenesisData Dummy.dummyConfig)
+  $ fromMaybe (error "Hardcoded genesis lacks secrets.") (configGeneratedSecrets Dummy.dummyConfig)  
+
+runCommand
+  KeyMaterialOps{..}
+  (PrintGenesisHash
+    secretPath)
+  =
+  putStrLn =<< F.format CCr.hashHexF . unGenesisHash . snd . either (error . show) id <$> (runExceptT $ readGenesisData secretPath)
+  where
```
ksaric: Ha? `where` what?

app/genesis-tool/Run.hs
```diff
@@ -144,6 +130,70 @@ runCommand
...
+
+runCommand
+  KeyMaterialOps{..}
+  (PrintSigningKeyAddress
+    networkMagic
+    secretPath)
+  =
+  putStrLn =<< T.unpack . prettyAddress . CC.makeVerKeyAddress networkMagic . CCr.toVerification . kmoDeserialiseDelegateKey <$> LB.readFile secretPath
+  where
```
ksaric: `where` what?

app/genesis-tool/Run.hs
```diff
@@ -144,6 +130,70 @@ runCommand
...
+
+runCommand
+  KeyMaterialOps{..}
+  (PrintGenesisHash
+    secretPath)
+  =
+  putStrLn =<< F.format CCr.hashHexF . unGenesisHash . snd . either (error . show) id <$> (runExceptT $ readGenesisData secretPath)
+  where
+
+runCommand
```
ksaric: No type signatures?

app/genesis-tool/Run.hs
```diff
@@ -144,6 +130,70 @@ runCommand
...
+  KeyMaterialOps{..}
+  (PrintSigningKeyAddress
+    networkMagic
+    secretPath)
+  =
+  putStrLn =<< T.unpack . prettyAddress . CC.makeVerKeyAddress networkMagic . CCr.toVerification . kmoDeserialiseDelegateKey <$> LB.readFile secretPath
```
ksaric: I wouldn't force it this much, but it's a matter of style, so fine.

wallet-client/Main.hs
```diff
@@ -54,26 +58,28 @@ instance Exception PartialConfigError
...
 
-initializeAllFeatures :: ArgParser -> CardanoConfiguration -> CardanoEnvironment -> IO ([CardanoFeature], NodeLayer)
-initializeAllFeatures (ArgParser logCli cli) cardanoConfiguration cardanoEnvironment = do
+initializeAllFeatures :: ArgParser -> PartialCardanoConfiguration -> CardanoEnvironment -> IO ([CardanoFeature], NodeLayer)
+initializeAllFeatures (ArgParser logCli cli) partialConfig cardanoEnvironment = do
+
+    finalConfig <- case finaliseCardanoConfiguration $
+                        mergeConfiguration partialConfig (cliGenesis cli) (cliKeyMaterial cli)
+                   of
+      Left err -> throwIO $ ConfigurationError err
```
ksaric: :scream:

wallet-client/CLI.hs
```diff
@@ -8,21 +8,30 @@ module CLI
...
+    cliGenesis      :: Last PartialGenesis,
+    cliKeyMaterial  :: Last PartialStaticKeyMaterial
   }
 
 parseCLI :: Parser CLI
 parseCLI = CLI
     <$> parseCoreNodeId
     <*> parseNumCoreNodes
     <*> parseProtocol
+    <*> (Last . Just <$> configGenesisCLIParser)
```
ksaric: That is something I'm fixing currently in `cardano-shell`.

cardano-node/Cardano/Node/CLI.hs
```diff
@@ -93,34 +117,76 @@ data SomeProtocol where
...
+
+-- | Perform merging of layers of configuration, but for now, only in a trivial way,
+--   just for Cardano.Shell.Constants.Types.{Genesis,StaticKeyMaterial}.
+--   We expect this process to become generic at some point.
+mergeConfiguration
+  :: PartialCardanoConfiguration
+  -> Last PartialGenesis
+  -> Last PartialStaticKeyMaterial
+  -> PartialCardanoConfiguration
+mergeConfiguration pcc@PartialCardanoConfiguration{pccCore} pg pskm =
```
ksaric: That is something I'm fixing currently in `cardano-shell`.
This pattern is not that well suited for nested configurations as you can see so we have to keep our configuration at a maximum level of 2 layers if we want it to be somewhat understandable.

dcoutts: Or make structures like `PartialGenesis` etc not be `Last` fields, since the leaves of those types are also monoids. Then it should work fine with nested records.

You only need `Last` on a record when that record itself is not a product of monoids. We don't need multiple levels of wrapping. Then merging from the top works very nicely.

Config that sets a single field, and then `<>` at the top level.
```
mainConfig <> mempty { foo = mempty { bar = x } }
```

app/genesis-tool/Run.hs
coot: The line is also pretty long, could we break it so it fits into `80` characters?

app/genesis-tool/Run.hs
coot: This line is too long.  If you fit in 80 characters, those with smaller monitors (laptos with high resolution) or bigger monitors can fit splits of 2 (or more). This is really handy when working with code.

PR #63, opened 2019-07-16
Update to HEAD of ouroboros-network

mrBliss:
Changes required due to https://github.com/input-output-hk/ouroboros-network/pull/727, https://github.com/input-output-hk/ouroboros-network/pull/776, and https://github.com/input-output-hk/ouroboros-network/pull/754.


cardano-node/Cardano/Node/CLI.hs
```diff
@@ -61,30 +71,47 @@ data Protocol =
...
+    case Consensus.runProtocol p of
       Dict -> return $ SomeProtocol p
   where
-    p = DemoMockPBFT defaultDemoPBftParams
+    p = ProtocolMockPBFT defaultDemoPBftParams
 fromProtocol RealPBFT =
-    case runDemo p of
-      Dict -> return $ SomeProtocol p
+    case Consensus.runProtocol p of
+      Dict -> return $ undefined -- SomeProtocol p
```
mrBliss: @nc6 I get:
```
[-Wdeferred-type-errors]
    • Could not deduce (reflection-2.1.4:Data.Reflection.Given
                          cardano-ledger-0.1.0.0:Cardano.Chain.Block.Header.HeaderHash)
        arising from a use of ‘SomeProtocol’
      from the context: Ouroboros.Consensus.Node.Run.Abstract.RunNode
                          (Ouroboros.Consensus.Ledger.Byron.ByronBlock ByronConfig)
        bound by a pattern with constructor:
                   Dict :: forall (a :: Constraint). a => Dict a,
                 in a case alternative
    • In the second argument of ‘($)’, namely ‘SomeProtocol p’
      In the expression: return $ SomeProtocol p
      In a case alternative: Dict -> return $ SomeProtocol p
```

So `ByronGiven`. I don't know why I get this error now and not before.

mrBliss: @nc6 Any progress on this?

PR #64, opened 2019-07-17
Added tracking of transactions processed by node.

365andreas:
The number of transactions which were send to node whether they were valid or not will be recorded as `transactions processed`.

### DIFF
--- a/app/Run.hs
+++ b/app/Run.hs
@@ -409,6 +409,14 @@ handleSimpleNode p NodeCLIArguments{..} myNodeAddress (TopologyInfo myNodeId top
           let logValue = LogValue "txsInMempool" $ PureI $ fromIntegral $ _txsInMempool mempoolEvent
           meta <- mkLOMeta Info Confidential
           traceNamedObject tr (meta, logValue)
+          case mempoolEvent of
+            TraceMempoolAddTxs      txs _ ->
+                let logValue' = LogValue "txsProcessed" $ PureI $ fromIntegral $ length txs in
+                traceNamedObject tr (meta, logValue')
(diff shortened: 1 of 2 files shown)


app/Run.hs
```diff
@@ -409,6 +409,14 @@ handleSimpleNode p NodeCLIArguments{..} myNodeAddress (TopologyInfo myNodeId top
           let logValue = LogValue "txsInMempool" $ PureI $ fromIntegral $ _txsInMempool mempoolEvent
           meta <- mkLOMeta Info Confidential
           traceNamedObject tr (meta, logValue)
+          case mempoolEvent of
+            TraceMempoolAddTxs      txs _ ->
+                let logValue' = LogValue "txsProcessed" $ PureI $ fromIntegral $ length txs in
+                traceNamedObject tr (meta, logValue')
+            TraceMempoolRejectedTxs txs _ ->
```
ksaric: Are these two cases equal?

365andreas: Yes the first one is about valid and the second about invalid transactions. We will keep the total number of txs sent to the node.

ksaric: So we simply correct the number of transactions depending on whether the two events occurred - `TraceMempoolAddTxs` or `TraceMempoolRejectedTxs`?
In both cases, we don't have any additional information but simply the number of transactions?

Do we even need two events than now?
Not saying we should change it, but something to keep in mind in the future.

365andreas: We can track only the valid ones.
We do not use the 2 separate events at the time being.
Warning messages could be produced by the rejected in the future.

PR #65, opened 2019-07-17
Tip of chain (pointHash).

denisshevchenko:
Now we show `tip of chain` as a `pointHash` (7 symbols) taken from `TVar Point`.


app/LiveView.hs
```diff
@@ -193,7 +196,17 @@ instance IsEffectuator LiveViewBackend Text where
                                          }
                     _ ->
                         return ()
-            LogObject _ _ (LogMessage msg) ->
+            LogObject _ _ (LogMessage msg) -> do
+                when ("[" `isPrefixOf` msg) $ do
+                    let pointHashInBrackets = T.take 9 msg
```
365andreas: You could reuse `when` here since you imported it.

PR #66, opened 2019-07-17
add source-pins

cleverca22:


nc6:
@cleverca22 Is this still relevant? Labelled WIP but many months old.

cleverca22:
it is still relevant, but has needs the conflicts fixed up first

PR #70, opened 2019-07-19

karknu:
2af66dc1 fixes a bug which caused cardano-node to not open connections
to all configured producers.


karknu:
Drat, will sort it out.

PR #71, opened 2019-07-19
Update cardano-ledger

nc6:
This bumps a bunch of dependencies, and updates the code to work with cardano-shell again. The way in which this is done involved a partial pattern match - I expect a proper updating to arrive with https://github.com/input-output-hk/cardano-node/pull/60

As a consequence, this fixes the `--real-pbft` demo.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -6,19 +6,19 @@ package cardano-node
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-base
-  tag: 501d91e426ae84ce0ae056be38bd3db594af9fc2
+  tag: 6b808ad5506cb097cdf5832e1cd5cad0c83c58d6
   subdir: binary
 
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-base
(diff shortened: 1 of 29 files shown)


cabal.project
```diff
@@ -29,94 +29,94 @@ source-repository-package
...
   location: https://github.com/input-output-hk/iohk-monitoring-framework
   subdir:   contra-tracer
-  tag: b2022cf7f5925e486f9e0a31f6f6d02145d5ebda
+  tag: bd31cd2f3922010ddb76bb869f29c4e63bb8001b
 
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/ouroboros-network
-  tag: 0fb0db07e8c3fdd60aa8e22492782f2afcf74788
+  tag: d0f4b732d34a1aa8693fc45f22c9689569c25a91
```
mrBliss: This is the hash of the commit in #798, which is not the same as the one on master, which is ca270835199406a06be590d2391bd87b072e7537. This is why CI is failing.

nc6: Oh, I thought I changed that. One sec...

PR #72, opened 2019-07-19
`cardano-node` service definitions & ghcid in Nix shell

deepfire:


jbgi:
I am not clear about the use-case for the instanced service. Could you explicit them? It it to reduce the cost of dev clusters running on aws? or for local testing, in which case, a docker based solution might be more appropriate? 
Due to operational concerned, I believe the monitored clusters will only run one instance per machine.

deepfire:
@jbgi, indeed, the intent is to have a single container/VM to run a local cluster.

deepfire:
Deprecated in favor of a different approach.

(Although, I think this still has certain merit, I'd say.  It's nice to be able to run a bunch of nodes as distinct services on a single machine..)

PR #76, opened 2019-07-24
Changed license to Apache 2.0

coot:


nix/.stack.nix/cardano-node.nix
```diff
@@ -4,8 +4,8 @@
     package = {
       specVersion = "1.10";
       identifier = { name = "cardano-node"; version = "3.0.1.87"; };
-      license = "MIT";
-      copyright = "2019 IOHK";
+      license = "NONE";
```
mrBliss: Same here.

cardano-node.cabal
```diff
@@ -4,8 +4,10 @@ description:    The cardano full node
 author:         IOHK
 maintainer:     operations@iohk.io
 copyright:      2019 IOHK
-license:        MIT
```
mrBliss: Missing:
```
license:             Apache-2.0
```

nix/.stack.nix/cardano-node.nix
coot: That's wired, let me check.

coot: ok, I missed the `license` field in `cardano-node.cabal` file.

cardano-node.cabal
```diff
@@ -4,8 +4,11 @@ description:    The cardano full node
 author:         IOHK
 maintainer:     operations@iohk.io
 copyright:      2019 IOHK
```
mrBliss: I just got this warning:
```
Cabal file warning in/home/thomas/work/cardano-node/cardano-node.cabal@6:1: The field "copyright" is specified more than once at positions 6:1, 11:1
```

PR #78, opened 2019-07-25
LiveView: stop the TUI & node itself.

denisshevchenko:
Now pressing `q` (in a live view) stops the TUI as well as the node itself.

### DIFF
--- a/app/LiveView.hs
+++ b/app/LiveView.hs
@@ -10,6 +10,7 @@ module LiveView (
     , effectuate
     , captureCounters
     , setTopology
+    , setNodeThread
     ) where
 
 import           Control.Concurrent (threadDelay)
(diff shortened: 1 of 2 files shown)


denisshevchenko:
So, my idea is to make `q` more powerful. We display:

> Press Q to quit

and user expects that `quit` means complete exit (stop the node). So we pass `nodeThread` to `LiveView`, and after user pressed `q` - just `cancel` node's working thread (function `handleSimpleNode`).

ksaric:
That is fine, but we need it to work for other signals as well.
We cannot have a terminal that doesn't respond to outside signals.

One of these signals is CTRL+C and it's currently being swallowed by that same thread (I presume).
You can simply rethrow the exception or throw an asynchronous one.

PR #79, opened 2019-07-25
[DONT-MERGE] Friday demo umbrella

deepfire:


wallet-client/CLI.hs
```diff
@@ -24,14 +24,12 @@ data CLI = CLI {
     cliCoreNodeId   :: CoreNodeId,
     cliNumCoreNodes :: NumCoreNodes,
     cliProtocol     :: Protocol,
-    cliGenesis      :: Last PartialGenesis,
-    cliKeyMaterial  :: Last PartialStaticKeyMaterial
```
coot: You should also remove `Partialgenesis` and `PartailStaticKeyMaterial` from the import list; I should not pass the CI. If this did, file a ticket to update the nix part to run with `-Werror`; `-Wall` is already enabled in the `cabal` file.

app/CLI.hs
```diff
@@ -49,6 +56,8 @@ nodeParser :: Parser NodeCLIArguments
 nodeParser = NodeCLIArguments
     <$> parseSystemStart
     <*> parseSlotDuration
+    <*> (pure <$> configGenesisCLIParser)
+    <*> (pure <$> configStaticKeyMaterialCLIParser)
```
coot: It would be better to use the specific constructor than `pure`

configuration/Test.Cardano.Chain.Genesis.Dummy.dummyConfig.configGenesisData.json
```diff
@@ -0,0 +1,114 @@
+{ "bootStakeholders":
```
coot: is this file only used in some tests, if so which one?

scripts/demo2.sh
```diff
@@ -8,18 +8,42 @@
 
 ALGO="--real-pbft"
 NOW=`date "+%Y-%m-%d 00:00:00"`
-NETARGS="--system-start '${NOW}' --slot-duration 2 node -t configuration/simple-topology.json ${ALGO}"
-ACCARGS="--system-start '${NOW}' --slot-duration 2 trace-acceptor"
+NETARGS=(
+        --system-start "'${NOW}'"
+        --slot-duration 2
+        --genesis-file "configuration/Test.Cardano.Chain.Genesis.Dummy.dummyConfig.configGenesisData.json"
```
coot: I see it's hardcoded here, why it is called `Test`?

PR #80, opened 2019-07-26
Chairman

coot:
Example usage, run the demo and in another terminal:
```
cabal new-run chairman -- -n 0 -n 1 -n 2 --real-pbft -k 5 -s 20 -t 40
```
This will evaluate first 20 blocks, and it will timeout after 40 seconds. (Currently this will timeout as the demo is not working)

or run the demo with `--bft` and
```
 cabal new-run chairman -- -n 0 -n 1 -n 2 --bft -k 5
```
it will continuously track if there is consensus (`--bft` runs fine).


chairman/Chairman.hs
```diff
@@ -0,0 +1,434 @@
+{-# LANGUAGE BangPatterns        #-}
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE GADTs               #-}
+{-# LANGUAGE KindSignatures      #-}
+{-# LANGUAGE NamedFieldPuns      #-}
```
erikd: :tada:

chairman/Chairman.hs
```diff
@@ -0,0 +1,434 @@
...
+                      ( NotFoundCommonBlock tips
+                      , chains'
+                      )
+                    -- there is an intersection and all forks are short
+                    Just headSlot ->
+                      ( WitnessedConsensusAt headSlot (map Block.blockPoint $ catMaybes tips)
+                      , chains'
+                      )
+            -- there is a long fork
+            else throw $ NodeMisconduct tips
```
erikd: What monad is this throwing in?

chairman/Main.hs
```diff
@@ -0,0 +1,111 @@
...
+      (Just timeout, SomeProtocol protocol) ->
+        runChairman protocol
+                    caCoreNodeIds
+                    (NumCoreNodes $ length caCoreNodeIds)
+                    caSecurityParam
+                    caMaxBlockNo
+                    stdoutTracer
+        `race_` do
+          threadDelay (timeout * 1_000_000)
+          throwIO Timeout
```
erikd: At least here I know its `IO`.

chairman/Chairman.hs
```diff
@@ -0,0 +1,434 @@
...
+instance Condense blk => Show (ChairmanError blk) where
+    show (NodeMisconduct blks) = "NodeMisconduct " ++ condense blks
+
+instance (Condense blk, Typeable blk) => Exception (ChairmanError blk)
+
+
+-- | Check if the oldest immutable tip agree, keep forks from the last common
+-- block (including it).
+--
+checkAndSubtract
```
mrBliss: `Subtract` here is a bit confusing (in the helper too). Isn't there a clearer name?

chairman/Chairman.hs
coot: here it throws in pure code, it's wrapped in `MonadSTM m`, which at runtime will be `STM`.

mrBliss: There is a comment `it isn't really pure as it throws the 'NodeMisconduct' exception`. I would advise using `Except` or `Either` for this function and then throwing the exception in `STM` (use `MonadThrow (STM m`).

chairman/Chairman.hs
```diff
@@ -0,0 +1,434 @@
...
+    return res
+  where
+    -- it isn't really pure as it throws the 'NodeMisconduct' exception
+    subtractPure :: Map CoreNodeId (ChainFragment blk)
+                 -> ( ChairmanTrace blk
+                    , Map CoreNodeId (ChainFragment blk)
+                    )
+    subtractPure chains =
+      let tips :: [Maybe blk]
+          tips = foldl' (\tips' cf -> CF.head cf : tips') [] chains
```
mrBliss: This is just `map` in disguise, I would prefer:
```haskell
tips = map CF.head $ Map.elems chains
```

chairman/Chairman.hs
```diff
@@ -0,0 +1,434 @@
...
+          -- find intersection of all the chains; the CLI guarantees that
+          -- there's at least one entry in @chains@, thus @fold1@ is safe here.
+          common :: ChainFragment blk
+          common =
+            foldl1
+              (\common' cf ->
+                case CF.intersectChainFragments common' cf of
+                  Nothing -> CF.Empty
+                  Just (_, common'', _, _)
+                          -> common'')
```
mrBliss: Having this on a separate line just looks weird to me :slightly_smiling_face:. I would put it on the previous line and align the `->`

chairman/Chairman.hs
```diff
@@ -0,0 +1,434 @@
...
+              chains
+
+          mbHeadSlot = CF.headSlot common
```
mrBliss: `mbCommonHeadSlot`?

See the next comment: use a `Point` instead!

chairman/Chairman.hs
```diff
@@ -0,0 +1,434 @@
...
+      txSubmissionClient txv
+
+--
+-- Chain-Sync client
+--
+
+-- | 'ChainSyncClient' which build chain fragment; on every roll forward it will
+-- check if there is consensus on immutable chain.
+--
+chainSyncClient
```
mrBliss: Can you move this up a bit, closer to the implementation of the helper functions it uses (`addBlock`, etc.)? Because I constantly had to scroll back and forth over unrelated code :slightly_smiling_face:

chairman/Chairman.hs
coot: Done.

cardano-node/Cardano/Node/CLI.hs
```diff
@@ -113,7 +113,7 @@ fromProtocol RealPBFT =
     case Consensus.runProtocol p of
       Dict -> return $ SomeProtocol p
   where
-    p = ProtocolRealPBFT defaultDemoPBftParams genesisConfig
+    p = ProtocolRealPBFT defaultDemoPBftParams genesisConfig Nothing
```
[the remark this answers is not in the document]
coot: Right, we need to merge #60 first.

chairman/Main.hs
```diff
@@ -0,0 +1,111 @@
...
+    ChairmanArgs { caProtocol
+                 , caCoreNodeIds
+                 , caSecurityParam
+                 , caMaxBlockNo
+                 , caTimeout
+                 } <- execParser opts
+    p <- fromProtocol caProtocol
+    case (caTimeout, p) of
+      (Nothing, SomeProtocol protocol) ->
+        runChairman protocol
```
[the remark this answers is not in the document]
coot: Done

chairman/Main.hs
```diff
@@ -0,0 +1,111 @@
...
+
+import           Cardano.Node.CLI
+
+import           Chairman (runChairman)
+
+
+data ChairmanArgs = ChairmanArgs {
+      caProtocol      :: Protocol
+    , caCoreNodeIds   :: [CoreNodeId]
+    , caSecurityParam :: Int
```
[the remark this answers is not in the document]
coot: Done

chairman/Chairman.hs
```diff
@@ -0,0 +1,434 @@
...
+                           ptcl
+
+        in connectTo
+            (,)
+            (localInitiatorNetworkApplication
+              coreNodeId
+              chainsVar
+              securityParam
+              maxBlockNo
+              (contramap show tracer)
```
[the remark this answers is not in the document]
coot: Nice, I wasn't aware.

chairman/Chairman.hs
coot: > See the next comment: use a Point instead!

I don't see it; Either will work, because the slot is occupied by the same block in each chain.  But using `Point` might be more instructive.

chairman/Chairman.hs
coot: Right, it was much longer in a previous version...

coot:
@mrBliss I pushed fixes for all your comments, except breaking the `demo`. Let's merge #60 first.

mrBliss:
> @mrBliss I pushed fixes for all your comments, except breaking the `demo`. Let's merge #60 first.

There a still some unresolved comments, I think because GitHub collapsed them. There should be a button "7 hidden conversations **Load more...**"

coot:
Are you sure, the only non comment which I haven't addressed is the one which breaks the `demo`, but I prefer to wait for #60 before merging this one.

mrBliss:
@coot:
https://github.com/input-output-hk/cardano-node/pull/80#discussion_r307648434
https://github.com/input-output-hk/cardano-node/pull/80#discussion_r307648963
https://github.com/input-output-hk/cardano-node/pull/80#discussion_r307650188
https://github.com/input-output-hk/cardano-node/pull/80#discussion_r307650463
https://github.com/input-output-hk/cardano-node/pull/80#discussion_r307651334
https://github.com/input-output-hk/cardano-node/pull/80#discussion_r307651714

chairman/Chairman.hs
```diff
@@ -0,0 +1,434 @@
...
+--
+-- Submission client
+--
+
+-- | A 'LocalTxSubmissionClient' that submits transactions reading them from
+-- a 'TMVar'.  A real implementation should use a better synchronisation
+-- primitive.  This demo creates and empty 'TMVar' in
+-- 'muxLocalInitiatorNetworkApplication' above and never fills it with a tx.
+--
+txSubmissionClient
```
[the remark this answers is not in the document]
coot: Because local client needs both chain sync and tx submission. It's a limitation of the `ouroboros-network` interface, and partially a limitation of the multiplexer (if any mini-protocol exists, the multiplexer will kill all other mini-protocol threads as well).

chairman/Chairman.hs
```diff
@@ -0,0 +1,434 @@
...
+import           Ouroboros.Consensus.Protocol
+import           Ouroboros.Consensus.Util.Condense
+
+import           Network.TypedProtocol.Codec
+import           Network.TypedProtocol.Codec.Cbor
+import           Network.TypedProtocol.Driver
+import           Ouroboros.Network.Mux
+import           Ouroboros.Network.Block (BlockNo, HasHeader, HeaderHash, Point, SlotNo)
+import qualified Ouroboros.Network.Block as Block
+import           Ouroboros.Network.ChainFragment (ChainFragment)
```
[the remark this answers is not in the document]
coot: but `AnchoredFragment` requires that I know the `anchorPoint`; I start with an empty chain, which is not anchored, then I will have to use `Maybe AnchoredFragment`. I feel that I will only have additional layer of `Maybe`s to deal with.

chairman/Chairman.hs
```diff
@@ -0,0 +1,434 @@
...
+       )
+    => CoreNodeId
+    -> ChainsVar m blk
+    -> blk
+    -> STM m ()
+addBlock coreNodeId chainsVar blk =
+    modifyTVar' chainsVar (Map.adjust (CF.addBlock blk) coreNodeId)
+
+
+data ChairmanError blk = NodeMisconduct [blk]
```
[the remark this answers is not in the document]
coot: I'll add a comment

chairman/Chairman.hs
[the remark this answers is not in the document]
mrBliss: The `anchorPoint` of an empty chain is ... genesis (`GenesisPoint`) 🙂. No `Maybe`s needed.

chairman/Chairman.hs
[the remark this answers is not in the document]
mrBliss: I see, that's unfortunate.

chairman/Chairman.hs
```diff
@@ -0,0 +1,434 @@
...
+              chains
+
+          mbHeadSlot = CF.headSlot common
+
+          -- remove common chain fragment from all the chains, but preserve the
+          -- common intersection point, this way we avoid leaving empty chains.
+          !chains'  = case mbHeadSlot of
+            Nothing       -> chains
+            Just headSlot ->
+                  (CF.takeWhileNewest (\blk -> Block.blockSlot blk >= headSlot))
```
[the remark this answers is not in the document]
coot: Good point, I'll use a point here.

chairman/Chairman.hs
coot: what about `prune`?

mrBliss: Yeah, that's better 👍

coot:
@mrBliss done, using `AnchoredChains` actually simplified things. I also added a commit which handles `MuxError` thrown when reading from closed bearer.

cardano-node/Cardano/Node/CLI.hs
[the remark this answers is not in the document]
coot: Fixed

chairman/Chairman.hs
```diff
@@ -0,0 +1,468 @@
...
+      traceWith tracer (show err)
+      atomically $ modifyTVar' chainsVar (Map.delete coreNodeId)
+
+
+data ChairmanTrace blk
+  = NotFoundCommonBlock [Point blk]
+  -- ^ the common block was not found, present list of head points.  Might be
+  -- empty if the chain was empty.
+  | WitnessedConsensusAt (Point blk) [Point blk]
+  -- ^ witness consenss at a given point.  The list is a list of tip points of
```
mrBliss: `s/consenss/consensus/`

chairman/Chairman.hs
```diff
@@ -0,0 +1,468 @@
...
+    -> STM m (ChairmanTrace blk)
+checkAndPrune chainsVar (SecurityParam securityParam) = do
+    chains <- readTVar chainsVar
+    case checkAndPrunePure chains of
+      Left err -> throwM err
+      Right (res, chains') -> do
+        writeTVar chainsVar $! chains'
+        return res
+  where
+    -- it isn't really pure as it throws the 'NodeMisconduct' exception
```
mrBliss: This comment is outdated.

chairman/Chairman.hs
coot: Good catch

coot:
Done, and rebased on top of merged #60.  I couldn't check weather it works since `script/demo.sh` got broken.

PR #81, opened 2019-07-26
Rename demo

coot:


365andreas:
@coot Added a commit to update the scripts. These changes are essential after the changes into master.

PR #93, opened 2019-08-01
ci/check-dependencies-merged-to-master.sh

deepfire:
This adds a new required Buildkite CI check -- as can be seen below, there's now a new entry -- `buildkite/cardano-node/dependencies-at-master`: https://buildkite.com/input-output-hk/cardano-node/builds/1494#d91bcbbd-8b4a-49ee-912e-0abec8926b25

When it fails, it provides output of the following kind:
```
bimap: master
cardano-binary: master
cardano-binary-test: master
cardano-crypto-class: cardano-crypto-class 0e26a4f00324100b4c34f2a9c7606ae03e94d7a1 not on master of https://github.com/input-output-hk/cardano-base
check:
curl --silent https://github.com/input-output-hk/cardano-base/branch_commits/0e26a4f00324100b4c34f2a9c7606ae03e94d7a1
```

# Implementation
- a _mish-mash-of-bash_ is not the prettiest way to do it, can certainly be improved -- but it's a start

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -25,3 +25,8 @@ steps:
     command: 'nix-shell scripts/buildkite/stack-cabal-sync.nix --run scripts/buildkite/stack-cabal-sync.sh'
     agents:
       system: x86_64-linux
+
+  - label: 'dependencies-at-master'
+    command: 'ci/check-dependencies-merged-to-master.sh'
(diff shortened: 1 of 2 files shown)


erikd:
I think we should merge something like this, make sure it works and then move it to iohk-nix whne we are happy with it.

Issue #96, opened 2019-08-06
Move configuration to cardano-node


ksaric:
Based on https://github.com/input-output-hk/cardano-shell/issues/238, this issue is all about moving the configuration-specific parts of cardano-shell into the node.

PR #97, opened 2019-08-06
[GH-96] Move the configuration into cardano-node.

ksaric:
https://github.com/input-output-hk/cardano-node/issues/96

The first step to moving the configuration from `cardano-shell` to `cardano-node`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,30 @@
+PROJECT_NAME = cardano-node
+NUM_PROC     = $(nproc --all)
+
+
+help: ## Print documentation
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
+
+stylish-haskell: ## Apply stylish-haskell on all *.hs files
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -0,0 +1,30 @@
...
+test-ghci: ## Run repl on test suites
+	@stack ghci $(PROJECT_NAME):lib $(PROJECT_NAME):test:$(PROJECT_NAME)-test --ghci-options=-fobject-code --nix
+
+test-ghcid: ## Run ghcid on test suites
+	@ghcid --command "stack ghci --nix $(PROJECT_NAME):lib $(PROJECT_NAME):test:$(PROJECT_NAME)-test --ghci-options=-fobject-code"
+
+test-ghcid-nix: ## Run ghcid on test suites with Nix
+	@ghcid --command="stack ghci --test --main-is $(PROJECT_NAME):test:$(PROJECT_NAME)-test --nix -j$(NUM_PROC)"
+
+.PHONY: stylish-haskell ghcid ghcid-test run-test test-ghci test-ghcid help
```
deepfire: I love the Makefile!

It always struck me as a particularly efficient way of providing canned scenarios in a repository!

ksaric: This is from @HirotoShioi.

deepfire:
This PR will help a lot to reduce friction between repositories!

One thing we should benefit from is changing the types for the parsers to be specific -- capitalising on https://github.com/input-output-hk/cardano-node/blob/master/cardano-node/Cardano/Node/CLI.hs -- but this is clearly out of the scope of this PR.

PR #100, opened 2019-08-07
Fix cabal new-build and stack build

denisshevchenko:
Now it can compiles with `cabal new-build`.

### DIFF
--- a/app/TxSubmission.hs
+++ b/app/TxSubmission.hs
@@ -146,6 +146,7 @@ submitTx :: ( RunDemo blk
          -> IO ()
 submitTx pInfoConfig nodeId tx tracer =
     connectTo
+      nullTracer
       (,)
       (localInitiatorNetworkApplication tracer pInfoConfig tx)
       Nothing
--- a/cardano-node.cabal
+++ b/cardano-node.cabal
@@ -24,7 +24,7 @@ library
                        base >=4.12 && <5
                      , bytestring
                      , canonical-json
-                     , cborg
+                     , cborg == 0.2.2.0
                      , formatting
                      , mtl
                      , optparse-applicative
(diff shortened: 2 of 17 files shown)


app/TxSubmission.hs
```diff
@@ -15,13 +15,13 @@ module TxSubmission (
...
-import           Data.Void (Void)
 import           Data.ByteString.Lazy (ByteString)
+import           Data.Proxy
 import qualified Data.Set as Set
+import           Data.Void (Void)
 import           Options.Applicative
-import           Data.Proxy
 
-import qualified Codec.Serialise as Serialise (encode, decode)
+import qualified Codec.Serialise as Serialise (decode, encode)
```
cleverca22: order was swapped

app/TxSubmission.hs
```diff
@@ -15,13 +15,13 @@ module TxSubmission (
     , localSocketAddrInfo
     ) where
 
-import           Data.Void (Void)
 import           Data.ByteString.Lazy (ByteString)
+import           Data.Proxy
 import qualified Data.Set as Set
+import           Data.Void (Void)
```
cleverca22: void was moved

app/TxSubmission.hs
```diff
@@ -15,13 +15,13 @@ module TxSubmission (
...
 import           Options.Applicative
-import           Data.Proxy
```
cleverca22: proxy was moved

cardano-node.cabal
```diff
@@ -24,7 +24,7 @@ library
                        base >=4.12 && <5
                      , bytestring
                      , canonical-json
-                     , cborg
+                     , cborg == 0.2.2.0
```
mrBliss: Do we want `==` or `>= && <`?

PR #104, opened 2019-08-08
Update after ouroboros-network#903

mrBliss:
The point of #903 is to trace the progress of the ledger initialisation. Some pointers on how to do this:
* Note the new line: `LedgerDB.ReplayedBlock {} -> ignore`
* See the new `ChainDB.TraceEvent.TraceLedgerReplayEvent` constructor: https://github.com/input-output-hk/ouroboros-network/pull/903/files#diff-f1834e0128e7b50336516db371f62579R302
* See https://github.com/input-output-hk/ouroboros-network/pull/903/files#diff-1b6e41536e4311c26facb6a658fdb160R388
* The above `TraceReplayEvent` is decorated with more info here: https://github.com/input-output-hk/ouroboros-network/pull/903/files#diff-da0a5a3d801a2f66b948bf6b372a6d82R272
  This includes:
  - The point and epoch of the replayed block
  - The slot of the first block in the epoch of the replayed block. If the block's `pointSlot` equals this slot, it is the first block in the epoch (note that the EBB has the same slot as the first block in the epoch)
  - The point and epoch of the last block that has to be replayed (the tip of the ImmutableDB).
  Using this information, it should be possible to trace  user-friendly progress of the ledger initialisation process. Note that showing progress per epoch might be too slow, progress per 500 blocks/slots or so seems reasonable.

### DIFF
--- a/app/Run.hs
+++ b/app/Run.hs
@@ -85,6 +85,7 @@ import           Ouroboros.Consensus.Util.ThreadRegistry
 
 import           Ouroboros.Storage.ChainDB (ChainDB)
 import qualified Ouroboros.Storage.ChainDB as ChainDB
+import           Ouroboros.Storage.EpochInfo (EpochInfo, newEpochInfo)
 import           Ouroboros.Storage.ImmutableDB (ValidationPolicy (..))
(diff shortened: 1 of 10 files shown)


mrBliss:
Can we merge this so that we can merge #105 afterwards?

PR #116, opened 2019-08-14
nix scripts for cardano-node

disassembler:


nix/cardano-node-scripts.nix
```diff
@@ -0,0 +1,66 @@
...
+      hostAddr = "127.0.0.1";
+      port = 3001;
+      signingKey = null;
+      delegationCert = null;
+      loggingConfig = ../configuration/log-configuration.yaml;
+      pbftThreshold = null;
+      nodeId = 0;
+    } // envConfig
+      // customConfig;
+      topologyFile = if (config ? topologyFile) then topologyFile else mkEdgeTopology {
```
johnalotoski: Indenting on L41-46 and empty on L45.

default.nix
```diff
@@ -1,4 +1,10 @@
-{ customConfig ? {}, ... }:
+let
+  lib = (import ./lib.nix).pkgs.lib;
+  commitIdFromGitRepo = import ./nix/commit-id.nix { inherit lib; };
```
johnalotoski: ./nix/commit-id.nix does not exist?

jbgi:
Merged with #111

Issue #115, opened 2019-08-14
Introduce conditional tracing and tracing based on verbosity level


CodiePP:
based on new version of iohk-monitoring-framework
(epic: https://github.com/input-output-hk/iohk-monitoring-framework/issues/277)

control verbosity
--------------------

`--tracing-verbosity-minimal`
`--tracing-verbosity-normal`  (the default)
`--tracing-verbosity-maximal`

switch textual representation / structured logging
-------------------------------------------------------------
by default, structured logging is turned on: the rendering of the captured observables is controlled by their `ToObject` instances.

`--trace-chain-db`    turns on textual representation in the ChainDB tracer

PR #121, opened 2019-08-15
[#115] added cli args for indicating tracing verbosity

CodiePP:
* updated iohk-monitoring to new version
* added cli args for indicating tracing verbosity
* `toLogObject` -> `toLogObject' TracingVerbosity`

### DIFF
--- a/cardano-node/src/Cardano/Node/Run.hs
+++ b/cardano-node/src/Cardano/Node/Run.hs
@@ -136,10 +136,10 @@ runNode nodeCli@NodeCLIArguments{..} loggingLayer cc = do
         let tracer      = contramap pack $ toLogObject trace'
 
         traceWith tracer $ "tracing verbosity = " ++
-                             (case (traceVerbosity traceOptions) of
+                             case traceVerbosity traceOptions of
                                  NormalVerbosity -> "normal"
                                  MinimalVerbosity -> "minimal"
-                                 MaximalVerbosity -> "maximal")
+                                 MaximalVerbosity -> "maximal"
         SomeProtocol p  <- fromProtocol cc protocol
         let tracers     = mkTracers traceOptions trace'
         case viewMode of
--- a/cardano-node/src/Cardano/Node/Run.hs
+++ b/cardano-node/src/Cardano/Node/Run.hs
@@ -18,7 +18,7 @@
 module Cardano.Node.Run (
       runNode
     ) where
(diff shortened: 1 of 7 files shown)


cardano-node/Cardano/Node/Run.hs
```diff
@@ -18,15 +18,14 @@
 module Cardano.Node.Run (
       runNode
     ) where
-import           Cardano.Prelude hiding (ByteString, atomically, throwIO, wait)
+import           Cardano.Prelude hiding (ByteString, atomically, throwIO, trace, wait)
```
Jimbo4350: Does `trace` need to be hidden here? I'm not seeing it used in the module.

intricate: I see a parameter name being specified [here](https://github.com/input-output-hk/cardano-node/blob/5dd7682d91612bde0f8538506749c7663c5bebcf/cardano-node/Cardano/Node/Run.hs#L183), called `trace`. Perhaps `trace` was hidden in order to prevent name shadowing.

CodiePP: an argument is also named "trace" so there was a warning.

CodiePP: we should turn on `-Werror` once.

intricate: @CodiePP: I agree on the note of `-Werror`. I quite like how it's done in `cardano-ledger`, where there's a `development` cabal flag for enabling/disabling the usage of `-Werror`:
```
flag development
    description: Disable `-Werror`
    default: False
    manual: True

...

  if (!flag(development))
    ghc-options:         -Werror

```

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -134,7 +133,13 @@ runNode nodeCli@NodeCLIArguments{..} loggingLayer cc = do
 
       SimpleNode topology myNodeAddress protocol viewMode traceOptions -> do
         let trace'      = appendName (pack $ show $ node topology) tr
+        let tracer      = contramap pack $ toLogObject trace'
 
+        traceWith tracer $ "tracing verbosity = " ++
+                             (case (traceVerbosity traceOptions) of
```
mrBliss: I think you can remove both pairs of parentheses.

CodiePP: cleaned up

PR #122, opened 2019-08-15
[#115] structured logging part 1

CodiePP:
adapted ChainDBTracer

### DIFF
--- a/cardano-node/src/Cardano/Node/ToObjectOrphans.hs
+++ b/cardano-node/src/Cardano/Node/ToObjectOrphans.hs
@@ -13,7 +13,7 @@ module Cardano.Node.ToObjectOrphans
   ) where
 
 import           Cardano.Prelude hiding (atomically, show)
-import           Prelude (String, show)
+import           Prelude (String, show, id)
 
 import           Data.Aeson (Value (..), toJSON, (.=))
 import           Data.Text (Text, pack)
@@ -47,14 +47,22 @@ showWithTip :: Condense (HeaderHash blk)
             => (a -> String)
             -> WithTip blk a
             -> String
-showWithTip customShow (WithTip tip a) = "[" ++ (showTip tip) ++ "] " ++ customShow a
+showWithTip customShow (WithTip tip a) = "[" ++ (showTip MinimalVerbosity tip) ++ "] " ++ customShow a
 
 showTip :: Condense (HeaderHash blk)
-        => Point blk
+        => TracingVerbosity
+        -> Point blk
         -> String
-showTip tip = case pointHash tip of
-    GenesisHash -> "genesis"
-    BlockHash h -> take 7 (condense h)
+showTip verb tip =
+    case pointHash tip of
+        GenesisHash -> "genesis"
+        BlockHash h -> trim $ condense h
+  where
+    trim :: [a] -> [a]
+    trim = case verb of
+        MinimalVerbosity -> take 7
+        NormalVerbosity  -> take 7
+        MaximalVerbosity -> id
 
 instance ( Show a
          , Condense (HeaderHash blk)
@@ -64,13 +72,15 @@ instance ( Show a
 
 
 
--- | instances of ToObject
+-- | instances of @ToObject@
 
 instance (Condense (HeaderHash blk), ProtocolLedgerView blk)
             => Transformable Text IO (WithTip blk (ChainDB.TraceEvent blk)) where
     trTransformer = trStructured -- structure required, will call 'toObject'
+
 instance (Condense (HeaderHash blk), ProtocolLedgerView blk)
             => ToObject (WithTip blk (ChainDB.TraceEvent blk)) where
+    -- example: turn off any tracing of @TraceEvent@s when minimal verbosity level is set
     -- toObject MinimalVerbosity _ = emptyObject -- no output
     toObject verb (WithTip tip ev) =
         let evobj = toObject verb ev
@@ -78,7 +88,7 @@ instance (Condense (HeaderHash blk), ProtocolLedgerView blk)
         if evobj == emptyObject
         then emptyObject
         else mkObject [ "kind" .= String "TraceEvent"
-                      , "tip" .= showTip tip
+                      , "tip" .= showTip MinimalVerbosity tip
                       , "event" .= evobj
                       ]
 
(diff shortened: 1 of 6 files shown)


cardano-node/src/Cardano/Node/Tracers.hs
```diff
@@ -132,6 +80,104 @@ data Tracers peer blk = Tracers {
...
+                                    ChainDB.TryAddToCurrentChain pt ->
+                                      "Block fits onto the current chain: " <> condense pt
+                                    ChainDB.TrySwitchToAFork pt _   ->
+                                      "Block fits onto some fork: " <> condense pt
+                                    ChainDB.SwitchedToChain _ c     ->
+                                      "Chain changed, new tip: " <> condense (AF.headPoint c)
+                                    ChainDB.AddBlockValidation ev'  -> case ev' of
+                                      ChainDB.InvalidBlock err pt
+                                        -> "Invalid block " <> condense pt <> ": " <> show err
+                                      _ -> "ignore"
```
mrBliss: I don't think we want to be tracing `"ignore"` here

cardano-node/src/Cardano/Node/Tracers.hs
```diff
@@ -132,6 +80,104 @@ data Tracers peer blk = Tracers {
...
+            => ToObject (WithTip blk (ChainDB.TraceEvent blk)) where
+    toObject MinimalVerbosity _ = emptyObject -- no output
+    toObject verb (WithTip tip ev) =
+        mkObject [ "tip" .= showTip tip
+                 , "TraceEvent" .= toObject verb ev
+                 ]
+instance (Condense (HeaderHash blk), ProtocolLedgerView blk)
+            => ToObject (ChainDB.TraceEvent blk) where
+    toObject MinimalVerbosity _ = emptyObject -- no output
+    toObject _ (ChainDB.TraceAddBlockEvent ev) =
```
mrBliss: A sketch of less verbose (pun intended) alternative approach:
```haskell
    toObject verbosity ev = case verbosity of
        MinimalVerbosity -> emptyObject -- no output
        MaximalVerbosity -> mkObj (constrName ev) (show ev) -- NEW: useful for debugging
        NormalVerbosity  -> case ev of
          ChainDB.TraceAddBlockEvent ev' -> mkObj "TraceAddBlockEvent" $ \case ev' of
            ChainDB.StoreButDontChange   pt ->
              "Ignoring block: " <> condense pt
            ChainDB.TryAddToCurrentChain pt ->
              "Block fits onto the current chain: " <> condense pt
        ..
      where
        mkObj kind event = mkObject
          [ "kind"  .= String kind
          , "event" .= String $ pack event
          ]
```

CodiePP: 👍 Thanks! I will revise.
now it looks like code. before, it was just a verbatim copy of the existing case-split.

cardano-node/src/Cardano/Node/Tracers.hs
CodiePP: yes, these should become `emptyObject` as you proposed below.

CodiePP:
bors try

cardano-node/src/Cardano/Node/Tracers.hs
```diff
@@ -217,16 +156,13 @@ mkTracers traceOptions tracer = Tracers
...
-              traceNamedObject tr (meta, logValue')
-          TraceMempoolRejectedTxs txs _ ->
-              let logValue' :: LOContent a
-                  logValue' = LogValue "txsProcessed" $ PureI $ fromIntegral $ length txs in
-              traceNamedObject tr (meta, logValue')
-          _                             -> return ()
+        let txs = case mempoolEvent of
+                  TraceMempoolAddTxs      txs0 _ -> txs0
+                  TraceMempoolRejectedTxs txs0 _ -> txs0
+                  _                             -> []
```
mrBliss: My OCD would be happier if you align the last arrow with the two above :slightly_smiling_face:

scripts/trace-acceptor.sh
```diff
@@ -2,9 +2,9 @@
```
mrBliss: I don't think changes to this file need to be included in this PR.

cardano-node/src/Cardano/Node/ToObjectOrphans.hs
```diff
@@ -0,0 +1,188 @@
...
+                MaximalVerbosity -> 99
+    in
+    case pointHash tip of
+        GenesisHash -> "genesis"
+        BlockHash h -> pack $ take n (condense h)
+
+instance (Condense (HeaderHash blk), ProtocolLedgerView blk)
+            => ToObject (ChainDB.TraceEvent blk) where
+
+    toObject verb (ChainDB.TraceAddBlockEvent ev) =
```
mrBliss: If you want to keep your lines shorter:
```haskell
toObject verb (ChainDB.TraceAddBlockEvent ev) = case ev of
  ChainDB.StoreButDontChange pt -> mkObject
      [ "kind"  .= String "TraceAddBlockEvent.StoreButDontChange"
      , "block" .= toObject verb pt ]
  ..
```

cardano-node/src/Cardano/Node/ToObjectOrphans.hs
```diff
@@ -0,0 +1,188 @@
...
+    toObject verb p =
+        mkObject [ "kind" .= String "Tip"
+                , "tip" .= String (showTip' verb p) ]
+showTip' :: (Condense (HeaderHash blk))
+        => TracingVerbosity -> Point blk -> Text
+showTip' verb tip =
+    let n :: Int = case verb of
+                MinimalVerbosity -> 7
+                NormalVerbosity -> 7
+                MaximalVerbosity -> 99
```
mrBliss: Instead of using 99 as a proxy for "the whole hash" (what if the hash length is 128? I also don't like the idea of using `maxBound`), I would do the following:
```haskell
showTip' :: Condense (HeaderHash blk)
         => TracingVerbosity -> Point blk -> Text
showTip' verb tip = case pointHash tip of
    GenesisHash -> "genesis"
    BlockHash h -> pack $ trim $ condense h
  where
    trim = case verb of
      MinimalVerbosity -> take 7
      NormalVerbosity  -> take 7
      MaximalVerbosity -> id
```

cardano-node/src/Cardano/Node/ToObjectOrphans.hs
```diff
@@ -0,0 +1,188 @@
...
+    toObject _verb slot =
+        mkObject [ "kind" .= String "SlotNo"
+                 , "slot" .= toJSON (unSlotNo slot) ]
+
+instance (Condense (HeaderHash blk), ProtocolLedgerView blk)
+    => ToObject (Point blk) where
+    toObject MinimalVerbosity p = toObject NormalVerbosity p
+    toObject verb p =
+        mkObject [ "kind" .= String "Tip"
+                , "tip" .= String (showTip' verb p) ]
```
mrBliss: Alignment :slightly_smiling_face:

cardano-node/src/Cardano/Node/ToObjectOrphans.hs
```diff
@@ -0,0 +1,188 @@
...
+-- | instances of ToObject
+
+instance (Condense (HeaderHash blk), ProtocolLedgerView blk)
+            => Transformable Text IO (WithTip blk (ChainDB.TraceEvent blk)) where
+    trTransformer = trStructured -- structure required, will call 'toObject'
+instance (Condense (HeaderHash blk), ProtocolLedgerView blk)
+            => ToObject (WithTip blk (ChainDB.TraceEvent blk)) where
+    -- toObject MinimalVerbosity _ = emptyObject -- no output
+    toObject verb (WithTip tip ev) =
+        let evobj = toObject verb ev
```
mrBliss: This is a very curious indentation style ("align on keyword"?) :slightly_smiling_face:

cardano-node/src/Cardano/Node/ToObjectOrphans.hs
```diff
@@ -0,0 +1,188 @@
...
+        mkObject [ "kind" .= String "SlotNo"
+                 , "slot" .= toJSON (unSlotNo slot) ]
+
+instance (Condense (HeaderHash blk), ProtocolLedgerView blk)
+    => ToObject (Point blk) where
+    toObject MinimalVerbosity p = toObject NormalVerbosity p
+    toObject verb p =
+        mkObject [ "kind" .= String "Tip"
+                , "tip" .= String (showTip' verb p) ]
+showTip' :: (Condense (HeaderHash blk))
```
mrBliss: It would be nicer to unify this function with `showTip`.

I think you could remove `showTip`, rename `showTip'` to `showTip`, and replace all calls to (the old) `showTip` with `showTip MinimalVerbosity`.

cardano-node/src/Cardano/Node/ToObjectOrphans.hs
CodiePP: yes, looks like BASIC! :-)

cardano-node/src/Cardano/Node/ToObjectOrphans.hs
CodiePP: :+1:

cardano-node/src/Cardano/Node/ToObjectOrphans.hs
CodiePP: :+1:

scripts/trace-acceptor.sh
CodiePP: it is just to make sure this and `shelley-testnet.sh` are using both `cabal`.
makes it much easier to launch the demo.

cardano-node/src/Cardano/Node/ToObjectOrphans.hs
```diff
@@ -0,0 +1,188 @@
...
+
+
+-- | instances of ToObject
+
+instance (Condense (HeaderHash blk), ProtocolLedgerView blk)
+            => Transformable Text IO (WithTip blk (ChainDB.TraceEvent blk)) where
+    trTransformer = trStructured -- structure required, will call 'toObject'
+instance (Condense (HeaderHash blk), ProtocolLedgerView blk)
+            => ToObject (WithTip blk (ChainDB.TraceEvent blk)) where
+    -- toObject MinimalVerbosity _ = emptyObject -- no output
```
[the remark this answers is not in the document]
CodiePP: > Looking good, just some misc comments.
> 
> It would be great if you pasted the trace output for all three verbosity levels. Maybe we spot things that are too verbose for the given verbosity level, etc.

yes, let's do this in a second PR. This one should be integrated when ready, to lower the burden of rebasing with master (other changes coming in).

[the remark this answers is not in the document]
CodiePP: this serves as an example of how to turn off tracing.

scripts/trace-acceptor.sh
CodiePP: restored

cardano-node/src/Cardano/Node/ToObjectOrphans.hs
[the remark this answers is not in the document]
mrBliss: > > Looking good, just some misc comments.
> > It would be great if you pasted the trace output for all three verbosity levels. Maybe we spot things that are too verbose for the given verbosity level, etc.
> 
> yes, let's do this in a second PR. This one should be integrated when ready, to lower the burden of rebasing with master (other changes coming in).

No, I mean: can you paste the output (for all three verbosities) that the node produces running *this PR*? (before we merge it) Otherwise, we're blindly merging something that might trace something totally  wrong.

I don't mind fine-tuning the trace output in follow-up PRs.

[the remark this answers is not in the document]
mrBliss: > this serves as an example of how to turn off tracing.

Ah, in that case a comment mentioning this would be helpful, otherwise other people will also suspect that it is a stray comment.

cardano-node/src/Cardano/Node/ToObjectOrphans.hs
```diff
@@ -0,0 +1,181 @@
...
+                     , "immtip" .= toObject verb immTip
+                     , "tip" .= toObject verb tip' ]
+        _ -> emptyObject
+
+    toObject _ _ = emptyObject -- no output
+
+instance ToObject LedgerDB.DiskSnapshot where
+    toObject MinimalVerbosity snap = toObject NormalVerbosity snap
+    toObject NormalVerbosity _ = mkObject [ "kind" .= String "snapshot" ]
+    toObject MaximalVerbosity snap =
```
mrBliss: Is this the only place where `MaximalVerbosity` does something different than `NormalVerbosity`?

I was expecting that `MaximalVerbosity` would just trace *all* trace events. This would make debugging much easier: just enable enable maximal verbosity with a flag, instead of modifying the code (to use, e.g., `showTracing debugTracer`), and recompiling. Or is this not the purpose of `MaximalVerbosity`?

cardano-node/src/Cardano/Node/ToObjectOrphans.hs
```diff
@@ -0,0 +1,181 @@
...
+-- STM transaction we risk reporting  wrong tip.
+--
+data WithTip blk a =
+    WithTip
+      (Point blk)
+      -- ^ current tip point
+      a
+      -- ^ data
+
+showWithTip :: Condense (HeaderHash blk)
```
mrBliss: Ah, if you don't want the log output to change when we change our `Condense` instance for `HeaderHash blk` (which, however unlikely, might happen), then you should roll your own version of it (`pprintPoint`?)

cardano-node/src/Cardano/Node/ToObjectOrphans.hs
CodiePP: done

cardano-node/src/Cardano/Node/ToObjectOrphans.hs
CodiePP: yes, probably. This needs to be discussed, then can be changed.

cardano-node/src/Cardano/Node/ToObjectOrphans.hs
CodiePP: removed the `Condense` constraint
what do you think?

Issue #84, opened 2019-07-29
Clean up


mrBliss:
I think this repo is in dire need for some clean up. We have been focussing too much on adding as much as we can in a short time, without thinking about the long term.

Examples:
* Most code lives in the executable folder `app/` instead of the library folder `cardano-node`. IMO, `app/` should just contain a single file `Main.hs` containing:
  ```haskell
  module Main (main) where
  import  Cardano.Node.Main (main)
  ```
  The [`genesis-tool`](https://github.com/input-output-hk/cardano-node/tree/master/app/genesis-tool) even lives inside this directory! I expect this is because it depends on the code in the `app` directory. It should live in a separate directory containing a similar `Main.hs` file (or only containing minimal code).
* We should extract code in helper functions and/or separate modules, e.g., https://github.com/input-output-hk/cardano-node/blob/e42d7396f77e24b03379bc61f668ab526a0e062d/app/Run.hs#L258-L416, this is just making this file way too long and unreadable (I'm guilty of these tings myself).
* In the `scripts` folder, there are too many demo scripts, some of which are not working anymore. The numbering might also make people think they match the demos that we gave, which is untrue.

Also, synching the different dependencies (`cardano-ledger`, `ouroboros-network`, ..) has been an endless time sink so far: to update the dependency on A, we often need to first get a PR updating the dependency on A merged in B before we can do it in `cardano-ledger` (we can't submit two PRs at once, because the hashes will change and have to be updated after the merge). These PRs quickly go stale or are stalled until someone that made some breaking change can help fix it.

christophergreenwood:
@nc6 @dcoutts @coot @avieth @deepfire Could you all please share the effort for this

intricate:
Related: #120

PR #123, opened 2019-08-16
Add `--database-path` parameter to `CommonCLI` parser

Jimbo4350:
There is also some further restructuring on this branch: #84 

 ~I have placed comments next to changes pertaining to the addition of the `--database-path` parameter~

The PR is split into two commits as per @mrBliss 's suggestion.


cardano-node/src/Cardano/Node/ConfigCLI.hs
```diff
@@ -73,250 +70,13 @@ data TraceOptions = TraceOptions
...
-  <*> parseConsensusTraceOptions
-  <*> parseProtocolTraceOptions
-  <*> parseTraceIpSubscription
-  <*> parseTraceDnsSubscription
-  <*> parseTraceDnsResolver
-
-data Command =
-    SimpleNode  TopologyInfo NodeAddress Protocol ViewMode TraceOptions
+data NodeCommand =
+    SimpleNode  TopologyInfo NodeAddress Protocol ViewMode TraceOptions FilePath
```
Jimbo4350: Addition of database `FilePath` to the `NodeCommand`

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -143,7 +143,7 @@ runNode nodeCli@NodeCLIArguments{..} loggingLayer cc = do
         SomeProtocol p  <- fromProtocol cc protocol
         let tracers     = mkTracers traceOptions trace'
         case viewMode of
-          SimpleView -> handleSimpleNode p nodeCli myNodeAddress topology trace' tracers cc
+          SimpleView -> handleSimpleNode p nodeCli myNodeAddress topology trace' tracers (cc { ccDBPath = dbFp })
```
Jimbo4350: Insertion of the db filepath into `CardanoConfiguration`

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -236,6 +238,7 @@ handleSimpleNode p NodeCLIArguments{..}
                           (withTip varTip $ chainDBTracer nodeTraces)
                           slotDuration
                           epochInfo
+                          (path <> "-" <> show nid )
```
Jimbo4350: db filepath added as argument to `ChainDb.defaultArgs`

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -228,6 +228,8 @@ handleSimpleNode p NodeCLIArguments{..}
 
       varTip <- atomically $ newTVar GenesisPoint
       epochInfo <- newEpochInfo $ nodeEpochSize (Proxy @blk)
+      -- Database filepath
+      path <- makeAbsolute ccDBPath
```
mrBliss: I think that `makeAbsolute` is not enough to resolve any `..` inside the path, you'll need `canonicalizePath`:

```
$ pwd
/home/thomas
$ ghci
> import System.Directory
> makeAbsolute "../../home/thomas"
"/home/thomas/../../home/thomas"
> canonicalizePath =<< makeAbsolute "../../home/thomas"
"/home/thomas"
```

cardano-node/src/Cardano/Node/Run.hs
mrBliss: Space before the `)`.

cardano-node/src/Cardano/Node/Run.hs
deepfire: Hmm, I'm not sure it's the best place for the configuration composition.  But it's hard to tell, with all the code movement.

deepfire: Ideally, the `cc` -- `CardanoConfiguration` -- should have the CLI args transparently `<>`'ed in the Partial Options Monoid style, in a single place -- so that we don't have to do one-off splicings like here.

PR #124, opened 2019-08-16
Clean up: Part 1

Jimbo4350:
Some further removal and condensation of modules: #84


cardano-node/src/Cardano/Node/Configuration/CLI.hs
```diff
@@ -1,433 +0,0 @@
...
-        <*> lastStrOption
-           ( long "th-period"
-          <> metavar "TH-PERIOD"
-          <> help "Throttle period."
-           )
-        <*> lastIntOption
-           ( long "th-burst"
-          <> metavar "TH-BURST"
-          <> help "Throttle burst."
-           )
```
mhuesch: Where did the contents of this file go? I'm not seeing them relocated anywhere.

Jimbo4350: The exported functions weren't being used anywhere. I need @deepfire to confirm if this is dead code.

mhuesch: Ah ok, nice!

PR #125, opened 2019-08-19
Add `LICENSE` and `NOTICE` files to root.

Jimbo4350:
Following  input-output-hk/plutus#635:

- Add `LICENSE` and `NOTICE` in the root.
    - This gets picked up by Github and covers things which aren't
      licensed separately (like docs).
- Remove Cabal `copyright` field.
    - There is a copyright notice in `NOTICE` which this would conflict
      with.


cardano-node/cardano-node.cabal
```diff
@@ -7,7 +7,6 @@ license:        Apache-2.0
 license-files:
   LICENSE
   NOTICE
-copyright:      2019 Input Output (Hong Kong) Ltd.
```
intricate: In some of our other repos, we appear to keep this line. Is it actually an issue for it to remain?

Jimbo4350: Apparently: 

input-output-hk/plutus#635

> Remove Cabal copyright field.
> There is a copyright notice in NOTICE which this would conflict
> with.

PR #128, opened 2019-08-20
Simplify node start-up

mrBliss:
Following https://github.com/input-output-hk/ouroboros-network/pull/938

### DIFF
--- a/cardano-node/src/Cardano/Node/Run.hs
+++ b/cardano-node/src/Cardano/Node/Run.hs
@@ -18,79 +18,51 @@
 module Cardano.Node.Run (
       runNode
     ) where
-import           Cardano.Prelude hiding (ByteString, atomically, throwIO, trace, wait)
-import           Prelude (id, error)
+import           Cardano.Prelude hiding (ByteString, atomically, throwIO, trace,
+                                  wait)
+import           Prelude (error, id, unlines)
 
-import           Codec.CBOR.Decoding (Decoder)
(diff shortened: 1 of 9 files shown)


stack.yaml
```diff
@@ -57,7 +57,8 @@ extra-deps:
     #Ouroboros-network dependencies
 
   - git: https://github.com/input-output-hk/ouroboros-network
-    commit: d69ef321766b33a8d1152b7bfa039717b543f162
+    # TODO this commit points to a branch, never merge it!
```
mrBliss: This is the only blocker (and updating the cabal and nix files)

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -180,313 +152,99 @@ handleSimpleNode p NodeCLIArguments{..}
...
 
+    removeStaleLocalSocket myLocalSockPath
+
+    dbPath <- canonicalizePath =<< makeAbsolute ccDBPath
+
+    varTip <- atomically $ newTVar GenesisPoint
+
+    Node.run
+      (consensusTracers nodeTracers)
+      -- (withTip varTip $ chainDBTracer nodeTracers)
```
Jimbo4350: Left over comment?

mrBliss: Good catch! The line below was tracing using `Show`, which we certainly wouldn't want.

PR #135, opened 2019-08-21
cardano-cli refactoring & submit-tx subcommand

deepfire:
0. New `submit-tx` subcommand. 
0. Drop `SystemVersion` in favor of `Protocol`.
0. Move some command parsers to `Cardano.Node.Parsers`.
0. Rename and segregate `CLIOps` (ex-`KeyMaterialOps`)
0. Add `ByronLegacy` to the `Protocol` type
0. Remove `submit` subcommand from `cardano-node`
0. Factor orphans.
0. Factor `mkConfiguration`
0. Eliminate all warnings.
0. Documentation & updated the `scripts/submit-tx.sh` script.

# TODO

- [ ] Separate the first megacommit ?
- [ ] Testing -- this requires a way to create transactions. WIP

### DIFF
--- a/cardano-node/app/cardano-cli.hs
+++ b/cardano-node/app/cardano-cli.hs
@@ -2,19 +2,26 @@
 module Main (main) where
 
 import           Cardano.Prelude hiding (option)
-import           Prelude (String)
+import           Prelude (String, error, id)
+
+import           Options.Applicative
+
+import           Control.Exception.Safe (catchIO)
+import           Data.Time.Clock.POSIX (posixSecondsToUTCTime)
+import           Data.Time (UTCTime)
+import           System.Exit (ExitCode(..), exitWith)
 
 import           Cardano.Binary (Annotated (..))
 import           Cardano.Chain.Common
-import qualified Cardano.Chain.Genesis as Genesis
-import           Cardano.Chain.Slotting (EpochNumber(..))
-import           Cardano.CLI.Run (ClientCommand(..), SystemVersion(..), decideKeyMaterialOps, runCommand)
-import           Cardano.Common.CommonCLI (command')
+import           Cardano.Chain.Genesis
+import           Cardano.Chain.Slotting
 import           Cardano.Crypto (AProtocolMagic(..), ProtocolMagic, ProtocolMagicId(..), RequiresNetworkMagic(..))
 
-import           Options.Applicative
-import           Control.Exception.Safe (catchIO)
-import           System.Exit (ExitCode(..), exitWith)
+import           Cardano.Common.CommonCLI
+import           Cardano.Common.Protocol
+import           Cardano.Node.Parsers
+import           Cardano.CLI.Ops (decideCLIOps)
+import           Cardano.CLI.Run (ClientCommand(..), runCommand)
 
 main :: IO ()
 main = do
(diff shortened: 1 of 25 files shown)


mrBliss:
Just curious (I haven't read through the whole PR yet), what's `ByronLegacy` going to be?

deepfire:
@mrBliss, the `cardano-cli` has to deal with legacy Byron formats, so we need to represent those somehow.  We used to have the separate `SystemVersion` type for this in `cardano-cli`, but now that `cardano-cli` is using more of the shared code, it makes less sense to maintain the extra type.

cardano-node/src/Cardano/Node/TxSubmission.hs
```diff
@@ -93,11 +93,9 @@ handleTxSubmission ptcl tinfo mocktx tracer = do
                        (CoreNodeId nid)
                        ptcl
 
-        tx :: GenTx blk
-        tx = demoMockTx pInfoConfig mocktx
-
-    traceWith tracer $
-      "The Id for this transaction is: " <> condense (H.hash @ShortHash mocktx)
+    -- TODO:  generalised Tx hash
```
Jimbo4350: Left over comment?

cardano-node/app/cardano-cli.hs
```diff
@@ -1,208 +1,39 @@
...
 
-import           Cardano.Binary (Annotated (..))
-import           Cardano.Chain.Common
-import qualified Cardano.Chain.Genesis as Genesis
-import           Cardano.Chain.Slotting (EpochNumber(..))
-import           Cardano.CLI.Run (ClientCommand(..), SystemVersion(..), decideKeyMaterialOps, runCommand)
-import           Cardano.Crypto (AProtocolMagic(..), ProtocolMagic, ProtocolMagicId(..), RequiresNetworkMagic(..))
-import           Cardano.Node.CLI (command')
+import           Cardano.CLI.Ops (decideCLIOps)
+import           Cardano.CLI.Run (ClientCommand(..), runCommand, parseClientCommand)
```
Jimbo4350: `parseClientCommand` belongs with the other parsers used only in `cardano-cli.hs`. It's less overhead when the parsers are just below where they are called. If the parser is used in multiple places then it can go into `Cardano.Node.Parsers`.

cardano-node/app/cardano-cli.hs
```diff
@@ -1,208 +1,39 @@
...
-      <*> parseFilePath    "certificate"              "Non-existent file to write the certificate to."
-    , command' "check-delegation"               "Verify that a given certificate constitutes a valid delegation relationship betwen keys." $
-      CheckDelegation
-      <$> parseProtocolMagicId "protocol-magic"
-      <*> parseFilePath    "certificate"              "The certificate embodying delegation to verify."
-      <*> parseFilePath    "issuer-key"               "The genesis key that supposedly delegates."
-      <*> parseFilePath    "delegate-key"             "The operation verification key supposedly delegated to."
-    ])
-
-parseTestnetBalanceOptions :: Parser Genesis.TestnetBalanceOptions
```
Jimbo4350: See comment about `parseClientCommand`. These should stay here.

cardano-node/app/cardano-node.hs
```diff
@@ -184,68 +177,15 @@ parseNodeCommand :: Parser NodeCommand
...
-  , command' "submit" "Submit a transaction." $
-      TxSubmitter <$> parseTopologyInfo <*> parseMockTx <*> parseProtocol
   , command' "trace-acceptor" "Spawn an acceptor." $
       pure TraceAcceptor
   ]
 
-parseTopologyInfo :: Parser TopologyInfo
-parseTopologyInfo = TopologyInfo <$> parseNodeId <*> parseTopologyFile
-
-parseNodeId :: Parser NodeId
```
Jimbo4350: See comment about `parseClientCommand`. These should stay here.

cardano-node/src/Cardano/Node/CLI.hs
```diff
@@ -4,29 +4,26 @@
...
 
 {-# OPTIONS_GHC -Wno-all-missed-specialisations #-}
 
 module Cardano.Node.CLI (
   -- * Untyped/typed protocol boundary
     Protocol(..)
   , SomeProtocol(..)
-  , ViewMode(..)
   , fromProtocol
+  , parseProtocol
```
Jimbo4350: `parseProtocol` is used in multiple places. This belongs in `Cardano.Node.Parsers`. Likewise for other parsers used in multiple places.

cardano-node/app/cardano-cli.hs
deepfire: The thing is, it's also less overhead to have the only three places that mention `ClientCommand` in one place. The `main` in `cardano-cli.hs` only just refers to it -- where as `CLI.Run` has both the definition of `ClientCommand` and `runCommand` that interprets it.  Adding in the parser simply completes the triumvirate.

I.e. to add/modify one branch, you can stay entirely within a single file.

What do you think, @Jimbo4350?

cardano-node/src/Cardano/Node/TxSubmission.hs
deepfire: Done.

cardano-node/src/Cardano/Node/CLI.hs
deepfire: Maybe then we also want to move `Protocol` -- to a new `Cardano.Node.Types` module, perhaps?

cardano-node/app/cardano-cli.hs
deepfire: Done.

cardano-node/src/Cardano/Node/CLI.hs
deepfire: Moved `Protocol` and `parseProtocol` into `Cardano.Node.Types`.

Jimbo4350: > Maybe then we also want to move `Protocol` -- to a new `Cardano.Node.Types` module, perhaps, or even, `Cardano.Node.Protocol`?

https://github.com/input-output-hk/cardano-node/pull/143

cardano-node/app/cardano-cli.hs
Jimbo4350: I understand your point but for now I think the style is clearer and more consistent if _all_ the exec's have their parsers in their respective 'main' file.

deepfire: @Jimbo4350, done!

cardano-node/app/cardano-cli.hs
```diff
@@ -122,87 +109,22 @@ parseClientCommand =
...
 parseTestnetBalanceOptions =
-  Genesis.TestnetBalanceOptions
+  TestnetBalanceOptions
   <$> parseIntegral        "n-poor-addresses"         "Number of poor nodes (with small balance)."
   <*> parseIntegral        "n-delegate-addresses"     "Number of delegate nodes (with huge balance)."
   <*> parseLovelace        "total-balance"            "Total balance owned by these nodes."
   <*> parseLovelacePortion "delegate-share"           "Portion of stake owned by all delegates together."
   <*> parseFlag            "use-hd-addresses"         "Whether generate plain addresses or with hd payload."
 
-parseLovelace :: String -> String -> Parser Lovelace
```
Jimbo4350: @deepfire   These parsers also need to be put back here as they are only used in `parseClientCommand`

cardano-node/app/cardano-node.hs
```diff
@@ -57,11 +53,7 @@ main = do
 
 initializeAllFeatures :: CLIArguments -> PartialCardanoConfiguration -> CardanoEnvironment -> IO ([CardanoFeature], NodeLayer)
 initializeAllFeatures (CLIArguments logCli nodeCli) partialConfig cardanoEnvironment = do
-    finalConfig <- case finaliseCardanoConfiguration $
-                          mergeConfiguration partialConfig (commonCLI nodeCli)
-                   of
-      Left err -> throwIO $ ConfigurationError err
-      Right x  -> pure x
+    finalConfig <- mkConfiguration partialConfig (commonCLI nodeCli)
```
Jimbo4350: :+1:

cardano-node/app/cardano-node.hs
```diff
@@ -184,68 +176,15 @@ parseNodeCommand :: Parser NodeCommand
...
         <*> parseProtocol
         <*> parseViewMode
         <*> parseTraceOptions
-  , command' "submit" "Submit a transaction." $
-      TxSubmitter <$> parseTopologyInfo <*> parseMockTx <*> parseProtocol
   , command' "trace-acceptor" "Spawn an acceptor." $
       pure TraceAcceptor
   ]
 
-parseTopologyInfo :: Parser TopologyInfo
```
Jimbo4350: :+1:

cardano-node/app/cardano-cli.hs
deepfire: I was hoping that some of those parsers would come across as at least somewhat generic : -)

deepfire: Done!

PR #140, opened 2019-08-22
Use `ApplyTxErr blk` for tx errors.

Jimbo4350:
Following from: https://github.com/input-output-hk/cardano-node/pull/101


Jimbo4350:
> LGTM
> 
> One question: are all instances of `Show (ApplyTxErr blk)` necessary? For example, just looking at the last two, I don't directly see why.

Strange I force pushed my branch because GHC pointed this out but the changes aren't in Github. They are redundant however!

deepfire:
Github started occasionally acting weird in the last half a year..

PR #147, opened 2019-08-23
`cardano-cli`: Genesis UTxO expenditure

deepfire:
- adds a new subcommand: `issue-genesis-utxo-expenditure`, that allows issuing the initial transaction on the chain, using the genesis UTxO as inputs
- adds a canned testnet -oriented script to easily exercise the Tx creation
- a couple script fixes


cardano-node/src/Cardano/CLI/Run.hs
```diff
@@ -125,9 +132,16 @@ data ClientCommand
     !FilePath
     !FilePath
   | SubmitTx
-    { stTopology :: TopologyInfo
-    , stTx       :: FilePath
-    , stCommon   :: CommonCLI
+    { stTopology      :: TopologyInfo
```
Jimbo4350: These need to be removed because they are partial record accessors.

cardano-node/src/Cardano/CLI/Run.hs
```diff
@@ -125,9 +132,16 @@ data ClientCommand
...
+    , stTx            :: FilePath
+    , stCommon        :: CommonCLI
+    }
+  | SpendGenesisUTxO
+    { ctTx            :: FilePath
```
Jimbo4350: These need to be removed because they are partial record accessors.

cardano-node/src/Cardano/Node/Types.hs
```diff
@@ -1,22 +0,0 @@
-{-# LANGUAGE RankNTypes           #-}
-
-module Cardano.Node.Types
```
Jimbo4350: :+1:

cardano-node/src/Cardano/CLI/Run.hs
deepfire: There is a balance to be had, because legibility is also important.

A similar case was discussed in: https://github.com/input-output-hk/ouroboros-network/pull/838#discussion_r307967852

cardano-node/app/cardano-cli.hs
```diff
@@ -208,3 +218,41 @@ parseFlag optname desc =
             long optname
          <> help desc
     )
+_parseTxIn :: Parser TxIn
```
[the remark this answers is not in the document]
deepfire: Done

cardano-node/src/Cardano/CLI/Run.hs
deepfire: Done!

cardano-node/src/Cardano/CLI/Run.hs
deepfire: Done!

erikd:
With the aim of keeping the git commit history clean an usable (specifically so its not too hard to find what has changed and why) we should avoid having commits with titles like "Address review suggestions". The changes in that commit should be squashed down to into the commits whose comments they are addressing.

deepfire:
> With the aim of keeping the git commit history clean an usable (specifically so its not too hard to find what has changed and why) we should avoid having commits with titles like "Address review suggestions". The changes in that commit should be squashed down to into the commits whose comments they are addressing.

@erikd, done.

cardano-node/src/Cardano/CLI/Run.hs
```diff
@@ -125,10 +132,15 @@ data ClientCommand
...
   | SubmitTx
-    { stTopology :: TopologyInfo
-    , stTx       :: FilePath
-    , stCommon   :: CommonCLI
-    }
+    TopologyInfo
+    FilePath
+    CommonCLI
+  | SpendGenesisUTxO
+    FilePath
```
Jimbo4350: No `newtype` wrapper?

configuration/Test.Cardano.Chain.Genesis.Dummy.dummyConfig.configGenesisData.json
```diff
@@ -1,114 +0,0 @@
-{ "bootStakeholders":
-    { "1a1ff7035103d8a9e1e56a4e4e1d0031fb3f31a6a9953137dfe54c67": 1
-    , "281e5ae9e357970a0dd007dc9b21675ab07de3d67ec825d7363621f9": 1
```
Jimbo4350: :+1:

cardano-node/src/Cardano/Node/TxSubmission.hs
```diff
@@ -94,15 +95,16 @@ submitTx :: ( RunDemo blk
          -> GenTx blk
          -> Tracer IO String
          -> IO ()
-submitTx pInfoConfig nodeId tx tracer =
+submitTx pInfoConfig nodeId tx tracer = do
+    -- TODO: get socketDir from CardanoConfiguration
+    socketDir <- pure "./socket/"
```
Jimbo4350: This seems brittle as the user can now define where they want the socket to live.

deepfire: @Jimbo4350, we need to decide where the socket directory is, somehow.

In `cardano-node` it's now a configurable: https://github.com/input-output-hk/cardano-node/blob/master/cardano-node/src/Cardano/Node/Run.hs#L221

..and I just made `handleTx` respect the configuration as well.

cardano-node/src/Cardano/CLI/Run.hs
deepfire: That'd be a relative first of doing `newtype` around `FilePath` in our codebase, I think, but I also think that the gravitas of the situation merits the excursion.

deepfire: And, done!

deepfire:
@Jimbo4350, squashed all things that weren't logically separate & added shiny `FilePath` newtypes (just in `cardano-cli`, not `cardano-node`).

deepfire:
bors r-

Forgot one little bit -- TxId printing.

PR #149, opened 2019-08-26
More restructuring 

Jimbo4350:
https://github.com/input-output-hk/cardano-node/issues/84

### DIFF
--- a/cardano-node/cardano-node.cabal
+++ b/cardano-node/cardano-node.cabal
@@ -28,14 +28,14 @@ library
                        Cardano.Node.Configuration.Presets
                        -- Features
                        Cardano.Node.Features.Logging
-                       Cardano.Node.GitRev
-                       Cardano.Node.GitRevFromGit
                        Cardano.Node.Orphans
                        Cardano.Node.Parsers
                        Cardano.Node.Run
                        Cardano.Node.Topology
                        Cardano.Node.TraceAcceptor
                        Cardano.Node.Tracers
+                       Cardano.Node.TUI.GitRev
+                       Cardano.Node.TUI.GitRevFromGit
                        Cardano.Node.TxSubmission
                        Cardano.Wallet.Run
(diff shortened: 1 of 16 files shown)


cardano-node/cardano-node.cabal
```diff
@@ -24,24 +24,24 @@ library
                        -- Conversion
                        Cardano.Node.Configuration.Conversion
                        Cardano.Node.Configuration.Partial
+                       Cardano.Node.Configuration.Topology
                        Cardano.Node.Configuration.Types
                        Cardano.Node.Configuration.Presets
```
intricate: This should probably come before `Cardano.Node.Configuration.Topology`

cardano-node/src/Cardano/Node/TUI/GitRev.hs
```diff
@@ -13,7 +13,7 @@
 
 {-# OPTIONS_GHC -Wno-simplifiable-class-constraints #-}
 
-module Cardano.Node.GitRev (
+module Cardano.Node.TUI.GitRev (
```
intricate: What does TUI stand for?

Jimbo4350: https://en.wikipedia.org/wiki/Text-based_user_interface

intricate: Ah, okay. I wonder if `TerminalUI` or `TextUI` might be more clear. Unless `TUI` is pretty common and people generally understand it (@deepfire, what do you think?). Perhaps it's just me who hasn't seen this acronym before.

intricate: Either way, I wouldn't block the PR on this

deepfire: I think `TUI` is good.

PR #152, opened 2019-08-26
Non-genesis UTxO expenditure

deepfire:
1. Adds `issue-utxo-expenditure` subcommand to `cardano-cli`, with `scripts/issue-utxo-expenditure.sh`
2. Factors a couple of immediately related things:
  - `withRealPBFT`
  - `signTxId`
  - `genesisUTxOTxIn`

### DIFF
--- a/cardano-node/README.md
+++ b/cardano-node/README.md
@@ -206,14 +206,22 @@ transaction, in its raw wire format (see GenTx for Byron transactions).
 The canned `scripts/submit-tx.sh` script will submit the supplied transaction to a testnet
 launched by `scripts/shelley-testnet*.sh` family of scripts.
 
-### Issuing genesis UTxO expenditure Tx
+### Issuing UTxO expenditure (genesis and regular)
 
-To make a transaction spending genesis UTxO, you can either use the `issue-genesis-utxo-expenditure`
-subcommand directly, or, again use a canned script that will make a transaction tailored
-for the aforementioned testnet cluster -- `scripts/issue-genesis-utxo-expenditure.sh`.
+To make a transaction spending UTxO, you can either use the:
 
-The script requires the target file name to write the transaction to, and optionally
-allows specifying the target address and lovelace value to send.
+  - `issue-genesis-utxo-expenditure`, for genesis UTxO
+  - `issue-utxo-expenditure`, for normal UTxO
+  
+subcommands directly, or, again use canned scripts that will make transactions tailored
+for the aforementioned testnet cluster:
+
+  - `scripts/issue-genesis-utxo-expenditure.sh`.
+  - `scripts/issue-utxo-expenditure.sh`.
+
+The script requires the target file name to write the transaction to, input TxId
+(for normal UTxO), and optionally allows specifying the source txin output index,
+source and target signing keys and lovelace value to send.
 
 The target address defaults to the 1-st richman key (`configuration/delegate-keys.001.key`)
 of the testnet, and lovelace amount is almost the entirety of its funds.
--- a/cardano-node/app/cardano-cli.hs
+++ b/cardano-node/app/cardano-cli.hs
@@ -3,31 +3,36 @@
 import           Cardano.Prelude hiding (option)
 import           Prelude (String, error, id)
 
+import           Control.Arrow
+import qualified Data.List.NonEmpty as NE
+import           Data.Text
+import           Data.Time (UTCTime)
+import           Data.Time.Clock.POSIX (posixSecondsToUTCTime)
+import qualified Options.Applicative as Opt
+import           Options.Applicative (Parser, ParserInfo, ParserPrefs, auto,
+                                      commandGroup, flag, flag', help, long,
+                                      metavar, option, showHelpOnEmpty,
+                                      strOption, subparser, value)
+
+import           Control.Exception.Safe (catchIO)
+import           System.Exit (ExitCode (..), exitWith)
+
 import           Cardano.Binary (Annotated (..))
 import           Cardano.Chain.Common
 import           Cardano.Chain.Genesis
 import           Cardano.Chain.Slotting
 import           Cardano.Chain.UTxO
+import           Cardano.Crypto ( AProtocolMagic(..)
+                                , ProtocolMagic
+                                , ProtocolMagicId(..)
+                                , RequiresNetworkMagic(..)
+                                , decodeHash)
+
+import           Cardano.Config.CommonCLI
+import           Cardano.Common.Protocol
+import           Cardano.Node.Parsers
 import           Cardano.CLI.Ops (decideCLIOps)
 import           Cardano.CLI.Run
-import           Cardano.Common.Protocol
-import           Cardano.Config.CommonCLI
-import           Cardano.Crypto (AProtocolMagic (..), ProtocolMagic,
-                                 ProtocolMagicId (..),
-                                 RequiresNetworkMagic (..))
-import           Cardano.Node.Parsers
-
-import           Control.Arrow
-import           Control.Exception.Safe (catchIO)
-import qualified Data.List.NonEmpty as NE
-import           Data.Time (UTCTime)
-import           Data.Time.Clock.POSIX (posixSecondsToUTCTime)
-import           Options.Applicative (Parser, ParserInfo, ParserPrefs, auto,
-                                      commandGroup, flag, flag', help, long,
-                                      metavar, option, showHelpOnEmpty,
-                                      strOption, subparser, value)
-import qualified Options.Applicative as Opt
-import           System.Exit (ExitCode (..), exitWith)
 
 main :: IO ()
 main = do
@@ -228,6 +233,17 @@ parseTxRelatedValues =
                   "rich-addr-from"
                   "Tx source: genesis UTxO richman address (non-HD)."
             <*> (NE.fromList <$> some parseTxOut)
+            <*> parseCommonCLI,
+          command'
+            "issue-utxo-expenditure"
+            "Write a file with a signed transaction, spending normal UTxO."
+            $ SpendUTxO
+            <$> parseNewTxFile "tx"
+            <*> parseSigningKeyFile
+                  "wallet-key"
+                  "Key that has access to all mentioned genesis UTxO inputs."
+            <*> (NE.fromList <$> some parseTxIn)
+            <*> (NE.fromList <$> some parseTxOut)
             <*> parseCommonCLI
           ]
       )
@@ -285,16 +301,16 @@ parseNetworkMagic :: Parser NetworkMagic
 parseNetworkMagic =
   asum
     [ flag' NetworkMainOrStage
-        $ mconcat
-            [ long "main-or-staging",
-              help ""
-              ],
+      $ mconcat
+      [ long "main-or-staging",
+        help ""
+      ],
       option (fmap NetworkTestnet auto)
-        ( long "testnet-magic"
-            <> metavar "MAGIC"
-            <> help "The testnet network magic, decibal"
-          )
-      ]
+      ( long "testnet-magic"
+        <> metavar "MAGIC"
+        <> help "The testnet network magic, decibal"
+      )
+    ]
 
 parseProtocolMagicId :: String -> Parser ProtocolMagicId
 parseProtocolMagicId arg =
@@ -310,17 +326,17 @@ parseUTCTime :: String -> String -> Parser UTCTime
 parseUTCTime optname desc =
   option (posixSecondsToUTCTime . fromInteger <$> auto)
     ( long optname
-        <> metavar "POSIXSECONDS"
-        <> help desc
-      )
+      <> metavar "POSIXSECONDS"
+      <> help desc
+    )
 
 parseFilePath :: String -> String -> Parser FilePath
 parseFilePath optname desc =
   strOption
     ( long optname
-        <> metavar "FILEPATH"
-        <> help desc
-      )
+      <> metavar "FILEPATH"
+      <> help desc
+    )
 
 parseIntegral :: Integral a => String -> String -> Parser a
 parseIntegral optname desc =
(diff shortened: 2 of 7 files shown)


cardano-node/src/Cardano/CLI/Run.hs
```diff
@@ -397,25 +421,44 @@ withRealPBFT CLIOps{coProtocol} pcc common action = do
...
+txSpendGenesisUTxOByronPBFT gc sk genAddr outs =
+  let txattrs    = mkAttributes ()
       tx         = UnsafeTx (pure txIn) outs txattrs
       txIn      :: CC.UTxO.TxIn
-      txIn       = handleMissingAddr $ fst <$> Map.lookup genRichAddr initialUtxo
+      txIn       = genesisUTxOTxIn gc sk genAddr
+      cheat     :: CC.UTxO.TxWitness
+      -- Below, we have to cheat to spend a genesis UTxO entry:
+      -- sign our own TxId, not that of the input Tx.
+      -- Ledger knows.
```
Jimbo4350: Ledger knows what?

cardano-node/README.md
```diff
@@ -81,33 +81,31 @@ A CLI utility to support a variety of key material operations (genesis, migratio
 
 The general synopsis is as follows:
```
Jimbo4350: This commit should be removed and put in a separate PR.

deepfire: Yes, sure, that's the intent.

cardano-node/src/Cardano/CLI/Run.hs
deepfire: Ledger has a special rule to validate TXes that use genesis UTxO -- it's in on the "cheating".

deepfire: I'll clarify the comment.

cardano-node/README.md
```diff
@@ -81,33 +81,31 @@ A CLI utility to support a variety of key material operations (genesis, migratio
...
 
 NOTE: the exact invocation command depends on the environment.  If you have only
 built `cardano-cli`, without installing it, then you have to prepend `cabal
 new-run -- ` before `cardano-cli`.  We henceforth assume that the necessary
 environment-specific adjustment has been made, so we only mention `cardano-cli`.
 
-`SYSTEMVER` is one of the supported system generations: `byron-legacy`, `real-pbft` etc.
-The full list can be seen in the output of `cardano-cli --help`, along with the list of
-supported commands.
+The `--byron-legacy` and `--real-pbft` options immediately preceding the
```
Jimbo4350: :+1:

cardano-node/README.md
```diff
@@ -116,25 +114,36 @@ Available options:
 
 ## Genesis operations
 
-The canned `scripts/genesis.sh` example provides a nice set of defaults and
-illustrates available options.  Running it will produce a `./genesis.${systemStart}` directory,
-where `${systemStart}` will be chosen 15 minutes in the future.
+### Generation
 
-The directory will have the following content:
+The genesis generation operations will create a directory that contains:
```
Jimbo4350: :+1:

cardano-node/src/Cardano/CLI/Run.hs
```diff
@@ -19,6 +20,7 @@
 {-# OPTIONS_GHC -Wno-all-missed-specialisations #-}
 {-# OPTIONS_GHC -Wno-simplifiable-class-constraints #-}
 {-# OPTIONS_GHC -Wno-partial-fields #-}
+{-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
```
Jimbo4350: Just saw this. Please handle the other cases explicitly.

cardano-node/src/Cardano/CLI/Run.hs
```diff
@@ -203,6 +209,12 @@ data ClientCommand
     Address
     (NonEmpty TxOut)
     CommonCLI
+  | SpendUTxO
+    NewTxFile
```
Jimbo4350: These should be annotated similarly to `IssueDelegationCertificate` above.

cardano-node/src/Cardano/CLI/Run.hs
deepfire: I found a way to convince GHC that the impossible branch is indeed impossible (removed a type variable + a type equality that pinned a branch in the `ProtocolInfo` GADT), so this is now unnecessary.

Done.

cardano-node/src/Cardano/CLI/Run.hs
deepfire: @Jimbo4350, done.

deepfire:
All suggestions addressed @Jimbo4350 -- although I think that we could have postponed the reformatting, since it incurred some avoidable conflicts that had to be dealt with.

I also don't think we've reached consensus here, since in https://github.com/input-output-hk/cardano-node/pull/161#issuecomment-526576119 you said:

> I think for now we can use stylish-haskell for imports and use our discretion for everything else

cardano-node/src/Cardano/CLI/Run.hs
```diff
@@ -87,8 +89,12 @@ import           Cardano.Crypto (SigningKey (..), ProtocolMagic, ProtocolMagicId
 import qualified Cardano.Crypto.Hashing as Crypto
 import qualified Cardano.Crypto.Random as Crypto
 import qualified Cardano.Crypto.Signing as Crypto
+import           Cardano.Config.Partial (PartialCardanoConfiguration (..))
```
Jimbo4350: Imports here also out of order.

cardano-node/src/Cardano/CLI/Run.hs
```diff
@@ -196,12 +202,27 @@ data ClientCommand
   | SubmitTx
     TopologyInfo
     TxFile
+    -- ^ Filepath of transaction to submit.
```
Jimbo4350: :100: !

cardano-node/src/Cardano/CLI/Run.hs
```diff
@@ -356,64 +377,113 @@ runCommand CLIOps{..}
...
-      txattrs    = mkAttributes ()
-      tx         = UnsafeTx (pure txIn) outs txattrs
-      txIn      :: CC.UTxO.TxIn
-      txIn       = handleMissingAddr $ fst <$> Map.lookup genRichAddr initialUtxo
+withRealPBFT
+  :: CLIOps IO
+  -> PartialCardanoConfiguration
+  -> CommonCLI
+  -> (RunDemo (ByronBlockOrEBB ByronConfig)
+      => CardanoConfiguration
```
Jimbo4350: It's difficult for me to infer what is happening here. Can you document with comments on what `withRealPBFT` is doing and perhaps wrap 
```
(RunDemo (ByronBlockOrEBB ByronConfig)
      => CardanoConfiguration
      -> Consensus.Protocol (ByronBlockOrEBB ByronConfig)
      -> IO ())
``` 
in a `newtype` and also annotate with comments?

Jimbo4350: @intricate any suggestions?

cardano-node/README.md
```diff
@@ -206,14 +206,22 @@ transaction, in its raw wire format (see GenTx for Byron transactions).
 The canned `scripts/submit-tx.sh` script will submit the supplied transaction to a testnet
 launched by `scripts/shelley-testnet*.sh` family of scripts.
 
-### Issuing genesis UTxO expenditure Tx
+### Issuing UTxO expenditure (genesis and regular)
 
-To make a transaction spending genesis UTxO, you can either use the `issue-genesis-utxo-expenditure`
-subcommand directly, or, again use a canned script that will make a transaction tailored
-for the aforementioned testnet cluster -- `scripts/issue-genesis-utxo-expenditure.sh`.
+To make a transaction spending UTxO, you can either use the:
```
Jimbo4350: We don't need to do this in this PR, but I think we should migrate all instructions wrt scripts to `SCRIPTS.md` and just leave a short section in the top level `README.md` pointing to this.

cardano-node/app/cardano-cli.hs
```diff
@@ -372,6 +393,15 @@ parseAddress opt desc =
         <> help desc
       )
 
+parseTxIn :: Parser TxIn
+parseTxIn = option (uncurry TxInUtxo
```
Jimbo4350: This could be reformatted to:
```
parseTxIn :: Parser TxIn
parseTxIn =
  option
    ( uncurry TxInUtxo
        . Control.Arrow.first cliParseTxId
        <$> auto
      )
    ( long "txin"
        <> metavar "(TXID,INDEX)"
        <> help "Transaction input is a pair of an UTxO TxId and a zero-based output index."
      )
```

Jimbo4350:
> All suggestions addressed @Jimbo4350 -- although I think that we could have postponed the reformatting, since:
> 
> 1. It incurred some avoidable conflicts that had to be dealt with.
> 2. It completely invalidated the work I did (separate from this PR, but temporarily residing in it) in [6f65dc0](https://github.com/input-output-hk/cardano-node/commit/6f65dc05f9f25ffe82c01eb93de74f34fb966d3a) -- that we discussed.
> 
> I also don't think we've reached consensus here, since in [#161 (comment)](https://github.com/input-output-hk/cardano-node/pull/161#issuecomment-526576119) you said:
> 
> > I think for now we can use stylish-haskell for imports and use our discretion for everything else
> 
> ..and I thought that was the consensus -- but apparently, alas, it wasn't.

@deepfire your suggested formatting makes reading diff more difficult than it needs to be. Formatting is not only about how it looks locally.

> All suggestions addressed @Jimbo4350 -- although I think that we could have postponed the reformatting, since:
> 
> 1. It incurred some avoidable conflicts that had to be dealt with.
> 2. It completely invalidated the work I did (separate from this PR, but temporarily residing in it) in [6f65dc0](https://github.com/input-output-hk/cardano-node/commit/6f65dc05f9f25ffe82c01eb93de74f34fb966d3a) -- that we discussed.
> 
> I also don't think we've reached consensus here, since in [#161 (comment)](https://github.com/input-output-hk/cardano-node/pull/161#issuecomment-526576119) you said:
> 
> > I think for now we can use stylish-haskell for imports and use our discretion for everything else
> 
> ..and I thought that was the consensus -- but apparently, alas, it wasn't.

Unfortunately the reformatting on your branch makes it very difficult to view changes on the github diff. The local view is important but I think the format in my PR is a good compromise.

cardano-node/app/cardano-cli.hs
```diff
@@ -8,18 +8,22 @@ import           Cardano.Chain.Common
 import           Cardano.Chain.Genesis
 import           Cardano.Chain.Slotting
 import           Cardano.Chain.UTxO
-import           Cardano.CLI.Ops (decideCLIOps)
-import           Cardano.CLI.Run
-import           Cardano.Common.Protocol
+import           Cardano.Crypto ( AProtocolMagic(..)
```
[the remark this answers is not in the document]
deepfire: @Jimbo4350, frankly I found the new order of imports confusing.

We used to have roughly the following order to the module groups:

- external pure
- external impure
- external Cardano
- cardano-node local 

The https://github.com/input-output-hk/cardano-node/pull/161 has cancelled that order.

So I'm a bit lost to what is your intent here.

cardano-node/README.md
deepfire: I think we can address that in a future PR, what do you think?

cardano-node/app/cardano-cli.hs
deepfire: It could be, but it seems fairly arbitrary.

I mean, how does one arrive at such a change, independently?

deepfire:
> Unfortunately the reformatting on your branch makes it very difficult to view changes on the github diff. The local view is important but I think the format in my PR is a good compromise.

I don't think it looks too bad -- does it?:

Screenshot from 2019-08-30 15-26-42

Granted, it might be a problem in a side-by-side diff, but that's not the default view, so it's probably not used very often by most people.

Reading the code in a text editor, though, is a much more frequent operation, for most people,
and the tabular layout, I think, was quite beneficial to that use case.

cardano-node/src/Cardano/CLI/Run.hs
deepfire: Removed that piece altogether.

cardano-node/app/cardano-cli.hs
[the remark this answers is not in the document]
Jimbo4350: I meant alphabetically, I made a mistake in #161.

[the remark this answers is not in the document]
deepfire: So will you support the restoration of the grouping I mentioned above -- with alphabetic intra-group sorting, of course?

cardano-node/src/Cardano/CLI/Run.hs
intricate: I wouldn't agree with the `newtype` point but adding a quick one-liner of doc for `withRealPBFT` would be nice.

cardano-node/app/cardano-cli.hs
intricate: I'm also not a big fan of the formatting here. I think the usage of horizontal space in this context is what bothers me. In fact, I think I would have even been okay with a minor variant of @deepfire's formatting:
```
parseTxIn :: Parser TxIn
parseTxIn = 
  option 
    (uncurry TxInUtxo 
      . Control.Arrow.first cliParseTxId
      <$> auto
    )
    ( long       "txin"
      <> metavar "(TXID,INDEX)"
      <> help    "Transaction input is a pair of an UTxO TxId and a zero-based output index."
    )
```

deepfire: @intricate, would you find it acceptable to have a minor variation on your option:
```
parseTxIn :: Parser TxIn
parseTxIn = 
  option 
  ( uncurry TxInUtxo 
    . Control.Arrow.first cliParseTxId
    <$> auto
  )
  ( long       "txin"
    <> metavar "(TXID,INDEX)"
    <> help    "Transaction input is a pair of an UTxO TxId and a zero-based output index."
  )
```
The difference is that this is compatible with the default way of `haskell-mode` indentation in Emacs -- whereas the previous option isn't (it's unavailable in the sequence of indentation options it provides).

cardano-node/app/cardano-cli.hs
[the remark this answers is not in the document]
deepfire: Done!

cardano-node/src/Cardano/CLI/Run.hs
deepfire: Done!

cardano-node/app/cardano-cli.hs
erikd: Vertically aligning code leads to git diffs with semantically null changes when the length of an identifier changes and other things are brought into alignment with it.

I currently rely heavily on git history and therefore highly value clean minimal commits with good commit messages. Anything that degrades the quality of git commits should be highly discouraged. Whoever maintains this code in the future will also rely heavily on git history. We should try and keep that history as clean and as useful as possible. This matters because we hope this code and the system it runs will be around for 50 years or more.

deepfire: @erikd, for the record -- I completely support that noble objective.  It's just that the original comment was unclear about the goal we were aiming for.

PR #154, opened 2019-08-27
Add `chairman.sh` script

Jimbo4350:
- Add some additional comments to `Chairman.hs`.
- Before this commit, `chairman` was comparing the shortest potential fork to `k` when it should be comparing the longest potential fork to `k`. See comment below.

### DIFF
--- a/scripts/chairman.sh
+++ b/scripts/chairman.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+
+set -e
+
+genesis="33873"
+genesis_root="configuration/${genesis}"
+genesis_file="${genesis_root}/genesis.json"
+if test ! -f "${genesis_file}"
+then echo "ERROR: genesis ${genesis_file} does not exist!">&1; exit 1; fi
+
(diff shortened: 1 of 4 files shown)


scripts/chairman.sh
```diff
@@ -0,0 +1,32 @@
...
+genesis_root="configuration/${genesis}"
+genesis_file="${genesis_root}/genesis.json"
+if test ! -f "${genesis_file}"
+then echo "ERROR: genesis ${genesis_file} does not exist!">&1; exit 1; fi
+
+cabal new-build "exe:cardano-cli"
+genesis_hash="$(cabal new-exec -- cardano-cli real-pbft print-genesis-hash --genesis-json ${genesis_file})"
+
+CMD=`find dist-newstyle/ -type f -name "cardano-node"`
+test -n "${CMD}" || {
```
coot: I think that instead of `new-build` & `new-exec` one can just use `new-run`.  It rebuilds if necessary.

coot: ok

PR #161, opened 2019-08-29
Reformat `app/cardano-cli.hs`

Jimbo4350:


deepfire:
@Jimbo4350, I wonder if we can postpone this PR -- this will cause bloody conflicts that would be really, really painful to resolve. Moreover -- we have an alternative take on the CLI formatting in https://github.com/input-output-hk/cardano-node/pull/152/commits/fb9f1552ae050ea1e1d0398cef0840f91fb56a93

Jimbo4350:
> @Jimbo4350, I wonder if we can postpone this PR -- this will cause bloody conflicts that would be really, really painful to resolve. Moreover -- we have an alternative take on the CLI formatting in [fb9f155](https://github.com/input-output-hk/cardano-node/commit/fb9f1552ae050ea1e1d0398cef0840f91fb56a93). One way to resolve this would be to add a script which does the actual reformatting, and it could be run at the end of the outstanding PR that would conflict here.
> 
> Next thing to consider is how we are approching formatting.
> 
> I don't think it's really sustainable to just reformat things using a code formatter, if we want to keep that format.
> 
> I think we need editor settings (at least I'd want `emacs`, not sure about others) if we want that reformatting to not incur constant costs as we go forward.

Sure I can wait till your PR is merged. I'm happy for us to decide on a format however too many times I've seen formatting turn into bikeshedding. I think for now we can use `stylish-haskell` for imports and use our discretion for everything else.

deepfire:
> I think for now we can use stylish-haskell for imports and use our discretion for everything else.

That sounds like a good middle ground, indeed!

deepfire:
For the record -- this PR duplicates some of the work in https://github.com/input-output-hk/cardano-node/tree/serge/reformat

I'm still of the opinion that it was merged prematurely (and also, in violation of due process):

https://github.com/input-output-hk/cardano-node/pull/161#event-2597566282 seems to imply there was consensus, but there was not.

Jimbo4350:
We are currently discussing a style for the repo here: https://github.com/input-output-hk/cardano-node/pull/164

PR #164, opened 2019-09-02
Commit style document

Jimbo4350:
A document to (hopefully) save us from the perils of inconsistent, ugly code :slightly_smiling_face: .

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -1,0 +1,453 @@
+Haskell Style Guide
+============================
+> This style guide is mostly a copy of [Johan Tibell's guide](https://github.com/tibbe/haskell-style-guide/blob/master/haskell-style.md)
+> with some restructurization, elaboration on some topics and some additions.
+> This style guide's aims are code beauty, readability and understandability.
+
+
+Principles
+----------
+* The primary goal is readability/maintainability with performance as a close second.
+* Always use types to clarify intent and make invalid states unrepresentable.
+* Use type signatures as compiler checked documentation.
+* Break up large functions into smaller sub functions.
+* When modifying or adding code to existing code (either at the file, project or company level), try to match the style of the existing code.
+
+General guide lines
+-------------------
+### Line Length
+
+Maximum line length is *80 characters* or *100 characters* if necessary.
+
+Modern screens have high definition and big width.
+But with some tiling managers and two terminals on one screen you are not able to
+see many characters on one line.
+On the other hand, restricting line size to a very small number like 80 leads to
+some crazy indentation despite the fact that
+shorter lines should force you to write well structured code.
+That's why *100* is a reasonable compromise.
+
+
+### Indentation
+
+* Do not vertically align code. Mainly because the need to change identifiers and re-aligning the other lines causes semantically null changes in the diff.
+* Tabs are illegal.
+* Use spaces for indenting.
+* Indent your code blocks with *2 spaces*.
+* Indent the `where` keyword with one space to set it
+apart from the rest of the code and indent the definitions in a
+`where` clause with 1 space.
+
+Some examples:
+
+```haskell
+sayHello :: IO ()
+sayHello = do
+  name <- getLine
+  putStrLn $ greeting name
+ where
+  greeting :: String -> String
+  greeting name = "Hello, " ++ name ++ "!"
+
+filter :: (a -> Bool) -> [a] -> [a]
+filter _ [] = []
+filter p (x:xs)
+  | p x = x : filter p xs
+  | otherwise = filter p xs
+```
+
+### Blank Lines
+
+* One blank line between top-level definitions.
+* No blank lines between type signatures and function definitions.
+* Add one blank line between
+functions in a type class instance declaration if the function bodies
+are large.
+* You can add blank lines inside a big `do` block to separate logical
+parts of it.
+* You can also use blank lines to separate definitions inside `where`
+clause, if necessary.
+
+### Whitespace
+
+* Surround binary operators with a single space on either side. In case of
+currying add one space between the argument and the operation.
+
+* Use some tool to remove trailing spaces.
+
+### Naming convention
+
+**Casing**:
+
++ **_lowerCamelCase_** for function and variable names.
++ **_UpperCamelCase_** for types.
+
+* Don't use short names like `n`, `sk`, `f` unless their meaning is clear from
+context (function name, types, other variables, etc.).
+
+* For readability reasons, don't capitalize all letters when using an
+abbreviation.  For example, write `HttpServer` instead of
+`HTTPServer`.  Exception: two or three letter abbreviations, e.g. `IO`, `STM`.
+
+**Records name conventions**
+
+If data type has only one constructor then this data type name should be the same
+as constructor name (also applies to `newtype`).
+
+```haskell
+data User = User Int String
+```
+
+**Field names for record data types**
+
+Choose a reasonable prefix is used such that we can ensure the field name will remain unique throughout the package. Use your discretion.
+
+e.g
+```haskell
+data MempoolEnv m blk = MempoolEnv
+  { meLedger :: LedgerInterface m blk
+  , meLedgerCfg :: LedgerConfig blk
+  , meStateVar :: StrictTVar m (InternalState blk)
+  , meTracer :: Tracer m (TraceEventMempool blk)
+  }
+```
+
+### Comments
+
+#### Punctuation
+
+Write proper sentences; start with a capital letter and use proper
+punctuation.
+
+#### Top-Level Definitions
+
+* Comment top level functions (particularly exported functions)
+as much as possible. If it is not blindingly obvious what a function
+is doing, you should add a comment.
+* Always provide a type signature
+* Use Haddock syntax in the comments.
+* Comment every exported data type.
+
+Function example:
+
+```haskell
+-- | Send a message on a socket. The socket must be in a connected
+-- state. Returns the number of bytes sent. Applications are
+-- responsible for ensuring that all data has been sent.
+send
+  :: Socket      -- ^ Connected socket
+  -> ByteString  -- ^ Data to send
+  -> IO Int      -- ^ Bytes sent
+```
+
+* For functions, the documentation should give enough information to
+apply the function without looking at its definition.
+
+Record example:
+
+```haskell
+-- | Bla bla bla.
+data Person = Person
+  { age :: !Int     -- ^ Age
+  , name :: !String  -- ^ First name
+  }
+```
+
+For fields that require longer comments, format them this way:
+
+```haskell
+data Record = Record
+  { -- | This is a very very very long comment that is split over
+    -- multiple lines.
+    field1 :: !Text
+
+    -- | This is a second very very very long comment that is split
+    -- over multiple lines.
+  , field2 :: !Int
+  }
+```
+
+#### End-of-Line Comments
+
+Separate end-of-line comments from the code with 2 spaces. Align
+comments for data type definitions. Some examples:
+
+```haskell
+data Parser =
+  Parser
+    !Int         -- ^ Current position
+    !ByteString  -- ^ Remaining input
+
+foo :: Int -> Int
+foo n = salt * 32 + 9
+  where
+    salt = 453645243  -- Magic hash salt.
+```
+
+#### Links
+
+Use in-line links economically.  You are encouraged to add links for
+API names. It is not necessary to add links for all API names in a
+Haddock comment. We therefore recommend adding a link to an API name
+if:
+
+* The user might actually want to click on it for more information (in
+  your opinion), and
+
+* Only for the first occurrence of each API name in the comment (don't
+  bother repeating a link)
+
+Top-down guideline
+------------------
+
+### LANGUAGE extensions section
+
+Write each `LANGUAGE` pragma on its own line, sort them alphabetically.
+_Do not_ align by max width as future re-alignment when adding new pragmas
+will cause semantically null changes in the diff.
+
+```haskell
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE Rank2Types #-}
+{-# LANGUAGE TemplateHaskell #-}
+```
+
+### Module name
+
+Use singular when naming modules (e.g. use `Data.Map` and
+`Data.ByteString.Internal` instead of `Data.Maps` and
+`Data.ByteString.Internals`). Sometimes it's acceptable to use plural
+(e. g. `Types`, `Instances`).
+
+### Export Lists
+
+Format export lists as follows:
+
+```haskell
+module Data.Set
+  ( Set
+  , empty
+  , singleton
+  , member
+  ) where
+```
+
+Some clarifications:
+
+1. Use 2 spaces indentation for export list.
+2. You can split export list into sections or just write all as single section.
+3. It is strongly adviced to sort each section alpabetically. However,
+   classes, data types and type aliases should be written before
+   functions.
+
+### Imports
+
+Imports should be grouped in the following order:
+
+0. Import of custom prelude (for example [`cardano-prelude`](https://github.com/input-output-hk/cardano-prelude)) if used.
+1. External modules.
+2. External IOHK modules.
+3. Local library modules.
+4. Local test modules
+
+* Put a blank line between each group of imports.
+
+* The imports in each group should be sorted alphabetically, by module name.
+
+* For external modules, use explicit or `qualified` imports.
+* For local modules, use implicit imports.
+* Always use explicit import lists or `qualified` imports.
+  * **Exception**:
+   [`cardano-prelude`](https://github.com/input-output-hk/cardano-prelude) (import implicitly)
+
+If the `import` is unqualified then put _11 spaces_ between the `import` keyword and the module name (i.e. length of `qualified` + 2).
+
+Unqualified types (i.e. `Map` vs. `M.Map`) look pretty good and not so ugly.
+Prefer two-line imports for such standard containers.
+
+```haskell
+import           Data.Map (Map)
+import qualified Data.Map as Map
+```
+
+### Data Declarations
+
+Align the constructors in a data type definition.  Example:
+
+```haskell
+data HttpException
+  = InvalidStatusCode Int
+  | MissingContentHeader
+```
+
+Format records as follows:
+
+```haskell
+data Person = Person
+  { firstName :: !String  -- ^ First name
+  , lastName :: !String  -- ^ Last name
+  , age :: !Int     -- ^ Age
+  } deriving (Eq, Show)
+```
+
+Type classes in `deriving` section should be always surrounded by
+parentheses.
+
+_WARNING_: try to avoid aggressive autoderiving. Deriving instances can
+slowdown compilation
+(stated here: http://www.stephendiehl.com/posts/production.html)
+
+> Deriving instances of Read/Show/Data/Generic for largely recursive ADTs can
+> sometimes lead to quadratic memory behavior when the nesting gets deep.
+
+### Function declaration
+
+* All functions must have type signatures.
+
+* Specialize function type signature for concrete types if you're using this function
+with only one type for each argument. Otherwise you should use more polymorphic
+version. Compiler can optimize specialized functions better
+and meaning of this function may be clearer.
+Use this rule unless you are the library creator and want your library to be abstract
+as possible.
+
+* It is allowed to omit parentheses for only one type class constraint.
+
+* If a function type signature is very long then place some of the types on its own line.
+
+Example :
+```haskell
+putValueInState
+  :: MonadIO m
+  => UserState -> Maybe Int -> AppConfig
+  -> Integer -> (Int -> m ()) -> m ()
+```
+
+
+### List Declarations
+
+Align the elements in the list.  Example:
+
+```haskell
+exceptions =
+  [ InvalidStatusCode
+  , MissingContentHeader
+  , InternalServerError
+  ]
+```
+
+### If-then-else clauses
+
+Generally, guards and pattern matches should be preferred over _if-then-else_
+clauses, where possible.  Short cases should usually be put on a single line
+(when line length allows it).
+
+### Case expressions
+
+The alternatives in a case expression can be indented using either of
+the two following styles:
+
+```haskell
+foobar =
+  case something of
+    Just j -> foo
+    Nothing -> bar
+```
+
+or as
+
+```haskell
+foobar = case something of
+           Just j -> foo
+           Nothing -> bar
+```
+
+but not
+
+```haskell
+foobar = case something of
+    Just j  -> foo
+    Nothing -> bar
+```
+Subexpressions should always be indented to the right of their parent expression.
+
+Dealing with laziness
+---------------------
+
+By default, use strict data types and lazy functions.
+
+### Data types
+
+Constructor fields should be strict, unless there's an explicit reason
+to make them lazy.  This helps to avoid many common pitfalls caused by too much
+laziness and reduces the number of brain cycles the programmer has to
+spend thinking about evaluation order.
+
+```haskell
+-- Good
+data Point = Point
+  { pointX :: !Double  -- ^ X coordinate
+  , pointY :: !Double  -- ^ Y coordinate
+  }
+```
+
+```haskell
+-- Bad
+data Point = Point
+  { pointX :: Double  -- ^ X coordinate
+  , pointY :: Double  -- ^ Y coordinate
+  }
+```
+
+Misc
+----
+
+### Point-free style ###
+
+Avoid over-using point-free style. Again, use your discretion.
+
+For example, this is hard to read:
+
+```haskell
+-- Bad:
+f = (g .) . h
+```
+Whereas this is fine:
+
+```haskell
+unFooBar :: Foo -> Baz
+unFooBar = Baz . unBar . unFoo
+```
+The point-free version needs to have a clarity benefit but we realize
+that this is vague so use your discretion.
+
+Cabal file formatting
+---------------------
+
+### Modules & libraries
+
+Modules and libraries should go in alphabetical order inside corresponding
+sections. You can put blank lines between groups in each section.
+
+### Warnings ###
+
+Code should be compilable with `-Wall` without warnings.
+
+
+### Default extensions
+
+The only allowed default extension is: `NoImplicitPrelude`
+All other extensions need to be put in the relevant module.
+
+## Language Extensions to avoid
+
+1. Avoid `RecordWildCards` - use `NamedFieldPuns` instead. `RecordWildCards` makes it difficult to discern where record accessor functions come from, _especially_ when used with external libraries. This results in time wasted searching for the accessors. `NamedFieldPuns` forces you to be more explicit and is a better alternative.
+
+
+2. Avoid `ViewPatterns` - `ViewPatterns` obfuscates what could have been readable code and incurs more cognitive overhead.
+
+## OPTIONS_GHC pragma to avoid
+
+1. `-Wno-incomplete-uni-patterns` - This a bad idea. GHC will not warn you if you have not explicitly handled all possible cases in your function definitions. This opens up the possibility of a runtime error.
+
+2. `-Wno-partial-fields` - This will allow you to define a partial record accessor which opens you up to the possibility of a runtime error.


STYLE.md
```diff
@@ -0,0 +1,436 @@
...
+
+**Records name conventions**
+
+If data type has only one constructor then this data type name should be same
+as constructor name (also applies to `newtype`).
+
+```haskell
+data User = User Int String
+```
+TODO: Convention for the `newtype` field name?
```
Jimbo4350: @deepfire

STYLE.md
```diff
@@ -0,0 +1,432 @@
...
+
+```haskell
+-- Bad
+data Point = Point
+  { pointX :: Double  -- ^ X coordinate
+  , pointY :: Double  -- ^ Y coordinate
+  }
+```
+
+TODO: Talk to luke
```
Jimbo4350: @intricate I know you did some work here. Feel free to add.

STYLE.md
```diff
@@ -0,0 +1,432 @@
...
+Additionally, unpacking simple fields often improves performance and
+reduces memory usage:
```
intricate: Unpacking strict fields can be a good idea but sometimes it's not (see links below). I tend to only look at manually `UNPACK`ing when I'm specifically trying to reduce memory usage for a particular data type. This might be worth the mention but I'm not sure it's worth enforcing as part of the style. 

https://wiki.haskell.org/Performance/Data_types#Unpacking_strict_fields
https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#unpack-pragma

intricate: Maybe @dcoutts will want to weigh in on this as well.

Jimbo4350: @intricate  Perhaps we can leave this out of the style document as well and review on a case by case basis.

STYLE.md
```diff
@@ -0,0 +1,432 @@
...
+  , pointY :: {-# UNPACK #-} !Double  -- ^ Y coordinate
+  }
+```
+
+Misc
+----
+
+### Point-free style ###
+
+Avoid over-using point-free style.  For example, this is hard to read:
```
intricate: :100:

STYLE.md
```diff
@@ -0,0 +1,432 @@
...
+
+
+### Default extensions
+
+The only allowed default extension is: `NoImplicitPrelude`
+All other extensions need to be put in the relevant module.
+
+## Extensions to avoid
+
+Avoid `RecordWildCards`, use `NameFieldPuns` instead.
```
intricate: Would also avoid `ViewPatterns`

Jimbo4350: I'll include this in  the changes after I get some more feedback from others.

STYLE.md
```diff
@@ -0,0 +1,432 @@
...
+
+If data type has only one constructor then this data type name should be same
+as constructor name (also applies to `newtype`).
+
+```haskell
+data User = User Int String
+```
+TODO: Convention for the `newtype` field name?
+
+TODO: Convention for field names for record data types?
```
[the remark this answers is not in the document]
deepfire: What I saw in the rest of the codebase, is that the combination of capital letters of the type is used -- i.e. for `MempoolEnv` that would be `me`, not `mp`.

STYLE.md
deepfire: Do we need to mandate anything here?

I think that enforcing the `unThing` style often leads to overly long accessor names (which are OK for the newtype itself).

But on the other hand, not having to remember the name of the accessor is an upside.

Jimbo4350: I'm happy to not mandate anything and just deal with on a case by case basis.

STYLE.md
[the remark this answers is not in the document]
Jimbo4350: That works for me :)

STYLE.md
```diff
@@ -0,0 +1,432 @@
...
+### Comments
+
+#### Punctuation
+
+Write proper sentences; start with a capital letter and use proper
+punctuation.
+
+#### Top-Level Definitions
+
+* Comment every top level function (particularly exported functions),
```
deepfire: I think that this is a very good recommendation for library code.

But on the other hand, in code that's at the top of the call stack, that's less of a concern, and sometimes leads to non-functional comments like in https://github.com/input-output-hk/cardano-node/blob/master/cardano-node/app/wallet-client.hs#L34 (no offense to the authors!).

Whatever the function does is fairly obvious from its name and type -- so I believe the comment is a waste of screen estate.

I think we should relax the wording.

STYLE.md
```diff
@@ -0,0 +1,432 @@
...
+
+1. Use 2 spaces indentation for export list.
+2. You can split export list into sections or just write all as single section.
+3. It is strongly adviced to sort each section alpabetically. However,
+   classes, data types and type aliases should be written before
+   functions.
+
+### Imports
+
+Imports should be grouped in the following order:
```
deepfire: :+1:

STYLE.md
```diff
@@ -0,0 +1,432 @@
...
+1. External modules.
+2. External IOHK modules.
+3. Local library modules.
+4. Local test modules
+
+* Put a blank line between each group of imports.
+
+* The imports in each group should be sorted alphabetically, by module name.
+
+* Always use explicit import lists or `qualified` imports.
```
deepfire: What is the rationale for this?

STYLE.md
```diff
@@ -0,0 +1,432 @@
...
+Try to use `qualified` imports only if import list is
```
deepfire: What is the rationale for discouraging `qualified` imports?

Can you explain the robustness-against-change argument (if I read it correctly that it applies)?

STYLE.md
deepfire: Now, I think we could make this more helpful, if we set some guidelines about what constitutes over-use.

Otherwise it's overwhelmingly left to judgement, and judgements can differ between reasonable people.

STYLE.md
```diff
@@ -0,0 +1,432 @@
...
+============================
+> This style guide is mostly a copy of [Johan Tibell's guide](https://github.com/tibbe/haskell-style-guide/blob/master/haskell-style.md)
+> with some restructurization, elaboration on some topics and some additions.
+> This style guide's aims are code beauty, readability and understandability.
+
+General guide lines
+-------------------
+### Line Length
+
+Maximum line length is *80 characters* or *100 characters* if necessary.
```
deepfire: Now, I think we could further relax that restriction in one very specific case, namely:

1. Lines that end with long string literals.

The reason is that this allows using tabular layout in code that often gains from it a lot.

The relative downside is small -- after all, this is just prose, that overflows in constrained line scenarios, not code.

STYLE.md
intricate: Perhaps we could note it as a point to consider in the document.

STYLE.md
intricate: > Now, I think we could make this more helpful, if we set some guidelines about what constitutes over-use.
> 
> Otherwise it's overwhelmingly left to judgement, and judgements can differ between reasonable people.

I think that's fair.

STYLE.md
```diff
@@ -0,0 +1,432 @@
...
+context (function name, types, other variables, etc.).
+
+* For readability reasons, don't capitalize all letters when using an
+abbreviation.  For example, write `HttpServer` instead of
+`HTTPServer`.  Exception: two or three letter abbreviations, e.g. `IO`, `STM`.
+
+**Records name conventions**
+
+If data type has only one constructor then this data type name should be same
+as constructor name (also applies to `newtype`).
```
deepfire: Counter examples in core code that I've seen only just recently:

- https://github.com/input-output-hk/cardano-ledger/blob/master/cardano-ledger/src/Cardano/Chain/UTxO/Tx.hs#L51
- https://github.com/input-output-hk/cardano-ledger/blob/master/cardano-ledger/src/Cardano/Chain/UTxO/Tx.hs#L123

STYLE.md
```diff
@@ -0,0 +1,432 @@
...
+
+If `import` is unqualified then put _11 spaces_ between `import` keyword and
+module name (i.e. length of `qualified` + 2).
+
+Unqualified types (i.e. `Map` vs. `M.Map`) look pretty good and not so ugly.
+Prefer two-line imports for such standard containers.
+
+```haskell
+import           Data.Map (Map)
+import qualified Data.Map as M hiding (Map)
```
deepfire: Is the `hiding (Map)` bit adding value here?

STYLE.md
deepfire: I don't think that applies to preludes, does it?

STYLE.md
Jimbo4350: This is a good point, making changes.

STYLE.md
intricate: I would argue that it isn't. It's quite obvious from the unqualified `Data.Map` import that you're using `Map` unqualified so I see no reason to hide it from the qualified `Data.Map` import.

STYLE.md
intricate: Custom preludes seem to be noted as an exception a few lines below: https://github.com/input-output-hk/cardano-node/pull/164/files/027d248e2d99847f3a109c8ae9794be0d8e83e07#diff-32430379d7ba797ddf827c29a4eac4c1R236

deepfire: Ah, my apologies!

STYLE.md
intricate: Well actually... perhaps since it prevents you from using `M.Map`, that could add some value.

STYLE.md
Jimbo4350: `qualified` imports take away from code readability IMO. It's usually better to do an explicit import.

STYLE.md
Jimbo4350: Updated.

STYLE.md
Jimbo4350: Updated.

STYLE.md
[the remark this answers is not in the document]
Jimbo4350: Updated

STYLE.md
Jimbo4350: Updated.

STYLE.md
Jimbo4350: Updated.

STYLE.md
Jimbo4350: I disagree here because it makes looking at changes to that code unsightly and deviates from the repo's overall format guidelines.

but I'd like to hear @erikd 's thoughts & @intricate 's thoughts.

STYLE.md
Jimbo4350: I think so in terms in readability. Prefixing a `Map` related function with `M.` makes sense but having `M.Map` in  a type signature doesn't add value to me. It adds noise.

STYLE.md
```diff
@@ -0,0 +1,446 @@
...
+* Don't use short names like `n`, `sk`, `f` unless their meaning is clear from
+context (function name, types, other variables, etc.).
+
+* For readability reasons, don't capitalize all letters when using an
+abbreviation.  For example, write `HttpServer` instead of
+`HTTPServer`.  Exception: two or three letter abbreviations, e.g. `IO`, `STM`.
+
+**Records name conventions**
+
+If data type has only one constructor then this data type name should be same
```
intricate: the same*

STYLE.md
Jimbo4350: Updated.

STYLE.md
```diff
@@ -0,0 +1,446 @@
...
+Choose a reasonable prefix is used such that we can ensure the field name will remain unique throughout the package. Use your discretion.
+
+e.g
+```haskell
+data MempoolEnv m blk = MempoolEnv
+  { meEnvLedger    :: LedgerInterface m blk
+  , meEnvLedgerCfg :: LedgerConfig blk
+  , meEnvStateVar  :: StrictTVar m (InternalState blk)
+  , meEnvTracer    :: Tracer m (TraceEventMempool blk)
+  }
```
intricate: If following Serge's point (https://github.com/input-output-hk/cardano-node/pull/164#discussion_r320014458), these field names would actually be `meLedger`, `meLedgerCfg`, `meStateVar`, `meTracer`, etc.

STYLE.md
intricate: Perhaps format these as bullet points.

STYLE.md
```diff
@@ -0,0 +1,446 @@
...
+
+Maximum line length is *80 characters* or *100 characters* if necessary.
+
+Modern screens have high definition and big width.
+But with some tiling managers and two terminals on one screen you are not able to
+see many characters on one line.
+On the other hand, restricting line size to a very small number like 80 leads to
+some crazy indentation despite the fact that
+shorter lines should force you to write well structured code.
+That's why *100* is a reasonable compromize.
```
intricate: compromise*

STYLE.md
```diff
@@ -0,0 +1,446 @@
...
+* The user might actually want to click on it for more information (in
+  your opinion), and
+
+* Only for the first occurrence of each API name in the comment (don't
+  bother repeating a link)
+
+Top-down guideline
+------------------
+
+### LANGUAGE extensions section
```
intricate: Maybe we should also have a section for `OPTIONS_GHC` pragma

STYLE.md
deepfire: So it's not about robustness-against-change, I see.

But there is another side -- explicit imports add a lot of bookkeeping cost, which qualified imports help avoid.

Also, which scenarios do you envision where qualified imports reduce readability?

I often find it the other way -- having many names used in a module can make the trip to the import section all too frequent.  Qualified names hint at which part of the realm the name comes from.  I sometimes reserve unqualified imports to names that have unambiguous interpretation and don't require this trip.

STYLE.md
Jimbo4350: @intricate  Feel free to suggest them :)

STYLE.md
Jimbo4350: I blame Serokell :laughing:

STYLE.md
intricate: I would note where `OPTIONS_GHC` pragmas should be located in the module (in relation to `LANGUAGE` pragmas), how they should be aligned, etc.

And yeah, at the bottom of the document we could also list GHC options to avoid.

STYLE.md
deepfire: > I disagree here because it makes looking at changes to that code unsightly

First, I think it should be fair to note that this is not a general statement.
The default, non-side-by-side diff view on Github is 116 characters wide, and it does not overflow.  The cases that blow that character budget are sufficiently rare, it shouldn't be a big problem to scroll to the right.  As an example:

Screenshot from 2019-08-30 15-26-42

Second, we should consider that the Github diff view isn't the main avenue for reading code, and especially so not for writing code.

So pessimising the rare cases that gain a lot from tabular layout by enforcing an absolute requirement on line length seems overly harsh to me.

STYLE.md
deepfire: I think there are two separate questions:

1. lack of merit of using `M.Map`
2. enforcement of #1 through `hiding` -- which has an extra bookkeeping cost attached.

It's easy to agree on 1, but the 2 requires us to recognise the cost and balance it against the benefit of turning the discouragement into prohibition.

intricate: > I think there are two separate questions:
> 
> 1. lack of merit of using `M.Map`
> 2. enforcement of #1 through `hiding` -- which has an extra bookkeeping cost attached.
> 
> It's easy to agree on 1, but the 2 requires us to recognise the cost and balance it against the benefit of turning the discouragement into prohibition.

Yeah, I don't think I'd enforce `hiding` the symbols utilized in the unqualified import. Although, perhaps, it's a valid option if the dev wishes to do so. I'm indifferent on this point.

STYLE.md
intricate: > Now, I think we could make this more helpful, if we set some guidelines about what constitutes over-use.
> 
> Otherwise it's overwhelmingly left to judgement, and judgements can differ between reasonable people.

@deepfire: Do you have any opinions as to what would constitute over-use?

STYLE.md
intricate: I'd have to think about this. I typically only use qualified imports where I think they would increase readability (which includes resolving symbol conflicts), but that's quite vague.

Jimbo4350: Reading code with a lot of qualified imports makes less readable IMO. It's difficult to define a hard line and say exactly under what conditions you should qualify an import vs explicit importation. 

I'd say use qualified imports to mostly avoid name conflicts. @erikd

STYLE.md
Jimbo4350: @deepfire I agree we don't need to hide `Map`.

STYLE.md
deepfire: There is another angle -- patch conflicts.

Since code changes so often cause changes to the set of used names, explicit import lists increases the probability of conflicts -- relative to qualified imports.

This is because changes in the set of imports from any particular module happen more often than the change in the set of used modules -- quite naturally.

Remember how many times you were resolving those conflicts around import lists..

STYLE.md
deepfire: @intricate, it's admittedly hard to put up useful (as in actionable) criteria. One thing that's actionable and definite, I think is use of composition sections -- `(foo .)`, at least I find that opaque enough.

Another one is less actionable, but rather -- if the expression is literally more complex (more elements) when point-free, then it's definitely not worth it.

Another one is when you have to use too many `flip`'s -- I'd say two is borderline too many.

But those are mostly soft criteria, admittedly.

You could say "the point-free version needs to have a clarity benefit" -- and that's easy to support, but then people have different views on clarity..

STYLE.md
erikd: The `Tx = UnsafeTx` almost certainly came from the Serokell code. There is not reason to perpretuate Serokell's insanity in the new code base. If I constructor actually is unsafe maybe it could be private to the module or some other approach taken.

STYLE.md
erikd: I do not think there should be an exception for long string literals, especially not when the start at column 40.

STYLE.md
Jimbo4350: @erikd  `UnsafeTx` exists in `cardano-ledger`: https://github.com/input-output-hk/cardano-ledger/blob/583718d76bd9952336c588f9c48828756ba83254/cardano-ledger/src/Cardano/Chain/UTxO/Tx.hs#L51

Hiding the constructor did occur to me but I wasn't sure which was better/best practice as `UnsafeTx` was in `cardano-ledger`.

STYLE.md
erikd: Specifically, I don't like `get${newtypeName}` because `get*` is so often used other places.

However, I think conventions are good and the larger the code base the higher the payoff. For the case of `newtype`s, I would suggest one of two options:

* No accessor so the user has to pattern match.
* Use `un${newtypeName}

For long lived, maintainable code the "overly long accessor names" problem is irrelevant. We should be writing code for long term readability not for being easy to write.

STYLE.md
erikd: I can't actually think of a place where I thought point free style was overused (or rather where it could not have been made perfectly fine with some better formatting).

STYLE.md
erikd: The are certain libraries that make far more sense to import qualified. For instance, `Set` and `Map` which i usually use as:
```
import           Data.Map.Strict (Map)
import qualified Data.Map.Strict as Map
import           Data.Set (Set)
import qualified Data.Set as Set
```
Once they are often used like that, I even think its best to *always* use them like that. Modules that make sense to always use qualified include the strict and lazy versions of `ByteString`, `Text` and `List`.

I also agree with @deepfire 's suggestion that changes to import lists can cause changes with little to no semantic value.

STYLE.md
erikd: Hiding a constructor that can be misused is a good idea (if possible).

STYLE.md
erikd: I would say that for:
```
import qualified Data.Map as M
```
`M` is not a great qualified name. I think `Map` would be a far better qualified name.

STYLE.md
erikd: The code @deepfire links is pretty close to a example of how I would *not* format that file.

STYLE.md
Jimbo4350: > For long lived, maintainable code the "overly long accessor names" problem is irrelevant. We >should be writing code for long term readability not for being easy to write.

:100:

STYLE.md
deepfire: I can accept that style requirement, by regarding it as a decree, even though I fail to see the balance of arguments.

Let's do it!

STYLE.md
deepfire: > We should be writing code for long term readability not for being easy to write.

One would of course agree with that, but longer names do not necessarily translate to better readability -- simply because they use more space, so less code can be seen, and so because of the reduced density the big picture could become less accessible (forest, trees, etc.).

Generally, I'd cautiously agree on the `unThing` being a balanced approach, even despite the longer names.

deepfire:
Have we explicitly addressed diff-friendliness?

We had it come up in one recent review.

STYLE.md
intricate: @erikd I would say that the same example in this document (a few lines below) is one that I would personally consider "bad" utilization of point-free:
```
f = (g .) . h
```
I find this unnecessarily hard to read.

intricate: Would have to agree with @deepfire here.

STYLE.md
deepfire: I would agree with @erikd., despite having (ab)used `M` for `Data.Map` a lot in the past.

STYLE.md
deepfire: Another question -- how do we treat local (intra-package) module imports.

Do we apply the same policy of no implicit imports for them as well?

Relevant differences that can affect balance in that case:

1. Intra-package module changes happen more often, than cross-package, making explicit import list churn even worse.
1. Since the names are local, they arguably are less prone to ambiguous interpretation.
1. People seem to rarely use qualified import lists for local modules.

STYLE.md
Jimbo4350: I will update with: "the point-free version needs to have a clarity benefit"  and resolve this. We can deal with on a case by case basis until we figure out actionable criteria.

STYLE.md
Jimbo4350: > Since code changes so often cause changes to the set of used names, explicit import lists increase the probability of conflicts. This is especially bad, because the import section is rather dense, per-line change frequency -wise.
Fair enough. Should we do the following:
* Explicit or qualified imports from external packages.
@deepfire

Jimbo4350:
> Have we explicitly addressed diff-friendliness?
> 
> We had it come up in one recent review.

I believe so. 80-100 char line width and no long running strings?

deepfire:
> I believe so. 80-100 char line width and no long running strings?

Hmm, I'm not sure that's everything.

At a minimum we want a code layout such that name change causes reasonably least line change.

I'm referring to the explanation by @erikd: https://github.com/input-output-hk/cardano-node/pull/152#discussion_r319795803

Jimbo4350:
> > I believe so. 80-100 char line width and no long running strings?
> 
> Hmm, I'm not sure that's everything.
> 
> At a minimum we want a code layout such that name change causes reasonably least line change.
> 
> I'm referring to the explanation by @erikd: [#152 (comment)](https://github.com/input-output-hk/cardano-node/pull/152#discussion_r319795803)

Does that work?
https://github.com/input-output-hk/cardano-node/pull/164/files#diff-32430379d7ba797ddf827c29a4eac4c1R33

STYLE.md
deepfire: @Jimbo4350, frankly, I am in favor of that, but I'd want consensus between all of us.

cc @erikd, @intricate.

deepfire:
@Jimbo4350, honest question -- does that phrase help motivate the change in https://github.com/input-output-hk/cardano-node/pull/152#discussion_r319647206 ?

I don't think it's the entire explanation -- or is it?

Jimbo4350:
> @Jimbo4350, question -- does that phrase help motivate the change in [#152 (comment)](https://github.com/input-output-hk/cardano-node/pull/152#discussion_r319647206) ?
> 
> I don't think it's the entire explanation -- or is it?

I was using a haskell formatter `ormolu` as a panacea before this style document. The horizontal spacing @intricate pointed out was just an eye sore. I wasn't suggesting you had to format it exactly as I specified.

deepfire:
> I wasn't suggesting you had to format it exactly as I specified in that comment.

No, no, I meant that the suggested change was:
```
- parseTxIn = option (uncurry TxInUtxo
+ parseTxIn =
+  option
+   ( uncurry TxInUtxo
```
I.e. there was two extra line-breaks that do not obviously follow from anything in this style guide -- unless I'm missing something.

Jimbo4350:
> > I wasn't suggesting you had to format it exactly as I specified in that comment.
> 
> No, no, I meant that the suggested change was:
> 
> ```
> - parseTxIn = option (uncurry TxInUtxo
> + parseTxIn =
> +   option
> +   ( uncurry TxInUtxo
> ```
> 
> I.e. there were two extra line-breaks that do not obviously follow from anything in this style guide -- unless I'm missing something.

The suggested change was before the style document. You can ignore it.

STYLE.md
erikd: I agree `f = (g .) . h` is "overused pointfree".

STYLE.md
mhuesch: @erikd agree that saying `data Tx = Tx`, not exporting the data constructor, and exporting a smart constructor would be a good approach.

STYLE.md
```diff
@@ -0,0 +1,453 @@
...
+
+Write each `LANGUAGE` pragma on its own line, sort them alphabetically and
+align by max width among them.
```
mhuesch: In the below code block, aren't we *not* aligning by width? I think that's the right approach for minimizing semantically-null patches.
```suggestion
```

STYLE.md
```diff
@@ -0,0 +1,453 @@
...
+2. You can split export list into sections or just write all as single section.
+3. It is strongly adviced to sort each section alpabetically. However,
+   classes, data types and type aliases should be written before
+   functions.
+
+### Imports
+
+Imports should be grouped in the following order:
+
+0. Implicit import of custom prelude (for example [`cardano-prelude`](https://github.com/input-output-hk/cardano-prelude)) if used.
```
mhuesch:
```suggestion
0. Import of custom prelude (for example [`cardano-prelude`](https://github.com/input-output-hk/cardano-prelude)) if used.
```
Isn't it not implicit, if it is there? e.g.
```
import Cardano.Prelude
```
https://github.com/input-output-hk/cardano-ledger/blob/677b9377d4dc541eb3d7ffac079bdab7d489df1f/cardano-ledger/src/Cardano/Chain/Epoch/File.hs#L13

STYLE.md
mhuesch: I agree with @deepfire that qualified imports can reduce the PITA of explicit imports. I feel as though I've spent a lot of cumulative hours at IOHK refactoring explicit imports, massaging them around to meet line lengths (before we had auto-formatters, or when they don't work), and generally just thinking about them when I would've rather been doing actual work.

I haven't worked on a large project that used primarily qualified imports, so I don't know if there are other problems, but it seems to me that we might benefit from using more qualified and fewer explicit imports.

STYLE.md
```diff
@@ -0,0 +1,453 @@
...
+
+### Warnings ###
+
+Code should be compilable with `-Wall` without warnings.
+
+
+### Default extensions
+
+The only allowed default extension is: `NoImplicitPrelude`
+All other extensions need to be put in the relevant module.
```
mhuesch: 💯

STYLE.md
```diff
@@ -0,0 +1,453 @@
...
+
+## Language Extensions to avoid
+
+1. Avoid `RecordWildCards`, use `NamedFieldPuns` instead. `RecordWildCards` makes it difficult to discern where record accessor functions come from, _especially_ when used with external libraries. This results in time wasted figuring out where things come from. `NamedFieldPuns` forces you to be more explicit and is a better alternative.
```
mhuesch:
```suggestion
1. Avoid `RecordWildCards` - use `NamedFieldPuns` instead. `RecordWildCards` makes it difficult to discern where record accessor functions come from, _especially_ when used with external libraries. This results in time wasted figuring out where things come from. `NamedFieldPuns` forces you to be more explicit and is a better alternative.
```

STYLE.md
Jimbo4350: Good catch, I'll fix.

STYLE.md
```diff
@@ -0,0 +1,453 @@
...
+> This style guide's aims are code beauty, readability and understandability.
+
+
+Principles
+----------
+* The primary goal is readability/maintainability with performance as a close second.
+* Always use types to clarify intent and make invalid states unrepresentable.
+* Use type signatures as compiler checked documentation.
+* Break up large functions into smaller sub functions.
+* When modifying or adding code to existing code (either at the file, project or company level, try to match the style of the existing code.
```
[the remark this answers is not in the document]
Jimbo4350: Change made.

STYLE.md
```diff
@@ -0,0 +1,453 @@
...
+Some examples:
+
+```haskell
+sayHello :: IO ()
+sayHello = do
+    name <- getLine
+    putStrLn $ greeting name
+  where
+    greeting :: String -> String
+    greeting name = "Hello, " ++ name ++ "!"
```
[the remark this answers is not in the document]
Jimbo4350: I wasn't aware you could do that. I prefer what you suggested.

STYLE.md
Jimbo4350: :100:

STYLE.md
Jimbo4350: As we aren't using an auto-formatter I also agree here. I was spoiled by them and therefore favoured explicit imports. I'll mark this as resolved :).

STYLE.md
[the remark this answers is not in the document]
mhuesch: Cool! 😎

CodiePP:
would this need to be discussed with a wider audience? I think there are quite some devs involved. And, most still adhere to the CSL style guide. however this one is outdated, is this somewhat compatible or in which parts does it deviate from it?

Jimbo4350:
> would this need to be discussed with a wider audience? I think there are quite some devs involved. And, most still adhere to the CSL style guide. however this one is outdated, is this somewhat compatible or in which parts does it deviate from it?

It deviates somewhat. If there are other devs that want to give their input tag them here :slightly_smiling_face: I will hold off from merging if they have requests/changes.

erikd:
> would this need to be discussed with a wider audience? 

Since this document is about this repo, I would think not. Jordan is the head honcho 'round here!

Jimbo4350:
Further suggestions/changes can be done via future PRs if anyone has objections or requests @CodiePP

CodiePP:
I find this a positive development overall. It is good to have a style guide.
Two points: in CSL we have had a tabwidth of 4 spaces and some still use this. Indenting by only two spaces makes the code quite dense and, IMO, harder to read. And, I would encourage to put `-Werror` in there as well.

STYLE.md
```diff
@@ -0,0 +1,453 @@
...
+### Blank Lines
+
+* One blank line between top-level definitions.
+* No blank lines between type signatures and function definitions.
+* Add one blank line between
+functions in a type class instance declaration if the function bodies
+are large.
+* You can add blank lines inside a big `do` block to separate logical
+parts of it.
+* You can also use blank lines to separate definitions inside `where`
```
deepfire: @Jimbo4350, this conflicts with https://github.com/input-output-hk/cardano-node/pull/164/files#diff-32430379d7ba797ddf827c29a4eac4c1R38, which mandates the blank line separation.

STYLE.md
```diff
@@ -0,0 +1,453 @@
...
+punctuation.
+
+#### Top-Level Definitions
+
+* Comment top level functions (particularly exported functions)
+as much as possible. If it is not blindingly obvious what a function
+is doing, you should add a comment.
+* Always provide a type signature
+* Use Haddock syntax in the comments.
+* Comment every exported data type.
```
deepfire: Even the criteria of blinding obviousness applies here as well?  Say a newtype around a `FilePath`.

If it literally means a `data` ADT, then we perhaps want to clarify that, @Jimbo4350.

STYLE.md
```diff
@@ -0,0 +1,453 @@
...
+  ) where
+```
+
+Some clarifications:
+
+1. Use 2 spaces indentation for export list.
+2. You can split export list into sections or just write all as single section.
+3. It is strongly adviced to sort each section alpabetically. However,
+   classes, data types and type aliases should be written before
+   functions.
```
deepfire: I think there's another angle to the story:

1. The export list could reflect the order of definitions, so as to give reader an expectation of where they shall find what they seek for.
2. The definition order often has its own logic.

STYLE.md
deepfire: @erikd, what is your opinion on the 1-blank-line mandatory separation of where clauses?

erikd: I think hard rules don't make a lot of sense. In some cases a single blank line does make sense (eg to separate two 10 line functions, with type signatures) and in other cases it doesn't (eg 3 single line clauses without type sigs).

Sorry for not giving a straight yes/not answer :)

STYLE.md
erikd: I personally like the alphabetical ordering, grouped as types followed by functions and then optionally functions exported purely for testing (with a comment).

I do however recognize that as purely my own preference that I would have difficulty justifying over the alternatives.

STYLE.md
erikd: Yeah, I do not comment newtype wrappers around `FilePath` and other obvious one.

This is an example of where hard rules like "all types need comments" don't make much sense.

Jimbo4350:
> in CSL we have had a tabwidth of 4 spaces and some still use this. Indenting by only two spaces makes the code quite dense and, IMO, harder to read.
I remember in CSL this resulted in functions that took up a lot more vertical space (every newline was shifted 4 spaces over). This made respecting the 80-100 char width limit more difficult and led to arguably less readable code. We are trying to reduce this with the 2 space indent.
> And, I would encourage to put `-Werror` in there as well.
Done

STYLE.md
Jimbo4350: I'll remove it :+1:

STYLE.md
Jimbo4350: Will relax the wording here too.

STYLE.md
Jimbo4350: > I personally like the alphabetical ordering, grouped as types followed by functions and then optionally functions exported purely for testing (with a comment).

I also like this, @deepfire  what do you think?

deepfire: There is yet another angle -- sometimes module exports several themes of entities, that:
1. Related enough to come together in a single module,
2. Yet distinct enough to merit their own grouping in the export list.

Like, for example in: https://github.com/input-output-hk/cardano-node/blob/b783a886d3afc76326165dbaeb54e671d3feab48/cardano-node/src/Cardano/CLI/Key.hs#L7

Sometimes those subgroups are crystallising towards a separation into a module of their own.
In this case having the group clearly visible helps towards that goal.

It's absolutely worth keeping each group sorted alphabetically, with types coming first, though!

What do you think, @erikd, @Jimbo4350, @intricate  ?

Jimbo4350: I've never felt like I'd gained much from the grouping of exports. I can see the value of dividing the exports if some are used only in testing. If I want to know what a function does (or group of functions), I'll look at said function & comment to go along with it. This is strictly a preference however.

I would personally opt for:
- Alphabetical ordering, grouped as types followed by functions and then optionally functions exported purely for testing (with a comment).

intricate: @deepfire: I don't mind the grouping of exports. I rarely do so myself but I've seen a few places where it was nice (no examples, unfortunately.. can't remember right now).

STYLE.md
```diff
@@ -0,0 +1,453 @@
...
+All other extensions need to be put in the relevant module.
+
+## Language Extensions to avoid
+
+1. Avoid `RecordWildCards` - use `NamedFieldPuns` instead. `RecordWildCards` makes it difficult to discern where record accessor functions come from, _especially_ when used with external libraries. This results in time wasted searching for the accessors. `NamedFieldPuns` forces you to be more explicit and is a better alternative.
+
+
+2. Avoid `ViewPatterns` - `ViewPatterns` obfuscates what could have been readable code and incurs more cognitive overhead.
+
+## OPTIONS_GHC pragma to avoid
```
intricate: Perhaps just "GHC options to avoid".

STYLE.md
intricate: >Sometimes those subgroups are crystallising towards a separation into a module of their own.
In this case having the group clearly visible helps towards that goal.

This seems like a fair point. I wouldn't be opposed to this type of grouping, then. However, I probably wouldn't strictly enforce it.

PR #166, opened 2019-09-03
module reorganisation + cardano-cli repair

deepfire:
This is a second take on PR originally by @Jimbo4350 (thanks for the prodding Jordan, that was definitely long overdue!), but one that preserves the changes done in https://github.com/input-output-hk/cardano-node/pull/152.

Note that this PR is strictly a combination of `7e2b284f7dd13a813d65b18ce20e7a86d73ddf4b` and code motion, no functionality is affected.

Summary of code motion:

- subset of `Cardano.CLI.Run` dispersed to:
  - `Cardano.CLI.Genesis`
  - `Cardano.CLI.Key`
  - `Cardano.CLI.Tx`
- `Cardano.Legacy.Byron` -> `Cardano.CLI.Legacy.Byron`
- `Cardano.Node.Orphans` -> `Cardano.Common.Orphans`
- `Cardano.Node.Parsers` -> `Cardano.Common.Parsers`
- `Cardano.Node.TxSubmission` ->
  -  `Cardano.CLI.Tx.Submission`
  -  `Cardano.Common.LocalSocket`

I've contemplated moving `Cardano.Node.Configuration` to either:
- `Cardano.Common.Configuration` (because it's indeed common)
- `Cardano.Configuration` (because the main type is `CardanoConfiguration`)
..but decided it'd be too invasive for now.

***NOTE:*** *This PR is about code motion, so please do not judge it one the merits of the code being moved.  The improvements should come separately.*


cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -110,8 +110,6 @@ data ViewMode =
 runNode :: NodeCLIArguments -> LoggingLayer -> CardanoConfiguration -> IO ()
 runNode nodeCli@NodeCLIArguments{..} loggingLayer cc = do
     let !tr = llAppendName loggingLayer "node" (llBasicTrace loggingLayer)
-    -- If the user asked to submit a transaction, we don't have to spin up a
-    -- full node, we simply transmit it and exit.
```
deepfire: Ok, that's not code motion, I admit : -)

But that comment was outdated.

cardano-node/src/Cardano/CLI/Genesis.hs
```diff
@@ -0,0 +1,71 @@
+{-# LANGUAGE CPP                        #-}
+{-# LANGUAGE RecordWildCards            #-}
+
+module Cardano.CLI.Genesis
+  ( dumpGenesis
+  )
+where
```
Jimbo4350: I don't think a single function `dumpGenesis` should have its own module. I would put this in `Cardano.CLI.Key` and if possible change the name of the module to encompass this (not strictly necessary for now).

cardano-node/src/Cardano/CLI/Genesis.hs
```diff
@@ -0,0 +1,71 @@
...
+dumpGenesis CLIOps{..} outDir genesisData GeneratedSecrets{..} = do
+  exists <- doesPathExist outDir
+  if exists
+    then throwIO $ OutputMustNotAlreadyExist outDir
+    else createDirectory outDir
+
+  let genesisJSONFile = outDir <> "/genesis.json"
+  LB.writeFile genesisJSONFile =<< coSerialiseGenesis genesisData
+
+  let dlgCertMap = unGenesisDelegation $ gdHeavyDelegation genesisData
```
Jimbo4350: This needs a type signature.

cardano-node/src/Cardano/CLI/Genesis.hs
```diff
@@ -0,0 +1,71 @@
...
+      findDelegateCert :: SigningKey -> IO Certificate
+      findDelegateCert sk =
+        case flip find (Map.elems dlgCertMap) . isCertForSK $ sk of
+          Nothing -> throwIO $ NoGenesisDelegationForKey $ prettySigningKeyPub sk
+          Just x  -> pure x
+      wOut :: String -> String -> (a -> IO LB.ByteString) -> [a] -> IO ()
+      wOut = writeSecrets outDir
+  dlgCerts <- mapM findDelegateCert gsRichSecrets
+
+  wOut "genesis-keys"    "key"  coSerialiseGenesisKey     gsDlgIssuersSecrets
```
Jimbo4350: No vertical alignment :slightly_smiling_face:

cardano-node/src/Cardano/CLI/Genesis.hs
```diff
@@ -0,0 +1,71 @@
...
+  wOut "delegate-keys"   "key"  coSerialiseDelegateKey    gsRichSecrets
+  wOut "poor-keys"       "key"  coSerialisePoorKey        gsPoorSecrets
+  wOut "delegation-cert" "json" coSerialiseDelegationCert dlgCerts
+  wOut "avvm-seed"       "seed" (pure . LB.fromStrict)     gsFakeAvvmSeeds
+
+writeSecrets :: FilePath -> String -> String -> (a -> IO LB.ByteString) -> [a] -> IO ()
+writeSecrets outDir prefix suffix secretOp xs =
+  forM_ (zip xs $ [0::Int ..]) $
+  \(secret, nr)-> do
```
Jimbo4350: What is nr?

cardano-node/src/Cardano/CLI/Tx.hs
```diff
@@ -0,0 +1,114 @@
+{-# OPTIONS_GHC -Wno-all-missed-specialisations #-}
+
+module Cardano.CLI.Tx
```
Jimbo4350: Like wise I also think `Cardano.CLI.Tx` and `Cardano.CLI.Tx.Submission` should be condensed to one module. Having a module for a single function is overkill IMO.

cardano-node/src/Cardano/CLI/Tx.hs
```diff
@@ -0,0 +1,114 @@
...
+  <> F.format CC.Common.addressDetailedF addr
+
+txSpendUTxOByronPBFT
+  :: CC.Genesis.Config
+  -> SigningKey
+  -> NonEmpty TxIn
+  -> NonEmpty TxOut
+  -> GenTx (ByronBlockOrEBB ByronConfig)
+txSpendUTxOByronPBFT gc sk ins outs =
+  let txattrs  = mkAttributes ()
```
Jimbo4350: No vertical alignment.

cardano-node/src/Cardano/CLI/Tx.hs
```diff
@@ -0,0 +1,114 @@
...
+  let vk         = Crypto.toVerification genSk
+      handleMissingAddr :: Maybe CC.UTxO.TxIn -> CC.UTxO.TxIn
+      handleMissingAddr  = fromMaybe . error
+        $  "\nGenesis UTxO has no address\n"
+        <> (T.unpack $ prettyAddress genAddr)
+        <> "\n\nIt has the following, though:\n\n"
+        <> Cardano.Prelude.concat (T.unpack . prettyAddress <$> Map.keys initialUtxo)
+
+      initialUtxo :: Map Address (CC.UTxO.TxIn, CC.UTxO.TxOut)
+      initialUtxo =
```
Jimbo4350: There is a lot going on in this function. It may be a good idea to factor out these let bindings into separate functions with comments.

cardano-node/src/Cardano/CLI/Tx/Submission.hs
```diff
@@ -8,18 +8,15 @@
 
 {-# OPTIONS_GHC -Wno-all-missed-specialisations #-}
 
-module Cardano.Node.TxSubmission (
+module Cardano.CLI.Tx.Submission (
```
Jimbo4350: See comment above about merging this with `Cardano.CLI.TX`.

cardano-node/app/chairman.hs
```diff
@@ -20,7 +20,7 @@ import           Ouroboros.Consensus.NodeId (CoreNodeId)
 import           Cardano.Config.CommonCLI
 import           Cardano.Common.Protocol (Protocol, SomeProtocol(..), fromProtocol)
 import           Cardano.Config.Types (CardanoConfiguration(..))
-import           Cardano.Node.Parsers (parseCoreNodeId, parseProtocol)
+import           Cardano.Common.Parsers (parseCoreNodeId, parseProtocol)
```
Jimbo4350: :+1:

cardano-node/src/Cardano/CLI/Key.hs
```diff
@@ -0,0 +1,71 @@
+{-# LANGUAGE CPP                        #-}
+{-# LANGUAGE GeneralisedNewtypeDeriving #-}
+{-# LANGUAGE RecordWildCards            #-}
+{-# LANGUAGE ViewPatterns               #-}
+{-# OPTIONS_GHC -Wno-all-missed-specialisations #-}
+
+module Cardano.CLI.Key
+  ( SigningKeyFile(..)
```
Jimbo4350: :+1:

cardano-node/src/Cardano/CLI/Tx.hs
```diff
@@ -0,0 +1,114 @@
...
+  in mkByronTx $ ATxAux (reAnnotate atx) (reAnnotate awit)
+
+txSpendGenesisUTxOByronPBFT
+  :: CC.Genesis.Config
+  -> SigningKey
+  -> Address
+  -> NonEmpty TxOut
+  -> GenTx (ByronBlockOrEBB ByronConfig)
+txSpendGenesisUTxOByronPBFT gc sk genAddr outs =
+  let txattrs  = mkAttributes ()
```
Jimbo4350: No vertical alignment.

cardano-node/src/Cardano/Common/Orphans.hs
```diff
@@ -4,7 +4,7 @@
 {-# OPTIONS_GHC -Wno-all-missed-specialisations #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 
-module Cardano.Node.Orphans
+module Cardano.Common.Orphans
```
Jimbo4350: :+1:

cardano-node/src/Cardano/CLI/Genesis.hs
deepfire: This is a code motion PR, Jordan!  It's intended to be quick, so it doesn't block others.

cardano-node/src/Cardano/CLI/Genesis.hs
deepfire: Again, this is a code motion PR!

cardano-node/src/Cardano/CLI/Genesis.hs
deepfire: This is a code motion PR!

cardano-node/src/Cardano/CLI/Tx.hs
deepfire: We are going to have `Cardano.Tx.Generator` soon -- this is a preparation for that.

cardano-node/src/Cardano/CLI/Tx/Submission.hs
deepfire: Again, see above response.

cardano-node/src/Cardano/CLI/Tx.hs
deepfire: This is a code motion PR.

deepfire:
@Jimbo4350, please consider the PR description that says:

> NOTE: This PR is about code motion, so please do not judge it one the merits of the code being moved. The improvements should come separately.

cardano-node/src/Cardano/CLI/Tx.hs
deepfire: This is a code motion PR.

cardano-node/src/Cardano/CLI/Tx.hs
deepfire: This is a code motion PR.

cardano-node/src/Cardano/CLI/Genesis.hs
deepfire: Two things to consider:

1. I foresee the `Cardano.CLI.Key` module to expand in future, so it should retain its name, I think.
2. the `dumpGenesis` has its local attendant `writeSecrets` function, so it's not traveling alone
3. I would like to factor the genesis generation into a function as well, so the cutoff point in the `Genesis` subcommand will change -- but again, this PR isn't about factoring -- it's pure simple code motion.

Given this, I don't think it's reasonable to do the module merge.

deepfire:
> Also the commit name global: code motion is very ambiguous to me.

This is moving code globally, between modules.  I think that description is fairly on the spot.

But also, do you have a better suggestion?

cardano-node/src/Cardano/CLI/Genesis.hs
Jimbo4350: Re: 1. Fore-seeing the expansion of a module is not a good reason to create one. If there are multiple functions that make sense to put in a separate module, then you should create that module. I'm green lighting this PR for now to see where you are going with this. But as a fair warning, later on I may make changes of my own to this structure.

deepfire: First, I don't think that the language of "warnings" (fair or not) is helpful.

I find the language of rationales much less conflict-ful and much more useful.

Next, the export is a single function, indeed.  But so what?  Modules abstract over code.
What matters is that there are two functions in the module, one hidden, and there will be more, fairly soon, I hope.

If we want prior art -- there is a number of modules in `cardano-ledger` that export a single data type and nothing else.  `Cardano.Chain.UTxO.GenesisUTxO`, `Test.Cardano.Chain.MempoolPayload.Gen`, `Test.Cardano.Chain.Config` and `Test.Cardano.Chain.Delegation.Example` export a single function (that isn't called `tests`).

deepfire:
Waiting for review by @intricate.

cardano-node/src/Cardano/CLI/Genesis.hs
Jimbo4350: Adding a module for a single function adds unnecessary clutter to the directory tree. I'm not sure how else to explain that.

deepfire:
Agreed with Luke to merge now.

deepfire:
BORS is stuck for some unclear reason, and I don't think it's necessary at this point, merging manually.

cardano-node/src/Cardano/CLI/Genesis.hs
deepfire: It is correct that there is also the FS hierarchy resource that we expend, so that's a cost.

But also important is what we buy with that cost -- and we buy separation of concerns, abstraction -- so we can say "for this bunch of code all the outside world needs to care about is this name".

What are some other benefits?

1. Import statements are topically localised (i.e. import lists from different aspects are not inter-mixed)
1. Locality of code is underscored -- i.e. you can see more clearly the role of ancillary functions, such as `writeSecrets`.
1. We have a ready landing place for future related code.
1. We don't have to come up with painfully contrived or non-descriptive names for modules that combine more than one aspect (i.e. what would you call a module that would cover `CLI.Key` and `CLI.Genesis`?) -- such names are often harder to remember & associate with their content in one's head..
1. Less conflicts.

Yes there are costs.  But it is important to see the balance.

PR #168, opened 2019-09-04
control tracing with '--tracing-off': turn off globally

CodiePP:
CLI flag '--tracing-off' to turn off tracing globally.
(this has been extracted from an earlier PR)


CodiePP:
bors +

PR #169, opened 2019-09-04
cardano-cli:  improve the help facilities and the README

deepfire:
1. Arg parsing: make the mode less opaque by switching to an option, hide currently unused system versions.  Update scripts to the changes.
2. Improve the README.

### DIFF
--- a/cardano-node/README.md
+++ b/cardano-node/README.md
@@ -81,7 +81,7 @@ A CLI utility to support a variety of key material operations (genesis, migratio
 
 The general synopsis is as follows:
  ```
-   cardano-cli SYSTEMVER COMMAND
+   Usage: cardano-cli (--byron-legacy | --real-pbft) (GENESISCMD | KEYCMD | DELEGCMD | TXCMD)
 ```
 
 NOTE: the exact invocation command depends on the environment.  If you have only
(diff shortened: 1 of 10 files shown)


cardano-node/README.md
```diff
@@ -116,25 +114,36 @@ Available options:
...
 
-The canned `scripts/genesis.sh` example provides a nice set of defaults and
-illustrates available options.  Running it will produce a `./genesis.${systemStart}` directory,
-where `${systemStart}` will be chosen 15 minutes in the future.
+### Generation
+
+The genesis generation operations will create a directory that contains:
 
-The directory will have the following content:
+  - the genesis JSON file itself (`genesis.json`)
```
Jimbo4350: Can we document each of the items it creates? e.g Which keys are generated and what do they do etc. Doesn't have to be long. You could link to other documentation that explains this if possible. I'm not sure if this is up to date: https://cardanodocs.com/

cardano-node/src/Cardano/Common/Parsers.hs
```diff
@@ -46,19 +47,31 @@ parseNodeId desc =
...
   , fl MockPBFT "mock-pbft"
-    "Use the Permissive BFT consensus algorithm using a mock ledger"
+    "Permissive BFT consensus with a mock ledger"
 
   , fl RealPBFT "real-pbft"
-    "Use the Permissive BFT consensus algorithm using the real ledger"
+    "Permissive BFT consensus with a real ledger"
+  ]
+  where
+    fl :: forall a. a -> String -> String -> Parser a
```
Jimbo4350: `fl` is duplicated below. Factoring out with a comment would be nice :).

intricate: @deepfire: What does `fl` stand for?

deepfire: Done.

cardano-node/README.md
deepfire: Done.

PR #176, opened 2019-09-06
Implement `ProtocolExceptions`

Jimbo4350:
Implement `ProtocolExceptions` & Replace `throw` with `throwIO` in `Cardano.Common.Protocol.hs`


cardano-node/src/Cardano/Common/LocalSocket.hs
```diff
@@ -23,3 +24,7 @@ localSocketAddrInfo socketPath =
       Socket.defaultProtocol
       (Socket.SockAddrUnix socketPath)
       Nothing
+
+eitherThrow :: Exception e => (b -> e) -> Either b a ->  IO a
```
Jimbo4350: @deepfire I'm wondering if to put this in `cardano-prelude` or to make this module more generic.

Jimbo4350:
Rather than extending exception throwing, we opt to use `Either`/`ExceptT`

PR #178, opened 2019-09-06
Take socket filepath instead of socket directory

Jimbo4350:
You must specify the filepath of the socket. Previously the filenames were automatically created. (change requested by @disassembler )

### DIFF
--- a/scripts/shelley-testnet2.sh
+++ b/scripts/shelley-testnet2.sh
@@ -19,19 +19,9 @@ genesis_hash="$(cabal new-exec -- cardano-cli --real-pbft print-genesis-hash --g
 
 ALGO="--real-pbft"
 NOW=`date "+%Y-%m-%d 00:00:00"`
-NETARGS=(
-        --slot-duration 2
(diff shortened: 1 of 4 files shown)


scripts/shelley-testnet2.sh
```diff
@@ -19,19 +19,45 @@ genesis_hash="$(cabal new-exec -- cardano-cli --real-pbft print-genesis-hash --g
 
 ALGO="--real-pbft"
 NOW=`date "+%Y-%m-%d 00:00:00"`
-NETARGS=(
+NETARGS1=(
```
deepfire: @Jimbo4350, can we factor the common part & concatenate it with the desired variable part?  because it's sligtly unwieldy now..

deepfire:
@Jimbo4350, good change, just one comment!

scripts/shelley-testnet2.sh
Jimbo4350: Updated!

PR #179, opened 2019-09-09
Use cabal new-run instead of new-exec in scripts

coot:
`cabal new-run` rebuilds the binary if needed, this is much better
default than `new-exec`.


CodiePP:
Andreas is not available atm; added some other reviewers.

coot:
Yes, the problem was that `new-run` outputs not just the command output (even with `--verbose 0` flag).  It should work now.

PR #183, opened 2019-09-12
Fix `readFile` in `Cardano.Node.Configuration.Topology`

Jimbo4350:
1. We should catch and render `IOException`s that may be thrown from the use of `readFile`. There are other `readFile`s in the repo that I will get to when I start removing exception throwing.
2. Style changes.

### DIFF
--- a/cardano-node/src/Cardano/Node/Configuration/Topology.hs
+++ b/cardano-node/src/Cardano/Node/Configuration/Topology.hs
@@ -1,30 +1,31 @@
-{-# LANGUAGE BangPatterns      #-}
-{-# LANGUAGE NamedFieldPuns    #-}
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE TemplateHaskell   #-}
+{-# LANGUAGE TemplateHaskell #-}
 
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 {-# OPTIONS_GHC -Wno-all-missed-specialisations #-}
 
 module Cardano.Node.Configuration.Topology
   ( TopologyInfo(..)
-  , NodeAddress(..)
-  , nodeAddressToSockAddr
-  , nodeAddressInfo
-  , RemoteAddress(..)
-  , remoteAddressToNodeAddress
-  , NodeSetup(..)
   , NetworkTopology(..)
+  , NodeAddress(..)
+  , NodeSetup(..)
+  , RemoteAddress(..)
+  , nodeAddressInfo
+  , nodeAddressToSockAddr
   , readTopologyFile
+  , remoteAddressToNodeAddress
   )
 where
 
 import           Cardano.Prelude hiding (toS)
 import           Prelude (String, read)
 
+import           Control.Exception (IOException)
+import qualified Control.Exception as Exception
 import           Data.Aeson
 import           Data.Aeson.TH
-import qualified Data.ByteString as B
+import qualified Data.ByteString as BS
 import qualified Data.IP as IP
 import           Data.String.Conv (toS)
 import           Text.Read (readMaybe)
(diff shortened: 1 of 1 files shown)


cardano-node/src/Cardano/Node/Configuration/Topology.hs
```diff
@@ -35,98 +36,83 @@ import           Ouroboros.Consensus.Util.Condense (Condense (..))
...
-      , addrFamily     = case naHostAddress of
-                          IP.IPv4{} -> AF_INET
-                          IP.IPv6{} -> AF_INET6
-      , addrSocketType = Stream
-      , addrProtocol   = addrProtocol defaultHints
-      , addrAddress    = nodeAddressToSockAddr na
-      , addrCanonName  = Nothing
-    }
+nodeAddressInfo na@(NodeAddress hostAddr _) =
+  AddrInfo { addrFlags = addrFlags defaultHints
```
deepfire: I wonder if this way of formatting things is prone to reindentation, if `AddrInfo` is ever renamed -- although in this particular case that seems unlikely.

cardano-node/src/Cardano/Node/Configuration/Topology.hs
```diff
@@ -35,98 +36,83 @@ import           Ouroboros.Consensus.Util.Condense (Condense (..))
 
 -- | A data structure bundling together a node identifier and the path to
 -- the topology file.
-data TopologyInfo = TopologyInfo {
-    node         :: NodeId
-  , topologyFile :: FilePath
-  }
+data TopologyInfo = TopologyInfo {node :: NodeId, topologyFile :: FilePath}
```
deepfire: I think this contradicts https://github.com/input-output-hk/cardano-node/blob/master/STYLE.md#data-declarations

```
Format records as follows:

data Person = Person
  { firstName :: !String  -- ^ First name
  , lastName :: !String  -- ^ Last name
  , age :: !Int     -- ^ Age
  } deriving (Eq, Show)
```

cardano-node/src/Cardano/Node/Configuration/Topology.hs
```diff
@@ -35,98 +36,83 @@ import           Ouroboros.Consensus.Util.Condense (Condense (..))
...
 
 -- | IPv4 address with a port number.
---
-data NodeAddress = NodeAddress {
-      naHostAddress :: !IP.IP
-    , naPort        :: !PortNumber
-    }
+data NodeAddress = NodeAddress {naHostAddress :: !IP.IP, naPort :: !PortNumber}
```
deepfire: Same as above.

cardano-node/src/Cardano/Node/Configuration/Topology.hs
```diff
@@ -35,98 +36,83 @@ import           Ouroboros.Consensus.Util.Condense (Condense (..))
...
+remoteAddressToNodeAddress:: RemoteAddress-> Maybe NodeAddress
+remoteAddressToNodeAddress (RemoteAddress addrStr port val) =
+    case readMaybe addrStr of
       Nothing -> Nothing
-      Just naHostAddress | raValency /= 0 -> Just (NodeAddress {
-                                                      naHostAddress
-                                                    , naPort = raPort
-                                                    })
-                         | otherwise      -> Nothing
+      Just addr -> if val /= 0
```
deepfire: I think this somewhat contradicts https://github.com/input-output-hk/cardano-node/blob/master/STYLE.md#if-then-else-clauses

> Generally, guards and pattern matches should be preferred over if-then-else clauses, where possible.

cardano-node/src/Cardano/Node/Configuration/Topology.hs
```diff
@@ -35,98 +36,83 @@ import           Ouroboros.Consensus.Util.Condense (Condense (..))
...
 
 instance FromJSON RemoteAddress where
-    parseJSON = withObject "RemoteAddress" $ \v -> RemoteAddress
-      <$> v .: "addr"
-      <*> ((fromIntegral :: Int -> PortNumber) <$> v .: "port")
-      <*> (v .: "valency")
+  parseJSON = withObject "RemoteAddress" $ \v ->
+    RemoteAddress <$> v .: "addr"
+                  <*> ((fromIntegral :: Int -> PortNumber) <$> v .: "port")
+                  <*> (v .: "valency")
```
deepfire: I think we were preferring this style:
```suggestion
RemoteAddress
  <$> v .: "addr"
￼  <*> ((fromIntegral :: Int -> PortNumber) <$> v .: "port")
￼  <*> (v .: "valency")
```

cardano-node/src/Cardano/Node/Configuration/Topology.hs
Jimbo4350: Not sure how likely it will be renamed but, putting the fields on a newline is an easy fix :slightly_smiling_face:

cardano-node/src/Cardano/Node/Configuration/Topology.hs
```diff
@@ -35,98 +36,83 @@ import           Ouroboros.Consensus.Util.Condense (Condense (..))
...
 -- non zero return corresponding NodeAddress.
 --
-remoteAddressToNodeAddress
-  :: RemoteAddress
-  -> Maybe NodeAddress
-remoteAddressToNodeAddress RemoteAddress {raAddress, raPort, raValency} =
-    case readMaybe raAddress of
+remoteAddressToNodeAddress:: RemoteAddress-> Maybe NodeAddress
+remoteAddressToNodeAddress (RemoteAddress addrStr port val) =
+    case readMaybe addrStr of
```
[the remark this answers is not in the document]
Jimbo4350: Oversight on my behalf, thanks.

cardano-node/src/Cardano/Node/Configuration/Topology.hs
Jimbo4350: Hmm...I genuinely felt like the `if then else` was more readable than the previous guards. Guards make more sense when there are more than 2 possible branches. I can revert it and we can take the discussion on the style doc, however.

deepfire: It was better than the previous guards, indeed, but what about:
```
  case readMaybe addrStr of
￼	    Nothing -> Nothing
￼	    Just addr -> if val /= 0
￼	                   then Just $ NodeAddress addr port
￼	                   else Nothing
```
->
```
  case readMaybe addrStr of
    Nothing -> Nothing
    Just addr | val /= 0  -> Just $ NodeAddress addr port
              | otherwise -> Nothing
```

deepfire: Or maybe that just asks for a `fmap` with the conditional factored into a `where` binding?

Jimbo4350: Going to go with:
```
case readMaybe addrStr of
    Nothing -> Nothing
￼   Just addr -> if val /= 0
                 then Just $ NodeAddress addr port
￼                else Nothing
```

deepfire: ok

PR #186, opened 2019-09-13
Update `scripts/shelley-testnet.sh`

Jimbo4350:
1. Update  `scripts/shelley-testnet.sh` to use the now mandatory socket path.
2. Use `cabal new-run` in `scripts/shelley-testnet2.sh`.


deepfire:
Pushed a couple of commits into a sister branch: https://github.com/input-output-hk/cardano-node/tree/serge/fix-scripts-socketpath

With these changes Tx submission works again.

deepfire:
@Jimbo4350, do we want to preserve separation between the Haskell commit and shell ones?

PR #187, opened 2019-09-13
update for ouroboros-network

avieth:


cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -111,8 +111,8 @@ data TraceOptions = TraceOptions
   , traceDnsResolver     :: !Bool
   }
 
-type ConsensusTraceOptions = Consensus.Tracers' () ()    (Const Bool)
-type ProtocolTraceOptions  = ProtocolTracers'   () () () (Const Bool)
+type ConsensusTraceOptions = Consensus.Tracers' () ()    () (Const Bool)
```
Jimbo4350: Any reason for the spaces here?

avieth: I thought people here liked this kind of alignment. I don't really care either way.

PR #188, opened 2019-09-13
Refactor `runCommand` with `ExceptT`

Jimbo4350:
Why `ExceptT`?
- When refactoring the code that throws exceptions, exception handling can be accidentally refactored out.
- We should be using `Either` or `ExceptT` instead of throwing exceptions in plain IO. This is more explicit.

### DIFF
--- a/cardano-node/src/Cardano/CLI/Genesis.hs
+++ b/cardano-node/src/Cardano/CLI/Genesis.hs
@@ -17,6 +17,8 @@ import           Prelude (String)
 import           Cardano.Prelude hiding (option, show, trace)
 import           Test.Cardano.Prelude (canonicalDecodePretty)
 
+import           Control.Monad.Trans.Except (ExceptT)
+import           Control.Monad.Trans.Except.Extra (firstExceptT, left, right)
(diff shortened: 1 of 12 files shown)


cardano-node/src/Cardano/CLI/Genesis.hs
```diff
@@ -130,34 +128,39 @@ dumpGenesis
...
-  wOut "delegation-cert" "json" (coSerialiseDelegationCert co) dlgCerts
-  wOut "avvm-seed" "seed" (pure . LB.fromStrict) (gsFakeAvvmSeeds gs)
+  then left $ OutputMustNotAlreadyExist outDir
+  else liftIO $ createDirectory outDir
+
+  liftIO $ LB.writeFile genesisJSONFile =<< coSerialiseGenesis co genesisData
+
+  dlgCerts <- mapM findDelegateCert $ gsRichSecrets gs
+
+  liftIO $ wOut "genesis-keys" "key" (coSerialiseGenesisKey co) (gsDlgIssuersSecrets gs)
```
deepfire: I think that `wOut` itself can be lifted to `ExceptT`.

PR #191, opened 2019-09-16
update consensus for access to chain sync tracers

CodiePP:
the tracing of chain sync structures depends on a newer version of the consensus layer, and had a recursive call to itself in the rendering.

the rendering was wrong: `toObject verb ev = toObject verb ev` will obviously loop
(sorry for that! :-) )

----
the rendering of the server events depends on this PR: https://github.com/input-output-hk/ouroboros-network/pull/1032
(if this is fast we can include it here)

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -91,37 +91,37 @@ source-repository-package
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/ouroboros-network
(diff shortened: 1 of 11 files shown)


erikd:
Going to use my admin super powers to merge this because bors is borked.

PR #192, opened 2019-09-16
Cabal.project <-> stack.yaml correspondence

deepfire:
`cabal.project` and `stack.yaml` keep going out of sync.

This PR does:

1. restore their correspondence
2. reformats them, so that it's easier to maintain the correspondence manually.

The second part can be controversial, so let's discuss!


stack.yaml
```diff
@@ -9,7 +9,30 @@ ghc-options:
...
+
+  - git: https://github.com/input-output-hk/cardano-base
+    commit: 0e26a4f00324100b4c34f2a9c7606ae03e94d7a2
+    subdirs:
+      - binary
+      - binary/test
+      - cardano-crypto-class
+
+
+
```
erikd: What is all this white space?

stack.yaml
```diff
@@ -9,7 +9,30 @@ ghc-options:
...
+
+
+
+
+
+  # - git: https://github.com/input-output-hk/cardano-crypto
+  #   commit: 4590efa638397e952a51a8994b5543e4ea3c1ecd
```
erikd: If that is not needed, it should be removed.

cabal.project
```diff
@@ -124,6 +125,11 @@ source-repository-package
   tag: 2ad7dced59caece77b6a5e06914aa37d8a3bba23
   subdir: io-sim-classes
 
+source-repository-package
+  type: git
+  location: https://github.com/joelwilliamson/bimap
+  tag: 013a0639c222b3f5572ae20ce364815e5d8f2156
+
```
erikd: We should not be using the git version here. `bimap-0.4` should be sufficient.

deepfire:
Deemed unnecessary.

PR #193, opened 2019-09-16
fix forceNewEval

disassembler:
Allows hydra to pass the gitrev.

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -1,5 +1,8 @@
 let
   commonLib = import ./lib.nix;
+  disabled = [];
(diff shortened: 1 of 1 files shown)


release.nix
```diff
@@ -1,5 +1,8 @@
 let
   commonLib = import ./lib.nix;
+  disabled = [];
+in { cardano-node ? { outPath = ./.; rev = "abcdef"; }, ... }@args:
```
cleverca22: this doesnt do anything with the rev your receiving

erikd:
I think this can be closed as this was fixed in `ac63c74d4e` .

PR #194, opened 2019-09-16
Update dependencies

erikd:
Including ouroboros-network.


erikd:
This should only be merged after #191 and a rebase.

PR #197, opened 2019-09-18
Tx generator, beta-version.

denisshevchenko:
This is beta-version of the new tx generator. It's not finished yet, and still contain some hard-coded values (for development simplicity), but it must be a part of `master`, to avoid endless rebasing to the `master` because of massive changes.

### DIFF
--- a/cardano-node/app/cardano-cli.hs
+++ b/cardano-node/app/cardano-cli.hs
@@ -27,10 +27,15 @@ import           Cardano.Crypto ( AProtocolMagic(..)
                                 , decodeHash)
 
 import           Cardano.Config.CommonCLI
+import           Cardano.Config.Types (CardanoEnvironment (..))
+import           Cardano.Config.Presets (mainnetConfiguration)
+
 import           Cardano.Common.Parsers
 import           Cardano.Common.Protocol
 import           Cardano.CLI.Genesis
 import           Cardano.CLI.Key
+import           Cardano.Config.Logging (LoggingCLIArguments (..),
+                                         createLoggingFeature)
 import           Cardano.CLI.Ops (decideCLIOps)
 import           Cardano.CLI.Run
 
@@ -38,7 +43,13 @@ main :: IO ()
 main = do
   co <- Opt.customExecParser pref opts
   ops <- decideCLIOps (protocol co)
-  cmdRes <- runExceptT . runCommand ops $ mainCommand co
+  -- Initialize logging layer. Particularly, we need it for benchmarking (command 'generate-txs').
+  let cardanoConfiguration = mainnetConfiguration
+  let cardanoEnvironment   = NoEnvironment
+  finalConfig <- mkConfiguration cardanoConfiguration (commonCli co)
+  (loggingLayer, _loggingFeature) <- createLoggingFeature cardanoEnvironment finalConfig (loggingCli co)
+
+  cmdRes <- runExceptT $ runCommand ops (mainCommand co) loggingLayer
   case cmdRes of
     Right _ -> pure ()
     Left err -> do print $ renderCliError err
@@ -60,17 +71,20 @@ opts =
          \ pretty-printing..) for different system generations."
     )
 
-data CLI
-  = CLI
-      { protocol    :: Protocol,
-        mainCommand :: ClientCommand
-        }
+data CLI = CLI
+  { protocol    :: Protocol
+  , mainCommand :: ClientCommand
+  , commonCli   :: CommonCLI
+  , loggingCli  :: LoggingCLIArguments
+  }
 
 parseClient :: Parser CLI
 parseClient =
   CLI
     <$> parseProtocolActual
     <*> parseClientCommand
+    <*> parseCommonCLI
+    <*> loggingParser
 
 parseClientCommand :: Parser ClientCommand
 parseClientCommand =
@@ -250,6 +264,28 @@ parseTxRelatedValues =
                       "Key that has access to all mentioned genesis UTxO inputs."
                 <*> (NE.fromList <$> some parseTxIn)
                 <*> (NE.fromList <$> some parseTxOut)
+                <*> parseCommonCLI,
+          command'
+            "generate-txs"
+            "Launch transactions generator."
+            $ GenerateTxs
+                <$> parseTopologyInfo
+                      "PBFT node ID to submit generated Txs to."
+                <*> parseNumberOfTxs
+                      "num-of-txs"
+                      "Number of transactions generator will create."
+                <*> parseNumberOfOutputsPerTx
+                      "outputs-per-tx"
+                      "Number of outputs in each of transactions."
+                <*> parseFeePerTx
+                      "tx-fee"
+                      "Fee per transaction, in Lovelaces."
+                <*> parseTPSRate
+                      "tps"
+                      "TPS (transaction per second) rate."
+                <*> parseSigningKeysFiles
+                      "sig-key"
+                      "Path to signing key file, for genesis UTxO using by generator."
                 <*> parseCommonCLI
           ]
       )
@@ -475,3 +511,18 @@ parseNewTxFile opt =
   NewTxFile
     <$> parseFilePath opt
           "Non-existent file to write the signed transaction to."
+
+parseNumberOfTxs :: String -> String -> Parser NumberOfTxs
+parseNumberOfTxs opt desc = NumberOfTxs <$> parseIntegral opt desc
+
+parseNumberOfOutputsPerTx :: String -> String -> Parser NumberOfOutputsPerTx
+parseNumberOfOutputsPerTx opt desc = NumberOfOutputsPerTx <$> parseIntegral opt desc
+
+parseFeePerTx :: String -> String -> Parser FeePerTx
+parseFeePerTx opt desc = FeePerTx <$> parseIntegral opt desc
+
+parseTPSRate :: String -> String -> Parser TPSRate
+parseTPSRate opt desc = TPSRate <$> parseIntegral opt desc
+
+parseSigningKeysFiles :: String -> String -> Parser [SigningKeyFile]
+parseSigningKeysFiles opt desc = many $ SigningKeyFile <$> parseFilePath opt desc
--- a/cardano-node/src/Cardano/CLI/Run.hs
+++ b/cardano-node/src/Cardano/CLI/Run.hs
@@ -30,9 +30,13 @@ module Cardano.CLI.Run (
   , NewCertificateFile(..)
   , TxFile(..)
   , NewTxFile(..)
+  , NumberOfTxs(..)
+  , NumberOfOutputsPerTx(..)
+  , FeePerTx(..)
+  , TPSRate(..)
   ) where
 
-import           Cardano.Prelude hiding (option, show, trace)
+import           Cardano.Prelude hiding (option, trace)
 
 import           Codec.Serialise (serialise)
 import           Control.Monad.Trans.Except (ExceptT)
@@ -48,22 +52,29 @@ import qualified Cardano.Chain.Delegation as Delegation
 import qualified Cardano.Chain.Genesis as Genesis
 import           Cardano.Chain.Slotting (EpochNumber(..))
 import qualified Cardano.Chain.UTxO as UTxO
+
 import           Cardano.Crypto (ProtocolMagicId)
 import qualified Cardano.Crypto.Hashing as Crypto
 import qualified Cardano.Crypto.Signing as Crypto
+
 import qualified Test.Cardano.Chain.Genesis.Dummy as Dummy
+
+import qualified Ouroboros.Consensus.Protocol as Consensus
 
 import           Cardano.CLI.Delegation
 import           Cardano.CLI.Genesis
 import           Cardano.CLI.Key
 import           Cardano.CLI.Ops
 import           Cardano.CLI.Tx
-import           Cardano.Config.CommonCLI
-import           Cardano.Config.Presets (mainnetConfiguration)
+import           Cardano.CLI.Tx.Generation (NumberOfTxs (..), NumberOfOutputsPerTx (..),
+                                            FeePerTx (..), TPSRate (..),
+                                            genesisBenchmarkRunner)
 import           Cardano.Common.Orphans ()
 import           Cardano.Common.Protocol
-import           Cardano.Node.Configuration.Topology
-
+import           Cardano.Config.CommonCLI
+import           Cardano.Config.Logging (LoggingLayer (..))
+import           Cardano.Config.Presets (mainnetConfiguration)
+import           Cardano.Node.Configuration.Topology (TopologyInfo)
 
 -- | Sub-commands of 'cardano-cli'.
 data ClientCommand
@@ -137,25 +148,36 @@ data ClientCommand
     -- ^ Genesis UTxO output Address.
     CommonCLI
 
-runCommand :: CLIOps IO -> ClientCommand -> ExceptT CliError IO ()
-runCommand co (Genesis outDir params) = do
+    --- Tx Generator Command ----------
+
+  | GenerateTxs
+    TopologyInfo
+    NumberOfTxs
+    NumberOfOutputsPerTx
+    FeePerTx
+    TPSRate
+    [SigningKeyFile]
+    CommonCLI
+
+runCommand :: CLIOps IO -> ClientCommand -> LoggingLayer -> ExceptT CliError IO ()
+runCommand co (Genesis outDir params) _ = do
   gen <- mkGenesis params
   dumpGenesis co outDir `uncurry` gen
 
-runCommand co (DumpHardcodedGenesis dir) =
+runCommand co (DumpHardcodedGenesis dir) _ =
   dumpGenesis co dir (Genesis.configGenesisData Dummy.dummyConfig) Dummy.dummyGeneratedSecrets
 
-runCommand co (PrettySigningKeyPublic skF) = do
+runCommand co (PrettySigningKeyPublic skF) _ = do
   sK <- readSigningKey co skF
   liftIO . putTextLn . prettyPublicKey $ Crypto.toVerification sK
 
-runCommand co (MigrateDelegateKeyFrom fromVer (NewSigningKeyFile newKey) oldKey) = do
+runCommand co (MigrateDelegateKeyFrom fromVer (NewSigningKeyFile newKey) oldKey) _ = do
   ops <- liftIO $ decideCLIOps fromVer
   sk <- readSigningKey ops oldKey
   sDk <- liftIO $ coSerialiseDelegateKey co sk
   liftIO $ ensureNewFileLBS newKey sDk
 
-runCommand _ (PrintGenesisHash genFp) = do
+runCommand _ (PrintGenesisHash genFp) _ = do
   eGen <- readGenesis genFp
 
   let formatter :: (a, Genesis.GenesisHash)-> Text
@@ -163,23 +185,23 @@ runCommand _ (PrintGenesisHash genFp) = do
 
   liftIO . putTextLn $ formatter eGen
 
-runCommand co (PrintSigningKeyAddress netMagic skF) = do
+runCommand co (PrintSigningKeyAddress netMagic skF) _ = do
   sK <- readSigningKey co skF
   let sKeyAddress = prettyAddress . Common.makeVerKeyAddress netMagic $ Crypto.toVerification sK
   liftIO $ putTextLn sKeyAddress
 
-runCommand co (Keygen (NewSigningKeyFile skF) passReq) = do
+runCommand co (Keygen (NewSigningKeyFile skF) passReq) _ = do
   pPhrase <- liftIO $ getPassphrase ("Enter password to encrypt '" <> skF <> "': ") passReq
   sK <- liftIO $ keygen pPhrase
   serDk <- liftIO $ coSerialiseDelegateKey co sK
   liftIO $ ensureNewFileLBS skF serDk
 
-runCommand co (ToVerification skFp (NewVerificationKeyFile vkFp)) = do
+runCommand co (ToVerification skFp (NewVerificationKeyFile vkFp)) _ = do
   sk <- readSigningKey co skFp
   let vKey = Builder.toLazyText . Crypto.formatFullVerificationKey $ Crypto.toVerification sk
   liftIO $ ensureNewFile TL.writeFile vkFp vKey
 
-runCommand co (IssueDelegationCertificate magic epoch issuerSK delegateVK cert) = do
+runCommand co (IssueDelegationCertificate magic epoch issuerSK delegateVK cert) _ = do
   vk <- readVerificationKey delegateVK
   sk <- readSigningKey co issuerSK
   let byGenDelCert :: Delegation.Certificate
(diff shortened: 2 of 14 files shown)


cardano-node/app/cardano-cli.hs
```diff
@@ -25,20 +25,30 @@ import           Cardano.Crypto ( AProtocolMagic(..)
                                 , ProtocolMagicId(..)
                                 , RequiresNetworkMagic(..)
                                 , decodeHash)
-
```
deepfire: This separator between local imports and those from other repos should stay.

cardano-node/app/cardano-cli.hs
```diff
@@ -25,20 +25,30 @@ import           Cardano.Crypto ( AProtocolMagic(..)
...
 import           Cardano.CLI.Genesis
 import           Cardano.CLI.Key
+import           Cardano.Config.Logging (LoggingCLIArguments (..),
+                                         createLoggingFeature)
 import           Cardano.CLI.Ops (decideCLIOps)
 import           Cardano.CLI.Run
 
 main :: IO ()
 main = do
-  co <- Opt.customExecParser pref opts
```
deepfire: Why did you remove customExecParser?

cardano-node/app/cardano-cli.hs
```diff
@@ -25,20 +25,30 @@ import           Cardano.Crypto ( AProtocolMagic(..)
...
+  co <- Opt.execParser opts
   ops <- decideCLIOps (protocol co)
-  cmdRes <- runExceptT . runCommand ops $ mainCommand co
+  -- Initialize logging layer. Particularly, we need it for benchmarking (command 'generate-txs').
+  let cardanoConfiguration = mainnetConfiguration
+  let cardanoEnvironment   = NoEnvironment
+  finalConfig <- mkConfiguration cardanoConfiguration (commonCli co)
+  (loggingLayer, _loggingFeature) <- createLoggingFeature cardanoEnvironment finalConfig (loggingCli co)
+
+  cmdRes <- runExceptT $ runCommand ops (mainCommand co) loggingLayer
```
deepfire: Why those changes?

cardano-node/app/cardano-cli.hs
```diff
@@ -60,17 +67,20 @@ opts =
...
-  = CLI
-      { protocol    :: Protocol,
-        mainCommand :: ClientCommand
-        }
+data CLI = CLI
+  { protocol    :: Protocol
+  , mainCommand :: ClientCommand
+  , commonCli   :: CommonCLI
+  , loggingCli  :: LoggingCLIArguments
+  }
```
deepfire: Why this change of CLI?

cardano-node/src/Cardano/CLI/Run.hs
```diff
@@ -53,17 +61,29 @@ import qualified Cardano.Crypto.Hashing as Crypto
...
 import           Cardano.Common.Protocol
-import           Cardano.Node.Configuration.Topology
+import           Cardano.Config.CommonCLI
+import           Cardano.Config.Logging (LoggingLayer (..), Trace)
+import           Cardano.Config.Presets (mainnetConfiguration)
+import           Cardano.Node.Configuration.Topology (TopologyInfo)
 
+import           Ouroboros.Consensus.Ledger.Byron.Config (ByronConfig)
+import           Ouroboros.Consensus.Ledger.Byron (ByronBlockOrEBB (..), GenTxId) 
+import qualified Ouroboros.Consensus.Protocol as Consensus
```
deepfire: These external imports should come before local imports.

cardano-node/app/cardano-cli.hs
deepfire: After some thinking, this is a good idea, to avoid repeating the same init across different commands.

cardano-node/src/Cardano/CLI/Run.hs
```diff
@@ -137,78 +157,139 @@ data ClientCommand
...
-runCommand co (SpendUTxO (NewTxFile ctTx) ctKey ins outs common) = do
+runCommand co (SpendUTxO (NewTxFile ctTx) ctKey ins outs common) _ = do
   cc <- liftIO $ mkConfiguration mainnetConfiguration common
   sk <- readSigningKey co ctKey
   gTx <- liftIO $ issueUTxOExpenditure co ins outs cc sk
   liftIO . ensureNewFileLBS ctTx $ serialise gTx
 
+runCommand co
+           (GenerateTxs topology numOfTxs numOfOutsPerTx feePerTx tps sigKeysFiles common)
+           loggingLayer = do
```
deepfire: This should be a short case branch that calls out to the main function.

cardano-node/src/Cardano/CLI/Run.hs
```diff
@@ -137,78 +157,139 @@ data ClientCommand
...
+  cc <- liftIO $ mkConfiguration mainnetConfiguration common
+  SomeProtocol p <- liftIO $ fromProtocol cc (coProtocol co)
+  case p of
+    protocol@(Consensus.ProtocolRealPBFT _ _ _ _ _) -> do
+      let basicTr :: Trace IO Text
```
deepfire: We should replace this `let` with a `where` -- so that it's immediately clear what is the action it performs.

The long chain of `let`'s is really obscuring the big picture here.

cardano-node/src/Cardano/CLI/Tx/BenchmarkingNodeToNode.hs
```diff
@@ -0,0 +1,109 @@
...
+import           Network.Socket (AddrInfo, SockAddr)
+import           Network.TypedProtocol.Driver (TraceSendRecv, runPeer)
+import           Network.TypedProtocol.Driver.ByteLimit (DecoderFailureOrTooMuchInput)
+import           Ouroboros.Consensus.Block (BlockProtocol)
+import           Ouroboros.Consensus.Mempool.API (GenTxId, GenTx)
+import           Ouroboros.Consensus.Node.Run (RunNode)
+import           Ouroboros.Consensus.NodeNetwork (ProtocolCodecs(..), protocolCodecs)
+import           Ouroboros.Consensus.Protocol.Abstract (NodeConfig)
+import           Ouroboros.Network.Mux (OuroborosApplication(..))
+import           Ouroboros.Network.NodeToNode
```
deepfire: We avoid use of implicit imports -- this should be wither qualified (`as Network`, for example) or explicit.

cardano-node/src/Cardano/CLI/Tx/BenchmarkingTxSubClient.hs
```diff
@@ -0,0 +1,53 @@
...
+{-# LANGUAGE RankNTypes          #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Cardano.CLI.Tx.BenchmarkingTxSubClient
+  ( txSubmissionClient
+  ) where
+
+import Prelude
+
+import Control.Monad.Class.MonadSTM
```
deepfire: Should be an explicit import, I suppose..

cardano-node/src/Cardano/CLI/Tx/BenchmarkingTxSubmission.hs
```diff
@@ -0,0 +1,350 @@
...
+  ( ROEnv(..)
+  , RPCTxSubmission(..)
+  , TraceBenchTxSubmit(..)
+  , bulkSubmission
+  )
+where
+
+import           Prelude
+import           Cardano.Prelude (MonadIO, Word16, Seq, StateT
+                                 , evalStateT, gets, isJust, lift, {- liftIO, -} modify, toList, unless, when)
```
deepfire: Preludes are exempt from the explicit import list requirement -- so this one can be removed.

cardano-node/src/Cardano/CLI/Tx/BenchmarkingTxSubmission.hs
```diff
@@ -0,0 +1,350 @@
...
+--   configurable latency hiding (number of transactions to try to
+--   keep ahead of actual submission) and has tracing that both
+--   reports observables related to transaction submission and whether
+--   the tx generation is able to keep up the pace.
+--
+--   For the details of the protocol description see
+--   `TxSubmit.TxSubmission` and its associated instances
+--   (e.g. `Message`).
+
+bulkSubmission :: forall m blk txid tx .
```
deepfire: It's better to move the `:: forall m blk txid tx .` to the next line and reindent -- so that the offset of the entire block doesn't depend on the length of the `bulkSubmission` name.

cardano-node/src/Cardano/CLI/Tx/Submission.hs
```diff
@@ -9,6 +9,9 @@
 
 module Cardano.CLI.Tx.Submission (
       handleTxSubmission
+    , localSocketFilePath
+    , localSocketAddrInfo
```
deepfire: We maintain alphabetic sort order in the export list.

cardano-node/src/Cardano/CLI/Tx/BenchmarkingTxSubmission.hs
denisshevchenko: Neil made it to avoid conflicts with `Prelude`.

cardano-node/src/Cardano/CLI/Run.hs
denisshevchenko: Sorry, didn't understand it...

cardano-node/src/Cardano/CLI/Run.hs
```diff
@@ -30,14 +30,22 @@ module Cardano.CLI.Run (
...
   , TxFile(..)
   , NewTxFile(..)
+  , NumberOfTxs(..)
+  , NumberOfOutputsPerTx(..)
+  , FeePerTx(..)
+  , TPSRate(..)
   ) where
 
-import           Cardano.Prelude hiding (option, show, trace)
+import Data.String(String)
```
deepfire: Here we currently mandate vertical alignment.

cardano-node/src/Cardano/CLI/Run.hs
```diff
@@ -48,22 +56,37 @@ import qualified Cardano.Chain.Delegation as Delegation
 import qualified Cardano.Chain.Genesis as Genesis
 import           Cardano.Chain.Slotting (EpochNumber(..))
 import qualified Cardano.Chain.UTxO as UTxO
+
```
deepfire: I wouldn't say the intra-external-module import white space is necessary, but it's not critical.

cardano-node/src/Cardano/CLI/Run.hs
deepfire: Because we don't want functions that are too long.

`runCommand` went through refactoring to avoid precisely that.

cardano-node/src/Cardano/CLI/Tx/Generation.hs
```diff
@@ -0,0 +1,850 @@
...
+  -> Maybe CC.Common.Address              -- the address to associate with the 'change',
+                                          -- if different from that of the first argument
+  -> Set (r, CC.UTxO.TxOut)               -- each recipient and their payment details
+  -> ( Maybe (Word32, CC.Common.Lovelace) -- the 'change' index and value (if any)
+     , CC.Common.Lovelace                 -- the associated fees
+     , Map r Word32                       -- the offset map in the transaction below
+     , GenTx (ByronBlockOrEBB cfg)
+     )
+mkTransaction cfg ((txinFrom, txoutFrom), signingKey) mChangeAddress payments =
+  let paymentsList = toList payments
```
deepfire: I think this would also benefit if rewritten with a `where` clause instead of a `let`.

cardano-node/src/Cardano/CLI/Tx/Generation.hs
```diff
@@ -0,0 +1,850 @@
...
+  -> Tracer IO String
+  -> CardanoConfiguration
+  -> Consensus.Protocol (ByronBlockOrEBB ByronConfig)
+  -> TopologyInfo
+  -> NumberOfTxs
+  -> NumberOfOutputsPerTx
+  -> FeePerTx
+  -> TPSRate
+  -> [FilePath]
+  -> IO ()
```
deepfire: I think many of the arguments should be grouped into a `data` type, like `BenchmarkConfig` or somesuch.
Same for the tracers, they should also benefit from grouping -- `data BencmarkTracers` maybe..

This would help with being able to just pass it down to `runBenchmark`.

cardano-node/src/Cardano/CLI/Tx/Generation.hs
```diff
@@ -0,0 +1,850 @@
...
+  -> NodeConfig ByronEBBExtNodeConfig
+  -> CC.Common.Address
+  -> Crypto.SigningKey
+  -> CC.Common.Address
+  -> TopologyInfo
+  -> NumberOfTxs
+  -> NumberOfOutputsPerTx
+  -> FeePerTx
+  -> TPSRate
+  -> IO ()
```
deepfire: Oh, yes, a dedicated `data` type would help here..

cardano-node/src/Cardano/CLI/Tx/Generation.hs
```diff
@@ -0,0 +1,850 @@
...
+        :: ROEnv (Byron.GenTxId (ByronBlockOrEBB ByronConfig)) (GenTx (ByronBlockOrEBB ByronConfig))
+        -> ROEnv (Byron.GenTxId (ByronBlockOrEBB ByronConfig)) (GenTx (ByronBlockOrEBB ByronConfig))
+      updROEnv defaultROEnv =
+        ROEnv { targetBacklog     = targetBacklog defaultROEnv
+              , txNumServiceTime  = Nothing
+              , txSizeServiceTime = Nothing
+              }
+
+  -- Run generator.
+  txGenerator tracer
```
deepfire: Maybe a `data` type for `txGenerator` as well? `TxGeneratorParams` or somesuch..

cardano-node/src/Cardano/CLI/Tx/BenchmarkingTxSubmission.hs
```diff
@@ -0,0 +1,353 @@
...
+    -- Update terminating, if needed 
+    -- lift . traceWith tr . TraceBenchTxSubDebug $ "go1, update terminating, if needed"
+    when (nowTerminating && not terminating') $ do
+      -- lift . traceWith tr . TraceBenchTxSubDebug $ "go1, update terminating: True"
+      modify (\x -> x {terminating = True})
+
+    -- incorporate new transactions from the generator into
+    -- NotYetSent
+    -- lift . traceWith tr . TraceBenchTxSubDebug $ "go1, incorporate new transactions from the generator into NotYetSent"
+    flip (maybe (pure ())) newTxs $ \xs -> do
```
deepfire: Hmm.. is it just me, or is this somewhat hard to read? 

Do you think we could have a simpler presentation -- maybe a dedicated function?

cardano-node/src/Cardano/CLI/Tx/BenchmarkingTxSubmission.hs
```diff
@@ -0,0 +1,353 @@
...
+          reqTxIds <- takeTMVar rpcIn -- no other message should occur as nothing is in flight
+          case reqTxIds of
+            RPCRequestTxIds _ resp -> putTMVar resp Nothing
+            _ -> return ()
+    else do -- process next interaction
+      -- lift . traceWith tr . TraceBenchTxSubDebug $ "go, process next interaction"
+      go1 env
+
+  go1 :: ROEnv txid tx -> StateT (RWEnv m txid tx) m ()
+  go1 env = do
```
deepfire: This function feels like it could greatly benefit from being split into parts, at least as a pipeline.

cardano-node/src/Cardano/CLI/Run.hs
denisshevchenko: Done. Now I just pass `loggingLayer` in my function, all tracers will be created on my side.

cardano-node/src/Cardano/CLI/Tx/BenchmarkingTxSubmission.hs
denisshevchenko: This code was written by Neil. To be honest, I don't want to touch it right now (actually, only Neil really knows how this code works).

cardano-node/src/Cardano/CLI/Tx/BenchmarkingTxSubmission.hs
denisshevchenko: This code was written by Neil. To be honest, I don't want to touch it right now (actually, only Neil really knows how this code works).

cardano-node/src/Cardano/CLI/Run.hs
```diff
@@ -137,78 +148,107 @@ data ClientCommand
...
   liftIO . ensureNewFileLBS ctTx $ serialise gTx
 
+runCommand co
+           (GenerateTxs topology numOfTxs numOfOutsPerTx feePerTx tps sigKeysFiles common)
+           loggingLayer = do
+  cc <- liftIO $ mkConfiguration mainnetConfiguration common
+  SomeProtocol p <- liftIO $ fromProtocol cc (coProtocol co)
+  case p of
+    protocol@(Consensus.ProtocolRealPBFT _ _ _ _ _) ->
+      liftIO $ genesisBenchmarkRunner loggingLayer
```
deepfire: Here you can use `withRealPBFT`.

denisshevchenko: Done.

deepfire:
I've asked @intricate to have an independent go at this PR.

cardano-node/src/Cardano/CLI/Tx/Generation.hs
```diff
@@ -0,0 +1,893 @@
...
+                    pInfoConfig
+                    topologyInfo
+                    genesisUtxo
+                    genesisAddress
+                    targetAddress = do
+  let (mFunds, _, _initGenTx) = extractInitialFunds pInfoConfig
+                                                    (genesisUtxo Map.! 0) -- corresponds to 'genesisAddress'.
+                                                    targetAddress
+
+  let ((_, _), signingKey) = genesisUtxo Map.! 0
```
intricate: I wonder if it'd be nicer to utilize `Map.!?`, pattern match on the result, and then explicitly call `error` if it's `Nothing`. This way we could also provide a more helpful error message detailing the module and function where the error occurred (i.e. `Cardano.CLI.Tx.Generation.prepareInitialFunds`). 

Also, it seems like we utilize `genesisUtxo Map.! 0` twice in this function so we could probably give it its own `let` binding.

@deepfire: think I'm nitpicking too much?

cardano-node/src/Cardano/CLI/Tx/Generation.hs
```diff
@@ -0,0 +1,893 @@
...
+               remoteAddr
+               updROEnv
+               txsForSubmission
+
+-- | At this moment 'sourceAddress' contains a huge amount of money (lets call it A).
+--   Now we have to split this amount to N equal parts, as a result we'll have
+--   N UTxO entries, and alltogether these entries will contain the same amount A.
+--   E.g. (1 entry * 1000 ADA) -> (10 entries * 100 ADA).
+--   Technically all splitting transactions will send money back to 'sourceAddress'.
+createMoreFundCoins
```
intricate: Nice comments btw :+1:. Appreciate you documenting much of what is going on.

cardano-node/src/Cardano/CLI/Tx/Generation.hs
```diff
@@ -0,0 +1,893 @@
...
+                                          -- input and the key to spend the
+                                          -- associated value
+  -> Maybe CC.Common.Address              -- the address to associate with the 'change',
+                                          -- if different from that of the first argument
+  -> Set (r, CC.UTxO.TxOut)               -- each recipient and their payment details
+  -> ( Maybe (Word32, CC.Common.Lovelace) -- the 'change' index and value (if any)
+     , CC.Common.Lovelace                 -- the associated fees
+     , Map r Word32                       -- the offset map in the transaction below
+     , GenTx (ByronBlockOrEBB cfg)
+     )
```
intricate: Not a big deal but, typically, once I start dealing with anything more than a triple, I like to define a separate data type.

cardano-node/src/Cardano/CLI/Tx/BenchmarkingTxSubmission.hs
```diff
@@ -0,0 +1,353 @@
...
+  --   peer.
+  | TraceBenchTxSubServAck [txid]
+  -- ^ An ack (window moved over) received for these transactions.
+  | TraceBenchTxSubIdle
+  -- ^ Remote peer requested new transasctions but none were
+  --   available, generator not keeping up?
+  | TraceBenchTxSubRateLimit DiffTime
+  -- ^ Rate limiter bit, this much delay inserted to keep within
+  --   configured rate.
+  | TraceBenchTxSubDebug String
```
intricate: Can we doc this constructor too?

cardano-node/src/Cardano/CLI/Tx/BenchmarkingTxSubmission.hs
```diff
@@ -0,0 +1,353 @@
...
+  , txSizeServiceTime = Nothing
+  }
+
+data RWEnv m txid tx = RWEnv
+  { terminating     :: Bool
+  , activityState   :: ActivityState
+  , proceedAfter    :: Time m
+  , availableOp     :: Maybe (Int, LazyTMVar m (Maybe [(txid, TxSubmit.TxSizeInBytes)]))
+  -- ^ the window and the response action
+  , inFlight
```
intricate: What's going on with this `inFlight` field? Is this valid code?

intricate:
> It has been agreed with @denisshevchenko that this PR is merged as a WIP, to reduce the rebase pain, and that further improvements will be made.

Hopefully, going forward, we can try to avoid raising large PRs like this in the future. It's quite difficult to actually give it a solid review and then there's the aforementioned rebase pain.

cardano-node/src/Cardano/CLI/Tx/BenchmarkingTxSubmission.hs
intricate: The CI checks are all passing so I'm guessing it's valid, but I'm wondering why there's no type signature?

denisshevchenko:
> Hopefully we can try to avoid raising large PRs like this in the future. It's quite difficult to actually give it a solid review and then there's the aforementioned rebase pain.

@intricate Agree. Tx generator is a quite big piece of code, but after this branch will be merged in the `master`, following PRs will be much, much smaller.

cardano-node/src/Cardano/CLI/Tx/BenchmarkingTxSubmission.hs
denisshevchenko: I'm not sure if we need this constructor at all. It was used as a fastest way to get raw debug messages..

cardano-node/src/Cardano/CLI/Tx/Generation.hs
```diff
@@ -0,0 +1,893 @@
...
+
+launchTxPeer
+  :: forall m block txid tx.
+     ( RunNode block
+     , m ~ IO
+     , txid ~ Byron.GenTxId block
+     , tx ~ GenTx block
+     )
+  => Tracer m (TraceBenchTxSubmit txid)
+  -- Tracer carrying the benchmarking events
```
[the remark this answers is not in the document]
denisshevchenko: I did it because parsing problems from Haddock. I don't know why...

cardano-node/src/Cardano/CLI/Tx/BenchmarkingTxSubClient.hs
```diff
@@ -0,0 +1,59 @@
...
+                                                                 TxSubmissionClient(..))
+import           Ouroboros.Network.Protocol.TxSubmission.Type (BlockingReplyList(..),
+                                                               TokBlockingStyle(..))
+
+txSubmissionClient
+  :: forall m block txid tx .
+     ( MonadSTM m
+     , txid ~ GenTxId block
+     , tx ~ GenTx block)
+  => LazyTMVar m (RPCTxSubmission m txid tx)
```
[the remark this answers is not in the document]
denisshevchenko: > Do we specifically need a LazyTMVar here 

Honestly, I don't know (this code was written by Neil).

cardano-node/src/Cardano/CLI/Tx/Generation.hs
denisshevchenko: > I wonder if it'd be nicer to utilize `Map.!?`, pattern match on the result, and then explicitly call `error` if it's `Nothing`.

It cannot be `Nothing` here, because of `extractGenesisFunds` function (which already called).

cardano-node/src/Cardano/CLI/Tx/BenchmarkingTxSubmission.hs
deepfire: Wow.

PR #199, opened 2019-09-19
Reinstate the local socket naming policy & fix all the scripts

deepfire:
1. `--socket-path` -> `--socket-dir`
2. Reinstate the global policy for socket naming that was undone in https://github.com/input-output-hk/cardano-node/pull/178
3. Avoid socket path specification everywhere, but the Nix service.
4. Fix & test all the scripts.


deepfire:
After a discussion with @disassembler, we extended scope to fix the socket path business.

The changes in https://github.com/input-output-hk/cardano-node/pull/178 removed a singular scheme for deriving socket paths from node IDs, which has a fractal of repercussions we'd rather avoid.

PR #201, opened 2019-09-19
Chairman:  allow it to fail, and add an extended failure criterion.

deepfire:
1. Make Chairman fail on time out.
2. Factor consensus validation spec and context data structures.
2. Make him enforce a stricter check on fork length:
   - fork lengths must remain under --maximum-fork-length
   - ..for at least --slots-within-tolerance slots


CodiePP:
is this still to be merged into master at some time?

deepfire:
@CodiePP, I think this can be retired.

PR #205, opened 2019-09-24
Mux Tracing

karknu:


cardano-node/src/Cardano/Tracing/ToObjectOrphans.hs
```diff
@@ -162,6 +163,25 @@ instance DefineSeverity (WithDomainName DnsTrace) where
...
+    MuxTraceRecvHeaderEnd _      -> Debug
+    MuxTraceRecvPayloadStart _   -> Debug
+    MuxTraceRecvPayloadEnd _     -> Debug
+    MuxTraceRecvStart _          -> Debug
+    MuxTraceRecvEnd _            -> Debug
+    MuxTraceSendStart _          -> Debug
+    MuxTraceSendEnd              -> Debug
+    MuxTraceStateChange _ _      -> Info
+    MuxTraceCleanExit _          -> Info
+    MuxTraceExceptionExit _ _    -> Info
```
CodiePP: this could be `Error` ?

karknu: No, it is not an error for a bearer to exit due to an exception (examples of such exceptions are when the peer closes the connection or chainsync thinks that the peer is too out-of-sync for us to be talking to it).

cardano-node/src/Cardano/Tracing/ToObjectOrphans.hs
```diff
@@ -453,6 +482,12 @@ instance ToObject (WithDomainName DnsTrace) where
              , "domain" .= show dom
              , "event" .= show ev ]
 
+instance Show ptcl => ToObject (WithMuxBearer (MuxTrace ptcl)) where
+  toObject _verb (WithMuxBearer b ev) =
+    mkObject [ "kind" .= String "MuxTrace"
+             , "bearer" .= show b
+             , "event" .= show ev ]
```
CodiePP: this could probably be extended later to provide more details for the events. OK for now.

PR #207, opened 2019-09-25
Logging disablement changes & script fixes

deepfire:
0. Restore the IPv6 topology in the testnet scripts
0. Refactor & update scripts to CLI changes (they break oh so often..)
0. `--tracing-off` and `--log-config` are now mutually exclusive -- one important implication is that `--log-config` is no longer a mandatory option.
0. All logging users now respect `--tracing-off` -- i.e. it is no longer specific to `cardano-node`
0. Some code shuffling for better locality

# Discussion

0. Perhaps, we might want a third option, that supplies some kind of non-disabled default for the tracing infrastructure. Maybe @CodiePP has some suggestions for how we can get this.

### DIFF
--- a/cardano-config/src/Cardano/Config/Logging.hs
+++ b/cardano-config/src/Cardano/Config/Logging.hs
@@ -1,13 +1,19 @@
 {-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE Rank2Types          #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections #-}
 
 {-# OPTIONS_GHC -Wno-all-missed-specialisations #-}
 
 module Cardano.Config.Logging
   ( LoggingLayer (..)
+  , TraceOptions (..)
+  , LoggingFlag (..)
+  , ConsensusTraceOptions (..)
+  , ProtocolTraceOptions (..)
   , LoggingConfiguration (..)
   , createLoggingFeature
+  , loggingCLIConfiguration
   -- re-exports
   , Trace
   , Configuration
@@ -20,10 +26,13 @@ module Cardano.Config.Logging
   , LoggingCLIArguments (..)
   ) where
 
+import           Prelude (error)
 import           Cardano.Prelude hiding (trace)
 
 import qualified Control.Concurrent.Async as Async
 import           Control.Exception.Safe (MonadCatch)
+import           Options.Applicative
+import qualified Options.Applicative as Opt
 
 import           Cardano.BM.Backend.Aggregation (plugin)
 import           Cardano.BM.Backend.Editor (plugin)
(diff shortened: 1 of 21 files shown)


deepfire:
> this duplicates another flag --tracing-off

@CodiePP, `--tracing-off` is specific to the node -- I was actually thinking of removing that one.

Would you support its removal?

deepfire:
@CodiePP, reworked!

1. --tracing-off and --log-config are now mutually exclusive
2. all users now respect --tracing-off
3. some code shuffling for better locality

cardano-config/src/Cardano/Config/Logging.hs
```diff
@@ -31,13 +37,15 @@ import           Cardano.BM.Backend.Monitoring (plugin)
 import qualified Cardano.BM.Backend.Switchboard as Switchboard
 import           Cardano.BM.Configuration (Configuration)
 import qualified Cardano.BM.Configuration as Config
+import qualified  Cardano.BM.Configuration.Model as Config
```
CodiePP: double space!

cardano-node/src/Cardano/Common/Parsers.hs
```diff
@@ -9,19 +9,25 @@ module Cardano.Common.Parsers
   , parseProtocolActual
   , parseProtocolAsCommand
   , parseTopologyInfo
+  , parseTraceOptions
   ) where
 
 
-import           Prelude (String)
+import           Prelude (String, error)
```
CodiePP: `error` is considered dangerous!

cardano-config/src/Cardano/Config/Logging.hs
```diff
@@ -110,34 +137,46 @@ type LoggingCardanoFeature = CardanoFeatureInit
...
   }
 
+data LoggingFlag = LoggingEnabled | LoggingDisabled
+  deriving (Eq, Show)
+
+-- | Interpret main logging CLI controls into a tuple of:
+--   - a designation of whether logging was disabled,
+--   - a valid 'LoggingConfiguration' (still necessary, even if logging was disabled)
+loggingCLIConfiguration :: LoggingCLIArguments -> IO (LoggingFlag, LoggingConfiguration)
+loggingCLIConfiguration (LoggingCLIArguments Nothing _) =
```
CodiePP: I see a slight problem with this pattern matching. What if we add a field to LoggingCLIArguments ?
:-) you know, in prep..

cardano-node/src/Cardano/Common/Parsers.hs
deepfire: In general, yes -- but its usage is in the CLI processing of a strict data structure, so it's not thunked and forced immediately.

More over -- the condition should never ever arise (lest there is a bug in `optparse-applicative`), but we can't prove that at the type level.

cardano-config/src/Cardano/Config/Logging.hs
deepfire: Very good point, fixed!

cardano-config/src/Cardano/Config/Logging.hs
deepfire: Fixed!

PR #208, opened 2019-09-26
Tx gen: real TPS and node network parameters

denisshevchenko:
Now tx generator doesn't contain hardcoded values of TPS and target node's host and port.


denisshevchenko:
@deepfire @Jimbo4350 Script `./scripts/shelley-testnet.sh` doesn't work again:

1. Invalid argument `node` (we don't need anymore?).
2. Invalid argument `-n` (now accepts only full form `--node-id`).

Moreover:

```
cardano-node: handleSimpleNode: own address NodeAddress {naHostAddress = ::1, naPort = 3000} not found in topology                                                                             
  error, called at src/Cardano/Node/Run.hs:157:22 in cardano-node-3.0.1.87-inplace:Cardano.Node.Run
```

deepfire:
@denisshevchenko, the script fixes are in https://github.com/input-output-hk/cardano-node/pull/207

PR #223, opened 2019-10-01
Stake pool registry validator

deepfire:
1. Drop `--tracing-off` -- it was implemented with a hard-to-fix bug.
2. `validate-registry-submission`:  new subcommand for the sake of https://github.com/input-output-hk/shelley-testnet/issues/49


cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
```diff
@@ -0,0 +1,301 @@
...
+-- | Stake pool ID, also its 32-byte public key.
+newtype Id = Id { unId :: Text }
+  deriving (Eq)
+
+-- | Stake pool ticker, a 3-4 all-ASCII-uppercase character ID.
+newtype Ticker = Ticker { unTicker :: Text }
+  deriving (Eq, Ord, Show)
+
+-- | Non-negative Scientific.
+newtype Cost = Cost { unCost :: Sci.Scientific }
```
KtorZ: Wouldn't a `Natural` be better-suited here?

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
```diff
@@ -0,0 +1,301 @@
...
+      -- All valid.
+      pure entry
+
+ where
+   checks :: [Text] -> ExceptT CliError IO ()
+   checks [] = ExceptT . pure . Right $ ()
+   checks es = err es
+
+   err :: [Text] -> ExceptT CliError IO a
+   err es = ExceptT . pure . Left  $ PoolRegistryValidationErrors fp es
```
KtorZ: Wouldn't the use of [throwE](https://hackage.haskell.org/package/transformers-0.5.6.2/docs/Control-Monad-Trans-Except.html#v:throwE) makes the above code a bit simpler / more idiomatic?

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
```diff
@@ -0,0 +1,301 @@
...
+     pure . Set.toList $ duplicates tickers
+     where
+       selectEntries :: [FilePath] -> [FilePath]
+       selectEntries paths =
+         flip filter paths $ (==".json") . Path.takeExtension
+
+       entryTicker :: FilePath -> IO (Either String Ticker)
+       entryTicker f =
+         (bimap (("While parsing '"<>f<>"': ")<>) sTicker) . AE.eitherDecode . BSL.fromStrict
+           <$> BS.readFile f
```
KtorZ: The logic here seems a bit backward to me :thinking: 
What about: 

- Get the list of all existing tickers

```hs 
    listTickers :: FilePath -> ExceptT Test IO [Ticker]
```

- Check whether the new ticker is part of that list 

```hs
tickers <- listTickers dir; 
when (myTicker `elem` tickers) $ 
    throwE (validationError "ticker already known")
```

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
```diff
@@ -0,0 +1,301 @@
...
+
+--------------------------------------------------------------------------------
+-- * Ancillary
```
KtorZ: now you're teaching me some vocabulary :smile:

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
```diff
@@ -0,0 +1,301 @@
...
+-- | Submission identifier is actually a 32-byte base16-encoded public key.
+--   We additionally require that the hexadecimal must be all-lowercase.
+validatePublicKey :: Text -> Either Text Id
+validatePublicKey text =
+  if      not (length s == 64 && all Char.isHexDigit s)
+  then fail $ "Public key not a 64-char-long base16: " <> take 128 s
+  else if not ((Char.toLower <$> s) == s)
+  then fail $ "Public key not all-lowercase: " <> s
+  else Right $ Id text
+  where s = unpack text
```
KtorZ: I'd suggest using [convertFromBase](https://hackage.haskell.org/package/memory-0.15.0/docs/Data-ByteArray-Encoding.html#v:convertFromBase):

```hs
convertPublicKey :: Text -> Either String Id
convertPublicKey txt = do
    bytes <- convertFromBase Base16 (T.encodeUtf8 txt)
    when (BS.length bytes /= 32) $ fail ...
    ...
```

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
```diff
@@ -0,0 +1,301 @@
...
+            where
+              keys, mandatory, optional', missing, unexpected :: Set Text
+              keys = Set.fromList $ HMap.keys v
+              mandatory = Set.fromList ["id", "ticker", "cost", "homepage", "margin"]
+              optional' = Set.fromList ["pledge_address"]
+              missing = mandatory `Set.difference` keys
+              unexpected = keys `Set.difference` (mandatory <> optional')
+
+              commaList :: Set Text -> String
+              commaList = List.intercalate ", " . (show . unpack <$>) . Set.toList
```
KtorZ: What about using [genericParseJSON](https://hackage.haskell.org/package/aeson-1.4.5.0/docs/Data-Aeson.html#v:genericParseJSON)? The structure looks pretty straightforward, so we could rely on generics here unless I am missing something :thinking: ?

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
```diff
@@ -0,0 +1,301 @@
...
+  parseJSON = AE.withText "Id" $
+    (either (fail . unpack) pure) . validatePublicKey
+
+instance FromJSON Ticker where
+  parseJSON = AE.withText "Ticker" $ \v ->
+    let s = unpack v
+    in if (length s >= 3 && length s <= 4
+           && all Char.isAsciiUpper s)
+    then pure $ Ticker v
+    else fail $ "Not an uppercase ASCII of length 3 or 4: " <> take 32 s
```
KtorZ: :+1:

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
```diff
@@ -0,0 +1,301 @@
...
+  parseJSON = AE.withText "Text" $ \v ->
+    case URI.parseAbsoluteURI (unpack v) of
+      Nothing -> fail $ "Not an absolute URI: " <> unpack v
+      Just x -> pure x
+
+instance FromJSON Cost where
+  parseJSON = _withBoundedScientific "Cost" $ \x ->
+    if not (x >= 0)
+    then fail "Value of cost must be: (x >= 0)"
+    else pure $ Cost x
```
KtorZ: Using `Natural` or `WordXX` would give us a generic instance for free :+1:

```hs
parseJSON = parseJSON >=> fmap Cost
```

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
```diff
@@ -0,0 +1,301 @@
...
+
+instance FromJSON Margin where
+  parseJSON = _withBoundedScientific "Margin" $ \x ->
+    if  not (x >= 0.0 && x <= 1.0)
+    then fail "Value of margin must be: (x >= 0.0 && x <= 1.0)"
+    else pure $ Margin x
+
+instance FromJSON PledgeAddress where
+  parseJSON = AE.withText "PledgeAddress" $ \v ->
+    case Bech32.decode v of
```
KtorZ: We will need [decodeLenient](https://hackage.haskell.org/package/bech32-1.0.0/docs/Codec-Binary-Bech32.html#v:decodeLenient) here! Cardano addresses are typically longer than 90 characters :man_shrugging:

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
deepfire: But `Natural` is integral, and the document didn't say it is, so a broader type was picked -- can we clarify that part?

Besides, we do validation, of course.

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
deepfire: Good point, will do!

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
deepfire: That would assume that there are no other duplicate tickers -- which is a flaky assumption I'd rather avoid.

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
deepfire: I'd say we want to be stricter than that -- we want to enforce that all:

1. filenames, and
2. `id` fields are lowercase

That simply avoids uniqueness issues.

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
deepfire: The choice to do manual decode is deliberate -- we want better error messages for human consumption.

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
deepfire: Again, the manual decoder is a deliberate choice for better human-oriendted messages.

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
deepfire: Ack, will do!

deepfire:
Will get to it tomorrow, on leave today.

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
KtorZ: I am not sure to follow you on this? Even if there are duplicates ticker already in the list, it will still work fine, no?

deepfire: You are suggesting to only validate the newly-added ticker for duplication -- that wouldn't find other duplicates, would it?

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
```diff
@@ -0,0 +1,301 @@
...
+    | parentDir /= "registry"] <>
+    ["Submission not properly relative the registry root"
+    | Path.dropTrailingPathSeparator root
+      /= Path.dropTrailingPathSeparator parentParentDir]
+
+  let (id, ext) = Path.splitExtension file
+      sigPath = dir </> id <.> "sig"
+
+  unlessM (lift $ Dir.doesFileExist sigPath) $
+    checks ["Missing signature file: " <> pack sigPath]
```
Jimbo4350: I think you could use `left` here (and in place of other `checks` that are meant to be failures): http://hackage.haskell.org/package/transformers-except-0.1.1/docs/Control-Monad-Trans-Except-Extra.html

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
Jimbo4350: You _may_ also be able to make use of `firstExceptT` in the above code.

cardano-node/src/Cardano/CLI/Run.hs
```diff
@@ -155,6 +156,14 @@ data ClientCommand
     TPSRate
     [SigningKeyFile]
 
+    --- Stake pool registry validation ----------
+
+  | ValidateSubmission
+    RegistryRoot
+    -- ^ Root directory of the registry.
```
Jimbo4350: :+1:

cardano-node/app/cardano-cli.hs
```diff
@@ -290,6 +292,21 @@ parseTxRelatedValues =
                   "Path to signing key file, for genesis UTxO using by generator."
       ]
 
+parseMiscCommands :: Parser ClientCommand
```
Jimbo4350: Should the README also be updated?

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
```diff
@@ -0,0 +1,301 @@
...
+validateRegistrySubmission
+  :: RegistryRoot
+  -> SubmissionFile
+  -> ExceptT CliError IO Submission
+validateRegistrySubmission (RegistryRoot root) (SubmissionFile fp) = do
+  -- Directory structure/access checks.
+  --
+  afp <- lift $ Dir.makeAbsolute fp
+  unlessM (lift $ Dir.doesFileExist afp) $
+    checks ["Submission doesn't exist: " <> pack afp]
```
Jimbo4350: Eg: 
```
-     checks ["Submission doesn't exist: " <> pack afp]
+     left $ PoolRegistryValidationErrors fp ["Submission doesn't exist: " <> pack afp]
```

Jimbo4350: You can maybe even get more specific with your error type `SubmissionDoesNotExist $ pack afp`. Up to you :).

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
```diff
@@ -0,0 +1,301 @@
...
+
+-- | Stake pool ticker, a 3-4 all-ASCII-uppercase character ID.
+newtype Ticker = Ticker { unTicker :: Text }
+  deriving (Eq, Ord, Show)
+
+-- | Non-negative Scientific.
+newtype Cost = Cost { unCost :: Sci.Scientific }
+
+-- | Rational in range [0, 1]
+newtype Margin = Margin { unMargin :: Sci.Scientific }
```
Jimbo4350: Pool Margin? Maybe add a comment to make this a bit more clear?

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
Jimbo4350: Cost of?

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
KtorZ: Indeed.

deepfire: So as I said, I intended to look for other duplicates too.

We already paid to the heavy cost of going to the kernel to find the files and open them and read them -- the extra check that there is no validation in the pre-existing set is peanuts, and it adds value -- I believe.

cardano-node/app/cardano-cli.hs
deepfire: @Jimbo4350, this code moves out to a different repo -- and there will be docs etc.

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
deepfire: Yes, I was weighing that option, indeed..

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
deepfire: I'd love to know myself :joy: -- no info in the requirements docs I have.

We should improve that before delivery.

cardano-node/src/Cardano/CLI/Delegation/PoolRegistry.hs
deepfire: Same as for the margin.

deepfire:
Ok, I think we can close this PR now.

Thanks everybody for the review!

The validator now lives in https://github.com/input-output-hk/testnet-stake-pool-registry-validator

deepfire:
Reopening, as this has non-validator changes that almost got lost : -)

PR #227, opened 2019-10-03
`cardano-cli`: drop `--tracing-off` & improve error printing

deepfire:
`cardano-cli`:

1. Drop `--tracing-off` -- it was implemented with a hard-to-fix bug.
2. Fix error printing -- it was over-quoting.


cardano-node/src/Cardano/Common/Parsers.hs
```diff
@@ -13,7 +13,7 @@ module Cardano.Common.Parsers
   ) where
 
 
-import           Prelude (String, error)
```
Jimbo4350: :+1:

PR #228, opened 2019-10-03
Remove remaining usage of `read`

Jimbo4350:
Entering an incorrectly formatted IP address for the `--host-addr` argument resulted in:
`cardano-node: Prelude.read: no parse`. 

Now there are more informative errors when there is an IP formatting error.

### DIFF
--- a/cardano-node/app/cardano-node.hs
+++ b/cardano-node/app/cardano-node.hs
@@ -4,13 +4,13 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RankNTypes #-}
 
-import           Prelude (String, read)
+import           Prelude (String)
 
 import qualified Data.IP as IP
 import           Data.Semigroup ((<>))
 import           Network.Socket (PortNumber)
-import           Options.Applicative ( Parser, auto, flag, help, long
-                                     , metavar, option, str, value
+import           Options.Applicative ( Parser, auto, eitherReader, flag,
+                                       help, long, metavar, option
                                      )
 import qualified Options.Applicative as Opt
 
@@ -28,7 +28,8 @@ import           Cardano.Shell.Types (CardanoApplication (..),
 
 import           Cardano.Config.CommonCLI
 import           Cardano.Common.Help
-import           Cardano.Common.Parsers
+import           Cardano.Common.CommandLineParsers
+import           Cardano.Common.Parsers (parseIP)
 import           Cardano.Node.Run
 import           Cardano.Node.Configuration.Topology (NodeAddress (..))
 import           Cardano.Tracing.Tracers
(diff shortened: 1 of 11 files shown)


cardano-node/src/Cardano/Common/NewParsers.hs
```diff
@@ -0,0 +1,153 @@
...
+-------------------------------------------------------------------------------
+-- IP parsers: Taken from http://hackage.haskell.org/package/iproute-1.7.7/docs/Data-IP.html
+-------------------------------------------------------------------------------
+
+parseIP :: String -> Either String IP
+parseIP cs = case A.runParser ip4 cs of
+    (Just ip,_) -> Right $ IPv4 ip
+    (Nothing,_) -> case A.runParser ip6 cs of
+        (Just ip,_) -> Right $ IPv6 ip
+        (Nothing,rest) -> Left $ "IP parsing error: " <> rest
```
Jimbo4350: Substitute `rest` with `cs`

cardano-node/app/cardano-node.hs
```diff
@@ -155,13 +156,12 @@ parseNodeArgs =
 parseNodeAddress :: Parser NodeAddress
 parseNodeAddress = NodeAddress <$> parseHostAddr <*> parsePort
 
-parseHostAddr :: Parser (Maybe IP.IP)
```
deepfire: Do we really want to make the `--host-addr` option mandatory?

I think recent work by @karknu was about going in the opposite direction?

cardano-node/src/Cardano/Common/CommandLineParsers.hs
```diff
@@ -0,0 +1,364 @@
+{-# LANGUAGE RankNTypes #-}
```
deepfire: We have apparent semantic changes in this PR @Jimbo4350, and I'm afraid with this code motion it's hard to see the picture..

I'd rather see semantics and code motion in separate PRs, since their conflation makes review harder -- at least the whole picture.

cardano-node/app/cardano-node.hs
karknu: --host-addr should not be mandatory, keeping it blank makes it possible to have a node who use both IPv4 and IPv6 addresses at the same time.
Users never had to specify the source address when running cardano-sl.

Jimbo4350: @karknu  I have to re-implement this with `attoparsec` so I will undo the change.

cardano-node/src/Cardano/Common/CommandLineParsers.hs
Jimbo4350: @deepfire this was due to the renaming, I'll fix it when I re-implement the parser with `attoparsec`

Jimbo4350:
Closed in favour of #256

Issue #233, opened 2019-10-07
Useful tracing of TraceChainSyncServerEvent


mrBliss:
The following TODO can be resolved as https://github.com/input-output-hk/ouroboros-network/pull/1032 has been merged: https://github.com/input-output-hk/cardano-node/blob/3276f8b115cd802828cff36f7291cffa3a9f393e/cardano-node/src/Cardano/Tracing/ToObjectOrphans.hs#L663
Now, it just traces `ChainSyncServerEvent`, which is pretty useless :slightly_smiling_face:.

Jimbo4350:
Closing due to #234.

PR #234, opened 2019-10-07
Reduce severity of slot number and implement `TraceChainSyncServerEvent`'s `ToObject` instance

Jimbo4350:
This PR changes the following

- Slot number was being reported as `Critical` severity and is now reduced to `Debug`
- Implements `TraceChainSyncServerEvent`'s `ToObject` instance #233

### DIFF
--- a/cardano-node/src/Cardano/Tracing/ToObjectOrphans.hs
+++ b/cardano-node/src/Cardano/Tracing/ToObjectOrphans.hs
@@ -35,7 +35,7 @@ import           Ouroboros.Consensus.BlockFetchServer
 import           Ouroboros.Consensus.ChainSyncClient
                    (TraceChainSyncClientEvent (..))
 import           Ouroboros.Consensus.ChainSyncServer
-                   (TraceChainSyncServerEvent)
+                   (TraceChainSyncServerEvent(..))
 import           Ouroboros.Consensus.Ledger.Abstract
 import           Ouroboros.Consensus.Mempool.API (GenTx, GenTxId)
 import qualified Ouroboros.Consensus.Node.Tracers as Consensus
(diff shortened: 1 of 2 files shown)


disassembler:
this should be debug unless it's switched to only print if it's either fully synced, or summary output at end of epoch.

cardano-node/src/Cardano/Tracing/ToObjectOrphans.hs
```diff
@@ -658,22 +658,18 @@ instance (Condense (HeaderHash blk), ProtocolLedgerView blk, SupportedBlock blk,
...
-  --     mkObject [ "kind" .= String "ChainSyncServerEvent.TraceChainSyncServerReadBlocked"
-  --              , "block" .= toObject verb pt ]
+instance StandardHash blk => ToObject (TraceChainSyncServerEvent blk b) where
+    toObject _ ev = case ev of
+      TraceChainSyncServerRead tip _ ->
+        mkObject [ "kind" .= String "ChainSyncServerEvent.TraceChainSyncServerRead"
+                 , "block" .= (String (pack . show $ tipPoint tip)) ]
+      TraceChainSyncServerReadBlocked tip _ ->
+        mkObject [ "kind" .= String "ChainSyncServerEvent.TraceChainSyncServerReadBlocked"
+                 , "block" .= (String (pack . show $ tipPoint tip)) ]
```
mrBliss: This is incorrect (same on line 665), the `tip` reported by the `TraceChainSyncServerRead(Blocked)` is the current tip of the server's chain. The actually interesting bit is the `ChainUpdate`, which is either a new block that is added or a roll back to a point in the past.

Jimbo4350: @mrBliss  I'll make the change in a follow up PR

Jimbo4350: https://github.com/input-output-hk/cardano-node/pull/238

PR #237, opened 2019-10-08
Tx gen fix: check target node IP addrFamily

denisshevchenko:
Previously I've hardcoded `AF_INET6`, because of local cluster settings. Now we check IP address family explicitly.

### DIFF
--- a/cardano-node/src/Cardano/CLI/Tx/Generation.hs
+++ b/cardano-node/src/Cardano/CLI/Tx/Generation.hs
@@ -37,6 +37,7 @@ import           Data.ByteString (ByteString)
 import qualified Data.ByteString.Lazy as LB
 import           Data.Either (isLeft)
 import           Data.Foldable (find, foldl', foldr, toList)
+import qualified Data.IP as IP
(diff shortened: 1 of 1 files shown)


cardano-node/src/Cardano/CLI/Tx/Generation.hs
```diff
@@ -646,15 +647,18 @@ runBenchmark benchTracer
...
 
-  let targetNodeHost = maybe (panic "Target node's IP-address is undefined!")
-                             show
-                             $ naHostAddress targetNodeAddress
-      targetNodePort = show $ naPort targetNodeAddress
+  let (anAddrFamily, targetNodeHost) =
+        case naHostAddress targetNodeAddress of
+          Just (IP.IPv4 ipv4) -> (AF_INET,  show ipv4)
+          Just (IP.IPv6 ipv6) -> (AF_INET6, show ipv6)
+          Nothing -> panic "Target node's IP-address is undefined!"
```
Jimbo4350: :+1:

PR #240, opened 2019-10-09
add a daedalus-bridge attribute

cleverca22:


disassembler:
This should be named something other than daedalus bridge because it would only be ran by people using the node as a stake pool (or just playing around with it) on Windows. The actual daedalus-bridge will go in the wallet.

PR #246, opened 2019-10-11
Update dependencies

mrBliss:
In particular: cardano-ledger#660


mrBliss:
> LGTM, scripts appear to be working. I would let @deepfire look at the `txSpendUTxOByronPBFT` before merging.

Can somebody push this forward? I'm afraid somebody is soon going to try to update the dependencies, run into the compilation errors I fix in this PR, try to fix them themselves, and finally ask on Slack about it. This happened a few times already :grin:.

PR #249, opened 2019-10-14
Finish implementation of partial options monoid pattern in `cardano-node.hs`

Jimbo4350:
A continuation of https://github.com/input-output-hk/cardano-node/pull/243


cardano-config/src/Cardano/Config/Logging.hs
```diff
@@ -149,26 +125,25 @@ data LoggingFlag = LoggingEnabled | LoggingDisabled
 -- | Interpret main logging CLI controls into a tuple of:
 --   - a designation of whether logging was disabled,
 --   - a valid 'LoggingConfiguration' (still necessary, even if logging was disabled)
-loggingCLIConfiguration :: LoggingCLIArguments -> IO (LoggingFlag, LoggingConfiguration)
-loggingCLIConfiguration  lca@LoggingCLIArguments{logConfigFile = Nothing} =
+loggingCLIConfiguration :: Maybe FilePath -> Bool -> IO (LoggingFlag, LoggingConfiguration)
```
Jimbo4350: I'm thinking of implementing a sum type for `captureMetrics` (isomorphic to bool)

cardano-config/src/Cardano/Config/Presets.hs
```diff
@@ -35,6 +39,12 @@ mainnetConfiguration =
     , pccLogConfig = pure "./configuration/log-configuration.yaml"
     , pccDBPath = pure "./db/"
     , pccApplicationLockFile = pure ""
```
Jimbo4350: Comments on sensible defaults? I would actually like to further separate the `cardano-node` and `cardano-cli`'s configurations in a follow up PR. Essentially the same POM design pattern in each separately. I don't see why they have to share the same configuration. @deepfire

cardano-node/app/trace-acceptor.hs
```diff
@@ -28,7 +28,11 @@ main = do
                    Left err -> throwIO err
                    Right x -> pure x
   (,) loggingLayer
-      loggingFeature <- createLoggingFeature env finalConfig loggingCLI
+      loggingFeature <- createLoggingFeature
+                          env
+                          finalConfig { ccLogConfig = logConfigFile loggingCLI
+                                      , ccLogMetrics = captureMetrics loggingCLI
+                                      }
```
Jimbo4350: Temporary fix. Will smooth out in a follow up PR.

cardano-config/src/Cardano/Config/Orphanage.hs
```diff
@@ -1,14 +1,85 @@
-{-# LANGUAGE StandaloneDeriving #-}
```
deepfire: @Jimbo4350, the `*.Orphanage` modules are supposedly intended for orphan instances, not for just about everything -- this is so they can:

1. have `-Wno-orphans`, and
2. be imported as `import *.Orphanage ()`

I think this is confusing this purpose..

cardano-config/src/Cardano/Config/Orphanage.hs
```diff
@@ -1,14 +1,85 @@
...
+    (Const True)
+
+protocolTraceDefault :: ProtocolTraceOptions
+protocolTraceDefault =
+  ProtocolTracers
+    (Const True)
+    (Const True)
+    (Const True)
+    (Const True)
+    (Const True)
```
deepfire: Do we really want the default trace options trace _everything_?

My impression was that the defaults should be somewhat reasonable, or at least not overwhelming.. : -)

cc @CodiePP

cardano-config/src/Cardano/Config/Presets.hs
deepfire: I'd say that it's hard to reason about this generally.

Let's decide on a case-by-basis -- sometimes sharing makes sense, sometimes it doesn't.

cardano-config/src/Cardano/Config/Protocol.hs
```diff
@@ -27,32 +29,44 @@ import qualified Cardano.Chain.Update as Update
...
--- TODO:  we don't want ByronLegacy in Protocol.  Let's wrap Protocol with another
--- sum type for cases where it's required.
-data Protocol =
-    ByronLegacy
-  | BFT
-  | Praos
-  | MockPBFT
-  | RealPBFT
-  deriving Show
-
```
deepfire: :+1:

cardano-node/app/cardano-node.hs
```diff
@@ -123,24 +117,22 @@ initializeAllFeatures (NodeCLI topInfo nodeAddr protocol vMode logCLI traceOpts
...
--- TODO: Condense `NodeCLI` into one big `PartialCardanoConfiguration`
-data NodeCLI = NodeCLI
-                !TopologyInfo
-                !NodeAddress
-                !Protocol
-                !ViewMode
-                !LoggingCLIArguments
-                !TraceOptions
-                !PartialCardanoConfiguration
+data NodeCLI = NodeCLI !PartialCardanoConfiguration
```
deepfire: Now, do we need this ADT anymore? : -)

cardano-node/src/Cardano/Common/Parsers.hs
```diff
@@ -56,20 +63,46 @@ parseProtocol :: Parser Protocol
...
+  flag' (Last $ Just MockPBFT) $ mconcat [ long "mock-pbft"
+                                         , help "Permissive BFT consensus with a mock ledger"
+                                         ]
+
+parseProtocolRealPBFT :: Parser (Last Protocol)
+parseProtocolRealPBFT =
+  flag' (Last $ Just RealPBFT) $ mconcat [ long "real-pbft"
+                                         , help "Permissive BFT consensus with a real ledger"
+                                         ]
+
```
deepfire: Now, this is a very sparse encoding of something that is really very simple and compact -- and this sparseness is really hiding the simplicity.

Do you think we can optimise the presentation here?

cardano-config/src/Cardano/Config/Orphanage.hs
Jimbo4350: @deepfire Agreed. Let me think about where to put some of this stuff.

cardano-node/app/cardano-node.hs
Jimbo4350: @deepfire If we decide that `cardano-cli` and `cardano-node` should have separate configurations then no. I'm leaning towards this and I was planning to remove it in a follow up PR.

cardano-node/src/Cardano/Common/Parsers.hs
Jimbo4350: @deepfire Yep this can be condensed. I'll use `fl`.

cardano-config/src/Cardano/Config/Presets.hs
Jimbo4350: I strongly disagree with sharing configuration types between `cardano-cli` and `cardano-node`.  I believe @erikd made a comment expressing a similar view. Having simpler cli parsers gives each cli more flexibility and less coupling.

cardano-node/src/Cardano/Common/Parsers.hs
Jimbo4350: @deepfire pushed changes

cardano-config/src/Cardano/Config/Orphanage.hs
Jimbo4350: @deepfire Pushed changes.

cardano-config/src/Cardano/Config/Presets.hs
deepfire: @Jimbo4350, so you say we don't want to share `CardanoConfiguration`, right?

cardano-node/src/Cardano/Common/Parsers.hs
```diff
@@ -49,33 +57,69 @@ parseNodeId desc =
     )
 
 -- | Flag parser, that returns its argument on success.
-fl :: a -> String -> String -> Parser a
-fl val opt desc = flag' val $ mconcat [long opt, help desc]
+flagParser :: a -> String -> String -> Parser a
```
ksaric: You have `lastStrOption`, and you have others you can check here - https://github.com/input-output-hk/cardano-node/blob/7c1ba8014ba62ec562dd2cd538a6f769e148c30a/cardano-config/src/Cardano/Config/CommonCLI.hs#L218-L223

cardano-node/src/Cardano/Common/Parsers.hs
```diff
@@ -49,33 +57,69 @@ parseNodeId desc =
...
+    "BFT consensus"
+  , flagParser Praos "praos"
+    "Praos consensus"
+  , flagParser MockPBFT "mock-pbft"
+    "Permissive BFT consensus with a mock ledger"
+  , flagParser RealPBFT "real-pbft"
+    "Permissive BFT consensus with a real ledger"
+  ]
+
+parseProtocolByron :: Parser (Last Protocol)
```
ksaric: Not really sure why you need this and the `parseProcotol` separately, really.

cardano-node/src/Cardano/Common/Parsers.hs
```diff
@@ -104,6 +148,21 @@ parseTopologyFile =
...
+  strOption
+    ( long "log-config"
+    <> metavar "LOGCONFIG"
+    <> help "Configuration file for logging"
+    <> completer (bashCompleter "file")
+    )
+
+parseLogMetrics :: Parser (Last Bool)
+parseLogMetrics =
+  (Last . Just) <$> switch
```
ksaric: Existing functions like `lastBoolOption`? Formatting? :pray:

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -87,72 +87,59 @@ instance NoUnexpectedThunks Peer where
...
-  -> LoggingLayer
-  -> TraceOptions
+  :: LoggingLayer
   -> CardanoConfiguration
   -> IO ()
-runNode topology myNodeAddress protocol viewMode loggingLayer traceOptions cc = do
+runNode loggingLayer cc = do
     let !tr = llAppendName loggingLayer "node" (llBasicTrace loggingLayer)
-    let trace'      = appendName (pack $ show $ node topology) tr
+    let trace'      = appendName (pack $ show $ node $ ccTopologyInfo cc) tr
```
ksaric: `toS` maybe from `Cardano.Prelude`?

scripts/lib-node.sh
```diff
@@ -39,6 +43,6 @@ function nodeargs () {
         printf -- "--pbft-signature-threshold 0.7 "
         printf -- "--require-network-magic "
         printf -- "--database-path db "
-        printf -- "--topology configuration/simple-topology.json "
+        topology
```
ksaric: Erm. Ok, nice, but not really sure if this is actually doing something?

cardano-node/app/cardano-node.hs
```diff
@@ -215,9 +229,9 @@ parsePort =
     )
 
 -- Optional flag for live view (with TUI graphics).
-parseViewMode :: Parser ViewMode
+parseViewMode :: Parser (Last ViewMode)
 parseViewMode =
-    flag SimpleView LiveView $ mconcat
+    flag (Last $ Just SimpleView) (Last $ Just LiveView) $ mconcat
```
ksaric: This looks kind of wordy, maybe try using existing functions from `CommonCLI`?

cardano-node/app/cardano-node.hs
```diff
@@ -152,33 +144,54 @@ nodeCliParser = do
...
+    traceOptions
     slotLength = mempty { pccDBPath = dbPath
+                        , pccNodeAddress = nAddr
                         , pccSocketDir = socketDir
+                        , pccTopologyInfo = topInfo
+                        , pccProtocol = ptcl
+                        , pccViewMode = vMode
+                        , pccLogConfig = logConfigFp
+                        , pccLogMetrics = logMetrics
                         , pccCore = mempty { pcoGenesisFile = genPath
```
ksaric: This looked kind of strange the first time I saw it, but now (since you kept adding things to it) looks even stranger.
Doesn't `mergeConfiguration` does most of these things?

cardano-config/src/Cardano/Config/Presets.hs
Jimbo4350: @deepfire Exactly.

cardano-config/src/Cardano/Config/Defaults.hs
```diff
@@ -0,0 +1,48 @@
...
+    (Const True)
+    (Const True)
+    (Const True)
+    (Const True)
+    (Const True)
+
+traceOptionsDefault :: TraceOptions
+traceOptionsDefault =
+  TraceOptions
+   { traceVerbosity = NormalVerbosity
```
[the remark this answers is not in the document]
Jimbo4350: @ksaric fixed

cardano-config/src/Cardano/Config/Logging.hs
```diff
@@ -149,26 +125,25 @@ data LoggingFlag = LoggingEnabled | LoggingDisabled
...
+loggingCLIConfiguration Nothing captureMetrics' =
   (LoggingDisabled,)
```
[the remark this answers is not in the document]
Jimbo4350: @ksaric I took the liberty of making it even clearer (IMO :slightly_smiling_face: )

cardano-config/src/Cardano/Config/Partial.hs
```diff
@@ -41,6 +42,12 @@ data PartialCardanoConfiguration = PartialCardanoConfiguration
     , pccDBPath              :: !(Last FilePath)
     , pccSocketDir           :: !(Last FilePath)
     , pccApplicationLockFile :: !(Last FilePath)
+    , pccTopologyInfo        :: !(Last TopologyInfo)
```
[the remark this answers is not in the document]
Jimbo4350: @ksaric Could you elaborate? Keep them more organized how?

scripts/lib-node.sh
Jimbo4350: @ksaric For some reason the script was not registering the `--topology` argument. This change fixed that for some reason :man_shrugging:.

cardano-config/src/Cardano/Config/Orphanage.hs
```diff
@@ -1,14 +1,32 @@
-{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GeneralisedNewtypeDeriving #-}
```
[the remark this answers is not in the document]
ksaric: Weird, but ok.

cardano-config/src/Cardano/Config/Partial.hs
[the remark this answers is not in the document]
ksaric: I mean, maybe some of these added params could be placed into some of the `Partial` types. For example, the `TopologyInfo` looks like something for networking (`PartialNetworkInfo`?).
Then you could move some of the things inside. 
And like that for other params.

cardano-node/src/Cardano/Node/Run.hs
Jimbo4350: @ksaric  unfortunately `No instance for (StringConv NodeId Cardano.BM.Data.LogItem.LoggerName)` when using `toS`

ksaric: I meant replace `pack $ show` with `toS`.

cardano-node/src/Cardano/Common/Parsers.hs
Jimbo4350: We want the various protocols to be flags though. `--praos` or `--real-pbft` is more succinct than say `--protocol praos`

cardano-node/app/cardano-node.hs
Jimbo4350: @ksaric I will move `flagParser` to `CommonCLI` and import it here.

cardano-node/src/Cardano/Node/Run.hs
Jimbo4350: @ksaric I did and ghc complained with the above error.

cardano-node/src/Cardano/Common/Parsers.hs
Jimbo4350: @ksaric I will clean up and condense all the of `Last` parsing functions in a follow up PR.

scripts/lib-node.sh
Jimbo4350: @ksaric You're right, I reverted and it still works. Strange.

cardano-config/src/Cardano/Config/Partial.hs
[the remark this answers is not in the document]
Jimbo4350: @ksaric If I understand you correctly you want to factor out things e.g all network related things into a `PartialNetworkInfo` ADT. The issue with this is that you would then need to write a custom `Monoid` instance for a "super type".

E.g: `SuperPartialCardanoConfiguration = PartialCardanoConfiguration | PartialNetworkInfo | ...`
And then you would need to define a custom `Monoid` instance to facilitate the following: `PartialCardanoConfiguration <> PartialNetworkInfo <> ...`.

At least at this stage I think this complicates matters and doesn't buy us much.

cardano-node/app/cardano-node.hs
Jimbo4350: @ksaric https://github.com/input-output-hk/cardano-node/pull/249#discussion_r335533540

cardano-node/src/Cardano/Node/Run.hs
ksaric: Ah, yes, missing `show`, my bad. Should be `toS == pack`.

Jimbo4350:
> Most of my comments are going to be addressed in the next PR, so I guess, LGTM?

In order to reduce the noise in this PR the remaining outstanding issues will be addressed in a follow up PR:
https://github.com/input-output-hk/cardano-node/pull/249#discussion_r335387320
https://github.com/input-output-hk/cardano-node/pull/249#discussion_r335385824
https://github.com/input-output-hk/cardano-node/pull/249#discussion_r335385414

Essentially refactoring and movement of parsers.

cardano-config/src/Cardano/Config/Orphanage.hs
Jimbo4350: @deepfire Currently even when I don't list a `--log-config` option and I make the default values all `False` the logging output does not change. I will have to investigate this in a separate PR.

PR #251, opened 2019-10-15
Update ouroboros-network dependency in cardano-node.

denisshevchenko:


mrBliss:
#254 will subsume this PR.

PR #252, opened 2019-10-15
Remove `pvss` dependency

Jimbo4350:
#242


cardano-node/src/Cardano/CLI/Legacy/Byron.hs
```diff
@@ -85,9 +63,9 @@ matchSize requestedSize lbl actualSize =
 -- | Encoder for a Byron/Classic signing key.
 --   Lifted from cardano-sl legacy codebase.
 encodeLegacyDelegateKey :: LegacyDelegateKey -> E.Encoding
-encodeLegacyDelegateKey dk@LegacyDelegateKey{lrkSigningKey=(SigningKey sk)}
+encodeLegacyDelegateKey (LegacyDelegateKey (SigningKey sk))
   =  E.encodeListLen 4
-  <> E.encodeListLen 1 <> encodeBinary (lrkVSSKeyPair dk)
+  <> E.encodeListLen 1 <> E.encodeNull
```
mrBliss: Does this encoder need to backwards compatible with existing code? If so, I'm not sure it will be.

Jimbo4350: @mrBliss If I understand correctly, I should use a dummy value instead of a null value here?

deepfire: @Jimbo4350, we need to verify that key migration still works -- just take a delegate key from `cardano-sl` and feed it to the migration subcommand.

Jimbo4350: @mrBliss @deepfire I can confirm key migration still works.

deepfire: Thank you!

PR #254, opened 2019-10-16
Configurable Protocol Magic

karknu:
Breaks backwards compatibility.
Depends on https://github.com/input-output-hk/ouroboros-network/pull/1142/

### DIFF
--- a/cardano-node/src/Cardano/CLI/Tx.hs
+++ b/cardano-node/src/Cardano/CLI/Tx.hs
@@ -43,10 +43,10 @@ import qualified Cardano.Crypto.Hashing as Crypto
 import qualified Cardano.Crypto.Signing as Crypto
 import qualified Ouroboros.Consensus.Demo.Run as Demo
 import qualified Ouroboros.Consensus.Ledger.Byron as Byron
-import           Ouroboros.Consensus.Ledger.Byron ( GenTx(..), GenTxId(..)
-                                                  , ByronBlockOrEBB)
+import           Ouroboros.Consensus.Ledger.Byron ( GenTx(..),ByronBlockOrEBB)
 import           Ouroboros.Consensus.Ledger.Byron.Config (ByronConfig)
 import qualified Ouroboros.Consensus.Protocol as Consensus
+import           Ouroboros.Consensus.Mempool.API (txId)
 
 import           Cardano.CLI.Ops
 import           Cardano.CLI.Tx.Submission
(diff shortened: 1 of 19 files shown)


cardano-node/src/Cardano/CLI/Tx.hs
```diff
@@ -144,6 +144,15 @@ withRealPBFT co cc action = do
     proto@Consensus.ProtocolRealPBFT{} -> action proto
     _ -> throwIO $ ProtocolNotSupported (coProtocol co)
 
+-- | Return an id of GenTx.
+byronGenTxId
```
mrBliss: @denisshevchenko This function exists already, it is called `txId`

karknu: I've modified the code to use `txId`.

cardano-node/src/Cardano/CLI/Tx.hs
```diff
@@ -43,10 +43,10 @@ import qualified Cardano.Crypto.Hashing as Crypto
 import qualified Cardano.Crypto.Signing as Crypto
 import qualified Ouroboros.Consensus.Demo.Run as Demo
 import qualified Ouroboros.Consensus.Ledger.Byron as Byron
-import           Ouroboros.Consensus.Ledger.Byron ( GenTx(..), GenTxId(..)
-                                                  , ByronBlockOrEBB)
+import           Ouroboros.Consensus.Ledger.Byron ( GenTx(..),ByronBlockOrEBB)
```
mrBliss:
```suggestion
import           Ouroboros.Consensus.Ledger.Byron (GenTx(..), ByronBlockOrEBB)
```

karknu:
This PR MUST be merged together with updates in `cardano-byron-broxy` and `cardano-explorer` since it breaks backwards compatibility.

mrBliss:
> This PR MUST be merged together with updates in `cardano-byron-broxy` and `cardano-explorer` since it breaks backwards compatibility.

Is that true for the `cardano-byron-proxy`? It does not depend on `cardano-node` (but on `ouroboros-network`).

karknu:
If `cardano-byron-proxy` isn't updated at the same time `cardano-node` would not be able to sync with it which would make for unhappy testers and developers who depend on it.

mrBliss:
> If `cardano-byron-proxy` isn't updated at the same time `cardano-node` would not be able to sync with it which would make for unhappy testers and developers who depend on it.

Ah, indeed, good point :+1:

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -182,18 +182,19 @@ handleSimpleNode p trace nodeTracers cc = do
...
+          , rnaDnsResolverTracer       = dnsResolverTracer     nodeTracers
+          , rnaMuxTracer               = muxTracer             nodeTracers
+          , rnaMuxLocalTracer          = nullTracer
+          , rnaMkPeer                  = Peer
+          , rnaMyAddrs                 = addrs
+          , rnaMyLocalAddr             = myLocalAddr
+          , rnaIpProducers             = ipProducers
+          , rnaDnsProducers            = dnsProducers
+          , rnaHandshakeTracer         = nullTracer
+          , rnaHandshakeLocalTracer    = nullTracer
```
mrBliss: There are some extra unnecessary spaces before the `=`. These changes can be undone, only the line below needs to be added.

PR #256, opened 2019-10-17
Replace `iproute` with `ip`

Jimbo4350:
- `iproute` uses `read` to parse IPs (this results in poor errors).
- `ip` has a more suitable `decode` function that at least returns a `Maybe`.
- We want to use one ip package


Jimbo4350:
@deepfire You're right. I will then need to remove `ip` from the various dependencies.

PR #258, opened 2019-10-21
Remove `CLIOps` datatype

Jimbo4350:
Having several modular pure functions is preferable to the larger "swiss army knife" style `CLIOps m`.

### DIFF
--- a/cardano-node/app/cardano-cli.hs
+++ b/cardano-node/app/cardano-cli.hs
@@ -37,13 +37,11 @@ import           Cardano.Config.Protocol
 import           Cardano.CLI.Genesis
 import           Cardano.CLI.Key
 import           Cardano.Config.Logging (createLoggingFeature)
-import           Cardano.CLI.Ops (decideCLIOps)
 import           Cardano.CLI.Run
 
 main :: IO ()
 main = do
   co <- Opt.customExecParser pref opts
-  ops <- decideCLIOps (protocol co)
   -- Initialize logging layer. Particularly, we need it for benchmarking (command 'generate-txs').
   let cardanoConfiguration = mainnetConfiguration
   let cardanoEnvironment   = NoEnvironment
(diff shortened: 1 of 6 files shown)


cardano-node/app/cardano-cli.hs
```diff
@@ -52,8 +50,10 @@ main = do
     finalConfig <- withExceptT ConfigError $ ExceptT $ pure $
       mkConfiguration cardanoConfiguration (commonCli co) (commonCliAdv co)
     (loggingLayer, _loggingFeature) <- liftIO $
-      createLoggingFeature cardanoEnvironment (finalConfig {ccLogConfig = logConfigFile $ loggingCli co})
-    (runCommand ops finalConfig loggingLayer (mainCommand co) :: ExceptT CliError IO ())
+      createLoggingFeature cardanoEnvironment (finalConfig { ccLogConfig = logConfigFile $ loggingCli co
+                                                           , ccProtocol = protocol co
+                                                           })
```
Jimbo4350: Temporary fix. Will be smoothed out in a follow up PR.

deepfire:
> Having several modular pure functions is preferable to the larger "swiss army knife" style `CLIOps m` data type that has to be instantiated in `IO`.

Can you explain more about this preference, @Jimbo4350 ?

From my perspective, we previously had a pre-decided dictionary, and one had to go to extra lengths to get behavior for a different protocol -- you couldn't pick up the methods directly.

So I think we actually slightly regressed the modularity / separation of concerns here, because the user code now has to ensure it picks the correct protocol value, which is a simple enumeration type.

Also, regarding purity -- If we wanted to switch to pure functions, we could have simply changed the signature of the particular methods -- the definitions of the functions were already pure, in fact.

erikd:
@deepfire Avoiding the swiss army knife approach is needed because the explorer node also needs to use this config code and the explorer does not need the vast majority of the config options the node uses. Having the swiss army knife approach makes is a huge pain in the neck for the explorer. If on the other hand the node uses a bunch of smaller configuration options, then the explorer can use the ones it needs and ignore the rest.

Jimbo4350:
> So I think we actually slightly regressed the modularity / separation of concerns here, because the user code now has to ensure it picks the correct protocol value, which is a simple enumeration type.

I think this is a reasonable requirement in that the user should specify the right protocol to use the relevant functions (`serialiseSigningKey`, `serialiseDelegateKey` etc)



> can I ask you to comment on how this is related to the Partial Options Monoid?
It's not directly related. I'll revise the commit message.

Jimbo4350:
The `chairman` continues to fail in CI. I'm still not sure why so I have re-opened this PR here: https://github.com/input-output-hk/cardano-node/pull/266

PR #260, opened 2019-10-22
Tx generator: tx additional size.

denisshevchenko:
Now it's possible to define additional size of generated transaction without increasing the number of inputs/outputs.

### DIFF
--- a/cardano-node/app/cardano-cli.hs
+++ b/cardano-node/app/cardano-cli.hs
@@ -286,6 +286,11 @@ parseTxRelatedValues =
             <*> parseTPSRate
                   "tps"
                   "TPS (transaction per second) rate."
+            <*> optional (
+                  parseTxAdditionalSize
+                    "add-tx-size"
+                    "Additional size of transaction, in bytes."
+                )
             <*> parseSigningKeysFiles
                   "sig-key"
                   "Path to signing key file, for genesis UTxO using by generator."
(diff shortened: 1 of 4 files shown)


cardano-node/src/Cardano/CLI/Tx/Generation.hs
```diff
@@ -115,6 +116,10 @@ newtype TPSRate =
   TPSRate Int
   deriving (Eq, Ord, Show)
 
+newtype TxAdditionalSize =
```
Jimbo4350: Can you add a comment elaborating what `TxAdditionalSize` is?

cardano-node/app/cardano-cli.hs
```diff
@@ -286,6 +286,11 @@ parseTxRelatedValues =
             <*> parseTPSRate
                   "tps"
                   "TPS (transaction per second) rate."
+            <*> optional (
+                  parseTxAdditionalSize
+                    "add-tx-size"
+                    "Additional size of transaction (optional value)."
```
deepfire: I think it's extraneous: `(optional value)`.  The help syntax already shows it's optional.

denisshevchenko: Yep, you're right!

cardano-node/src/Cardano/CLI/Tx/Generation.hs
denisshevchenko: Yes, I'll add.

cardano-node/src/Cardano/CLI/Tx/Generation.hs
```diff
@@ -507,11 +517,49 @@ mkTransaction cfg ((txinFrom, txoutFrom), signingKey) mChangeAddress payments =
...
+  emptyAttributes :: CC.UTxO.TxAttributes
+  emptyAttributes = CC.Common.mkAttributes ()
+
+  blobAttributes :: Int -> CC.UTxO.TxAttributes
+  blobAttributes aSize =
+    emptyAttributes {
+      CC.Common.attrRemain = CC.Common.UnparsedFields $
+        Map.singleton k $ LB.replicate finalSize byte
+    }
+   where
```
Jimbo4350: I think removing this nested `where` clause by converting `finalSize` to 
```
finalSize :: Int -> Int64
finalSize aSize = fromIntegral (aSize - sizeOfKey)
```
would make this a lot neater :slightly_smiling_face:

denisshevchenko: Fixed.

PR #267, opened 2019-10-25
Implement POM design pattern in `cardano-cli.hs`

Jimbo4350:
Following from #249 & #243

### DIFF
--- a/cardano-node/app/cardano-cli.hs
+++ b/cardano-node/app/cardano-cli.hs
@@ -1,5 +1,3 @@
-{-# LANGUAGE ApplicativeDo #-}
-
 import           Cardano.Prelude hiding (option)
 import           Prelude (String, error, id)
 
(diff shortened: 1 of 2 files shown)


cardano-node/app/cardano-cli.hs
```diff
@@ -76,21 +84,81 @@ main = do
     renderCliError = show
 
 data CLI = CLI
-  { protocol    :: Protocol
+  { totalConfig :: PartialCardanoConfiguration
```
deepfire: Hmm, `totalConfig` vs.  `Partial*` type? : -)

What is our picture here?  Maybe we want a better name?

cardano-node/app/cardano-cli.hs
```diff
@@ -76,21 +84,81 @@ main = do
...
+  socketDir <- parseSocketDir
+  pbftSigThresh <- parsePbftSigThreshold
+  reqNetMagic <- parseRequireNetworkMagic
+  slotLength <- parseSlotLength
+  logConfigFp <- lastOption parseLogConfigFile
+  logMetrics <- parseLogMetrics
+  pure $ CLI (createPcc ptcl dbPath genPath genHash delCert sKey socketDir
+                        pbftSigThresh reqNetMagic slotLength logConfigFp
+                        logMetrics)
+              cmd
```
deepfire: Don't you feel that with the `ApplicativeDo`'s introduction of the unnecessary names we've lost some of the clarity here?

I think previously it was much easier on the eyes..

cardano-node/src/Cardano/Common/Parsers.hs
```diff
@@ -13,8 +13,6 @@ module Cardano.Common.Parsers
   , parseProtocolMockPBFT
   , parseProtocolPraos
   , parseProtocolRealPBFT
-  , parseProtocolActual
-  , parseProtocolAsCommand
```
deepfire: :+1:

cardano-node/app/cardano-cli.hs
Jimbo4350: @deepfire I'm not opposed to removing `ApplicativeDo`. You have a good point.

cardano-node/app/cardano-cli.hs
Jimbo4350: @deepfire Yep, this was a poor name choice. I'll change it.

cardano-node/app/cardano-cli.hs
Jimbo4350: @deepfire Thinking about it some more, the issue here is then I would need `CLI Protocol FilePath...` etc. `ApplicativeDo` allows me to keep the `CLI` datatype more concise.

deepfire: Just pushed an example of how to do it without any type changes: https://github.com/input-output-hk/cardano-node/commit/a86716311c0d74860140fe61beda91a4565e4da1#diff-deb1de1e23655df9067b46df271864c6L109

PR #270, opened 2019-10-29
Implement ability to turn individual tracers on/off

Jimbo4350:
This PR adds the following functionality:
- Not specifying a tracer on the command line will now turn it off.
-  `cardano-cli` mutes all tracing by default

Relevant: #409


cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -129,31 +125,40 @@ mkTracers :: forall peer blk.
            -> Tracers peer blk
 mkTracers traceOptions tracer = Tracers
     { chainDBTracer
-        = annotateSeverity $ filterSeverity (pure . const (tracingSeverity $ traceChainDB traceOptions))
-          $ teeTraceChainTip (tracingFormatting $ traceChainDB traceOptions) tracingVerbosity
-          $ addName "ChainDB" tracer
+        = tracerOnOff (traceChainDB traceOptions)
+            $ annotateSeverity $ filterSeverity (pure . const (tracingSeverity $ traceChainDB traceOptions))
```
CodiePP: filtering based on `Severity` (tracingSeverity) and controlling the verbosity of the output (tracingVerbosity)
could also be retained and directed by the new configuration.
and in this particular case, we need to have this `teeTraceChainTip` to remain here, as this will be hooking into the trace and update the LiveView and output metrics to monitoring.

Jimbo4350: @CodiePP Will do. I'll make the changes.

cardano-config/src/Cardano/Config/Types.hs
```diff
@@ -455,15 +453,39 @@ instance FromJSON ViewMode where
...
+  , traceBlockFetchServer :: !Bool
+  , traceTxInbound :: !Bool
+  , traceTxOutbound :: !Bool
+  , traceLocalTxSubmissionServer :: !Bool
+  , traceMempool :: !Bool
+  , traceForge :: !Bool
+    -----------------------
+
+    -- Protocol Tracers --
+  , traceChainSyncProtocol :: Bool
```
CodiePP: not all fields are strict.

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -117,35 +113,45 @@ mkTracers :: forall peer blk.
            -> Tracers peer blk
 mkTracers traceOptions tracer = Tracers
     { chainDBTracer
-        = annotateSeverity $ filterSeverity (pure . const (tracingSeverity $ traceChainDB traceOptions))
+        = tracerOnOff (traceChainDB traceOptions)
+          $ annotateSeverity $ filterSeverity (pure . const (tracingSeverity True))
```
CodiePP: the calls to `filterSeverity` on something that is "True" can be removed. previously, the flag was used to control the filter.

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -218,95 +224,85 @@ mkTracers traceOptions tracer = Tracers
...
-    mkConsensusTracers = Consensus.Tracers
+      traceWith (showTracing $ withName "Mempool" tracer) ev
+
+    mkConsensusTracers :: TraceOptions -> Consensus.Tracers' peer blk (Tracer IO)
+    mkConsensusTracers traceOpts = Consensus.Tracers
       { Consensus.chainSyncClientTracer
-        = annotateSeverity $ filterSeverity (pure . const (tracingSeverity $ hasConsensusTraceFlag Consensus.chainSyncClientTracer))
-          $ toLogObject' (tracingFormatting $ hasConsensusTraceFlag Consensus.chainSyncClientTracer) tracingVerbosity
+        = tracerOnOff (traceChainSyncClient traceOpts)
+          $ annotateSeverity $ filterSeverity (pure . const (tracingSeverity True))
```
CodiePP: maybe we don't clal "filterSeverity" anywhere, so it can also be removed

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -218,95 +224,85 @@ mkTracers traceOptions tracer = Tracers
...
+          $ toLogObject' (tracingFormatting True) tracingVerbosity
```
CodiePP: also "tracingFormatting" does not depend on the flag anymore. do we have another cli arg to distinguish between textual and structured output? if not I think we should go for "StructuredLogging" in all cases.

Jimbo4350: @CodiePP at the moment I'm trying to establish a process for adding/removing cli flags, I've talked to @i-o-m about this so for now I will default it to `StructuredLogging`

CodiePP:
you indicate that issue #409 is relevant for this work item. But, that issue does not contain anything, is as empty as it can be.
Where is the plan for doing this work item?

Jimbo4350:
> you indicate that issue #409 is relevant for this work item. But, that issue does not contain anything, is as empty as it can be.
> Where is the plan for doing this work item?

@CodiePP  :+1:  Updated the ticket with more context.

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -281,92 +286,75 @@ mkTracers traceOptions tracer = do
...
-        = annotateSeverity $ filterSeverity (pure . const (tracingSeverity $ hasConsensusTraceFlag Consensus.chainSyncServerBlockTracer))
-          $ toLogObject' (tracingFormatting $ hasConsensusTraceFlag Consensus.chainSyncServerBlockTracer) tracingVerbosity
+        = tracerOnOff (traceChainSyncBlockServer traceOpts)
+          $ toLogObject' StructuredLogging tracingVerbosity
           $ addName "ChainSyncBlockServer" tracer
       , Consensus.blockFetchDecisionTracer
-        = annotateSeverity $ filterSeverity (pure . const (tracingSeverity $ hasConsensusTraceFlag Consensus.blockFetchDecisionTracer))
-          $ teeTraceBlockFetchDecision (tracingFormatting $ hasConsensusTraceFlag Consensus.blockFetchDecisionTracer) tracingVerbosity
+        = tracerOnOff (traceBlockFetchDecisions traceOpts)
+          $ annotateSeverity $ filterSeverity (pure . const (tracingSeverity True))
```
CodiePP: `filterSeverity` can be removed everywhere since the tracer is either on or off.

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -141,35 +137,45 @@ mkTracers traceOptions tracer = do
...
+          $ annotateSeverity $ filterSeverity (pure . const Info)  -- filter out everything below this level
           $ toLogObject' (tracingFormatting $ traceMux traceOptions) tracingVerbosity
           $ addName "Mux" tracer
     }
   where
+    -- Turn on/off a tracer depending on what was parsed from the command line.
+    tracerOnOff :: Bool -> Tracer IO a -> Tracer IO a
+    tracerOnOff False _ = nullTracer
+    tracerOnOff True tracer' = tracer'
     tracingFormatting :: Bool -> TracingFormatting
```
CodiePP: also, remove all `tracingFormatting` since its always True. Replace with `StructuredLogging`.

PR #276, opened 2019-10-31
add node id as hostname in federated logging

CodiePP:
the node id is removed from the context naming.
we can nevertheless indicate the source of the tracer with setting the hostname to the node id in federated logging.

- [x] depends on upstream PR https://github.com/input-output-hk/iohk-monitoring-framework/pull/448

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -104,43 +104,43 @@ source-repository-package
   type: git
   location: https://github.com/input-output-hk/iohk-monitoring-framework
   subdir: iohk-monitoring
-  tag: bea0e079fc32ed316ce352d17d14199a680e3f6c
+  tag: a71addb2c45a8e116c2a57385b67812d51352a7a
 
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/iohk-monitoring-framework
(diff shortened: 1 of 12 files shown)


CodiePP:
ready; now refers to commit hash on `master` in iohk-monitoring

CodiePP:
> It's ok to use the system's real hostname, but the node id is not particulary meaningful. In particular it only means something now in the BFT era for core nodes involved in creating blocks. It means nothing in Shelley with Praos rather than BFT.
> 
> But do we need the host name? Does @disassembler think it is useful? Don't the monitoring aggregation systems do this automatically, do we need to add it ourselves?

this gives us the ability to name the node and in the aggregated log indicate the source of the log item. Moreover, this will also fill in the field 'hostname' in katip's log item header, which was empty previously.
Per default, this information is empty. It made sense to indicate the node id in the demo cluster and will also make sense in other setups such as benchmarking.

CodiePP:
addressed @dcoutts 's issues and rebased on `master`.
finally `CI` is happily compiling and testing again.

PR #277, opened 2019-11-01
clean up log config

CodiePP:
clean up logging configuration:
metrics should not show up in the logs anymore.

### DIFF
--- a/configuration/log-config-0.yaml
+++ b/configuration/log-config-0.yaml
@@ -62,23 +62,27 @@ options:
       # Change the `subtrace` value to `Neutral` in order to log
       # `UTxO`-related messages during epoch validation.
       subtrace: NoTrace
-    '#messagecounters.aggregation':
+    'cardano.#messagecounters.aggregation':
       subtrace: NoTrace
-    '#messagecounters.ekgview':
+    'cardano.#messagecounters.ekgview':
       subtrace: NoTrace
-    '#messagecounters.switchboard':
+    'cardano.#messagecounters.switchboard':
       subtrace: NoTrace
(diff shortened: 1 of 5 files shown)


scripts/shelley-testnet.sh
```diff
@@ -13,8 +13,6 @@ set -e
 # CMD="stack exec --nix --"
 CMD="cabal new-run --"
 
-. $(dirname $0)/lib-node.sh
-
```
deepfire: @CodiePP, I think that change should break the script -- this lib defines `nodeargs`, which is used below.

configuration/log-config-0.yaml
```diff
@@ -62,23 +62,27 @@ options:
       # Change the `subtrace` value to `Neutral` in order to log
       # `UTxO`-related messages during epoch validation.
       subtrace: NoTrace
-    '#messagecounters.aggregation':
+    'cardano.#messagecounters.aggregation':
```
deepfire: Just a stray thought, not really critical -- do we really need this `cardano` prefix everywhere?

It seems that everything in IOHK is about Cardano, somehow, so it seems.. implicit? : -)

CodiePP: at the moment, yes. there will be changes in the framework on the context naming ... (to be expected)

scripts/shelley-testnet.sh
CodiePP: it's included twice. one time would be enough?

deepfire: Oh, right, my bad! : -)

PR #293, opened 2019-11-11
Transaction generator: multiple target nodes, WIP.

denisshevchenko:
https://github.com/input-output-hk/cardano-node/issues/273


denisshevchenko:
> There is a requirement that each target node is delivered its own sequence of (dependent) transactions, but I can't see the logic for this being present in this code.

@njd42 It's not ready yet (I've made some force-pushes, but didn't finish yet). Moreover, currently all transactions we're generating are _independent_ (because, after splitting phase, we already have enough outputs that will be used as inputs for all "main transactions").

PR #313, opened 2019-11-20
Bump dependencies

edsko:
Most of this is "coding by analogy".

@coot , the most important thing to verify here is whether the various severities and defaults make sense.


cardano-node/src/Cardano/Tracing/ToObjectOrphans.hs
```diff
@@ -163,6 +167,20 @@ instance DefineSeverity (WithDomainName DnsTrace) where
     DnsTraceLookupAResult {} -> Debug
     DnsTraceLookupAAAAResult {} -> Debug
 
+instance DefinePrivacyAnnotation (WithAddr Socket.SockAddr ErrorPolicyTrace)
+instance DefineSeverity (WithAddr Socket.SockAddr ErrorPolicyTrace) where
+  defineSeverity (WithAddr _ ev) = case ev of
+    ErrorPolicySuspendPeer {} -> Debug
```
coot: ^ this should be `Error`

Issue #310, opened 2019-11-19
Make genesis file more resilient to bad `NetworkMagic`


nc6:
@edsko has just spent day trying to debug a validation failure in the new node which resulted from the following chain of events:

1. The `RequiresMagic` was set incorrectly for mainnet.
2. When interpreting the AVVM addresses, `makeRedeemAddress` (Which lives in `cardano-ledger`) uses the network magic to construct the address. Thus it was incorporating the incorrect network magic.
3. When we got to the first transactions, they failed to lookup inputs correctly because of the differently encoded address. This happened before the network magic in their own outputs was checked and found not to conform.

There are a couple of ways we could improve this situation:
1. Rather than having the `RequiresNetworkMagic` as a separate configuration item, encode it directly as a flag in the config file. To preserve the integrity of the mainnet hash, the absence of this should be interpreted as `RequiresNoMagic`. Any other genesis file would be required to have this field present and set to a value.

2. AVVM addresses in the genesis file could be considered to have the network magic pre-encoded. This approach will work if the format in the genesis file is equal to the on-chain format with no network magic. All new files would need to have the network magic pre-encoded (and we would need to check that it was the same for each address)

mhuesch:
> Rather than having the RequiresNetworkMagic as a separate configuration item, encode it directly as a flag in the config file. To preserve the integrity of the mainnet hash, the absence of this should be interpreted as RequiresNoMagic. Any other genesis file would be required to have this field present and set to a value.

To clarify, is the suggested change here allowing the `RequiresNetworkMagic:` key in the `.yaml` config file to be absent? Currently I see, for example:

```
...
NumCoreNodes: 1
RequiresNetworkMagic: RequiresNoMagic
PBftSignatureThreshold: 0.7
...
```

are we proposing that we allow 

```
...
NumCoreNodes: 1
PBftSignatureThreshold: 0.7
...
```

and have it be behaviorally equivalent (`RequiresNoMagic`)?

***

> AVVM addresses in the genesis file could be considered to have the network magic pre-encoded. This approach will work if the format in the genesis file is equal to the on-chain format with no network magic. All new files would need to have the network magic pre-encoded (and we would need to check that it was the same for each address)

Does this equate to decoding the addresses as though they include a NetworkMagic? For mainnet addresses, if we do that, we will get out a `NetworkMainOrStage`. It's then suggested (I think) that we check all `NetworkMagic` values of these AVVM addresses for equality, which seems like a good idea.

[...]

PR #315, opened 2019-11-20
Mainet doesn't require magic

karknu:


karknu:
#310 may improve on magic handling, but the problem in this file needs to be addressed now.

Issue #311, opened 2019-11-19
Make `cardano-cli` sane again


Jimbo4350:
Currently `cardano-cli` has a lot of additional unnecessary arguments (in an effort to keep the PR from becoming too large) due to a significant restructure in #289.

This has broken the scripts `issue-genesis-utxo-expenditure.sh` and `issue-utxo-expenditure.sh`.

Solution:
`fromProtocol` demands `NodeConfiguration` & `NodeCLI` which forces the relevant parsers on `cardano-cli`. I am currently amending this so that `fromProtocol` takes exactly what it needs.

PR #318, opened 2019-11-20
Simpler topology

deepfire:
1. Simplify topology file format: `s/NetworkTopology/NodeSetup/`
2. Eliminate `NetworkTopology` and `TopologyInfo` types.
3. Update scripts.

# TODO

- [ ] Update the Chairman test

PR #326, opened 2019-11-21
Refactor `fromProtocol`

Jimbo4350:
This PR: 
- Refactors `fromProtocol` to accept exactly what it needs.
- Removes extraneous command line arguments as a result of the above refactor.

Relevant: #311


CodiePP:
> The node should not determine its IP address from a topology file, it should use a topology file to determine its peer's IP addresses.

@denisshevchenko is working on issue: https://github.com/input-output-hk/cardano-node/issues/329 which should eliminate the need for a topology file.

cardano-config/src/Cardano/Config/CommonCLI.hs
```diff
@@ -95,6 +99,14 @@ parseGenesisPathLast =
         <> help "The filepath to the genesis file."
     )
 
+parseGenesisHash :: Parser Text
```
CodiePP: looking forward to have this option on the command line

Jimbo4350:
> this PR touches 32 files! can it be broken in smaller chunkes?
> and, there is work underway to remove the topology. It might make sense to push this other PRs first and then rebase.

I can cut it in half. I think @deepfire is still trying to get the chairman to work on #318 so I'd prefer to get this merged as I'm not sure how long that will take. It's not a big difference in terms of rebase conflict & I've reduced the PR by ~50%.

Jimbo4350:
> The node should not determine its IP address from a topology file, it should use a topology file to determine its peer's IP addresses.

@karknu I've reduced this PR, so in a separate PR I will address this. You wanted to specify the IP via the command line correct?

karknu:
> > The node should not determine its IP address from a topology file, it should use a topology file to determine its peer's IP addresses.
> 
> @karknu I've reduced this PR, so in a separate PR I will address this. You wanted to specify the IP via the command line correct?

Yeah, an optional command line argument for the IP address will work fine.

PR #330, opened 2019-11-25
Transaction generator: specify target nodes without topology.

denisshevchenko:
https://github.com/input-output-hk/cardano-node/issues/329

Since https://github.com/input-output-hk/cardano-node/pull/318 is going to eliminate `TopologyInfo` type, tx generator has to avoid such a dependency. Currently tx generator uses `--target-node-id`s only for extracting corresponding IPs/ports from topology file, so now it should receive `--target-node` CLI parameter with not `NodeId`, but with explicit IP/port, for example: `--target-node ("127.0.0.1",3002)`.

### DIFF
--- a/cardano-node/app/cardano-cli.hs
+++ b/cardano-node/app/cardano-cli.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE OverloadedStrings #-}
+
 import           Cardano.Prelude hiding (option)
 import           Prelude (String, error, id)
 
@@ -7,12 +9,14 @@ import qualified Data.List.NonEmpty as NE
 import           Data.Text
 import           Data.Time (UTCTime)
 import           Data.Time.Clock.POSIX (posixSecondsToUTCTime)
+import           Network.Socket (PortNumber)
 import qualified Options.Applicative as Opt
 import           Options.Applicative (Parser, ParserInfo, ParserPrefs, auto,
                                       commandGroup, flag, flag', help, long,
                                       metavar, option, showHelpOnEmpty,
                                       strOption, subparser, value)
 import           System.Exit (exitFailure)
+import           Text.Read (readMaybe)
 
 import           Cardano.Binary (Annotated (..))
 import           Cardano.Chain.Common
@@ -30,6 +34,7 @@ import           Cardano.Config.Partial (PartialCardanoConfiguration (..),
                                          mkCardanoConfiguration)
 import           Cardano.Config.Presets (mainnetConfiguration)
 import           Cardano.Config.Protocol (Protocol)
+import           Cardano.Config.Topology (NodeAddress (..), NodeHostAddress (..))
 import           Cardano.Config.Types (CardanoEnvironment (..), GenesisFile(..),
                                        RequireNetworkMagic)
 import           Cardano.Crypto ( AProtocolMagic(..)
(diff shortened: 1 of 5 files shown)


cardano-node/app/cardano-cli.hs
```diff
@@ -461,6 +469,19 @@ parseFlag def active optname desc =
   flag def active $ long optname <> help desc
 
 
+parseTargetNodeAddress :: String -> String -> Parser NodeAddress
```
CodiePP: how can we indicate multiple target hosts? it is not "many" anymore.

cardano-node/app/cardano-cli.hs
```diff
@@ -461,6 +469,19 @@ parseFlag def active optname desc =
...
+parseTargetNodeAddress optname desc =
+  option
+    ( uncurry NodeAddress
+      . Control.Arrow.first cliParseHostAddress
+      . Control.Arrow.second cliParsePort
```
CodiePP: looks like it is going to be "(hostname, portnumber)"
would it be possible to have it closer to: "hostname:portnumber" ?

cardano-node/app/cardano-cli.hs
denisshevchenko: Yes, it's `some` now (one or more): https://github.com/input-output-hk/cardano-node/pull/330/files#diff-deb1de1e23655df9067b46df271864c6R342-R343

cardano-node/src/Cardano/CLI/Tx/Generation.hs
```diff
@@ -185,29 +167,11 @@ genesisBenchmarkRunner loggingLayer
...
 
-  when (null targetNodeIds) $
-    left $ NeedMinimumOneTargetNodeId targetNodeIds
-
   let (benchTracer, connectTracer, submitTracer, lowLevelSubmitTracer) = createTracers loggingLayer
 
   liftIO . traceWith benchTracer . TraceBenchTxSubDebug
     $ "******* Tx generator, tracers are ready *******"
 
-  liftIO . traceWith benchTracer . TraceBenchTxSubDebug
```
Jimbo4350: :+1:

cardano-node/src/Cardano/CLI/Tx/Generation.hs
```diff
@@ -854,7 +809,7 @@ createMoreFundCoins llTracer
                                              txOut
                                              []
   liftIO $ forM_ splittingTxs $ \(tx, txDetailsList) -> do
-    submitTx nCli pInfoConfig (node topologyInfo) tx llTracer
```
Jimbo4350: Wouldn't it make sense to supply a `NodeId` on the command line and pipe it here rather than always submit the transaction to the same node?

cardano-node/app/cardano-cli.hs
denisshevchenko: Yes, if such a form isn't suitable - I could change it.

cardano-node/app/cardano-cli.hs
Jimbo4350: This could be made arguably simpler by breaking it up into something like the following:

`NodeAddress <$> parseHostAddress <*> parsePort`

cardano-node/src/Cardano/CLI/Tx/Generation.hs
denisshevchenko: > always submit the transaction to the same node?

This is about genesis transaction and splitting transactions only. Main transactions will be send to multiple nodes (if user defined more than one `--target-node`).

Moreover, `--node-id` CLI parameter is already removed. Later we'll change this code: probably `submitTx` won't be used at all.

PR #331, opened 2019-11-25
Limited chain sync benchmark

CodiePP:
this is a benchmark that runs for a limited time (e.g. one hour) and syncs blocks from a local byron-proxy. Alternatively, these blocks could also be downloaded from a remote proxy.
the reason for limiting the time is that the block chain grows every day, and thus the times of different runs cannot be compared directly. Using the obtained result we can at least estimate how much time it may need to sync a chain of certain size (i.e. 3.5*10^6 blocks).


benchmarking/chain-sync/prepare-byron-proxy.sh
```diff
@@ -0,0 +1,17 @@
...
+
+set -e
+
+BASEDIR=$(realpath $(dirname $0))
+#GITURL="https://github.com/input-output-hk/cardano-byron-proxy.git"
+TARGETDIR=cardano-byron-proxy.git
+
+#rm -rf ${TARGETDIR}
+#git clone -b master ${GITURL} ${TARGETDIR}
+git submodule update --init --remote ${TARGETDIR}
```
deepfire: In https://github.com/input-output-hk/cardano-node/pull/269 we'll get a different way of pinning the proxy, so we'd have to change this -- but that's good for now.

PR #332, opened 2019-11-26
Run three node benchmark

CodiePP:
this introduces benchmarking of the new node on local machines with a cluster of three nodes, an explorer, and a tx generator. 
the setup requires a Postgres instance for the explorer, and the explorer compiled before launching benchmarks.
then the one-button approach to benchmarking is starting the script `./start.sh`


benchmarking/cluster3nodes/benchmark.sh
```diff
@@ -0,0 +1,106 @@
...
+
+BASEDIR=$(realpath $(dirname $0))
+
+# clean
+if [ -d db ]; then rm -rf db; fi
+if [ -d logs ]; then rm -rf logs; fi
+if [ -d socket ]; then rm -rf socket; fi
+if [ -d db-0 ]; then rm -rf db-0; fi
+if [ -d db-1 ]; then rm -rf db-1; fi
+if [ -d db-2 ]; then rm -rf db-2; fi
```
deepfire: I'd say that we don't need the tests -- just a single unconditional `rm -rf logs socket db db-{0,1,2}` would do.

benchmarking/cluster3nodes/configuration/log-config-generator.yaml
```diff
@@ -0,0 +1,149 @@
+# global filter; messages must have at least this severity to pass:
```
deepfire: @CodiePP, I think it's no longer appropriate to call these `log-config`'s -- they're general configuration since the recent changes.

benchmarking/cluster3nodes/run-3node-cluster.sh
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env bash
```
deepfire: I'm a little afraid this is going to break as the configuration changes happen to the node..

If we shared more with `scripts/` (which even has `lib-node.sh` for this purpose), I'd say we might have better chances of surviving these changes..

scripts/shelley-testnet.sh
```diff
@@ -54,8 +54,8 @@ tmux select-pane -t 4
 tmux send-keys "cd '${PWD}'; ${CMD} trace-acceptor $(acceptorargs)" C-m
 sleep 2
 tmux select-pane -t 1
-tmux send-keys "cd '${PWD}'; ${CMD} exe:cardano-node $(nodeargs 0 "${ALGO} $(echo -n ${EXTRA})") > /tmp/NODE0" C-m
+tmux send-keys "cd '${PWD}'; ${CMD} exe:cardano-node $(nodeargs 0 "${ALGO} $(echo -n ${EXTRA})") " C-m
```
deepfire: :heart:

benchmarking/cluster3nodes/run-3node-cluster.sh
CodiePP: I wanted to keep this separate so we can make changes/adaptations without interfering with other uses. IMO, the scripts will not be used in the future.

benchmarking/cluster3nodes/configuration/log-config-generator.yaml
CodiePP: yes, we might rename these in the future

PR #338, opened 2019-11-27
cabal changes

coot:
* disable `-Werror` from cabal.project & stack.yaml files
* enable `-Werror` on CI
* use `-O1` instead of `-O2` in `cardano-node` executables
* enable `-rtsopts`
* use `cborg >= 0.2.2 && < 0.3` bounds (as in `ouroboros-consensus`).

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -3,10 +3,10 @@ packages:
     cardano-node
 
 package cardano-config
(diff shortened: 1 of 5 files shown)


coot:
It's not clear to me why the cross compilation build failed.

Issue #336, opened 2019-11-26
Review severities of `MuxTrace`


coot:
https://github.com/input-output-hk/cardano-node/blob/master/cardano-node/src/Cardano/Tracing/ToObjectOrphans.hs#L185

PR #339, opened 2019-11-27
MuxTraceHandshakeClientEnd should be at Info

karknu:
Change MuxTraceHandshakeClientEnd to info so that the initial rtt
sample is visible in the logs.


karknu:
Addresses #336.

PR #343, opened 2019-11-27
adds IP's for testnet/mainnet relays

disassembler:


deepfire:
Investigating this in https://github.com/input-output-hk/cardano-node/issues/346

PR #351, opened 2019-11-28
Unshare the module list

deepfire:
1. Unshares `nix/nixos/module-list.nix` -- leaving it to only be used for deployments.
2. Repairs `extraArgs` to be additive 
3. Drops `extraOptions` -- with apologies to @cleverca22 for me screwing it up

### DIFF
--- a/nix/nixos/cardano-node-service.nix
+++ b/nix/nixos/cardano-node-service.nix
@@ -55,11 +55,6 @@ in {
         type = types.str;
         default = mkScript cfg;
       };
-      extraOptions = mkOption {
-        type = types.listOf types.str;
-        default = [];
-        description = "extra command line arguments for cardano-node";
-      };
 
       package = mkOption {
         type = types.package;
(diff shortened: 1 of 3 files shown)


nix/nixos/cardano-node-service.nix
```diff
@@ -24,7 +24,7 @@ let
             "--signing-key ${cfg.signingKey}"}"
           "${lib.optionalString (cfg.delegationCertificate != null)
             "--delegation-certificate ${cfg.delegationCertificate}"}"
-          "${cfg.extraArgs}"
+          "${toString cfg.extraArgs}"
```
cleverca22: i think `] ++ cfg.extraArgs` is better, because of how `cmd` gets processed to pretty-print in the bash script

deepfire: A very good point @cleverca22.. how sloppy am I -- having introduced that mechanism in the first place..

deepfire: Fixed!

PR #357, opened 2019-11-29
Make public key extraction more robust in the cluster test

deepfire:
This should improve robustness against logging changes in the cluster test.

The root of the problem is the impossibility to completely disable logging in `cardano-node` -- we used to have this mechanism, but at some point it was removed, making any usage of `cardano-cli`'s standard output more fragile.

### DIFF
--- a/nix/svclib.nix
+++ b/nix/svclib.nix
@@ -188,7 +188,7 @@ let
       signing-key-public
       --secret ${pbftSK}
       )
-      ${cardano-node}/bin/cardano-cli "''${args[@]}" | tail -n -2 | head -n 1 | cut -d: -f2 | xargs echo -n > $out
+      ${cardano-node}/bin/cardano-cli "''${args[@]}" | fgrep 'public key (base64):' | cut -d: -f2 | xargs echo -n > $out
     '';
 
   ## extractDelegateCertificate


Jimbo4350:
> The root of the problem is the impossibility to completely disable logging in `cardano-node` -- we used to have this mechanism, but at some point it was removed, making any usage of `cardano-cli`'s standard output more fragile.

I removed the functionality as part of a refactor. I have a more direct way of disabling logging (vs what was previously implemented to disable logging) on a branch that I will PR next week.

deepfire:
@Jimbo4350, I don't think there's anything to squash..

deepfire:
> I removed the functionality as part of a refactor however I was not aware that it had this effect on the cli until very recently.

It's generally not safe to remove functionality, unless you are sure you know all the implications.

Jimbo4350:
> It's generally not safe to remove functionality, unless you are sure you know all the implications.

It's also good practice to inform others when you are **aware** of implications like this.

deepfire:
> It's also good practice to inform others when you are aware of implications like this.

I'm sorry, do I understand it correctly, that you imply that I didn't inform you about it?

Because I did.

deepfire:
@Jimbo4350, I have a weird feeling that we are going through a discussion we already had a couple weeks ago, didn't we?

Jimbo4350:
> @Jimbo4350, I don't think there's anything to squash..

Interesting, maybe a git malfunction :man_shrugging:

deepfire:
@Jimbo4350, at the time I've made the commit a top of `bors/staging`, as it was en route to merging.

`master` was updated since then, so that's no longer visible..

Issue #362, opened 2019-12-02
Investigate why https://github.com/input-output-hk/cardano-node/pull/352 fixed the issue


ksaric:
After the PR - https://github.com/input-output-hk/cardano-node/pull/348
We noticed that the PR actually failed in the node with:
```
cardano-node: user error ((line 1, column 1):
unexpected end of input
expecting white space or JSON value)
```

This fixed it - https://github.com/input-output-hk/cardano-node/pull/352

The question was - why is that fixing the actual issue with the CI?

ksaric:
I thought that was because of the `ToLogObject IO` instance:
```
instance ToLogObject IO where
    toLogObject :: (MonadIO m, ToObject a, Transformable a m b)
                => Tracer m (LogObject a) -> Tracer m b
    toLogObject tr = trTransformer StructuredLogging NormalVerbosity tr
    toLogObject' :: (MonadIO m, ToObject a, Transformable a m b)
                 => TracingFormatting -> TracingVerbosity -> Tracer m (LogObject a) -> Tracer m b
    toLogObject' hint verb tr = trTransformer hint verb tr
```

Since the transformer definition was not complete (well, it was complete, but not exhaustive I guess), I presumed that was causing it.
Notice how it was defined before - https://github.com/input-output-hk/cardano-node/pull/352/files#diff-d1ea55d7138fc8b6df9635128887767aL353-L354

My assumption was that `trStructured` was called on `TextualRepresentation` where the default implementation was:
```
trStructured :: (ToObject b, MonadIO m, DefinePrivacyAnnotation b, DefineSeverity b)
             => TracingVerbosity -> Tracer m (LogObject a) -> Tracer m b
trStructured verb tr = Tracer $ \arg ->
        let obj = toObject verb arg
            tracer = if obj == emptyObject then nullTracer else tr
        in
        traceWith tracer =<<
            LogObject <$> pure mempty
                      <*> (mkLOMeta (defineSeverity arg) (definePrivacyAnnotation arg))
                      <*> pure (LogStructured $ encode $ obj)
```

Where the default mechanism is:
```
class ToObject a where
    toObject :: TracingVerbosity -> a -> Object
    default toObject :: ToJSON a => TracingVerbosity -> a -> Object
    toObject _ v = case toJSON v of
        Object o     -> o
        s@(String _) -> HM.singleton "string" s
        _            -> mempty
```
So it tries to convert `v` to JSON.

That's why I figured the error was appearing:
```
[...]

PR #363, opened 2019-12-02
[GH-362] Test if the ToObject instance is fixed.

ksaric:
https://github.com/input-output-hk/cardano-node/issues/362


ksaric:
This doesn't seem to be the cause.

PR #368, opened 2019-12-03
[GH-362]

ksaric:


ksaric:
This PR shows that the issue was not in the `Transformable` instance.

Issue #356, opened 2019-11-29
`./scripts/issue-genesis-utxo-expenditure` is broken


Jimbo4350:
`./scripts/issue-genesis-utxo-expenditure` is currently broken on `master`.

PR #375, opened 2019-12-04
Fix `/scripts/issue-genesis-utxo-expenditure.sh`

Jimbo4350:
- Remove `GenesisHash` from config `.yaml` files and specify it on the command line instead.
Relevant: #356


scripts/issue-genesis-utxo-expenditure.sh
```diff
@@ -1,15 +1,27 @@
 #!/usr/bin/env bash
 
-RUNNER=${RUNNER:-cabal new-run -v0 --}
+if [ -e configuration/GenesisFiles/ ]
+then
+    :
+else
+    echo "configuration/GenesisFiles/GENHASH does not exist"
+	/bin/bash ./scripts/genesis.sh
```
CodiePP: /bin/bash does not exist on every platform, that's why this script is launched via `/usr/bin/env bash`
isn't the `script/genesis.sh` executable?

scripts/lib-node.sh
```diff
@@ -1,9 +1,15 @@
...
+then
+    :
+else
+    echo "configuration/GenesisFiles/GENHASH does not exist"
+    /bin/bash ./scripts/genesis.sh
+fi
 
-genesis="$(echo ${genesis_hash} | cut -c-5 | xargs echo -n)"
-genesis_root="configuration/${genesis}"
+genesis_hash="$(echo $(<configuration/GenesisFiles/GENHASH))"
```
CodiePP: why not `cat` ?

scripts/shelley-testnet2.sh
```diff
@@ -15,6 +15,7 @@ EXTRA=""
 
 . $(dirname $0)/lib-node.sh
 
+
```
CodiePP: noise.

scripts/get-default-key-address.sh
```diff
@@ -1,10 +1,21 @@
 #!/bin/sh
 
+if [ -e configuration/GenesisFiles/GENHASH ]
+then
+    :
+else
+    echo "configuration/GenesisFiles/GENHASH does not exist"
+	/bin/bash ./scripts/genesis.sh
```
CodiePP: /bin/bash does not exist on every platform, that's why this script is launched via /usr/bin/env bash
isn't the script/genesis.sh executable?

scripts/lib-node.sh
Jimbo4350: No particular reason, I can change it.

Issue #376, opened 2019-12-05
Make shelley scripts generate genesis files automatically


Jimbo4350:

deepfire:
Fixed.

PR #396, opened 2019-12-11
Fix scripts

deepfire:
Issues: #376

Note, that this doesn't touch the benchmarking-specific scripts, which could be converted to using the shared generated genesis.

### DIFF
--- a/scripts/lib-node.sh
+++ b/scripts/lib-node.sh
@@ -1,7 +1,15 @@
-
-genesis_hash=`cat configuration/GenesisFiles/GENHASH`
-genesis_root="configuration/GenesisFiles"
+genesis_root="$(realpath configuration/genesis)"
 genesis_file="${genesis_root}/genesis.json"
+
+if test ! -f "${genesis_file}"
+then if test -x ./scripts/genesis.sh
+     then ./scripts/genesis.sh
+     else echo "ERROR: could not locate genesis generator!">&1; exit 1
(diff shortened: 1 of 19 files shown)


scripts/trace-acceptor.sh
```diff
@@ -9,6 +11,7 @@ ${CMD} \
     --topology "./configuration/simple-topology.json" \
     --database-path "./db/" \
     --genesis-file "configuration/mainnet-genesis.json" \
+    --genesis-hash "5f20df933584822601f9e3f8c024eb5eb252fe8cefb24d1317dc3d432e940ebb" \
```
Jimbo4350: Shouldn't the genesis hash be pulled from `lib-node.sh`?

scripts/lib-node.sh
```diff
@@ -1,7 +1,15 @@
-
-genesis_hash=`cat configuration/GenesisFiles/GENHASH`
-genesis_root="configuration/GenesisFiles"
+genesis_root="$(realpath configuration/genesis)"
 genesis_file="${genesis_root}/genesis.json"
+
```
Jimbo4350: 👍

scripts/trace-acceptor.sh
deepfire: @Jimbo4350, it's hard-coded to mainnet here, so no it can't be pulled from there.

PR #399, opened 2019-12-11
Update `fromProtocol` to use `ExceptT`

Jimbo4350:
This PR adds more explicit error handling to `fromProtocol` and its callers.
#404

### DIFF
--- a/cardano-node/src/Cardano/CLI/Ops.hs
+++ b/cardano-node/src/Cardano/CLI/Ops.hs
@@ -1,10 +1,11 @@
-{-# LANGUAGE DataKinds           #-}
-{-# LANGUAGE FlexibleContexts    #-}
-{-# LANGUAGE GADTs               #-}
-{-# LANGUAGE LambdaCase          #-}
-{-# LANGUAGE RankNTypes          #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TypeApplications    #-}
+{-# LANGUAGE TypeApplications #-}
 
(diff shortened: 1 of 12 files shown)


cardano-node/src/Cardano/CLI/Ops.hs
```diff
@@ -151,6 +170,57 @@ instance Show CliError where
...
+  -> Maybe DelegationCertFile
+  -> Maybe SigningKeyFile
+  -> Update
+  -> Protocol
+  -> (RunNode ByronBlock
+        => Consensus.Protocol ByronBlock
+        -> ExceptT RealPBFTError IO a)
+  -> ExceptT RealPBFTError IO a
+withRealPBFT gHash nId mNumNodes genFile nMagic sigThresh delCertFp sKeyFp update ptcl action = do
+  SomeProtocol p <- firstExceptT FromProtocolError $ fromProtocol
```
deepfire: This should be indented so the subsequent lines don't depend on the `SomeProtocol p <- firstExceptT FromProtocolError`'s length.

cardano-node/src/Cardano/CLI/Tx.hs
```diff
@@ -320,7 +289,10 @@ nodeSubmitTx
...
     withRealPBFT gHash (Just nId) mNumCoreNodes genFile nMagic sigThresh delCertFp sKeyFp update ptcl $
       \p@Consensus.ProtocolRealPBFT{} -> do
-        case gentx of
-          ByronTx txid _ -> putStrLn $ sformat ("TxId: "%Crypto.hashHexF) txid
-          _ -> pure ()
-        handleTxSubmission socketFp p topology gentx stdoutTracer
+        _ <- case gentx of
+               ByronTx txid _ -> pure . putTextLn
+                                      $ sformat ("TxId: "%Crypto.hashHexF) txid
+               otherTxType -> left . TransactionTypeNotHandledYet
```
Jimbo4350: @denisshevchenko do we want to fail here? I'm not sure.

denisshevchenko: @Jimbo4350 I'm not sure as well. :-) Probably yes.

Jimbo4350: I agree, was just double checking in case there was a reason why you wanted this to fail silently.

Issue #401, opened 2019-12-12
CAD-390 profiling chain-sync


CodiePP:
a memory leak has been observed.
the chain-sync benchmark runs for one hour; should be done with profiling.

karknu:
The memory leak seems to have been introduced with https://github.com/input-output-hk/ouroboros-network/commit/f2a050ba9ada3bf3ee2421f5e610947619d28337

mrBliss:
Fixed in https://github.com/input-output-hk/ouroboros-network/pull/1357

CodiePP:
needs to be reevaluated on new integration.

PR #385, opened 2019-12-09
Logging config mainnet

CodiePP:
adapting configuration for 'mainnet.sh' script.

PR #408, opened 2019-12-12
[#401] update chain sync config and scripts

CodiePP:
updates to chain sync config and scripts.
replaces PR #385


CodiePP:
for profiling add to the file `cabal.project.local` the following lines:
```
library-profiling: True
profiling: True
executable-profiling: True
```
and change the ghc-options of exe:cardano-node (add '-fprof-auto'):
```
executable cardano-node
                       -Wall
                       -fprof-auto
                       -rtsopts
                       "-with-rtsopts=-T -I0"
                       -fno-warn-unticked-promoted-constructors
```
and recompile.
Then, run the `benchmark-chain-sync-mainnet.sh` script with arguments (for example): `+RTS -N1 -p -hy -RTS`

Issue #412, opened 2019-12-13
genesis.sh returns error if directory above cardano-node have spaces in its name


ArturWieczorek:
I discovered that if your directory above `cardano-node` have spaces in it then the execution of `genesis.sh` script will break with following error:

`artur@artur-VirtualBox:~/Projects/Explorer Readme/cardano-node$ ./scripts/genesis.sh`

> ++ cabal v2-run -v0 -- cardano-cli --log-config '/home/artur/Projects/Explorer Readme/cardano-node/configuration/configuration-silent.yaml' --real-pbft print-genesis-hash --genesis-json /tmp/tmp.HAVPtFP5k8.d/genesis.json
> ++ tail -1
> warning: redirecting to https://github.com/well-typed/canonical-json/
> remote: Enumerating objects: 1, done.
> remote: Counting objects: 100% (1/1), done.
> remote: Total 1 (delta 0), reused 0 (delta 0), pack-reused 0
> Unpacking objects: 100% (1/1), done.
> From https://github.com/input-output-hk/cardano-ledger-specs
> df2bea65..4f318ea4 master -> origin/master
> 
> GENHASH=d86c1f4e32a459a38c4fb07d097a7a2771f1ea502e801bd297ff08c3d7594cb4
> TARGETDIR='/home/artur/Projects/Explorer Readme/cardano-node/configuration/genesis'
> mkdir -vp '/home/artur/Projects/Explorer Readme/cardano-node/configuration/genesis'
> mkdir: created directory '/home/artur/Projects/Explorer Readme/cardano-node/configuration/genesis'
> cp -iav /tmp/tmp.HAVPtFP5k8.d/genesis.json /home/artur/Projects/Explorer Readme/cardano-node/configuration/genesis/
> cp: target 'Readme/cardano-node/configuration/genesis/' is not a directory

When I changed it to `ExplorerReadme` - without spaces it ended OK

ArturWieczorek:
This works now:
```
$ ./scripts/genesis.sh
...
'/tmp/tmp.Fu77k46o1n.d/delegation-cert.005.json' -> '/home/artur/Projects/Testing Fixes For Scripts/cardano-node/configuration/genesis/delegation-cert.005.json'
'/tmp/tmp.Fu77k46o1n.d/delegation-cert.006.json' -> '/home/artur/Projects/Testing Fixes For Scripts/cardano-node/configuration/genesis/delegation-cert.006.json'
+ set -
[...]

Issue #411, opened 2019-12-13
get-default-key-address.sh script return error


ArturWieczorek:
When running:
`$ ./scripts/get-default-key-address.sh configuration/genesis/delegat-keys.001.key`

I got:
> ./scripts/get-default-key-address.sh: 16: ./scripts/lib-node.sh: Syntax error: "(" unexpected

As Jordan Millar suggested - you need to change:
 `#!/bin/sh` to `#!/usr/bin/env bash` in first line for `get-default-key-address.sh`

ArturWieczorek:
This one works now:

```
./scripts/get-default-key-address.sh configuration/genesis/delegate-keys.001.key
2cWKMJemoBahs2gd4sQX1XgSLVLyAMrtNy2qgp4t2BYjBFygD3dLo4ejXC4p3C3pnc6qd
```

Closing as it was fixed and merged: Fix `scripts/get-default-key-address.sh` & `scripts/genesis.sh` #413

PR #413, opened 2019-12-13
Fix `scripts/get-default-key-address.sh` & `scripts/genesis.sh`

Jimbo4350:
#411  & #412

### DIFF
--- a/scripts/lib-node.sh
+++ b/scripts/lib-node.sh
@@ -8,7 +8,7 @@ then if test -x ./scripts/genesis.sh
      fi
 fi
 
-genesis_hash=`cat ${genesis_root}/GENHASH`
+genesis_hash="cat ${genesis_root}/GENHASH"
 
 if test ! -f "${genesis_file}"
 then echo "ERROR: genesis ${genesis_file} does not exist!">&1; exit 1; fi
--- a/scripts/lib-node.sh
+++ b/scripts/lib-node.sh
@@ -8,7 +8,7 @@ then if test -x ./scripts/genesis.sh
      fi
 fi
 
-genesis_hash="cat ${genesis_root}/GENHASH"
+genesis_hash=`cat "${genesis_root}/GENHASH"`
 
 if test ! -f "${genesis_file}"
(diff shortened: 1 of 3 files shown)


ArturWieczorek:
`genesis.sh` now works fine but the other one  still got issues:

> ./scripts/get-default-key-address.sh configuration/genesis/delegate-keys.001.key 
> cat: /home/artur/Projects/Fix: No such file or directory
> cat: Script/cardano-node/configuration/genesis/GENHASH: No such file or directory
> 2cWKMJemoBahs2gd4sQX1XgSLVLyAMrtNy2qgp4t2BYjBFygD3dLo4ejXC4p3C3pnc6qd

I managed to fix it locally by modifying lib-node.sh: 

`genesis_hash="cat ${genesis_root}/GENHASH"` - surrounded it with double quotes instead of backticks

scripts/lib-node.sh
```diff
@@ -8,7 +8,7 @@ then if test -x ./scripts/genesis.sh
      fi
 fi
 
-genesis_hash=`cat ${genesis_root}/GENHASH`
+genesis_hash="cat ${genesis_root}/GENHASH"
```
CodiePP: that will not work! try it in a shell and you will see that the $genesis_hash will contain a string: "cat configuration/Something/GENHASH", but not the genesis' hash.

CodiePP: the meaning of backticks is: run the command and get its output
whereas double quotes are only to enclose strings

ArturWieczorek: @CodiePP Hi, after branch checkout I ran those scripts again (`genesis.sh` and then `get-default-key-address.sh`)  and everything works fine... **HOWEVER** you are right - this line is still wrong. Thanks for pointing this.

Leaving it like it was before change: `` `cat ${genesis_root}/GENHASH` `` will not work for directory with white spaces in name but I think the change below should work for all cases:

`` genesis_hash=`cat "${genesis_root}/GENHASH"` ``

CodiePP: 👍

Issue #416, opened 2019-12-16
CAD-389 nightly run of chain-sync benchmark


CodiePP:
The tests should run in two flavours:

1) profiling (#401 #408)
2) compiled for release

WIP PR: https://github.com/input-output-hk/cardano-node/pull/427

# Requirements

* the test runs until the node process exhausts its limit of CPU time (e.g. 60 minutes)
* the 'analysis.sh' script outputs the slot number and timestamp of the first and last block integrated. this data can then be pasted into a spreadsheet to calculdate the total duration and the average rate of blocks synchronized per second.
* the "profiling" run should also process its output with 'hp2ps' and 'flamegraph'
* these resulting artifacts of the runs should be kept on disk for a number of days, such that they can be copied over to permanent store

# Implementation plan

0. [ ] `NEGOTIATE` with Devops for a _dedicated_ (i.e. nothing else running there) bare metal (likely Packet.net) machine for a benchmarking Buildkite agent.  Do this as a first step, so they can execute this in parallel.
0. [ ] `MODE` of operation: a command line option, driving the choice of the `cardano-node` build -- normal, or profiled.  Ask devops about how to do this effectively -- while providing them with the context of the script:  https://github.com/input-output-hk/cardano-node/blob/416-chainsync-bench/benchmarking/chain-sync/benchmark-chain-sync-mainnet.sh
0. [ ] `OUTPUTS`: Ensure that the script can run locally and produce the necessary results:
   - the slot numbers+timestamps of first/last blocks
   - the profiling `.hp` outputs:
   - processing of the outputs
       - https://mpickering.github.io/eventlog2html/
       - https://hackage.haskell.org/package/ghc-prof-flamegraph
       - ask people around (esp. Thomas and Duncan) for what else they might want, but don't block on that 
0. [ ] `DURATION` in seconds has to be added as a command-line parameter to the script.
0. [ ] `HOOK` that script to https://github.com/input-output-hk/cardano-node/blob/416-chainsync-bench/.buildkite/pipeline.yml#L30 as a two clauses (one per mode, `release`/`profiled`), limiting the agents to those which have `benchmarking` tag set, like this:
   ```
   - label: 'benchmarking'
    command: 'benchmarking/chain-sync/benchmark-chain-sync-mainnet.sh --mode MODE --duration SECONDS'
    agents:
      system: x86_64-linux
      benchmarking: true
   ```
[...]

PR #427, opened 2019-12-17
[#416] WIP chainsync bench

deepfire:
Issue: #416

### DIFF
--- a/benchmarking/chain-sync/analyse-logs.sh
+++ b/benchmarking/chain-sync/analyse-logs.sh
@@ -1,10 +1,19 @@
 #!/bin/sh
+set -x
+logdir="$1"
+if test ! -d "${logdir}"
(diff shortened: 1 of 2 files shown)


benchmarking/chain-sync/analyse-logs.sh
```diff
@@ -1,10 +1,19 @@
 #!/bin/sh
+set -x
+logdir="$1"
+if test ! -d "${logdir}"
+then echo "ERROR:  log directory not found at '${logdir}'" >&2; exit 1; fi
 
 # get first block copy time
-LOGFILE=`ls -1 state-node-mainnet/node-0-*.log | head -3`
```
CodiePP: the first reported slot might not be recorded in the first log file. that's why the script looks at the first three files.

Jimbo4350:
@deepfire can we close this?

PR #437, opened 2019-12-19
Tx generator: pure availableFunds.

denisshevchenko:
After #359 was done, we saw that `availableFunds` should not be `TVar` actually. Now we have to turn it from the mutable value to a pure one.

This PR fixes `log-config-explorer.yaml` for local cluster as well.

### DIFF
--- a/cardano-node/src/Cardano/CLI/Tx/Generation.hs
+++ b/cardano-node/src/Cardano/CLI/Tx/Generation.hs
@@ -26,7 +26,6 @@ import           Prelude (String)
 import           Codec.CBOR.Read (deserialiseFromBytes)
 import           Control.Concurrent (threadDelay)
 import qualified Control.Concurrent.STM as STM
-import qualified Control.Concurrent.STM.TVar as TVar
 import           Control.Exception (IOException)
 import qualified Control.Exception as Exception
 import           Control.Monad (forM, forM_, mapM, when)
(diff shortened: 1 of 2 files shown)


cardano-node/src/Cardano/CLI/Tx/Generation.hs
```diff
@@ -816,16 +770,22 @@ createMoreFundCoins llTracer
...
-  liftIO $ forM_ splittingTxs $ \(tx, txDetailsList) -> do
+                                         (genesisTxDetails, sourceKey)
+                                         numSplittingTxOuts
+                                         numOutsPerSplittingTx
+                                         42
+                                         txOut
+                                         []
+  -- Submit all splitting transactions sequentially.
+  liftIO $ forM_ splittingTxs $ \(tx, _) ->
     submitTx socketFp pInfoConfig (CoreId 0) tx llTracer
```
ksaric: So all the (splitting) transactions go into a single node (`CoreId 0`)?

denisshevchenko: Yes.

cardano-node/src/Cardano/CLI/Tx/Generation.hs
```diff
@@ -816,16 +770,22 @@ createMoreFundCoins llTracer
...
-    -- Update available fundValueStatus to reuse the numSplittingTxOuts TxOuts.
-    forM_ txDetailsList (addToAvailableFunds availableFunds)
+
+  -- Re-create availableFunds with information about all splitting transactions
+  -- (it will be used for main transactions).
+  right . Set.fromList $ map (\txDetails -> FundValueStatus { status    = Unknown
```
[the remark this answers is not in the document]
denisshevchenko: Done.

cardano-node/src/Cardano/CLI/Tx/Generation.hs
CodiePP: this is not subject to benchmarking; it's OK

PR #444, opened 2019-12-23
CAD-379 Update `nix/cardano-node-service.nix` with tracing arguments

Jimbo4350:
Previously not specifying a tracer did not turn it off, it just changed its severity. Currently if you do not specify a tracer, that tracer is turned off. Therefore this is a temporary fix until https://github.com/input-output-hk/cardano-node/issues/443 is implemented.

### DIFF
--- a/scripts/shelley-testnet2.sh
+++ b/scripts/shelley-testnet2.sh
@@ -11,7 +11,22 @@ set -e
 # CMD="stack exec --nix cardano-node --"
 CMD="cabal v2-run --"
 # EXTRA="--live-view"
-EXTRA=""
+EXTRA="
+  --trace-block-fetch-decisions
+  --trace-block-fetch-client
+  --trace-block-fetch-server
+  --trace-chain-db
+  --trace-tx-inbound
+  --trace-tx-outbound
+  --trace-local-tx-submission-server
+  --trace-mempool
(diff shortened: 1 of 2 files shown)


scripts/shelley-testnet2.sh
```diff
@@ -11,7 +11,21 @@ set -e
...
 # EXTRA="--live-view"
-EXTRA=""
+EXTRA="
+  --trace-block-fetch-decisions
+  --trace-block-fetch-client
+  --trace-block-fetch-server
+  --trace-tx-inbound
+  --trace-tx-outbound
+  --trace-local-tx-submission-server
+  --trace-mempool
```
CodiePP: missing: trace-chain-db

Jimbo4350:
> how can this be tested?

I'm not sure, the running the output of  `nix-build -A scripts.staging.node` on `master` gives the following error:
```cardano-node: Error in $: expected [a], encountered Object
  error, called at src/Cardano/Node/Run.hs:155:14 in cardano-node-1.2.0-LOpCdC5zsAm4JmQZiPafiR:Cardano.Node.Run%                                              
```
The output does look correct however.

The scripts successfully passing in hydra should confirm that I didn't break anything.

PR #446, opened 2019-12-27
Bump ouroboros-network (fix memory leak).

jbgi:


mrBliss:
Note that there's still a slow memory leak in the node, but after investigating it the past few days, I'm getting more and more confident that it is not in consensus (`ouroboros-network`) but in `cardano-node`.

jbgi:
bors retry

jbgi:
bors retry

jbgi:
bors retry

Issue #443, opened 2019-12-23
Move tracers into config `.yaml` file


Jimbo4350:
Tracers should be specified from the config `.yaml` file as having them on the command line is cumbersome.

CodiePP:
it probably would make sense to type them like: `Maybe Severity`. This way we can distinguish whether they are on or off, and what the minimum severity is that will pass the filter on this trace.
what do you think?

Jimbo4350:
@CodiePP  I was thinking of a dictionary structure like this:
```
TraceBlockFetchClient:
    On: True
    Severity: Debug
    Verbosity: MaximumVerbosity
```
Or we can add an additional constructor to `Severity` (maybe `Off` for example) and we can simplify the configuration to:
```
TraceBlockFetchClient:
    Severity: Off
    Verbosity: MaximumVerbosity
```
or 
```
TraceBlockFetchClient:
    Severity: Debug
    Verbosity: MaximumVerbosity
```

CodiePP:
> I was thinking of a dictionary structure per tracer in the yaml file like this:
> 
> ```
> TraceBlockFetchClient:
>     On: True
>     Severity: Debug
>     Verbosity: MaximumVerbosity
> ```

or `enabled: True` ?    can turn on/off the complete trace
currently, verbosity is a global switch. But, this would give more flexibility if it can be defined per trace.
and "Severity" could be renamed to `MinSeverity`? it is actually the minimum severity level that passes the filter.
you are right, a structure is much better to describe all these options.

Jimbo4350:
```
TraceBlockFetchClient:
    Enabled: True
    MinSeverity: Debug
    Verbosity: MaximumVerbosity
```
Made the changes you suggested above :+1: 

I'd like to make verbosity definable per trace unless there is a reason not to.

CodiePP:
could there also be a global default for MinSeverity? so if it is not indicated for a tracer, then the default will be applied.
same goes for the verbosity, which could even have an implicit default of the global default: NormalVerbosity.

Jimbo4350:
Yep we could define that via `Maybe Severity`/ `Maybe Verbosity`. If you get `Nothing` (parsing an empty field) then we can apply a default `Severity`/`Verbosity`.

PR #447, opened 2019-12-27
Move tracers to `.yaml` files

Jimbo4350:
Specifying the tracers on the command line is cumbersome and the consensus is to move to them to the configuration `.yaml` files. 

Relevant: #443

### DIFF
--- a/cardano-config/src/Cardano/Config/Types.hs
+++ b/cardano-config/src/Cardano/Config/Types.hs
@@ -104,7 +104,6 @@ data NodeCLI = NodeCLI
     , genesisHash :: !Text
     , nodeAddr :: !NodeAddress
     , configFp :: !ConfigYamlFilePath
-    , traceOpts :: !TraceOptions
     , validateDB :: !Bool
     } deriving Show
 
@@ -157,6 +156,7 @@ data NodeConfiguration =
       , ncPbftSignatureThresh :: Maybe Double
       , ncLoggingSwitch :: Bool
       , ncLogMetrics :: Bool
+      , ncTraceOptions :: TraceOptions
       , ncViewMode :: ViewMode
       , ncUpdate :: Update
       } deriving (Show)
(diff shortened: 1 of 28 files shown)


cardano-config/src/Cardano/Config/Types.hs
```diff
@@ -324,38 +352,27 @@ instance FromJSON ViewMode where
 --   which verbosity to the log output.
 data TraceOptions = TraceOptions
   { traceVerbosity :: !TracingVerbosity
+  , traceBlockFetchClient :: !Bool
```
Jimbo4350: This is fragile. Currently if you add a tracer, you would need to make sure that 'NodeConfiguration`'s `FromJSON` instance "aligns" with `TraceOptions`.

Potential solution: Create a data type isomorphic to bool for each tracer. I'd prefer a more elegant solution to this.

rvl: If you split this type, you should be able to use Generic FromJSON, and have a nicer config format. Something like this:

```haskell
data TraceOptions = TraceOptions
  { traceVerbosity :: !TracingVerbosity
  , traceEnable :: !TraceEnable
  }
data TraceEnable = TraceEnable
  { traceBlockFetchClient :: !Bool
  , traceBlockFetchDecisions :: !Bool
  , traceBlockFetchProtocol :: !Bool
  , ...
  }
```

Then the YAML file could look like:
```yaml
tracing:
  verbosity: brief
  enable:
  - BlockFetchClient: false
  - BlockFetchDecisions: false
  - BlockFetchProtocol: false
  - ...
```

Jimbo4350: @rvl the plan so far seems to be having the tracers formatted in the yaml file as follows:
(https://github.com/input-output-hk/cardano-node/issues/443#issuecomment-569678510)

```
TraceBlockFetchClient:
    Enabled: True
    MinSeverity: Debug
    Verbosity: MaximumVerbosity
```
Splitting `TraceOptions` doesn't alleviate the potential to mix up of tracers if the `FromJSON` instance of `NodeConfiguration` and `TraceOptions` records aren't aligned.

Jimbo4350:
Failing with:
```
machine# [   16.724792] bc50mcgyx2rxcw6a1n4dngsvp8mknp48-unit-script-cardano-node_-start[1183]: Error in $: key "TracingVerbosity" not present
```
Need to update the generation of the config `.yaml` files in the nix chairman cluster test.

jbgi:
https://github.com/input-output-hk/cardano-node/commit/95d9fb239ef4548149ef157d2163d3a10ca7a610 should probably be reverted as part of this PR.

nix/nixos/cardano-node-service.nix
```diff
@@ -25,11 +25,6 @@ let
             "--signing-key ${cfg.signingKey}"}"
           "${lib.optionalString (cfg.delegationCertificate != null)
             "--delegation-certificate ${cfg.delegationCertificate}"}"
-          "--tracing-verbosity-${cfg.tracingVerbosity}"
```
jbgi: the `tracingVerbosity` option definition could be removed too.

Jimbo4350: @jbgi removed. This should pass ci now. Once it does are we good to merge: https://github.com/input-output-hk/iohk-nix/pull/303

Jimbo4350:
Don't merge yet. I need to re-point `iohk-nix` at `master`.

PR #454, opened 2020-01-06
CAD-371 trace acceptor removed

CodiePP:
the logging framework does not contain the trace forwarder/acceptor anymore; kept in a plugin which can be loaded separately.

~depends on https://github.com/input-output-hk/iohk-monitoring-framework/pull/498~

^^ upstream PR merged


scripts/shelley-testnet.sh
```diff
@@ -38,24 +39,21 @@ EXTRA="
...
 
-# for acceptor logs:
+# for logs:
 mkdir -p logs/
 
 PWD=$(pwd)
 
-tmux split-window -v
-tmux select-pane -t 0
+#tmux split-window -v
```
ksaric: So opening 4 tmux windows instead of 5?

CodiePP:
@ksaric removed all the mentioning of '#messagecounters' in configurations.

PR #457, opened 2020-01-09
CAD-384:  add mempoolBytes metric

deepfire:
1. Expose the mempool size (in bytes) as a new metric.
2. Add the metric to the TUI LiveView interface
3. A bit more order in the trace hierarchy.

### DIFF
--- a/cardano-node/src/Cardano/Node/TUI/LiveView.hs
+++ b/cardano-node/src/Cardano/Node/TUI/LiveView.hs
@@ -203,6 +203,13 @@ instance IsEffectuator LiveViewBackend Text where
                         return $ lvs { lvsMempool = lvsMempool'
                                      , lvsMempoolPerc = percentage
                                      }
+            LogObject _ _ (LogValue "mempoolBytes" (PureI mempoolBytes)) ->
+                modifyMVar_ (getbe lvbe) $ \lvs -> do
+                        let lvsMempoolBytes' = fromIntegral mempoolBytes :: Word64
+                            percentage = fromIntegral lvsMempoolBytes' / fromIntegral (lvsMempoolCapacityBytes lvs) :: Float
+                        return $ lvs { lvsMempoolBytes = lvsMempoolBytes'
+                                     , lvsMempoolBytesPerc = percentage
+                                     }
             LogObject _ _ (LogValue "density" (PureD density)) ->
                 modifyMVar_ (getbe lvbe) $ \lvs ->
                         return $ lvs { lvsChainDensity = 0.05 + density * 100.0 }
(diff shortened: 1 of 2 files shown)


scripts/lib-node.sh
```diff
@@ -1,6 +1,42 @@
...
+while test -n "$1"
+do case "$1" in
+           --cabal )              mode='cabal';;
+           --stack )              mode='stack';;
+           --stack-nix )          mode='stack-nix';;
+           --help | "--"* ) usage; exit 1;;
+           * ) break;; esac; shift; done
+if   test "$mode" != 'default'
+then true
+elif test -d 'dist-newstyle'
```
CodiePP: I am not sure this will work. What if $CWD is not where 'dist-newstyle' is?
same for '.stack-work'.

scripts/lib-node.sh
```diff
@@ -1,6 +1,42 @@
...
+then mode='cabal'
+elif test -d '.stack-work'
+then mode='stack'
+else mode=$default_mode
+fi
+
+executable_runner() {
+        case ${mode} in
+                cabal )     echo "cabal v2-run -- exe:$1";;
```
CodiePP: I would put the '--' after the executable name, before its arguments.

deepfire: Done.

scripts/lib-node.sh
deepfire: Done.

scripts/lib-node.sh
```diff
@@ -1,9 +1,45 @@
...
+                stack )     echo "stack exec -- $1";;
+                stack-nix ) echo "stack exec --nix $1 --";;
```
CodiePP: still the '--' after the executable.

scripts/lib-node.sh
CodiePP: I meant: "cabal v2-run exe:$1 -- "

scripts/lib-node.sh
```diff
@@ -1,9 +1,45 @@
...
+then mode='cabal'
+elif test -d '.stack-work'
+then mode='stack'
+else mode=$default_mode
+fi
+
+executable_runner() {
+        case ${mode} in
+                cabal )     echo "cabal v2-run -- exe:$1";;
+                stack )     echo "stack exec -- $1";;
```
CodiePP: `"stack exec $1 -- "`

cardano-config/src/Cardano/Config/Logging.hs
```diff
@@ -255,7 +255,7 @@ loggingCardanoFeatureInit disabled' conf = do
 
 startCapturingMetrics :: Trace IO Text -> IO ()
 startCapturingMetrics trace0 = do
-  let trace = appendName "node-metrics" trace0
+  let trace = appendName "node.metrics" trace0
```
CodiePP: these should not be named the same as other metrics. we want to keep them in a separate name context so they can be routed differnt. please, don't change the name here.

scripts/lib-node.sh
deepfire: Done (also switched to `stack run`).

cardano-config/src/Cardano/Config/Logging.hs
deepfire: Done.

deepfire:
[liveview.txt](https://github.com/input-output-hk/cardano-node/files/4053090/liveview.txt)

deepfire:
Screenshot from 2020-01-13 13-26-41
..is the updated Live View.

PR #458, opened 2020-01-09
CAD-405:  add cabal/stack mode switches & detection

deepfire:
1. add `--cabal` / `--stack` / `--stack-nix` switches to `shelley-testnet` scripts
2. absent switches, detect mode from existing `dist-newstyle` `.stack-work` directories
3. absent directories, default to cabal
4. make all scripts position-independent


mrBliss:
Are these the only two scripts to modify? I think there are more, e.g., https://github.com/input-output-hk/cardano-node/blob/master/scripts/shelley-testnet-dns.sh

deepfire:
Converted all the scripts & also made them position-independent (they weren't previously).

cc @CodiePP @mrBliss.

PR #462, opened 2020-01-09
ToObjectOrphans: Fix typos in some keys

rvl:
- Fixes typos in some `toObject` keys

Is there a schema for the `toObject` values? I notice that the `"kind"` object key is used a lot. Also did you consider using generics or TH to reduce boilerplate in these instances?


mrBliss:
I think that module really needs some cleaning up. I have complained before that there were at least 3 (IIRC) ways of printing a slot number...

PR #467, opened 2020-01-13
[CAD-409] Update dependencies for ouroboros-network.

ksaric:
https://jira.iohk.io/browse/CAD-409

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -163,37 +163,37 @@ source-repository-package
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/ouroboros-network
-  tag: 408f3a3fc8be601075ee2bacdb7b7b9ebac9d2c2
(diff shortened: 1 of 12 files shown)


cabal.project
```diff
@@ -110,49 +110,49 @@ source-repository-package
   type: git
   location: https://github.com/input-output-hk/iohk-monitoring-framework
   subdir: iohk-monitoring
-  tag: c35f302ad39568b9813f20a0108049627bd7ace0
+  tag: e0eba52e8e7c81d3f002c8d36a3e5f6ed30bb216
```
mrBliss: It seems like you accidentally changed the hashes of `iohk-monitoring-framework` instead of `ouroboros-network` in this file.

ksaric: Yep, saw about the Nix deps, missed the wrong project. FIxed.

PR #483, opened 2020-01-20
Break up `ToObjectOrphans.hs`

Jimbo4350:
This PR breaks up the `ToObjectOrphans.hs` module (which was becoming unwieldy) in to smaller modules defined by the tracer.

### DIFF
--- a/cardano-node/src/Cardano/Tracing/Tracers.hs
+++ b/cardano-node/src/Cardano/Tracing/Tracers.hs
@@ -68,10 +68,12 @@ import qualified Ouroboros.Storage.ChainDB as ChainDB
 import           Cardano.Config.Protocol (TraceConstraints)
 import           Cardano.Config.Types
 import           Cardano.Tracing.ToObjectOrphans.ChainDBTrace ()
+import           Cardano.Tracing.ToObjectOrphans.DnsTrace ()
+import           Cardano.Tracing.ToObjectOrphans.ErrorPolicyTrace ()
+import           Cardano.Tracing.ToObjectOrphans.MuxTrace ()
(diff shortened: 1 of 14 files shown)


cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -67,8 +67,18 @@ import qualified Ouroboros.Storage.ChainDB as ChainDB
...
+import           Cardano.Tracing.ToObjectOrphans.ConsensusToObjectOrphans
+import           Cardano.Tracing.ToObjectOrphans.DnsTrace ()
+import           Cardano.Tracing.ToObjectOrphans.ErrorPolicyTrace ()
+import           Cardano.Tracing.ToObjectOrphans.ForgeEventTrace ()
+import           Cardano.Tracing.ToObjectOrphans.LabelPeerTrace ()
+import           Cardano.Tracing.ToObjectOrphans.MuxTrace ()
+import           Cardano.Tracing.ToObjectOrphans.SubscriptionTrace ()
+import           Cardano.Tracing.ToObjectOrphans.TxSubmissionTrace ()
 import           Cardano.Tracing.MicroBenchmarking
-import           Cardano.Tracing.ToObjectOrphans
```
CodiePP: I would just keep ToObjectOrphans.hs that imports/exports all the others

CodiePP:
not all of your commits are signed. why?

cardano-node/src/Cardano/Tracing/Tracers.hs
deepfire: Frankly, I second this.  The way the change currently looks like is it's adding work in future with no obvious net gain.

PR #484, opened 2020-01-20
CAD-425:  journald logging

deepfire:
Task: https://jira.iohk.io/browse/CAD-425

### DIFF
--- a/cardano-config/src/Cardano/Config/Logging.hs
+++ b/cardano-config/src/Cardano/Config/Logging.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP                 #-}
 {-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE Rank2Types          #-}
 {-# LANGUAGE ScopedTypeVariables #-}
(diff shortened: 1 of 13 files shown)


cardano-config/src/Cardano/Config/Logging.hs
```diff
@@ -227,6 +231,11 @@ loggingCardanoFeatureInit disabled' conf = do
   Cardano.BM.Backend.Monitoring.plugin logConfig trace switchBoard
       >>= loadPlugin switchBoard
 
+#if defined(linux_HOST_OS)
+  Cardano.BM.Scribe.Systemd.plugin logConfig baseTrace switchBoard
```
CodiePP: 'baseTrace' should be 'trace' so it can be turned on/off.

deepfire: @CodiePP, done.

configuration/log-configuration.yaml
```diff
@@ -76,6 +76,10 @@ options:
     cardano.node.peers.BlockFetchDecision:
       - kind: UserDefinedBK
         name: LiveViewBackend
+  mapScribes:
+    cardano:
+      - JournalSK
```
CodiePP: this should be added to 'defaultScribes'

deepfire: > this should be added to 'defaultScribes'

There is that old parsing problem popping up, which I mentioned previously:
```
defaultScribes:
  - - FileSK
    - "node-0.log"
  - - JournalSK
    - "foo"
```
..yields
```
AesonException "Error in $.defaultScribes[1][0]: The key \"JournalSK\" was not found"

cardano-node: Aeson exception:
Error in $.defaultScribes[1][0]: The key "JournalSK" was not found
```

deepfire:
@CodiePP, waiting for your feedback on the design question re: plugin separation .

CodiePP:
there was a way to integrate the lobemo-scribe-systemd plugin.
this PR is replaced with PR #502.

PR #493, opened 2020-01-23
CAD-430, CAD-429: live peer list & other TUI improvements

deepfire:
TUI changes:

1. give the TUI direct access to the `NodeKernel`, to allow easier & richer data queries
1. added a peer list screen, accessible by the 'p' key
1. minor string changes
1. compressed the mempool bars in one line
1. a bunch of minor refactoring

Screenshot from 2020-01-30 17-01-48

### DIFF
--- a/cardano-node/src/Cardano/Node/Run.hs
+++ b/cardano-node/src/Cardano/Node/Run.hs
@@ -106,34 +106,37 @@ runNode loggingLayer nc nCli = do
                                          (ncUpdate nc)
                                          (ncProtocol nc)
 
-    SomeProtocol p <- case eitherSomeProtocol of
-                        Left err -> (putTextLn . pack $ show err) >> exitFailure
-                        Right (SomeProtocol p) -> pure $ SomeProtocol p
+    SomeProtocol (p :: Consensus.Protocol blk) <-
+      case eitherSomeProtocol of
+        Left err -> (putTextLn . pack $ show err) >> exitFailure
+        Right (SomeProtocol p) -> pure $ SomeProtocol p
 
     tracers <- mkTracers (ncTraceOptions nc) trace
 
     case ncViewMode nc of
-      SimpleView -> handleSimpleNode p trace tracers nCli nc
+      SimpleView -> handleSimpleNode p trace tracers nCli nc (const $ pure ())
       LiveView   -> do
 #ifdef UNIX
         let c = llConfiguration loggingLayer
         -- We run 'handleSimpleNode' as usual and run TUI thread as well.
         -- turn off logging to the console, only forward it through a pipe to a central logging process
         CM.setDefaultBackends c [KatipBK, TraceForwarderBK, UserDefinedBK "LiveViewBackend"]
-        -- User will see a terminal graphics and will be able to interact with it.
-        nodeThread <- Async.async $ handleSimpleNode p trace tracers nCli nc
 
-        be :: LiveViewBackend Text <- realize c
+        be :: LiveViewBackend blk Text <- realize c
         let lvbe = MkBackend { bEffectuate = effectuate be, bUnrealize = unrealize be }
         llAddBackend loggingLayer lvbe (UserDefinedBK "LiveViewBackend")
         let nId = fromMaybe (panic "LiveView not possible for real protocols as yet") (ncNodeId nc)
         setTopology be nId
-        setNodeThread be nodeThread
         captureCounters be trace
+
+        -- User will see a terminal graphics and will be able to interact with it.
+        nodeThread <- Async.async $ handleSimpleNode p trace tracers nCli nc
+                       (setNodeKernel be)
+        setNodeThread be nodeThread
 
         void $ Async.waitAny [nodeThread]
 #else
-        handleSimpleNode p trace tracers nCli nc
+        handleSimpleNode p trace tracers nCli nc (const $ pure ())
 #endif
   where
     hostname = do
@@ -143,14 +146,19 @@ runNode loggingLayer nc nCli = do
 -- | Sets up a simple node, which will run the chain sync protocol and block
 -- fetch protocol, and, if core, will also look at the mempool when trying to
 -- create a new block.
-handleSimpleNode :: forall blk. RunNode blk
-                 => Consensus.Protocol blk
-                 -> Tracer IO (LogObject Text)
-                 -> Tracers ConnectionId blk
-                 -> NodeCLI
-                 -> NodeConfiguration
-                 -> IO ()
-handleSimpleNode p trace nodeTracers nCli nc = do
+handleSimpleNode
+  :: forall blk. RunNode blk
+  => Consensus.Protocol blk
+  -> Tracer IO (LogObject Text)
+  -> Tracers ConnectionId blk
+  -> NodeCLI
+  -> NodeConfiguration
+  -> (NodeKernel IO ConnectionId blk -> IO ())
+  -- ^ Called on the 'NodeKernel' after creating it, but before the network
+  -- layer is initialised.  This implies this function must not block,
+  -- otherwise the node won't actually start.
+  -> IO ()
+handleSimpleNode p trace nodeTracers nCli nc onKernel = do
     NetworkTopology nodeSetups <-
       either error id <$> readTopologyFile (unTopology . topFile $ mscFp nCli)
 
@@ -255,6 +263,7 @@ handleSimpleNode p trace nodeTracers nCli nc = do
         void $ onEachChange registry id Nothing
                             (ChainDB.getTipPoint chainDB) $ \tip ->
           atomically $ writeTVar varTip tip
+        onKernel nodeKernel
   where
     nid :: Int
     nid = case ncNodeId nc of
--- a/cardano-node/src/Cardano/Node/TUI/LiveView.hs
+++ b/cardano-node/src/Cardano/Node/TUI/LiveView.hs
@@ -1,12 +1,15 @@
 {-# LANGUAGE BangPatterns          #-}
+{-# LANGUAGE CPP                   #-}
 {-# LANGUAGE DataKinds             #-}
 {-# LANGUAGE DeriveAnyClass        #-}
+{-# LANGUAGE DeriveFunctor         #-}
+{-# LANGUAGE DeriveFoldable        #-}
 {-# LANGUAGE DeriveGeneric         #-}
+{-# LANGUAGE DeriveTraversable     #-}
 {-# LANGUAGE FlexibleInstances     #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings     #-}
 {-# LANGUAGE RankNTypes            #-}
-{-# LANGUAGE CPP                   #-}
 
 {-# OPTIONS_GHC -Wno-redundant-constraints #-}
 
@@ -17,6 +20,7 @@ module Cardano.Node.TUI.LiveView (
     , captureCounters
     , setTopology
     , setNodeThread
+    , setNodeKernel
     ) where
 
 import           Cardano.Prelude hiding (isPrefixOf, modifyMVar_, newMVar, on,
@@ -30,6 +34,9 @@ import           Control.Concurrent.MVar.Strict (MVar, modifyMVar_, newMVar, rea
 import           Control.DeepSeq (rwhnf)
 import           Control.Monad (forever, void)
 import           Control.Monad.IO.Class (liftIO)
+import qualified Control.Monad.Class.MonadSTM.Strict as STM
+import qualified Data.Map.Strict as Map
+import qualified Data.Set as Set
 import           Data.Text (Text, pack, unpack)
 import           Data.Time.Calendar (Day (..))
 import           Data.Time.Clock (NominalDiffTime, UTCTime (..), addUTCTime,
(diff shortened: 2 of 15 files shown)


CodiePP:
* description?
* screenshots?
* why was it necessary to upgrade ouroboros-network?
* something not on 'master' -> PR cannot be merged!

deepfire:
@CodiePP, I was operating under the impression was that we moved the book keeping to the other tracker.  How much information should we duplicate between them?

Regarding the ouroboros-network bump -- I believe you acked https://github.com/input-output-hk/ouroboros-network/issues/1510 on Slack.  Indeed that PR wasn't merged yet.

cardano-node/src/Cardano/Node/TUI/LiveView.hs
```diff
@@ -556,23 +652,26 @@ headerW p =
...
+systemStatsW :: LiveViewState blk a -> Widget ()
 systemStatsW p =
       padTop   (T.Pad 1)
     . padLeft  (T.Pad 2)
     . padRight (T.Pad 2)
-    $ vBox [ vBox [ hBox [ txt "Mempool txs:"
-                         , withAttr barValueAttr . padLeft T.Max . str . show $ lvsMempoolCapacity p
+    $ vBox [ hBox [ vBox [ hBox [ txt "Mempool (KBs):"
+                                , withAttr barValueAttr . padLeft T.Max . str .
+                                  show . (floor :: Float -> Int) . (/ 1000) . fromIntegral $ lvsMempoolCapacityBytes p
```
CodiePP: KB are kilobytes? don't they need to be divided by 1024? despite that this is not correct SI unit?

deepfire: I was trying to be correct re: the unit nomenclature, yes -- (/1000) is KB, (/1024) is KiB -- which one do we want, @CodiePP?

deepfire: Going for KiB, as per private discussion.

CodiePP: the file is full of "KB = 1024 bytes" assumptions. This is not SI-correct but to stay consistent within this file just use it like that.

karknu:
<3
Can you add uptime, and epoch/slot to the Peer view too?

deepfire:
@karknu, done! : -)

cardano-node/src/Cardano/Node/TUI/LiveView.hs
deepfire: @CodiePP, done.

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -149,8 +152,9 @@ handleSimpleNode :: forall blk. RunNode blk
                  -> Tracers ConnectionId blk
                  -> NodeCLI
                  -> NodeConfiguration
+                 -> (NodeKernel IO ConnectionId blk -> IO ())
```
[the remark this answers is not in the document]
deepfire: @Jimbo4350, two points:

1. I'm hesitant to introduce coupling like this into `handleSimpleNode`, which is currently abstract re: the TUI -- the latter only being introduced at the `runNode` level.
2. I've modelled this callback after `Ouroboros.Consensus.Node.run` -- as in https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs#L105

How do these arguments factor in your view of things?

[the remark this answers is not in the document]
dcoutts: You have to be careful with a callback like this. It is not allowed to block, or the node will not actually run. This is clearly documented for `Node.run` but is not documented here.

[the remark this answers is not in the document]
deepfire: @dcoutts, will it be acceptable, if I document that?

cardano-node/src/Cardano/Node/TUI/LiveView.hs
```diff
@@ -288,10 +324,43 @@ data LiveViewState a = LiveViewState
     , lvsUIThread            :: Maybe (Async.Async ())
     , lvsMetricsThread       :: Maybe (Async.Async ())
     , lvsNodeThread          :: Maybe (Async.Async ())
+    , lvsNodeKernel          :: Maybe (NodeKernel IO ConnectionId blk)
+    , lvsPeers               :: [PeerUI blk]
```
ksaric: As I mentioned, I would create separation from this `NodeKernel` mess here and just introduce the variables that you actually use in the LiveView.
Coupling everything with references to huge data structures is a no-no in my book.
That's just my view.

deepfire: From my standpoint there is another important question -- what is the intent/mandate of the `LiveViewState` data structure?

Is it for keeping all of the state that the `LiveView` needs?  So far it has been.

cardano-node/src/Cardano/Node/Run.hs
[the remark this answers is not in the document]
deepfire: @dcoutts, added the documentation.

cardano-node/src/Cardano/Node/TUI/LiveView.hs
ksaric: So we don't depend on the data structure that we don't use.
Why don't we have a single `GlobalState` passing around and be done with it?

deepfire: > So we don't depend on the data structure that we don't use.

But we _do_ depend on it!

ksaric: I'm not sure if there is some language barrier here or I'm not making sense but I'm not sure how you don't see that this question about `GlobalState` is exactly what you are describing above.

What  would be wrong if you had a field inside `LiveViewState`:
```
ppPeer :: Text
```

You would have to move the computation before assigning it to the data type and you wouldn't have to jump through all these hoops?

deepfire:
Building everything with profiling, to ensure we don't add any leaks here..

cardano-node/src/Cardano/Node/TUI/LiveView.hs
```diff
@@ -423,60 +475,61 @@ initLiveViewState = do
...
+                , lvsDiskUsageRLast      = 0
+                , lvsDiskUsageRNs        = 10000
+                , lvsDiskUsageWLast      = 0
+                , lvsDiskUsageWNs        = 10000
+                , lvsNetworkUsageInLast  = 0
+                , lvsNetworkUsageInNs    = 10000
+                , lvsNetworkUsageOutLast = 0
+                , lvsNetworkUsageOutNs   = 10000
+                , lvsMempoolCapacity     = mempoolCapacity
+                , lvsMempoolCapacityBytes = mempoolCapacity * maxBytesPerTx
```
CodiePP: oops!

cardano-node/src/Cardano/Node/TUI/LiveView.hs
```diff
@@ -750,7 +873,12 @@ systemStatsW p =
                  ) $ bar mempoolBytesLabel (lvsMempoolBytesPerc p)
     mempoolBytesLabel = Just $ (show . lvsMempoolBytes $ p)
                         ++ " / "
-                        ++ take 5 (show $ lvsMempoolBytesPerc p * 100) ++ "%"
+                        ++ trimDecimalPoints
+                           (take 5 (show $ lvsMempoolBytesPerc p * 100)) ++ "%"
+    trimDecimalPoints :: String -> String
+    -- NOTE: breaks badly, when the float is 'show'n in exponential form: "1.0e-3"
+    --       This is a tradeoff for not using 'printf'.
+    trimDecimalPoints x = fromJust $ take <$> ((+ 2) <$> elemIndex '.' x) <*> Just x
```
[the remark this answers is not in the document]
CodiePP: why not "withOneDecimal :: Double -> String" ?    and case split on the value?

cardano-node/src/Cardano/Node/TUI/LiveView.hs
deepfire: @ksaric, how would that be computed?

cardano-node/src/Cardano/Node/TUI/LiveView.hs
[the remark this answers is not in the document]
deepfire: @CodiePP, done.

cardano-node/src/Cardano/Node/TUI/LiveView.hs
deepfire: @CodiePP, done!

cardano-node/src/Cardano/Node/TUI/LiveView.hs
```diff
@@ -146,11 +167,25 @@ data LiveViewState a = LiveViewState
     , lvsMempoolCapacity      :: !Word64
     , lvsMempoolCapacityBytes :: !Word64
     , lvsMessage              :: !(Maybe a)
+    -- Async threads.
     , lvsUIThread             :: !LiveViewThread
     , lvsMetricsThread        :: !LiveViewThread
     , lvsNodeThread           :: !LiveViewThread
+
+    , lvsNodeKernel           :: Maybe (LVNodeKernel blk)
```
njd42: Shouldn't this be strict?

cardano-node/src/Cardano/Node/TUI/LiveView.hs
```diff
@@ -308,14 +342,28 @@ instance IsEffectuator LiveViewBackend Text where
...
+                            candidates <- fromMaybe mempty <$>
+                              sequence (atomically . getCandidates . getNodeCandidates . getNodeKernel <$> lvsNodeKernel lvs)
+
                             return $ lvs { lvsNetworkUsageInCurr = currentNetRate
                                          , lvsNetworkUsageInPerc = (currentNetRate / (maxNetRate / 100.0)) / 100.0
                                          , lvsNetworkUsageInLast = inBytes
                                          , lvsNetworkUsageInNs   = currentTimeInNs
                                          , lvsNetworkUsageInMax  = maxNetRate
                                          , lvsUpTime             = diffUTCTime (tstamp meta) (lvsStartTime lvs)
+                                         , lvsPeers = Map.elems . flip Map.mapMaybeWithKey candidates $
```
njd42: You might want to pull this calculation out into a supporting function and put it near the newtype above - then it could be clearly documented that this is dependent on details of the wrapped data structure - just for longer term maintenance

cardano-node/src/Cardano/Node/TUI/LiveView.hs
```diff
@@ -750,63 +869,68 @@ systemStatsW p =
...
     bar :: forall n. Maybe String -> Float -> Widget n
-    bar lbl pcntg = P.progressBar lbl pcntg
+    bar = P.progressBar
     lvsMemUsagePerc = lvsMemoryUsageCurr p / max 200 (lvsMemoryUsageMax p)
 
-nodeInfoW :: LiveViewState a -> Widget ()
+withOneDecimal :: (Fractional f, Ord f, Show f) => f -> String
+withOneDecimal x
+  | x < 0.1 = "0.0"
+  | otherwise = uncurry (<>) . (id *** take 2) $ span (/= '.') (show x)
```
njd42: this seems to be the same basic functionality as those in `Numeric` ie `ShowFFloat`

cardano-node/src/Cardano/Node/TUI/LiveView.hs
edsko: Careful that simply adding a `!` on a `Maybe` is pretty useless.

cardano-node/src/Cardano/Node/Run.hs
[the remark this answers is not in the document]
Jimbo4350: @deepfire I was waiting on some feedback from @erikd but I will dismiss my review as its a very small change if we decide what I suggested is favorable. I'm not sure what the benefits are of having this as a callback vs passing in the `MVar` itself (which to me is more readable).

cardano-node/src/Cardano/Node/TUI/LiveView.hs
njd42: I would agree - except - given that this data structure is also an instance of  `NoUnexpectedThunks` you might need it?

cardano-node/src/Cardano/Node/TUI/LiveView.hs
ksaric: Replied in DM on Slack.
If what I wrote doesn't make sense, it's fine this way.
The costs/benefits all mentioned in this conversation.

cardano-node/src/Cardano/Node/TUI/LiveView.hs
deepfire: Done, thanks!

cardano-node/src/Cardano/Node/TUI/LiveView.hs
deepfire: Done, good point!

cardano-node/src/Cardano/Node/TUI/LiveView.hs
deepfire: Done, by adding `SMaybe`.

deepfire:
@CodiePP, this is the plot I'm getting after 45 minutes of running `shelley-testnet-live.sh`:

Screenshot from 2020-01-30 20-11-10

cardano-node/src/Cardano/Node/Run.hs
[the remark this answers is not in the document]
erikd: Sorry, had a crazy week!

Documenting that something is unsafe is fine if it can't be made unsafe, but it if can, make it safe is always better than documenting that it may be unsafe.

mrBliss:
> ..so the definite leaks are:
> 
> [..]

> bors r+

I didn't follow this PR in detail, but your response to "definite leaks" is "bors it"? :confused:

PR #494, opened 2020-01-23
CAD-411 default tracers

CodiePP:
tracers which have severity annotation >= Warning shall be turned on by default.


cardano-config/src/Cardano/Config/Types.hs
```diff
@@ -134,25 +134,25 @@ instance FromJSON NodeConfiguration where
                   tOptions <- TraceOptions
                                 <$> v .: "TracingVerbosity"
```
Jimbo4350: We should default the remaining tracers to `False`

PR #496, opened 2020-01-27
CAD-378 Update ouroboros-network dependency.

denisshevchenko:
We have to update `ouroboros-network` dependency, because it contains bigger size of `mempoolCap` (by default - `maxBlockSize * 2`), but I need a help with this updating, because currently there's an error with other dependencies:

```
Error: While constructing the build plan, the following exceptions were encountered:

In the dependencies for network-mux-0.1.0.0:
    network-2.8.0.1 from stack configuration does not match >=3.1 && <3.2  (latest matching version is 3.1.1.0)
needed due to cardano-node-1.4.0 -> network-mux-0.1.0.0

In the dependencies for ouroboros-consensus-0.1.0.0:
    network-2.8.0.1 from stack configuration does not match >=3.1 && <3.2  (latest matching version is 3.1.1.0)
needed due to cardano-node-1.4.0 -> ouroboros-consensus-0.1.0.0

In the dependencies for ouroboros-network-0.1.0.0:
    network-2.8.0.1 from stack configuration does not match ==3.1.*  (latest matching version is 3.1.1.0)
needed due to cardano-node-1.4.0 -> ouroboros-network-0.1.0.0

Some different approaches to resolving this:

  * Set 'allow-newer: true' in /home/shevchenko/.stack/config.yaml to ignore all version constraints and build anyway.

  * Consider trying 'stack solver', which uses the cabal-install solver to attempt to find some working build configuration. This can be convenient when dealing with many complicated
    constraint errors, but results may be unpredictable.

  * Recommended action: try adding the following to your extra-deps in /home/shevchenko/Code/cardano-node/stack.yaml:

network-3.1.1.0@sha256:c60e754c878bf769476c8f717ca36dfa54a5151d8aa058778decf92680ed3841

Plan construction failed.
```


erikd:
To me looks like `ouroboros-network` needs to be fixed to use a single version of `network`.

coot:
`ouroboros-network` can be build with `stack`, `cabal` and `nix`.  We recently added bounds `network >=3.1 && <3.2` bounds in `ouroboros-network`, I'll go and check if other repositories compile with the against `ouroboros-network` master.

avieth:
`cardano-node` uses stack lts-13.26, which picks `network-2.8.0.1`, which is too old for `ouroboros-network`.

Solution: put `network-3.1.0.1` in the `extra-deps` section of the `stack.yaml` file, and similar for the `cabal.project`.

Jimbo4350:
`stack build cardano-node` is now working but I now have the following error when I run `./nix/update-iohk-nix.sh`

```
error: attribute '0.12.6.1' missing, at /home/jordan/Repos/Work/userepotool/cardano-node/nix/.stack.nix/default.nix:27:26
(use '--show-trace' to show detailed location information)
```

avieth:
> `stack build cardano-node` is working but I now have the following error when I run `./nix/update-iohk-nix.sh`
> 
> ```
> error: attribute '0.12.6.1' missing, at /home/jordan/Repos/Work/userepotool/cardano-node/nix/.stack.nix/default.nix:27:26
> (use '--show-trace' to show detailed location information)
> ```
> 
> `0.12.6.1` refers to the `websockets` library.
> 
> In https://raw.githubusercontent.com/input-output-hk/nixpkgs/iohk-18.09/pkgs/development/haskell-modules/hackage-packages.nix the `websockets` version is `0.12.5.3`. If we try to use `websockets-0.12.5.3` we have to use `network-3.0.1.1` but `ourboros-network` requires `>=3.1 && <3.2` as stated above ([#496 (comment)](https://github.com/input-output-hk/cardano-node/pull/496#issuecomment-578789749)).
> @avieth @coot

I think that means we have to update the hackage revision somewhere in the nix infrastructure. IIRC that's set by the `haskell.nix` program. Should be some way to override it without necessarily updating the revision of `haskell.nix` but I'm not the guy to ask about this.

avieth:
Alternatively, you could just add the git repo and commit sha of the websockets package to stack.yaml and cabal.project

denisshevchenko:
Now, `cardano-node` can be built using `stack build` as well as `cabal new-build all`.

coot:
I think this is about missing `websockets-0.12.6.1`

erikd:
This needs more dependency updates after https://github.com/input-output-hk/cardano-ledger/pull/701 and https://github.com/input-output-hk/ouroboros-network/pull/1542 are merged.

disassembler:
approving the nix changes in here.

erikd:
Might as well get this merged now. The other PRs are nowhere near ready.

denisshevchenko:
> I think this is about missing websockets-0.12.6.1

@coot Ok, how can we add it?

PR #497, opened 2020-01-27
Update transaction creation in `README.md`

Jimbo4350:
Relevant: #410

### DIFF
--- a/cardano-node/README.md
+++ b/cardano-node/README.md
@@ -244,6 +244,16 @@ options.
 
 ## Transactions
 
+### Creation
+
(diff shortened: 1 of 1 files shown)


cardano-node/README.md
```diff
@@ -244,6 +244,16 @@ options.
 
 ## Transactions
 
+### Creation
+
+Transactions can be created via the  `issue-genesis-utxo-expenditure` & `issue-utxo-expenditure` commands.
```
erikd: Can these only spend genesis utxos? Is there a general sub command to generate a plain tx?

PR #506, opened 2020-01-29
Use +RTS -N2 -A16m to improve bulk sync performance

karknu:


cardano-node/cardano-node.cabal
```diff
@@ -140,7 +140,7 @@ executable cardano-node
   ghc-options:         -threaded
                        -Wall
                        -rtsopts
-                       "-with-rtsopts=-T -I0"
+                       "-with-rtsopts=-T -I0 -N2 -A16m"
```
CodiePP: I am not sure this will help. The node is run in various scripts with different RTS flags.
Do we need this here?

karknu: Yes, it needs to be run with those flags.
But if it is bypassed in scripts somewhere it doesn't do much good.

oneEdoubleD:
This was _fast_ - 1h6m for a complete sync of Staging (170 epochs). Previously it was ~6h30m, and the best I've ever actually seen was ~2h30m.

Built with:
`nix-build -A scripts.staging.node`

cardano-node/cardano-node.cabal
mrBliss: In my experiments, having `-N2` (or greater) has been very important for performance.

PR #507, opened 2020-01-29
Haskell nix overhaul

disassembler:


lib.nix
```diff
@@ -1,20 +1,46 @@
...
+      ((hasPrefix "cardano-node" package.identifier.name) ||
+       (elem package.identifier.name [ "text-class" "bech32" ]));
+  filterCardanoPackages = pkgs.lib.filterAttrs (_: package: isCardanoNode package);
+  getPackageChecks = pkgs.lib.mapAttrs (_: package: package.checks);
+in lib // iohkNix.cardanoLib // {
+  inherit (pkgs.haskell-nix.haskellLib) collectComponents;
+  inherit
+    niv
+    sources
+    haskellPackages
```
jbgi: this `haskellPackages` can accidentally be hidden by nixpkgs `pkgs.haskellPackages` in imports like:
```
with ../lib.nix;
with pkgs;
```
should/could we avoid this situation?

PR #522, opened 2020-01-31
Fix windows build

hamishmack:


disassembler:
cherry picking to #524 because of numerous conflicts in both PRs.

PR #525, opened 2020-01-31
SRE-16: Add nix dockerTools image, build and push to dockerhub per commit

craigem:
This completes point 1:

1. base image with just the binaries

This PR provides a method to build docker binaries that cane be used with an appropriate set of CLI flags.

Images are tagged using the the commit in which they are built.

### DIFF
--- a/.buildkite/docker-build-push.nix
+++ b/.buildkite/docker-build-push.nix
@@ -1,0 +1,79 @@
+# This script will load nix-built docker images of cardano-node
+# into the Docker daemon (must be running), and then push to the
+# Docker Hub. Credentials for the hub must already be installed with
+# "docker login".
+#
+# There is a little bit of bash logic to replace the default repo and
+# tag from the nix-build (../nix/docker.nix).
+#
+# 1. The repo (default "inputoutput/cardano-node") is changed to match
+#    the logged in Docker user's credentials.
+#
+# 2. The tag (default "VERSION") is changed to reflect the
+#    branch which is being built under this Buildkite pipeline.
+#
+#    - All commits are tagged with commit hash.
+#    - If the branch is master then master tag is updated
+#    - If there is a tag associated with the commit then that tag is pushed as well
+
+{ nodePackages ?  import ../. {}
+
+# Build system's Nixpkgs. We use this so that we have the same docker
+# version as the docker daemon.
+, hostPkgs ? import <nixpkgs> {}
+
+# Dockerhub repository for image tagging.
+, dockerHubRepoName ? null
+}:
+
+with hostPkgs;
+with hostPkgs.lib;
+
+let
+  images = map impureCreated [
+    nodePackages.dockerImage.base
+    nodePackages.dockerImage.testnet
+    nodePackages.dockerImage.mainnet
+  ];
+
+  # Override Docker image, setting its creation date to the current time rather than the unix epoch.
+  impureCreated = image: image.overrideAttrs (oldAttrs: { created = "now"; }) // { inherit (image) version; };
+
+in
+  writeScript "docker-build-push" (''
+    #!${runtimeShell}
+
+    set -euo pipefail
+
+    export PATH=${lib.makeBinPath [ docker gnused ]}
+
+    ${if dockerHubRepoName == null then ''
+    reponame=cardano-node
+    username="$(docker info | sed '/Username:/!d;s/.* //')"
+    fullrepo="$username/$reponame"
+    '' else ''
+    fullrepo="${dockerHubRepoName}"
+    ''}
+
+  '' + concatMapStringsSep "\n" (image: ''
+    branch="''${BUILDKITE_BRANCH:-}"
+    tag="''${BUILDKITE_TAG:-}"
+    tagged="$fullrepo:$tag"
+    gitrev="${image.imageTag}"
+    echo "Loading $fullrepo:$gitrev"
+    docker load -i ${image}
+    echo "Pushing $fullrepo:$gitrev"
+    docker push "$fullrepo:$gitrev"
+    if [[ "$branch" = master ]]; then
+      echo "Tagging as master"
+      docker tag $fullrepo:$gitrev $fullrepo:$branch
+      echo "Pushing $fullrepo:$branch"
+      docker push "$fullrepo:$branch"
+    fi
+    if [[ "$tag" ]]; then
+      echo "Tagging as ${image.imageTag}"
+      docker tag $fullrepo:$gitrev $fullrepo:$tag
+      echo "Pushing $fullrepo:$tag"
+      docker push "$fullrepo:$tag"
+    fi
+  '') images)
--- a/nix/docker.nix
+++ b/nix/docker.nix
@@ -1,0 +1,119 @@
+############################################################################
+# Docker image builder
+#
+# To test it out, use:
+#
+#   docker load -i $(nix-build -A dockerImage.base --no-out-link)
+#   docker load -i $(nix-build -A dockerImage.latency-tests --no-out-link)
+#   docker load -i $(nix-build -A dockerImage.mainnet --no-out-link)
+#   docker load -i $(nix-build -A dockerImage.mainnet-ci --no-out-link)
+#   docker load -i $(nix-build -A dockerImage.shelley_staging --no-out-link)
+#   docker load -i $(nix-build -A dockerImage.shelley_staging_short --no-out-link)
+#   docker load -i $(nix-build -A dockerImage.staging --no-out-link)
+#   docker load -i $(nix-build -A dockerImage.testnet --no-out-link)
+#
+# For all output except base, run:
+#
+#   docker run inputoutput/cardano-node:<TAG>
+#
+# ie:
+#
+#   docker run inputoutput/cardano-node:6a5b233e5861aa0012287a5557b3a6a08afbf3ca-testnet
+#
+# For base, you will need to provide all variables. ie:
+#
+#   docker run -ti \
+#   --volume $PATH_TO/state-docker:/data \
+#   --volume $PATH_TO/configuration-mainnet.yaml:/config/config.yaml \
+#   --volume $PATH_TO/mainnet-topology.json:/config/topology.json \
+#   --volume $PATH_TO/mainnet-genesis.json:/config/genesis.json \
+#   inputoutput/cardano-node:<GIT_HASH>-custom
+#   --genesis-hash <GENESIS_HASH>
+#
+############################################################################
+
+{ iohkNix
+, commonLib
+, dockerTools
+
+# The main contents of the image.
+, cardano-node
+, scripts
+
+# Get the current commit
+, gitrev ? iohkNix.commitIdFromGitRepoOrZero ../.git
+
+# Other things to include in the image.
+, bashInteractive
+, cacert
+, coreutils
+, curl
+, glibcLocales
+, iana-etc
+, iproute
+, iputils
+, socat
+, utillinux
+
+, repoName ? "inputoutput/cardano-node"
+}:
+
+let
+
+  # Layer of tools which aren't going to change much between versions.
+  baseImage = dockerTools.buildImage {
+    name = "${repoName}-env";
+    contents = [
+      bashInteractive   # Provide the BASH shell
+      cacert            # X.509 certificates of public CA's
+      coreutils         # Basic utilities expected in GNU OS's
+      curl              # CLI tool for transferring files via URLs
+      glibcLocales      # Locale information for the GNU C Library
+      iana-etc          # IANA protocol and port number assignments
+      iproute           # Utilities for controlling TCP/IP networking
+      iputils           # Useful utilities for Linux networking
+      socat             # Utility for bidirectional data transfer
+      utillinux         # System utilities for Linux
+    ];
+    # set up /tmp (override with TMPDIR variable)
+    extraCommands = ''
+      mkdir -m 0777 tmp
+    '';
+  };
+  customImage = dockerTools.buildImage {
+    name = "${repoName}";
+    tag = "${gitrev}-base";
+    fromImage = baseImage;
+    contents = [
+      cardano-node
+    ];
+    created = "now";   # Set creation date to build time. Breaks reproducibility
+    extraCommands = ''
+      mkdir -p data config
+    '';
+    config = {
+      EntryPoint = [
+        "${cardano-node}/bin/cardano-node" "run" "--config"
+        "/config/config.yaml" "--database-path" "/data.db" "--genesis-file"
+        "/config/genesis.json" "--host-addr" "127.0.0.1" "--port" "3001"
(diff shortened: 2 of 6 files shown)


rvl:
In this case I don't see must benefit in supplying command-line parameters to cardano-node in the entrypoint script. Just make the entrypoint cardano-node and let the user do the rest.

Set up `/data` and `/config` using `extraCommands`.

Have you got it to automatically push images to Dockerhub from CI?

rhyslbw:
I agree on using `cardano-node` as the ENTRYPOINT (this was our intent), however the syntax of the  causing problems, and the script worked as expected. @craigem having another look at this now is a good idea.

I disagree on removing the arguments though. it's very nice to simply mount the volumes by convention.

nix/docker.nix
```diff
@@ -0,0 +1,83 @@
...
+, curl
+, socat
+
+# Used to generate the docker image names
+, repoName ? "inputoutput/cardano-node"
+}:
+
+let
+
+  startScript = writeScriptBin "start-cardano-node" ''
```
disassembler: I would prefer this to use scripts.<env>.node so parameters can be overridden with a customConfig parameter.

nix/docker.nix
```diff
@@ -68,14 +62,16 @@ in
...
       cardano-node
-      startScript
     ];
     created = "now";   # Set creation date to build time. Breaks reproducibility
     extraCommands = ''
       mkdir -p data config
     '';
     config = {
-      EntryPoint = [ "start-cardano-node" ];
+      EntryPoint = [ "${cardano-node}/bin/cardano-node" "--config"
```
rhyslbw: @rvl this is the intended design, as per the prototype https://github.com/input-output-hk/cardano-docker-stack/blob/master/services/node/Dockerfile#L19

nix/docker.nix
```diff
@@ -68,14 +62,16 @@ in
...
+      "/config/config.yaml" "--database-path" "/data.db" "--genesis-file"
+      "/config/genesis.json" "--host-addr" "127001" "--port" "3001"
```
rhyslbw:
```suggestion
      "/config/genesis.json" "--host-addr" "127.0.0.1" "--port" "3001"
```

craigem: I clearly need to pay more scrutising to your pastes :rofl: 

Resolved in commit [51b43a7](https://github.com/input-output-hk/cardano-node/pull/525/commits/51b43a73362eec741d4aa25f08ca5fade9b92e42)

nix/docker.nix
craigem: We've removed the script section all together now.

nix/docker.nix
```diff
@@ -0,0 +1,79 @@
...
+      bashInteractive coreutils utillinux iproute iputils curl socat
+    ];
+    # set up /tmp (override with TMPDIR variable)
+    extraCommands = "mkdir -m 0777 tmp";
+  };
+
+in
+  dockerTools.buildImage {
+    name = repoName;
+    ### tag = "${cardano-node.version}";
```
rhyslbw: Remove this line. The base will always have the git commit as the tag

nix/docker.nix
```diff
@@ -0,0 +1,79 @@
...
+    config = {
+      EntryPoint = [ "${cardano-node}/bin/cardano-node" "--config"
+      "/config/config.yaml" "--database-path" "/data.db" "--genesis-file"
+      "/config/genesis.json" "--host-addr" "127.0.0.1" "--port" "3001"
+      "--socket-dir" "/data/ipc/socket" "--topology" "/config/topology.json"];
+      ExposedPorts = {
+        "3001/tcp" = {};  # Cardano node to node port
+      };
+    };
+  }  #### // { inherit (cardano-node) version; }
```
rhyslbw: Remove the commented code

nix/docker.nix
craigem: Resolved in commit 5e4d3c8

nix/docker.nix
craigem: Resolved in commit 5e4d3c8

nix/docker.nix
```diff
@@ -47,8 +47,16 @@ let
   baseImage = dockerTools.buildImage {
     name = "${repoName}-env";
     contents = [
-      glibcLocales iana-etc cacert
-      bashInteractive coreutils utillinux iproute iputils curl socat
+      bashInteractive   # Provide the BASH shell
```
rhyslbw: Nice comments

disassembler:
> Also, to ensure the scope of the change is captured so the merge makes sense, maybe rename this PR to "Add nix dockerTools image, build and push to dockerhub per commit"

We should only push on tagged releases, not every commit.

rhyslbw:
@disassembler pushing every commit is part of the tagging strategy, as it's a requirement to target any commit for the purpose of pinning of dependencies in production or working with the services outside of the released versions.

nix/docker.nix
```diff
@@ -0,0 +1,86 @@
...
+    extraCommands = ''
+      mkdir -p data config
+    '';
+    config = {
+      EntryPoint = [ "${cardano-node}/bin/cardano-node" "--config"
+      "/config/config.yaml" "--database-path" "/data.db" "--genesis-file"
+      "/config/genesis.json" "--host-addr" "127.0.0.1" "--port" "3001"
+      "--socket-dir" "/data/ipc/socket" "--topology" "/config/topology.json"];
+      ExposedPorts = {
+        "3001/tcp" = {};  # Cardano node to node port
```
rhyslbw:
```suggestion
        "3001/tcp" = {};  # p2p
```

.buildkite/docker-build-push.nix
```diff
@@ -0,0 +1,94 @@
...
+
+    ${if dockerHubRepoName == null then ''
+    reponame=cardano-node
+    username="$(docker info | sed '/Username:/!d;s/.* //')"
+    fullrepo="$username/$reponame"
+    '' else ''
+    fullrepo="${dockerHubRepoName}"
+    ''}
+
+  '' + concatMapStringsSep "\n" (image: ''
```
rhyslbw: For reference, here is the conceptual flow for building with the commit hash, then applying conditional tags to allow commit, branch, or version pinning. 

https://github.com/input-output-hk/cardano-graphql/blob/master/Jenkinsfile#L62

nix/docker.nix
```diff
@@ -0,0 +1,86 @@
...
+, curl
+, glibcLocales
+, iana-etc
+, iproute
+, iputils
+, socat
+, utillinux
+
+# Used to generate the docker image names
+, repoName ? "inputoutput/cardano-node"
```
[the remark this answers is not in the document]
craigem: Agreed and committed.

nix/docker.nix
```diff
@@ -0,0 +1,86 @@
...
+#
+############################################################################
+
+{ system ? builtins.currentSystem
+, crossSystem ? null
+, config ? {}
+, profiling ? false
+
+, runtimeShell
+, writeScriptBin
```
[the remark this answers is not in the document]
craigem: Not any more. Nice catch. Removing.

nix/docker.nix
```diff
@@ -0,0 +1,86 @@
...
+      socat             # Utility for bidirectional data transfer
+      utillinux         # System utilities for Linux
+    ];
+    # set up /tmp (override with TMPDIR variable)
+    extraCommands = "mkdir -m 0777 tmp";
+  };
+
+in
+  dockerTools.buildImage {
+    name = repoName;
```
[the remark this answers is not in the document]
craigem: Done.

nix/docker.nix
craigem: How about "Cardano node peer to peer"?

craigem: Corrected.

nix/docker.nix
```diff
@@ -0,0 +1,86 @@
...
+, coreutils
+, curl
+, glibcLocales
+, iana-etc
+, iproute
+, iputils
+, socat
+, utillinux
+
+# Used to generate the docker image names
```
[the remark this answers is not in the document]
craigem: Removed.

disassembler:
If we do think this is fine for a first pass, maybe it makes sense to use forEnvironments to create docker config directories with nix for every environment that the user can map to a directory on the container and execute cardano-node against that if the directory exists.

Something like nix-build -A dockerConfigs.mainnet
docker run -v ./result:/config cardnao-node

Which would run:

rhyslbw:
> I leave it up to @rhyslbw if he thinks this is a decent first pass worth merging, but no end user will use this. 

Agree with you @disassembler, it's intended to be the base we build from, where now to make the supported network extensions is simply a matter of mounting files into the containers using a standard file structure. @craigem let's present the full offering in this PR.

> I really think it makes more sense to embed the configs for all the environments in the docker image and provide a way to override the default configs with your own.

I disagree because the container is only concerned with only the network it's configured for (See CAD-556). From a consumers perspective this makes sense to me and forms a decent descriptive standard:

```
dockerImage.cardano-node
dockerImage.cardano-node.mainnet
dockerImage.cardano-node.byron-testnet
dockerImage.cardano-node.some-other-future-public-network
```
eg
```
docker load -i $(nix-build -A dockerImage.cardano-node.mainnet --no-out-link)

docker run cardano-node
# or override complete config.yaml
docker run cardano-node -v ../custom-mainnet-config.yaml:config/config.yaml
```

disassembler:
yeah, if we build separate images for each network, I fully agree. I thought one of your requirements was a single image that could support multiple environments, but maybe I'm mistaken.

disassembler:
I made some progress on this. It now builds images for each environment in dockerImages. Although the tagging is only dependent on the gitrev which needs to be resolved. it should probably be env:gitrev || env:master or similar.

.buildkite/docker-build-push.nix
```diff
@@ -42,7 +42,8 @@ with hostPkgs.lib;
 
 let
   images = map impureCreated [
-    nodePackages.dockerImage
+    nodePackages.dockerImage.testnet
```
rhyslbw: This should be pushing the standard image too, not just the configured public networks

nix/docker.nix
```diff
@@ -72,12 +72,11 @@ in
       mkdir -p data config
     '';
     config = {
-      EntryPoint = [ "${cardano-node}/bin/cardano-node" "run" "--config"
-      "/config/config.yaml" "--database-path" "/data.db" "--genesis-file"
-      "/config/genesis.json" "--host-addr" "127.0.0.1" "--port" "3001"
-      "--socket-path" "/data/ipc/socket0" "--topology" "/config/topology.json"];
+      EntryPoint = [ scripts.${env.name}.node ];
```
rhyslbw: The foundation image was purposely free from environmental concerns, which is why the `ENTRYPOINT` was `cardano-node` with configuration options based on the file system conventions. Does this change cover the requirement and produce the same image as the previous version?

craigem:
> ```
>  docker load -i $(nix-build -A dockerImage --no-out-link)
> ```

The `commonLib` is removed wth commit 5ba30ad. However as it's now trying to build multiple images the load fails.

I'm unsure why we've gone backwards on all the other tests. Looking into them now.

nix/docker.nix
disassembler: I've added custom. It's not very user friendly. My preference would have been to have a wrapper that checked if a file existed in /config for topology, config and genesis and used that if it did, but @craigem said that doesn't match the spec you provided to him. So now you either run dockerImage.<env> or dockerImage.custom. env just works. custom you have to pass it everything (probably through a mounted volume). It's not something I'd want end users to use, and maybe we can follow up with a better script wrapper for environments to be able to override parts of an environment in a follow-up.

craigem: I've built and tested `dockerImage.custom` and I believe that as of commit 6a5b233 it works as expected.

.buildkite/docker-build-push.nix
craigem: I believe that as of commit 6a5b233 the standard image is provided now via `dockerImage.custom`.

nix/docker.nix
```diff
@@ -0,0 +1,123 @@
...
+      utillinux         # System utilities for Linux
+    ];
+    # set up /tmp (override with TMPDIR variable)
+    extraCommands = ''
+      mkdir -m 0777 tmp
+    '';
+  };
+  customImage = dockerTools.buildImage {
+    name = "${repoName}";
+    tag = "${gitrev}-base";
```
rhyslbw:
```suggestion
    tag = "${gitrev}";
```
The base image needs no network specifier

.buildkite/docker-build-push.nix
```diff
@@ -0,0 +1,90 @@
...
+# tag from the nix-build (../nix/docker.nix).
+#
+# 1. The repo (default "inputoutput/cardano-node") is changed to match
+#    the logged in Docker user's credentials.
+#
+# 2. The tag (default "VERSION") is changed to reflect the
+#    branch which is being built under this Buildkite pipeline.
+#
+#    - If this is a git tag build (i.e. release) then the docker tag
+#      is left as-is.
```
rhyslbw: I don't think this description is valid for the implementation here. The build is first tagged with the commit hash, and then the build tagged with the possible git tags and the master branch tag. We're also not tagging `latest`

nix/docker.nix
disassembler: that's why I thought custom was a better name. My concern is someone is going to run it with just the gitrev and then complain on social media our code sucks because they didn't read the instructions and that this image is strictly for debugging with local custom files. debug might be a good tag as well.

rhyslbw: It's **not** strictly for debugging! It's simply the base image that wraps the `cardano-node` service, which has similar requirements for the user, so should not come as a suprise. Please consider separating the psychological attachment with nix being the only source of config

.buildkite/docker-build-push.nix
```diff
@@ -0,0 +1,80 @@
...
+#
+# 2. The tag (default "VERSION") is changed to reflect the
+#    branch which is being built under this Buildkite pipeline.
+#
+#    - All commits are tagged with commit hash.
+#    - If the branch is master then master tag is updated
+#    - If there is a tag associated with the commit then that tag is pushed as well
+
+{ nodePackages ?  import ../default.nix {}
+, pkgs ? nodePackages.pkgs
```
jbgi: this line is dead code. (also `pkgs` does not exist in `nodePackages`)

nix/docker.nix
```diff
@@ -0,0 +1,123 @@
...
+, iana-etc
+, iproute
+, iputils
+, socat
+, utillinux
+
+, repoName ? "inputoutput/cardano-node"
+}:
+
+with pkgs;
```
jbgi: it's a bit strange to both require explicit packages arguments using `with pkgs;` which kind-of negate the point of explicit dependencies.
Can we remove `pkgs` from arguments?

PR #537, opened 2020-02-04
Update `TraceForgeEvent` logging details

Jimbo4350:
- Update tracer `TraceAdoptedBlock` to log the block hash and a list of txids in that block. 

Example output: `[cardano.node.Forge:Info:61] [2020-02-04 18:08:26.00 UTC] {"kind":"TraceAdoptedBlock","slot":6,"tx ids":"[txid: txid:6875a2f9]","block hash":"a6d8325832257989"}`

- Update the following tracers to `Error` severity:
 `TraceNoLedgerState`
 `TraceNoLedgerView`
 `TraceBlockFromFuture`
 `TraceDidntAdoptBlock`
 `TraceForgedInvalidBlock`

Relevant issue: #528

### DIFF
--- a/cardano-node/src/Cardano/Tracing/ToObjectOrphans.hs
+++ b/cardano-node/src/Cardano/Tracing/ToObjectOrphans.hs
@@ -37,7 +37,7 @@ import           Ouroboros.Consensus.ChainSyncClient
 import           Ouroboros.Consensus.ChainSyncServer
                    (TraceChainSyncServerEvent(..))
 import           Ouroboros.Consensus.Ledger.Abstract
-import           Ouroboros.Consensus.Mempool.API (GenTx, GenTxId)
+import           Ouroboros.Consensus.Mempool.API (GenTx, GenTxId, HasTxId, TxId, txId)
 import           Ouroboros.Consensus.Node.Tracers (TraceForgeEvent (..))
 import           Ouroboros.Consensus.TxSubmission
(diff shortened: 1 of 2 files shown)


Jimbo4350:
We may want to remove the "txid" string from the `Condense CC.TxId` instance in `ouroboros-network`

cardano-node/src/Cardano/Tracing/ToObjectOrphans.hs
```diff
@@ -842,11 +843,14 @@ instance ToObject (TraceLocalTxSubmissionServerEvent blk) where
   toObject _verb _ =
     mkObject [ "kind" .= String "TraceLocalTxSubmissionServerEvent" ]
 
-instance ProtocolLedgerView blk => ToObject (TraceForgeEvent blk tx) where
-  toObject _verb (TraceAdoptedBlock slotNo _blk _txs) =
+instance (HasTxId tx, ProtocolLedgerView blk, Condense (HeaderHash blk), Show (TxId tx))
+           => ToObject (TraceForgeEvent blk tx) where
+  toObject _verb (TraceAdoptedBlock slotNo blk txs) =
```
CodiePP: we should probably print the list of transaction ids only if _verb == 'MaximumVerbosity'
can potentially be huge!

Jimbo4350: @CodiePP changes made

PR #538, opened 2020-02-04
Mempool event tracing details at default log level

### LINKED ISSUE #527: tx submission tracing details at default log level

So at default logging level:

intricate:
Related: #527

### DIFF
--- a/cardano-node/src/Cardano/Tracing/ToObjectOrphans.hs
+++ b/cardano-node/src/Cardano/Tracing/ToObjectOrphans.hs
@@ -37,7 +37,8 @@ import           Ouroboros.Consensus.ChainSyncClient
 import           Ouroboros.Consensus.ChainSyncServer
                    (TraceChainSyncServerEvent(..))
 import           Ouroboros.Consensus.Ledger.Abstract
-import           Ouroboros.Consensus.Mempool.API (GenTx, GenTxId, HasTxId, TxId, txId)
+import           Ouroboros.Consensus.Mempool.API (ApplyTxErr, GenTx, GenTxId,
+                   HasTxId, TraceEventMempool (..), TxId, txId)
 import           Ouroboros.Consensus.Node.Tracers (TraceForgeEvent (..))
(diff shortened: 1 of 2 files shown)


cardano-node/src/Cardano/Tracing/ToObjectOrphans.hs
```diff
@@ -842,6 +851,34 @@ instance ToObject (TraceLocalTxSubmissionServerEvent blk) where
   toObject _verb _ =
     mkObject [ "kind" .= String "TraceLocalTxSubmissionServerEvent" ]
 
+instance (Show (GenTx blk), Show (GenTxId blk), Show (ApplyTxErr blk))
+      => ToObject (TraceEventMempool blk) where
+  toObject _verb (TraceMempoolAddTxs txs mpSz) =
+    mkObject
+        [ "kind" .= String "TraceMempoolAddTxs"
+        , "txsAdded" .= String (pack $ show $ txs)
+        , "mempoolSize" .= String (pack $ show $ mpSz)
```
intricate: So using `show` here kind of sucks. Preferably, we'd want objects like `GenTx`s, `ApplyTxErr`s, and `MempoolSize`s to all have `ToJSON`/`ToObject` instances for structured logging purposes.

@dcoutts @Jimbo4350: Should the instances for the aforementioned data types be defined in `ouroboros-consensus` or here in `cardano-node`? Immediately, I'm thinking to define orphans here in this module since logging output is more relevant to `cardano-node`.

intricate: Using `show` here kind of sucks, so I'm going to introduce some `ToObject` instances for the required types.

intricate: Using `show` here probably isn't so great. 

@CodiePP: Should I introduce some additional `ToObject` instances for the required types?

CodiePP: :+1: I think this is OK

PR #540, opened 2020-02-05
Create benchmarking command group & simplify  `cardano-cli` commands

Jimbo4350:
- Remove `--genesis-hash` from all `cardano-cli` commands as this can be computed from the genesis files
- Create a benchmarking command group for benchmarking related commands
- Remove `--protocol` from `generate-txs` as this can be read from the supplied configuration `.yaml` file

### DIFF
--- a/cardano-node/src/Cardano/CLI/Parsers.hs
+++ b/cardano-node/src/Cardano/CLI/Parsers.hs
@@ -2,6 +2,7 @@
 
 module Cardano.CLI.Parsers
   ( command'
+  , parseBenchmarkingCommands
   , parseDelegationRelatedValues
   , parseGenesisParameters
   , parseGenesisRelatedValues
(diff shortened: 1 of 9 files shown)


cardano-node/src/Cardano/CLI/Parsers.hs
```diff
@@ -421,53 +466,12 @@ parseTxRelatedValues =
         $ SpendUTxO
             <$> parseProtocol
             <*> (GenesisFile <$> parseGenesisPath)
-            <*> parseGenesisHash
```
CodiePP: I would prefer if we keep this argument as otherwise we need to adapt the configuration file if genesis is recreated.

cardano-node/src/Cardano/CLI/Run.hs
```diff
@@ -362,3 +365,8 @@ ensureNewFile writer outFile blob = do
 
 ensureNewFileLBS :: FilePath -> LB.ByteString -> IO ()
 ensureNewFileLBS = ensureNewFile LB.writeFile
+
+getGenesisHash :: GenesisFile -> ExceptT CliError IO Text
+getGenesisHash genFile = do
+  (_, Genesis.GenesisHash gHash) <- readGenesis genFile
+  return $ F.sformat Crypto.hashHexF gHash
```
CodiePP: is this computing the genesis' hash?
then, you can ignore my above comment about keeping the argument 'genesis-hash'

Jimbo4350: @CodiePP  Yes this is computing the genesis hash via https://github.com/input-output-hk/cardano-ledger/blob/a773ba71eb9a6fb91dda2cdf6c41b2280f2acf7d/cardano-ledger/src/Cardano/Chain/Genesis/Data.hs#L116

cardano-node/src/Cardano/CLI/Parsers.hs
Jimbo4350: See: https://github.com/input-output-hk/cardano-node/pull/540/files#r375498783

PR #548, opened 2020-02-06
Create `Benchmarking` folder for relevant modules

Jimbo4350:


cardano-node/cardano-node.cabal
```diff
@@ -22,29 +22,29 @@ library
 
   hs-source-dirs:      src
 
-  exposed-modules:     Cardano.Chairman
+  exposed-modules:     Cardano.Benchmarking.Tx.Generation
+                       Cardano.Benchmarking.Tx.NodeToNode
+                       Cardano.Benchmarking.Tx.TxSubmission
```
erikd: Why are benchmarking modules part of the library itself?

Jimbo4350: @erikd should I create a benchmark stanza in the cabal file?

erikd: Yeah, I think that would make sense.

Jimbo4350: Because these modules are used in `cardano-cli` as pointed out by @CodiePP I think it would make sense to move these to the `cardano-cli` executable and just move all remaining related commands under the benchmarking command group in the `cardano-cli`

CodiePP:
@denisshevchenko you should have a look at this. It will touch the modules you use in the tx gen.

PR #555, opened 2020-02-07
add buildkite job to fetch eval errors from hydra API

disassembler:
Instead of running hydra-eval-jobs in buildkite which constantly fails this queries hydra API for eval errors.

How it works:
1. wait for jobset to be created for PR (up to an hour)
2. query the jobset for evals
3. wait for jobset to contain an eval with the correct git rev (up to an hour)
4. query jobset for any eval errors and print them out if any exist

Known issues:

every branch without a PR will fail so first commit in a PR will fail as buildkite doesn't know the PR number. This is not a huge deal as long as it's communicated properly as pushing another commit, ammending current commit or merging/trying with bors will all run it correctly.


disassembler:
bors try

disassembler:
bors try

PR #573, opened 2020-02-17
Updated ouroboros-network dependencies

coot:


cardano-node/src/Cardano/Chairman.hs
```diff
@@ -89,7 +95,7 @@ runChairman :: forall blk.
 runChairman ptcl nids securityParam maxBlockNo socketDir tracer = do
 
     (chainsVar :: ChainsVar IO blk) <- newTVarM
-      (Map.fromList $ map (\coreNodeId -> (coreNodeId, AF.Empty Block.GenesisPoint)) nids)
+      (Map.fromList $ map (\coreNodeId -> (coreNodeId, AF.Empty undefined)) nids)
```
coot: @mrBliss could you help me filling the `undefined :: Header blk` here.  It is supposed to be the genesis header.

cardano-node/src/Cardano/CLI/Parsers.hs
```diff
@@ -208,8 +210,8 @@ parseGenesisParameters =
     <*> parseFakeAvvmOptions
     <*> parseLovelacePortionWithDefault
           "avvm-balance-factor"
-          "AVVM balances will be multiplied by this factor (defaults to 1)."
-          1
+          "AVVM balances will be multiplied by this factor (defaults to 1e-15)."
+          (toRational ((1 :: Int) % 1e15))
```
coot: @dcoutts changed exported terms in `Cardano.Chain.Common.LovelacePortion`.  This gives us only access to transform `LovelacePortion` to / from `Rational`. @Jimbo4350 is it fine for you to pass a rational on the command line?
It makes easier to pass large values but more difficutl to pass smaller ones; e.g. now `0.1` and `0.000000000000001` instead of `100000000000000` or `1` .

cardano-node/src/Cardano/Chairman.hs
mrBliss: Use `AF.Empty AF.AnchorGenesis`

PR #584, opened 2020-02-18
Add structured logging for TraceSendRecv TxSubmission objects

### LINKED ISSUE #527: tx submission tracing details at default log level

intricate:


cardano-node/src/Cardano/Tracing/ToObjectOrphans.hs
```diff
@@ -822,6 +837,47 @@ instance ToObject (FetchDecision [Point header]) where
...
+        [ "kind" .= String "MsgRequestTxIds"
+        ]
+  toObject _verb (MsgReplyTxIds _) =
+    mkObject
+        [ "kind" .= String "MsgReplyTxIds"
+        ]
+  toObject _verb MsgDone =
+    mkObject
+        [ "kind" .= String "MsgDone"
+        ]
```
intricate: Is this going to be too noisy being logged at info?

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -195,7 +195,7 @@ mkTracers traceOptions tracer = do
     , consensusTracers
         = mkConsensusTracers forgeTracers traceOptions
     , protocolTracers
-        = mkProtocolsTracers traceOptions
+        = mkProtocolTracers traceOptions
```
CodiePP: :+1:

cardano-node/src/Cardano/Tracing/ToObjectOrphans.hs
CodiePP: if these messages are useful to see in production then "Info" is right.

intricate: Since `MsgRequestTxs` and `MsgReplyTxs` are the only useful constructors right now, I added a transformer in `Cardano.Tracing.Tracers` (see `traceMsgReqAndReplyTxs`) to ensure that those are the only ones that are traced.

intricate:
Although this PR is no longer related to #527, it still adds structured logging for `TxSubmission` at the protocol tracer level. So I've instead changed this back to logging at debug severity.

PR #588, opened 2020-02-19

mrBliss:
Issue
------

Pushing through the changes made in https://github.com/input-output-hk/ouroboros-network/pull/1657

checklist
---------

- [ ] This PR results in breaking changes to upstream dependencies.


mrBliss:
@coot There are some breaking changes in the network code, can you fix these?

edsko:
This will need updating after https://github.com/input-output-hk/ouroboros-network/pull/1674 is merged (I don't think that should be too bad).

karknu:
@mrBliss your branch builds and runs fine for me. What was it that you needed help with?

mrBliss:
> @mrBliss your branch builds and runs fine for me. What was it that you needed help with?

I suppose the main node itself runs fine, but the other tools probably wont, as I left a bunch of `undefined`s in there. Can you have a look through at the diff and look for `undefined` and other commented out code?

cardano-node/src/Cardano/CLI/Benchmarking/Tx/NodeToNode.hs
```diff
@@ -204,15 +206,16 @@ benchmarkConnectTxSubmit
   -> TxSubmissionClient (GenTxId blk) (GenTx blk) m ()
   -- ^ the particular txSubmission peer
   -> m ()
-benchmarkConnectTxSubmit trs nc localAddr remoteAddr myTxSubClient = do
+benchmarkConnectTxSubmit trs nc localAddr remoteAddr myTxSubClient = withIOManager $ \iocp -> do
```
coot: I think this is wrong place to call `withIOManager`.  It should be started at the very top level when the application starts.  We don't want two IO manager threads to complete.

cardano-node/src/Cardano/Chairman.hs
```diff
@@ -131,25 +132,25 @@ createConnection
...
-        securityParam
-        maxBlockNo
-        (showTracing tracer)
-        nullTracer
-        nullTracer
-        pInfoConfig)
-      Nothing
-      addr
-    `catch` handleMuxError tracer chainsVar socketPath
+  socketPath = withIOManager $ \iocp -> do
```
coot: The same here, `withIOManager` must be called when chairman starts and then `iocp :: AssociateWithIOCP` should be passed down.

cardano-node/src/Cardano/Node/Submission.hs
```diff
@@ -106,15 +107,15 @@ submitTx :: ( RunNode blk
          -> GenTx blk
          -> Tracer IO TraceLowLevelSubmit
          -> IO ()
-submitTx targetSocketFp protoInfoConfig tx tracer = do
-    targetSocketFp' <- localSocketAddrInfo targetSocketFp
+submitTx targetSocketFp protoInfoConfig tx tracer = NodeToClient.withIOManager $ \iocp -> do
```
coot: The same applies here.

cardano-node/src/Cardano/Wallet/Client.hs
```diff
@@ -58,16 +59,17 @@ runWalletClient :: forall blk.
                 -> SocketPath
                 -> Tracer IO String
                 -> IO ()
-runWalletClient ptcl sockFp tracer = do
+runWalletClient ptcl sockFp tracer = withIOManager $ \iocp -> do
```
coot: This is ok, `runWalletClient` is where the `wallet-client` starts.

cardano-node/src/Cardano/CLI/Run.hs
```diff
@@ -53,6 +57,10 @@ import           Cardano.Crypto (ProtocolMagicId, RequiresNetworkMagic(..))
 import qualified Cardano.Crypto.Hashing as Crypto
 import qualified Cardano.Crypto.Signing as Crypto
 
+import           Ouroboros.Network.NodeToClient ( AssociateWithIOCP
+                                                , withIOManager
+                                                )
```
mrBliss: @Jimbo4350 What does the style guide say about this? Does the stylish-haskell config in this repo do the right thing?

Jimbo4350: I don't think it does however I need to update the style guide re: imports. We want this format if imports exceed 100 chars:

```
import           Ouroboros.Network.NodeToClient 
                       (AssociateWithIOCP,  withIOManager)
```

cardano-node/src/Cardano/CLI/Run.hs
```diff
@@ -190,57 +197,57 @@ data ClientCommand
     (Maybe ExplorerAPIEnpoint)
     [SigningKeyFile]
    deriving Show
-runCommand :: ClientCommand -> ExceptT CliError IO ()
```
Jimbo4350: Is there a better way to introduce `AssociateWithIOCP`? Only one command uses this parameter and I've spent a lot of time removing parameters like this.

edsko: /cc @karknu , @coot

cardano-node/src/Cardano/CLI/Tx.hs
```diff
@@ -277,8 +280,8 @@ nodeSubmitTx
       \p@Consensus.ProtocolRealPBFT{} -> liftIO $ do
```
Jimbo4350: So for instance here this could be:
```
 \p@Consensus.ProtocolRealPBFT{} -> liftIO $ withIOManager $ \iocp -> do
          -- TODO: Update submitGenTx to use `ExceptT`
          traceWith stdoutTracer ("TxId: " ++ condense (Consensus.txId gentx))
          submitTx iocp targetSocketFp
                   (pInfoConfig (Consensus.protocolInfo p))
                   gentx
                   nullTracer -- stdoutTracer
```

cardano-node/src/Cardano/CLI/Run.hs
Jimbo4350: See: https://github.com/input-output-hk/cardano-node/pull/588/files#r383898945

mrBliss:
I'm going to push another bump of `ouroboros-network` to this PR. It's already out of date, as it has been open for a week :slightly_frowning_face:

PR #594, opened 2020-02-24
CAD-328 context name in Trace; adaptations to upstream changes

CodiePP:
Issue
-----------

- adaptations to upstream changes in iohk-monitoring fw; context name is kept in 'Trace'

Checklist
---------

### DIFF
--- a/cardano-node/src/Cardano/CLI/Benchmarking/Tx/NodeToNode.hs
+++ b/cardano-node/src/Cardano/CLI/Benchmarking/Tx/NodeToNode.hs
@@ -177,11 +177,8 @@ instance Transformable Text IO (SendRecvTxSubmission ByronBlock) where
             -- Add a timestamp in 'ToObject'-representation.
             HM.insert "time" (String (T.pack . show $ currentTime)) obj
       tracer = if obj == emptyObject then nullTracer else tr
-    traceWith tracer =<<
-      LogObject
-        <$> pure mempty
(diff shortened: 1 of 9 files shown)


cardano-node/src/Cardano/CLI/Benchmarking/Tx/NodeToNode.hs
```diff
@@ -177,11 +177,8 @@ instance Transformable Text IO (SendRecvTxSubmission ByronBlock) where
             -- Add a timestamp in 'ToObject'-representation.
             HM.insert "time" (String (T.pack . show $ currentTime)) obj
       tracer = if obj == emptyObject then nullTracer else tr
-    traceWith tracer =<<
-      LogObject
-        <$> pure mempty
-        <*> mkLOMeta (defineSeverity arg) (definePrivacyAnnotation arg)
-        <*> pure (LogStructured updatedObj)
+    meta <- mkLOMeta (defineSeverity arg) (definePrivacyAnnotation arg)
+    traceWith tracer (mempty, LogObject mempty meta (LogStructured updatedObj))
```
ksaric: This pattern seems to be pretty common. As I mentioned, it seems to be missing the newtype.
The second thing it might be missing is some constructor function that provides this "out-of-the-box".

Like `createNoContextLogObject :: MetaInfo -> LogStructured -> LogObject`.

CodiePP: good point. yes, this can be further simplified as @deepfire already showed here.

PR #599, opened 2020-02-25
Add build instructions for x86_64 Linux platform

erikd:


README.md
```diff
@@ -11,3 +11,82 @@ repositories.
...
+git clone https://github.com/input-output-hk/cardano-node/
+cd cardano-node
+```
+
+#### Initiate `cardano-node` build with `stack`
+Use [Haskell Stack](https://docs.haskellstack.org/en/stable/README/) to build required `cardano-node` binaries, there are 4 executable targets defined for `cardano-node`, however only `cardano-cli` and `cardano-node` are required to connect to the Cardano Blockchain network. In open terminal window, run:
+
+```bash
+stack build \
+      --flag cardano-node:-with_wallet-client \
```
mark-stopka: @dcoutts requested change of a flag name to `with_wallet_client` in #597 which has been done, can you please make the two-liner change (also line 87)?

mark-stopka:
CentOS / RHEL / Fedora {7,8} are subject to a build difficulties related to missing non-versioned object file [libtinfo.so](https://bugs.centos.org/view.php?id=17085), which is required by [`vty-5.25.1`](https://hackage.haskell.org/package/vty). Currently symlink needs to be created manually.

The essential issue is inherited from [`terminfo`](https://hackage.haskell.org/package/terminfo), which is a dependency of both `vty-5.25.1` and [`ghc`](https://hackage.haskell.org/package/ghc).

This issue is discussed in [jtdaugherty/vty#65](https://github.com/jtdaugherty/vty/issues/65). I believe it's connected with the issue of using binary distribution of `ghc`. Ideal solution would be to modify `stack.yaml` to download binary `ghc` and then re-build `ghc` from source including all it's dependencies, but the time trade-off is probably significant.

README.md
erikd: @mark-stopka Sorry, where does that change go? Ad diff posted here would be very useful.

mark-stopka:
@erikd, can you give me rights to push to your branch, so I can make the appropriate adjustments directly there now that #597 has been merged into master? Or should I modify our repo and you pull from there? BTW it seems @dcoutts knows a way to merge directly from our repo =). Check the reference PR for instructions to your lovely `bors` bot...

erikd:
Sorry, I missed you response on this.

Allowing you to push directly to this repo is definitely the hard way to do this. It would be much eacier if you could either post a diff or an updated version in your repo.

erikd:
Superceded by https://github.com/input-output-hk/cardano-node/pull/599 .

erikd:
I am so confused.  Superceded by #640 .

PR #610, opened 2020-02-28
Remove cruft

erikd:


CodiePP:
please add a description to this PR explaining what you are aiming at.
removing these files will render the config editor non-functional, I guess.

PR #614, opened 2020-02-28
nix shell:  eliminate the need to build ouroboros-{consensus,network}

deepfire:
The status quo of the `nix-shell` is that one needs to remove the three lines this PR removes, to avoid re-building the packages they mention.

The underlying cause is that `haskell.nix`'s version of `shellFor` does not supply the packages mentioned on the list, even if they are dependencies of _other_ such packages.

Checklist
---------


deepfire:
@rvl, thanks for the idea!

PR #615, opened 2020-02-28
Remove `stack build` CI job

Jimbo4350:
Issue
-----------

- No linked issue. We are no longer building `cardano-node` in CI with `stack`.

Checklist
---------


Jimbo4350:
> Having no tests is a bit of a red flag. but removing coveralls for something without tests makes sense.

The stack rebuild job is error prone and I've needed to ask devops to wipe the `/stack-work` dir on a couple of occasions. It would be good to migrate to https://hackage.haskell.org/package/hpc-coveralls so that we don't have to use stack at all to generate code coverage.

> Although, if it takes basically zero time in CO, leaving it in so that its there when tests are added may make sense.

I've put back in the test suite with the aim to implement tests as recommended by you: https://github.com/erikd/jenga/blob/master/jenga.cabal#L107

PR #627, opened 2020-03-04
Update dependencies

erikd:
We should not merge this until I know that `db-sync` works with this version.


erikd:
Closing this as out-of-date.

PR #644, opened 2020-03-06

ksaric:


cardano-node/src/Cardano/CLI/Delegation.hs
```diff
@@ -58,15 +61,15 @@ checkByronGenesisDelegation
...
   -> Crypto.VerificationKey
   -> Crypto.VerificationKey
-  -> IO ()
+  -> ExceptT CliError IO ()
 checkByronGenesisDelegation (CertificateFile certF) magic issuer delegate = do
-  ecert <- canonicalDecodePretty <$> LB.readFile certF
+  ecert <- liftIO $ canonicalDecodePretty <$> LB.readFile certF
   case ecert of
-    Left e -> throwIO $ DlgCertificateDeserialisationFailed certF e
+    Left e -> left $ DlgCertificateDeserialisationFailed certF e
```
mrBliss: I prefer `throwError` to `left` :slightly_smiling_face:

cardano-node/src/Cardano/CLI/Delegation.hs
```diff
@@ -14,6 +14,9 @@ module Cardano.CLI.Delegation
 where
 
 import           Cardano.Prelude hiding (option, show, trace)
+
+import           Control.Monad.Trans.Except.Extra (left)
```
Jimbo4350: `import           Control.Monad.Trans.Except.Extra (left)` should be moved as follows:
```
import           Control.Monad.Trans.Except.Extra (left)
import qualified Data.ByteString.Lazy as LB
```

cardano-node/src/Cardano/CLI/Delegation.hs
erikd: I prefer `left` to `throwError`. The former suggests a connection to the `Left` of an `Either` whereas the later uses "throw' which (incorrectly) suggests a connection to exceptions. `

cardano-node/src/Cardano/CLI/Tx.hs
```diff
@@ -80,11 +80,11 @@ prettyAddress addr = sformat
   (Common.addressF %"\n"%Common.addressDetailedF)
   addr addr
 
-readByronTx :: TxFile -> IO (GenTx ByronBlock)
+readByronTx :: TxFile -> ExceptT CliError IO (GenTx ByronBlock)
 readByronTx (TxFile fp) = do
-  txBS <- LB.readFile fp
+  txBS <- liftIO $ LB.readFile fp
   case fromCborTxAux txBS of
```
erikd: Maybe [hoistEither](http://hackage.haskell.org/package/transformers-except-0.1.1/docs/Control-Monad-Trans-Except-Extra.html) can make that even nicer.

PR #657, opened 2020-03-10

ksaric:
Fixing the `FileNotFound` and propagating it properly up to the top level using `ExceptT`.

### DIFF
--- a/cardano-config/src/Cardano/Config/Logging.hs
+++ b/cardano-config/src/Cardano/Config/Logging.hs
@@ -26,11 +26,10 @@ module Cardano.Config.Logging
 
 import           Cardano.Prelude hiding (trace)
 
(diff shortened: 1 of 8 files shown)


cardano-config/src/Cardano/Config/Logging.hs
```diff
@@ -26,8 +26,10 @@ module Cardano.Config.Logging
 
 import           Cardano.Prelude hiding (trace)
 
+import           Control.Monad.Trans.Except.Extra (catchIOExceptT)
```
Jimbo4350: Should be:
```
import           Control.Exception (IOException)
import           Control.Exception.Safe (MonadCatch)
import           Control.Monad.Trans.Except.Extra (catchIOExceptT)
```

PR #662, opened 2020-03-11
[CAD-709] TUI not switching to peer list when P is hit.

ksaric:
https://jira.iohk.io/browse/CAD-709

### DIFF
--- a/cardano-node/src/Cardano/Node/TUI/LiveView.hs
+++ b/cardano-node/src/Cardano/Node/TUI/LiveView.hs
@@ -533,7 +533,7 @@ initLiveViewState = do
                 , lvsNetworkUsageOutLast    = 0
                 , lvsNetworkUsageOutNs      = 10000
                 , lvsMempoolCapacity        = mempoolCapacity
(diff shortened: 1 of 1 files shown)


cardano-node/src/Cardano/Node/TUI/LiveView.hs
```diff
@@ -533,7 +533,7 @@ initLiveViewState = do
                 , lvsNetworkUsageOutLast    = 0
                 , lvsNetworkUsageOutNs      = 10000
                 , lvsMempoolCapacity        = mempoolCapacity
-                , lvsMempoolCapacityBytes   = mempoolCapacity * maxBytesPerTx
+                , lvsMempoolCapacityBytes   = mempoolCapacity * maxBytesPerTx -- / 1024
```
deepfire: @ksaric, do we need this comment here?

ksaric: It's going to be fixed in the next PR, probably should have added `TODO(ks): ` before that.

PR #668, opened 2020-03-11
CAD-508: update traces for Tx throughput benchmarking

deepfire:
1. trace txid lists in the blockfetch protocol's `AddBlock` message
1. establish order among the `ToObjectOrphans`'s instances -- all now sorted alphabetically, keyed by the unqualified name of their outermost type.
1. a dependency for https://github.com/input-output-hk/cardano-benchmarking/pull/2

Checklist
---------

### DIFF
--- a/cardano-config/src/Cardano/Tracing/ToObjectOrphans.hs
+++ b/cardano-config/src/Cardano/Tracing/ToObjectOrphans.hs
@@ -133,129 +133,6 @@ instance ( Show a
 
   show = showWithTip show
 
--- instances of @DefinePrivacyAnnotation@ and @DefineSeverity@
-instance DefinePrivacyAnnotation (WithIPList (SubscriptionTrace Socket.SockAddr))
-instance DefineSeverity (WithIPList (SubscriptionTrace Socket.SockAddr)) where
-  defineSeverity (WithIPList _ _ ev) = case ev of
-    SubscriptionTraceConnectStart _ -> Info
-    SubscriptionTraceConnectEnd _ connectResult -> case connectResult of
-      ConnectSuccess -> Info
-      ConnectSuccessLast -> Notice
-      ConnectValencyExceeded -> Warning
-    SubscriptionTraceConnectException {} -> Error
-    SubscriptionTraceSocketAllocationException {} -> Error
-    SubscriptionTraceTryConnectToPeer {} -> Info
-    SubscriptionTraceSkippingPeer {} -> Info
-    SubscriptionTraceSubscriptionRunning -> Debug
-    SubscriptionTraceSubscriptionWaiting {} -> Debug
-    SubscriptionTraceSubscriptionFailed -> Error
-    SubscriptionTraceSubscriptionWaitingNewConnection {} -> Notice
-    SubscriptionTraceStart {} -> Debug
-    SubscriptionTraceRestart {} -> Info
-    SubscriptionTraceConnectionExist {} -> Notice
-    SubscriptionTraceUnsupportedRemoteAddr {} -> Error
-    SubscriptionTraceMissingLocalAddress -> Warning
-    SubscriptionTraceApplicationException {} -> Error
-    SubscriptionTraceAllocateSocket {} -> Debug
-    SubscriptionTraceCloseSocket {} -> Info
-
-instance DefinePrivacyAnnotation (WithDomainName (SubscriptionTrace Socket.SockAddr))
-instance DefineSeverity (WithDomainName (SubscriptionTrace Socket.SockAddr)) where
-  defineSeverity (WithDomainName _ ev) = case ev of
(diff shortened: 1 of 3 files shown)


cardano-config/src/Cardano/Tracing/ToObjectOrphans.hs
```diff
@@ -412,13 +412,16 @@ instance Show peer => Transformable Text IO [TraceLabelPeer peer
   trTransformer _ verb tr = trStructured verb tr
 
 -- transform @BlockFetchDecision@
-instance Show peer => Transformable Text IO (TraceLabelPeer peer
-                                (TraceFetchClientState header)) where
+instance {-# OVERLAPS #-}
```
CodiePP: why does this type overlap with another? and with which? can they be put together in the file?
what caused you to introduce overlapping types?

cardano-config/src/Cardano/Tracing/ToObjectOrphans.hs
```diff
@@ -450,6 +453,58 @@ instance ( Condense (HeaderHash blk)
   -- user defined formatting of log output
   trTransformer _ verb tr = trStructured verb tr
 
+instance DefinePrivacyAnnotation a => DefinePrivacyAnnotation (TraceLabelPeer peer a)
```
CodiePP: OK, because of these generic instances?
so what did you gain? I see 78 additions and only 21 deletions in this file.

shell.nix
```diff
@@ -22,9 +22,6 @@ let
        # in theory we should only have the above two packages (or better, they should be auto-detected),
        # but due to source-repository-package declarations being considered as local packages by cabal, we need the following packages as well.
        # cf. https://github.com/haskell/cabal/issues/6249 and https://github.com/haskell/cabal/issues/5444
-       ps.cardano-sl-x509
```
CodiePP: these changes are not related to this PR.

cardano-config/src/Cardano/Tracing/ToObjectOrphans.hs
```diff
@@ -450,6 +453,58 @@ instance ( Condense (HeaderHash blk)
...
+         , HasHeader blk
+         , HasTxs blk
+         , HasTxId (GenTx blk)
+         )
+ => ToObject (AnyMessage (BlockFetch blk)) where
+  toObject _verb (AnyMessage (MsgBlock blk)) =
+    mkObject
+      [ "kind" .= String "MsgBlock"
+      , "blkid" .= String (pack . condense $ blockHash blk)
+      , "txids" .= Array (V.fromList . fmap presentTx $ extractTxs blk)
```
CodiePP: the newly introduced dependency "vector" is used only once, exactly here. Is this necessary?
a list of "String"s would also be acceptable. we do not need "Array" here.

shell.nix
deepfire: @CodiePP, I need these changes when I work on the node, and basically no one else cares.

cardano-config/src/Cardano/Tracing/ToObjectOrphans.hs
deepfire: @CodiePP, I'm sorry, I don't understand -- there isn't an Aeson type with a `[]` -- only one with a vector.

cardano-config/src/Cardano/Tracing/ToObjectOrphans.hs
deepfire: I'll try to illustrate the difference using a commit that removes the generic instances.

cardano-config/src/Cardano/Tracing/ToObjectOrphans.hs
deepfire: That's a bit odd.. I just dropped the overlaps and all is well now.  I clearly remember needing them.

Gone, anyway.

cardano-config/src/Cardano/Tracing/ToObjectOrphans.hs
deepfire: The overlaps are gone.

cardano-config/src/Cardano/Tracing/ToObjectOrphans.hs
deepfire: There isn't an Aeson Value ADT variant with a [] -- only one with a vector -- but we can reuse `toJSON` to avoid the dependency

Good point.

shell.nix
deepfire: Dropped, as I finally got https://github.com/input-output-hk/cardano-node/pull/614 reviewed (was hanging since Feb 28).

deepfire:
@CodiePP, I think I've already done what we discussed, so to prevent further confusion I went to the town and sorted _all_ instances by the unqualified name of the outermost type.

PR #669, opened 2020-03-12
Add Windows build and tests to required hydra job.

jbgi:


nix/nixos/cardano-node-service.nix
```diff
@@ -266,10 +275,19 @@ in {
...
-      message =
-        "The option services.cardano-node.stateDir should have ${stateDirBase} as a prefix!";
-    }];
+    assertions = [
+      {
+        assertion = lib.hasPrefix stateDirBase cfg.stateDir;
+        message = "The option services.cardano-node.stateDir should have ${stateDirBase} as a prefix!";
+      }
+      {
+        assertion = (cfg.genesisHash == null) != (cfg.genesisHashPath == null);
```
jbgi: That assertion (XOR) also cover the case of the next one. So the next assertion can be removed (and the message of this one complexified) or change this assertion to not cover the case of next one:
```suggestion
        assertion = !((cfg.genesisHash != null) && (cfg.genesisHashPath != null));
```

PR #674, opened 2020-03-12

ksaric:
Last `throwIO` I found.

### DIFF
--- a/cardano-node/src/Cardano/Common/LocalSocket.hs
+++ b/cardano-node/src/Cardano/Common/LocalSocket.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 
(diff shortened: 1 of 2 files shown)


cardano-node/src/Cardano/Common/LocalSocket.hs
```diff
@@ -8,14 +9,26 @@ module Cardano.Common.LocalSocket
   )
 where
 
+import           Prelude (show)
 import           Cardano.Prelude
 
+import           Control.Monad.Trans.Except.Extra (firstExceptT, hoistEither)
```
Jimbo4350: Should be:
```
import           Control.Monad.Trans.Except.Extra (firstExceptT, hoistEither)
import           System.Directory (createDirectoryIfMissing, removeFile)
```

erikd:
@ksaric This need to be rebase against master and force pushed.

Issue #613, opened 2020-02-28
`cardano-node` should not require genesis file and genesis hash arguments


Jimbo4350:
After #611 is completed, the genesis file can be obtained from the config yaml file and therefore the genesis hash can be derived from this.

rhyslbw:
#611 is closed. Does #648 now unblock this?

Jimbo4350:
It wasn't blocked, however I will wait until the end of this sprint before starting on this issue.

PR #683, opened 2020-03-16
Remove genesis-hash and genesis-file arguments from cardano-node

Jimbo4350:
Issue
-----------

- #613 

Checklist
---------

### DIFF
--- a/benchmarking/chain-sync/benchmark-chain-sync-mainnet.sh
+++ b/benchmarking/chain-sync/benchmark-chain-sync-mainnet.sh
@@ -36,8 +36,6 @@ fi
 
 exec ${NODE} \
   run \
-  --genesis-file ${BASEDIR}/../../configuration/mainnet-genesis.json \
-  --genesis-hash "5f20df933584822601f9e3f8c024eb5eb252fe8cefb24d1317dc3d432e940ebb" \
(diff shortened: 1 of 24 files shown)


erikd:
I assume that these can still be overridden on the command line.

rhyslbw:
@erikd 
> I assume that these can still be overridden on the command line. If so
```
Invalid option `--genesis-file'

Usage: cardano-node run --topology FILEPATH --database-path FILEPATH
                        [--delegation-certificate FILEPATH]
                        [--signing-key FILEPATH] [--socket-path FILEPATH]
                        [--host-addr HOST-NAME] --port PORT
                        --config NODE-CONFIGURATION [--validate-db]
```

Jimbo4350:
@erikd @rhyslbw It was my understanding that we wanted to reduce the number of arguments needed for `cardano-node`. DevOps wanted still wanted the  ability to specify socket-path from the command line (which remains but defaults to what is in the configuration yaml file if nothing is specified). It doesn't make sense to specify genesis-hash from the command line because we can derive the genesis hash from the genesis file. 

So the question is, am I completely removing the ability to specify the genesis file from the command line (i.e it will be defined in the configuration yaml file) or am I leaving the ability to specify the genesis file on the command line (i.e if is it specified, it will override whatever is in the configuration yaml file)?

PR #699, opened 2020-03-20
[SRE-71] Hydra build artefacts for nodeConfig and topologyFile. 

craigem:
As per the title, add hydra build artifacts for nodeConfig and topologyFile.

### DIFF
--- a/nix/scripts.nix
+++ b/nix/scripts.nix
@@ -90,15 +90,21 @@ let
         pkgsModule
       ];
     }).config.services.cardano-node.script;
-  in pkgs.writeScript "cardano-node-${envConfig.name}" ''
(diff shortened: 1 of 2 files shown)


nix/scripts.nix
```diff
@@ -77,12 +77,15 @@ let
...
+      #!${pkgs.runtimeShell}
+      set -euo pipefail
+      mkdir -p "${config.stateDir}"
+      ${nodeScript} $@
+    '';
+    inherit topologyFile;
+  };
   scripts = forEnvironments (environment:
   {
     node = mkNodeScript environment;
```
manveru: You need to point to `.node` here, or the old behavior won't work anymore.

PR #744, opened 2020-03-31
Use approved codecs from Ouroboros.Consensus.NodeNetwork

MarcFontaine:
Issue
-----------

-

Checklist
---------

### DIFF
--- a/cardano-node/src/Cardano/Node/Submission.hs
+++ b/cardano-node/src/Cardano/Node/Submission.hs
@@ -28,9 +28,12 @@ import           Control.Monad.Class.MonadTimer (MonadTimer)
 import           Control.Tracer (Tracer, nullTracer, traceWith)
 
 import           Ouroboros.Consensus.Mempool (ApplyTxErr, GenTx)
+import           Ouroboros.Consensus.NodeNetwork (ProtocolCodecs (..), protocolCodecs)
+import           Ouroboros.Consensus.Node.NetworkProtocolVersion (mostRecentNetworkProtocolVersion)
 import           Ouroboros.Consensus.Node.Run (RunNode)
 import qualified Ouroboros.Consensus.Node.Run as Node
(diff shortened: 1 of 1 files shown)


cardano-node/src/Cardano/Node/Submission.hs
```diff
@@ -154,14 +158,16 @@ localInitiatorNetworkApplication tracer cfg tx =
...
+                          (pcLocalTxSubmissionCodec codecs)
                           channel
                           (LocalTxSub.localTxSubmissionClientPeer
                              (txSubmissionClientSingle tx))
               case result of
                 Nothing  -> traceWith tracer TraceLowLevelAccepted
                 Just msg -> traceWith tracer (TraceLowLevelRejected $ show msg)
       }
+    where
+      codecs = protocolCodecs cfg (mostRecentNetworkProtocolVersion (Proxy @ blk))
```
mrBliss: Is this correct?
https://github.com/input-output-hk/ouroboros-network/issues/1633 and https://github.com/input-output-hk/ouroboros-network/pull/1522 are not yet closed/merge. In other words: consensus supports newer protocol versions (to support the LocalStateQuery protocol and to include the block size in the headers sent across the ChainSync protocol), but I'm not sure that network has enabled them. So using the most recent protocol version might not be correct.

However, I think you happen to be safe because the two local protocols you're using are the same for all current protocol versions.

But if we add another version that *does affect* one of those two protocols, then this will *immediately* use that, which might not be what we want.

MarcFontaine: Thx for the references to the related issues.
At the moment I'm just looking at the places that use codes with are not defined in Ouroboros.Consensus.NodeNetwork and try to find out if it is possible to export all codecs from a single API.
Yes this will need carefully reviewing before it is ready.

Jimbo4350:
@MarcFontaine can I close this PR?

MarcFontaine:
Yes it is outdated

PR #757, opened 2020-04-03
CAD-790: Tracers for node version.

denisshevchenko:
Issue
-----------

- New tracers for node's release, version and commit. Currently, I'm creating the new service for a real-time view of metrics from the node. Since my service is a separate process, it receives these metrics via socket, using `TraceForwarder`/`TraceAcceptor` mechanism. So I need new tracers that will send `LogObject`s with node's release (for example, "Shelley"), version (for example, "1.8.0") and Git commit. It will allow me to show this information in my service.

- Please not that currently it doesn't work: I don't see version info in the log!

Checklist
---------

### DIFF
--- a/cardano-node/src/Cardano/Node/Run.hs
+++ b/cardano-node/src/Cardano/Node/Run.hs
@@ -35,25 +35,30 @@ import qualified Data.List as List
 import           Data.Proxy (Proxy (..))
 import           Data.Semigroup ((<>))
 import           Data.Text (Text, breakOn, pack, take)
+import           Data.Version (showVersion)
 import           Network.HostName (getHostName)
 import           Network.Socket (AddrInfo)
 import           System.Directory (canonicalizePath, makeAbsolute)
 
 import           Control.Monad.Class.MonadSTM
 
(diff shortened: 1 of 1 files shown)


cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -290,6 +301,17 @@ mkTracers traceOptions tracer = do
...
     tracingVerbosity = traceVerbosity traceOptions
 
+    tracerNodeVersion :: TracingVerbosity -> Trace IO Text -> Tracer IO (WithSeverity Text)
+    tracerNodeVersion _ tr = Tracer $ \_ -> do
+        meta <- mkLOMeta Critical Confidential
+        let tr' = appendName "nodeVersion" tr
+            rTr = appendName "release" tr'
+            vTr = appendName "version" tr'
+            cTr = appendName "commit"  tr'
+        traceNamedObject rTr (meta, LogMessage "Byron")
```
denisshevchenko, line 311: It's hardcoded now, where can I take the current release from?

PR #773, opened 2020-04-08
CAD-791 eliding addtocurrentchain messages

CodiePP:
Issue
-----------

- eliding AddToCurrentChain messages

depends on PR #760 (merged)

Checklist
---------

### DIFF
--- a/cardano-node/src/Cardano/Tracing/Tracers.hs
+++ b/cardano-node/src/Cardano/Tracing/Tracers.hs
@@ -146,18 +146,49 @@ instance ElidingTracer (WithSeverity (ChainDB.TraceEvent blk)) where
   isEquivalent (WithSeverity s1 (ChainDB.TraceGCEvent ev1))
                (WithSeverity s2 (ChainDB.TraceGCEvent ev2)) =
                   s1 == s2 && indexGCType ev1 == indexGCType ev2
-   
+  isEquivalent (WithSeverity _s1 (ChainDB.TraceAddBlockEvent _))
+               (WithSeverity _s2 (ChainDB.TraceAddBlockEvent _)) = True
+  isEquivalent (WithSeverity _s1 (ChainDB.TraceGCEvent _ev1))
+               (WithSeverity _s2 (ChainDB.TraceAddBlockEvent _)) = True
+  isEquivalent (WithSeverity _s1 (ChainDB.TraceAddBlockEvent _))
+               (WithSeverity _s2 (ChainDB.TraceGCEvent _ev2)) = True   
+  isEquivalent (WithSeverity _s1 (ChainDB.TraceGCEvent _ev1))
+               (WithSeverity _s2 (ChainDB.TraceCopyToImmDBEvent _)) = True
+  isEquivalent (WithSeverity _s1 (ChainDB.TraceCopyToImmDBEvent _))
+               (WithSeverity _s2 (ChainDB.TraceGCEvent _ev2)) = True   
+  isEquivalent (WithSeverity _s1 (ChainDB.TraceCopyToImmDBEvent _))
+               (WithSeverity _s2 (ChainDB.TraceAddBlockEvent _)) = True
+  isEquivalent (WithSeverity _s1 (ChainDB.TraceAddBlockEvent _))
+               (WithSeverity _s2 (ChainDB.TraceCopyToImmDBEvent _)) = True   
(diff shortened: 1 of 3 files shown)


CodiePP:
example output:
```
[nixos19:cardano.node.ChainDB:Notice:50] [2020-04-16 21:40:34.68 UTC] Chain extended, new tip: (Point 65998, 028b27d4063e06f7)
[nixos19:cardano.node.ChainDB:Notice:50] [2020-04-16 21:40:35.56 UTC] Chain extended, new tip: (Point 66248, 85e60103fc87234e)
[nixos19:cardano.node.BlockFetchDecision:Info:60] [2020-04-16 21:40:35.80 UTC] [String "ConnectionId {localAddress = 192.168.11.175:46279, remoteAddress = 18.177.103.105:3001}",String "FetchDecision declined",String "FetchDeclineConcurrencyLimit FetchModeBulkSync 1"]
[nixos19:cardano.node.ChainDB:Notice:50] [2020-04-16 21:40:36.41 UTC] Chain extended, new tip: (Point 66498, 06fd59fd496e0f64)
[nixos19:cardano.node.ChainDB:Notice:50] [2020-04-16 21:40:37.27 UTC] Chain extended, new tip: (Point 66748, 4d5c4cec902d64a5)
[nixos19:cardano.node.BlockFetchDecision:Info:60] [2020-04-16 21:40:37.52 UTC] [String "ConnectionId {localAddress = 192.168.11.175:46279, remoteAddress = 18.177.103.105:3001}",String "FetchDecision declined",String "FetchDeclineConcurrencyLimit FetchModeBulkSync 1"]
[nixos19:cardano.node.ChainDB:Notice:50] [2020-04-16 21:40:38.41 UTC] Chain extended, new tip: (Point 66998, 5c680cfff361df10)
[nixos19:cardano.node.ChainDB:Notice:50] [2020-04-16 21:40:39.51 UTC] Chain extended, new tip: (Point 67248, f2541b040fe78376)
```
* new blocks are reported every 250 blocks; can still be parsed for the chain-sync benchmark
* messages in "FetchDecision" tracer are reported every 1000 messages

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -145,18 +145,49 @@ instance ElidingTracer (WithSeverity (ChainDB.TraceEvent blk)) where
   isEquivalent (WithSeverity s1 (ChainDB.TraceGCEvent ev1))
                (WithSeverity s2 (ChainDB.TraceGCEvent ev2)) =
                   s1 == s2 && indexGCType ev1 == indexGCType ev2
-   
+  isEquivalent (WithSeverity _s1 (ChainDB.TraceAddBlockEvent _))
+               (WithSeverity _s2 (ChainDB.TraceAddBlockEvent _)) = True
+  isEquivalent (WithSeverity _s1 (ChainDB.TraceGCEvent _ev1))
+               (WithSeverity _s2 (ChainDB.TraceAddBlockEvent _)) = True
```
mrBliss, line 151: So you're saying that `TraceGCEvent` and `TraceAddBlockEvent` are equivalent so that when a `TraceGCEvent`, which is unrelated to the `TraceAddBlockEvent` in question, doesn't interrupt the elision and is elided as well? I guess that works, but it feels strange/ad hoc/hacky to me.

What about the following: instead of eliding *consecutive* "equivalent" messages (i.e., not interrupted by "un-equivalent" messages), just print those "un-equivalent" messages without interrupting the elision by making the elision a bit cleverer. Instead of looking at the equivalence with the previous message, remember whether an equivalent message was traced in the last x milliseconds or so. You'll have to be careful not to hold on to too much memory. This means that the elision doesn't restart when interrupted and the unrelated "un-equivalent" (that are incorrectly made "un-equivalent" in this PR) still show up in the output.

CodiePP: The logic of the message suppression is that simple now. I hesitate to make it much more complex, e.g. keep more state, and potentially run into problems. We can't compare timestamps as they are not part of the state. Here we only compare based on the traced types and the annotated severity.

deepfire: I think that if we want to use such broad equivalence for elision, then we probably want to make elision optional -- for the cases when we need the full message stream -- for debugging purposes.

What do you think @CodiePP ?

CodiePP: the 'doelide' functions can be commented out to disable eliding.

PR #776, opened 2020-04-09
CAD-815: shutdown API and node exit on ChainDB slot sync limit

deepfire:
1. Factor `Cardano.Node.Shutdown` as an API for graceful shutdown, either externally- or internally-mediated.
1. Provide a restricted run mode, where the node exits after syncing the chain to a specified slot no -- enabled by `--shutdown-on-slot-synced SLOTNO` option.

Checklist
---------


cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -219,8 +229,29 @@ handleSimpleNode p trace nodeTracers npm onKernel = do
...
+            ("will terminate upon reaching " <> (pack $ show maxSlot) :: Text)
+          spawnSlotLimitTerminator registry (Node.getChainDB nodeKernel) maxSlot shutFd
+        _ -> pure ()
+      onKernel nodeKernel
  where
+  spawnSlotLimitTerminator
+    :: ResourceRegistry IO -> ChainDB.ChainDB IO blk -> SlotNo -> Fd -> IO ()
+  spawnSlotLimitTerminator registry chainDB maxSlot (Fd shutFd) =
+    void $ onEachChange registry identity Nothing
+      (getPoint <$> ChainDB.getTipPoint chainDB) $
```
mrBliss, line 246: Use `pointSlot` here so you can pattern match on `WithOrigin SlotNo`

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -219,8 +229,29 @@ handleSimpleNode p trace nodeTracers npm onKernel = do
...
+        \case
+          Origin -> pure ()
+          At pt -> when (blockPointSlot pt >= maxSlot) $ do
+            traceWith (trTransformer TextualRepresentation MaximalVerbosity $
+                       severityNotice trace)
+              ("Cardano.Node.Run.spawnSlotLimitTerminator: reached " <> show (blockPointSlot pt) <> ", ringing the shutdown doorbell" :: Text)
```
mrBliss, line 252: This line is a bit too long

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -219,8 +229,29 @@ handleSimpleNode p trace nodeTracers npm onKernel = do
...
+            IO.hClose =<< IO.fdToHandle shutFd
```
mrBliss, line 253: Can this be encapsulated in a function, e.g., `triggerShutdown`?

cardano-config/src/Cardano/Config/Types.hs
```diff
@@ -87,6 +88,7 @@ data NodeCLI = NodeCLI
   , configFp :: !ConfigYamlFilePath
   , validateDB :: !Bool
   , shutdownIPC :: !(Maybe Fd)
+  , shutdownOnSlotSynced :: !MaxSlotNo
```
mrBliss, line 91: This could also be useful for mock protocols. Is there a common place for shut arguments?

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -326,20 +357,49 @@ handleSimpleNode p trace nodeTracers npm onKernel = do
...
+  -- ^ We shall watch this externally-supplied FD for termination EOF.
+  | InternalShutdown Fd Fd
+  -- ^ Pair of FDs for intra-process graceful shutdown termination messaging.
+  --   First FD is the read side as above,
+  --   second FD is the shutdown doorbell to be closed.
+
+sfdsReadSide :: ShutdownFDs -> Maybe Fd
+sfdsReadSide (ExternalShutdown r) = Just r
+sfdsReadSide (InternalShutdown r _w) = Just r
+sfdsReadSide _ = Nothing
```
mrBliss, line 386: I recently saw a PR about cleanly shutting down the node go by, but I'm not familiar with the details. Can somebody more knowledgeable about it review this part?

deepfire: @mrBliss, that was @dcoutts 's work, and I'm reusing & extending his machinery a bit.

cardano-node/src/Cardano/Node/Run.hs
deepfire: Done.

cardano-node/src/Cardano/Node/Run.hs
deepfire: Done.

cardano-node/src/Cardano/Node/Run.hs
deepfire: Done, although it is unable to reuse the pattern match, so we now have to deal with "escaped" partiality.

cardano-config/src/Cardano/Config/Types.hs
deepfire: Done.

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -326,20 +354,62 @@ handleSimpleNode p trace nodeTracers npm onKernel = do
...
+    pure $ InternalShutdown (Fd r) (Fd w)
+
+data ShutdownFDs
+  = NoShutdownFDs
+  | ExternalShutdown Fd
+  -- ^ We shall watch this externally-supplied FD for termination EOF.
+  | InternalShutdown Fd Fd
+  -- ^ Pair of FDs for intra-process graceful shutdown termination messaging.
+  --   First FD is the read side as above,
+  --   second FD is the shutdown doorbell to be closed.
```
deepfire, line 382: I'm not 100% sure on the name.  Maybe we could call it either `Shutdown` or `ShutdownHandle`.

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -326,20 +354,62 @@ handleSimpleNode p trace nodeTracers npm onKernel = do
...
+
+sfdsReadSide :: ShutdownFDs -> Maybe Fd
+sfdsReadSide = \case
+  ExternalShutdown r -> Just r
+  InternalShutdown r _w -> Just r
+  _ -> Nothing
+
+triggerShutdown :: ShutdownFDs -> Trace IO Text -> Text -> IO ()
+triggerShutdown (InternalShutdown _ (Fd shutFd)) trace reason = do
```
mrBliss, line 391: What if you just pass it the `Fd`, then you wouldn't have the partiality?

deepfire: Well, then we wouldn't be operating at the level of the abstraction.. and we would introduce ability to express an invalid situation -- we could pass an `Fd` that is not associated with the shutdown doorbell.

mrBliss: What if you wrap that `Fd` in a newtype? Or you can make this function a local helper instead of a top-level one.

deepfire: Good point!  I've actually thought of the `newtype` but was a bit afraid of over-engineering -- wasn't sure if this abstraction is worth adding _two_ types..

deepfire: Done.

deepfire: @mrBliss, done -- also newtyped the listener side as well, for symmetry and completeness.  Three types for the shutdown API..

cardano-node/src/Cardano/Node/Run.hs
deepfire: @mrBliss, also starting to think that if we specialise the `InternalShutdown` to serve directly the specific cause of slot-limited termination, then we get rid of partiality.  So that would be the DSL-based approach for shutdown -- for each internal shutdown usecase we add an ADT branch.

Plus side:
 - no partiality
 - we mean exactly what we say by the ADT, so it's harder to misuse/get wrong

Minus side:
 - ADT clauses are exclusive (with the current type design, unless we introduce yet another type wrapper..)
 - we'll have to spell out each internal shutdown context separately

mrBliss:
I defer to @dcoutts on the whole shutdown business. (A thought: maybe that code deserves a separate module?)

deepfire:
Almost completed refactoring & documenting the `Cardano.Node.Shutdown` module.

cc @mrBliss @dcoutts .

cardano-node/src/Cardano/Node/Shutdown.hs
```diff
@@ -0,0 +1,191 @@
...
+-- | We provide an optional cross-platform method to politely request shut down.
+--
+-- For the duration of 'action', we gracefully handle shutdown requests,
+-- external or internal, as requested by configuration in 'NodeProtocolMode',
+-- while allocating corresponding 'ShutdownFDs', and providing them to the 'action'.
+withShutdownHandling
+  :: NodeProtocolMode
+  -> Trace IO Text
+  -> (ShutdownFDs -> IO ())
+  -> IO ()
```
deepfire, line 139: Note how this is specifically not merged with `withShutdownHandler`:

- `withShutdownHandler` remains the mechanism
- `withShutdownHandling` is the policy bit around the mechanism

cardano-node/src/Cardano/Node/Shutdown.hs
```diff
@@ -0,0 +1,191 @@
...
+--   - by internal call of 'triggerShutdown' on 'ShutdownFDs', or
+--   - by parent process
+-- ..or automatically, because the parent process itself terminated,
+-- then we initiate a clean shutdown.
+data ShutdownFDs
+  = NoShutdownFDs
+  | ExternalShutdown !ShutdownListener
+  -- ^ Extra-processually signalled shutdown.
+  | InternalShutdown !ShutdownListener !ShutdownDoorbell
+  -- ^ Intra-processually signalled shutdown.
```
deepfire, line 62: Not sure about the name -- it reflects content, not intent.

Maybe `ShutdownChan` or just `Shutdown` instead?

deepfire:
@mrBliss, seeing how the review takes this long, maybe we can find another way?

deepfire:
bors merge

PR #779, opened 2020-04-09
Isolate node socket in Docker image

rhyslbw:
Issue
-----------
The existing design for cross-container IPC requires mounting a single `data` volume, including the node db. 

This PR isolates the node socket to allow independent volume management so that the `ipc` volume can be accessed by clients, without giving them access to the database.

@disassembler This is WIP to show the intended interface. The image still results in the node creating the socket in `data/node.socket`, but I'm not sure where that is defined.

- This PR **results** in a breaking change to the Docker run interface

Checklist
---------

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -31,7 +31,7 @@ let
     defaultConfig = rec {
       stateDir = "/data";
       dbPrefix = "db";
-      socketPath = stateDir + "/node.socket";
+      socketPath = "/ipc/node.socket";
     };
     customConfig' = defaultConfig // customConfig;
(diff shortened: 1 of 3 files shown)


default.nix
```diff
@@ -31,7 +31,7 @@ let
     defaultConfig = rec {
       stateDir = "/data";
       dbPrefix = "db";
-      socketPath = stateDir + "/node.socket";
+      socketPath = "/ipc/node.socket";
```
rhyslbw, line 34: This is not enough to set the path passed to `cardano-node`

PR #785, opened 2020-04-12
CAD-828 scripts:  fixes; factor; profiling; verbosity; --nix mode (new default)

deepfire:
1. `--nix` mode added to all scripts, and has been made the default. For the end users this will effectively make the scripts work out of the box (if they have Nix installed).

1. A Nix evaluation caching mechanism has been added to avoid costly re-evaluation, in cases where we run a certain executable several times.

1. `--profile` flag has been added;  this is end-to-end integrated with Nix, but in Cabal/Stack mode the user has to provide correspondingly built binaries for the flag to work.

1. All executable invocations now go through `scripts/lib.sh`, which means all of them are subject to the build tool choice and profiling machinery.

1. The use of configuration profiles in `configuration/defaults/*` has been further enshrined in `scripts/lib-node.sh`, which is now the code path that is used for every single `cardano-node` execution.

1. It's no longer needed to run `tmux` manually -- the `shelley-testnet` family of scripts start it on their own.

1. Verbosity-controlled tracing has been systematically planted across all the scripts.

1. CLI usage has been librified.

1. Everything has been factored and uses common code paths.
    
1. Fixes has been made to make outdated scripts work.

Checklist
---------


erikd:
Have you runs `shellcheck` over that shell code?

deepfire:
@erikd, `scripts/*.sh` are now `shellcheck`-clean.

deepfire:
Squashed.

deepfire:
bors merge

PR #794, opened 2020-04-14
Cardano node systemd activated sockets

Jimbo4350:
Issue
-----------

- #498 

- Currently the cardano-node executable can be started via systemd and successfully inherits two TCP/IP sockets. However when trying to connect to the UNIX socket passed to cardano-node, the connection times out. I was testing the UNIX socket via `cardano-cli`'s `get-tip` command.

- Follow the following to reproduce the above

1) Create a socket in your `run` folder e.g `/run/mytestsocket.sk`
2) Create your `.socket` file in `/etc/systemd/system` (reference: https://www.freedesktop.org/software/systemd/man/systemd.socket.html)
3) Create your `.service` file in `/etc/systemd/system` (reference: https://www.freedesktop.org/software/systemd/man/systemd.service.html)
4) Connect to your socket file via: `nc -U /run/mytestsocket.sk`

This should spawn your specified executable (`cardano-node` in this instance) and pass the activated sockets specified in your `.socket` file to it.

Example setup instructions:

In `/etc/systemd/system`: `test.socket`
```
[Socket]
ListenStream=/run/mytestsocket.sk

#IPv4 socket.
[Socket]
ListenStream=0.0.0.0:7776
ReusePort=true

#IPv6 socket.
[Socket]
ListenStream=7776

#UNIX socket.
[Socket]
ListenStream=../cardano-node/socket/mainnet-socket

[Install]
WantedBy=sockets.target
```
NB: 
- You need to create the socket `/run/mytestsocket.sk` & all paths need to be absolute.
- The order of the sockets specified in the `.socket` file is the order in which `cardano-node` receives the socket. Currently changing this order will prevent the node from starting correctly via `systemd`.

In `/etc/systemd/system`: `test.service`
```
[Service]
Type=simple
ExecStart=../cardano-haskell/cardano-node/scripts/systemd_mainnet.sh
Restart=no
# time to sleep before restarting a service
RestartSec=1
StandardOutput=file:../cardano-node/systemdlogs/systemd-mainnent.log
WorkingDirectory=../cardano-haskell/cardano-node/systemdlogs
NonBlocking=true

[Unit]
# StartLimitIntervalSec in recent systemd versions
StartLimitInterval=0

[Install]
WantedBy=multi-user.target
```

NB: All paths in both files (.socket & .service) need to be absolute.

In `scripts/systemd_mainnet.sh`:

```
#!/usr/bin/env bash

exec /home/jordan/Repos/Work/cardano-haskell/cardano-node/cardano-node/cardano-node run --config /home/jordan/Repos/Work/cardano-haskell/cardano-node/configuration/defaults/mainnet/configuration.yaml --database-path db --socket-path socket/mainnet-socket --topology /home/jordan/Repos/Work/cardano-haskell/cardano-node/configuration/defaults/mainnet/topology.json --use-activated-sockets
```
(you can create the cardano-node binary with the following: `stack build cardano-node:exe:cardano-node --copy-bins --local-bin-path .`)

1) sudo systemctl enable test.socket
2) sudo systemctl enable test.service
3) sudo systemctl start test.socket
3) nc -U /run/mytestsocket.sk

This will start the executable specifed in `test.service` (`ExecStart`) and output the db & logs to `cardano-node/systemdlogs`. This will also pass 3 sockets to the executable in the order listed in `test.socket`.

This in theory should allow you to use the `get-tip` command like so: `cabal exec cardano-cli -- get-tip --config /cardano-node/configuration/defaults/mainnet/configuration.yaml --socket-path /cardano-node/socket/mainnet-socket` however currently this times out.

Checklist
---------


Jimbo4350:
Subsumed by #856

PR #799, opened 2020-04-14
Add Shelley keypair generation

intricate:


cardano-api/src/Cardano/Api/Crypto/Shelley.hs
```diff
@@ -0,0 +1,22 @@
...
+
+import           Crypto.Random (MonadRandom)
+
+import           Shelley.Spec.Ledger.Keys (SKey (..), pattern VKey)
+
+genKeyPair :: MonadRandom m
+           => m KeyPair
+genKeyPair = do
+  sk <- genKeyDSIGN
+  pure $ KeyPairShelley (VKey (deriveVerKeyDSIGN sk)) (SKey sk)
```
intricate, line 22: Perhaps this should be in `cardano-ledger-specs` instead.

cardano-api/src/Cardano/Api/Types.hs
```diff
@@ -35,8 +41,8 @@ data KeyPair
   -- An 'XPub' is 32 bytes of public key and 32 bytes of Chaincode which is used in the
   -- Byron address derivation scheme.
   = KeyPairByron !Crypto.VerificationKey !Crypto.SigningKey
-  | KeyPairShelley -- Exact crypto not ready yet.
-  deriving (Generic, NFData, Show)
+  | KeyPairShelley !(VKey TPraosStandardCrypto) !(SKey TPraosStandardCrypto)
```
intricate, line 44: Should I use the `cardano-ledger-specs` types (`VKey` and `SKey`) or stick with `VerKeyDSIGN (DSIGN TPraosStandardCrypto)` and `SignKeyDSIGN (DSIGN TPraosStandardCrypto)`?

cardano-api/src/Cardano/Api/Crypto/Shelley.hs
erikd: Is that a separate file? If so I think its so small and trivial i should just be in `Cardano.Api`.

cardano-api/src/Cardano/Api.hs
```diff
@@ -78,7 +82,7 @@ mkPublicKey :: KeyPair -> Network -> PublicKey
 mkPublicKey kp nw =
   case kp of
     KeyPairByron vk _ -> PubKeyByron nw vk
-    KeyPairShelley -> PubKeyShelley
+    KeyPairShelley _ _ -> PubKeyShelley
```
erikd, line 85: `PubKeyShelley` needs two fields.

cardano-api/src/Cardano/Api/CBOR.hs
```diff
@@ -54,15 +60,23 @@ keyPairFromCBOR bs =
       tag <- CBOR.decodeWord8
       case tag of
         172  -> KeyPairByron <$> fromCBOR <*> fromCBOR
-        173  -> pure KeyPairShelley
+        173  -> do
+          decodedVerKey <- decodeVerKeyDSIGN
+          decodedSignKey <- decodeSignKeyDSIGN
+          pure $ KeyPairShelley (VKey decodedVerKey) (SKey decodedSignKey)
```
erikd, line 66: No love for `Applicative`.

cardano-api/src/Cardano/Api/CBOR.hs
```diff
@@ -6,6 +8,11 @@ module Cardano.Api.CBOR
   , publicKeyFromCBOR
   , publicKeyToCBOR
 
+  , decodeShelleyVerificationKey
```
erikd, line 11: I find `'decode` and `encode` a little too imprecise. That is why I named these as `typenameToCBOR` and `typenameFromCBOR` .

cardano-api/src/Cardano/Api/CBOR.hs
```diff
@@ -6,6 +8,11 @@ module Cardano.Api.CBOR
...
+  , encodeShelleyVerificationKey
```
erikd, line 12: Probably do not need to export the `encode `and `decode` functions.

intricate: Sure, I'll drop them from the export list.

cardano-api/test/Test/Cardano/Api/Gen.hs
```diff
@@ -38,16 +51,61 @@ genAddress =
 
 genKeyPair :: Gen KeyPair
 genKeyPair =
-  -- When Shelly is sorted out, this should change to `Gen.choose`.
-  Gen.frequency
-    [ (9, genKeyPairByron)
-    , (1, pure KeyPairShelley)
+  Gen.choice
```
erikd, line 54: :+1:

cardano-api/src/Cardano/Api.hs
```diff
@@ -10,6 +11,8 @@ module Cardano.Api
   , KeyPair (..)
   , Network (..)
   , PublicKey (..)
+  , ShelleyKeyDiscriminator (..)
+  , ShelleyVerificationKey (..)
```
erikd, line 15: Its a pity that these leak from the API, but lets get the API finished and then see if we can clean it up.

PR #805, opened 2020-04-15
CAD-830:  refactor log/trace configuration

deepfire:
- don't enable tracers if logging is disabled
- factor TraceConfig as a single source of logger-side logging enabled-ness
- simplify logging init
- inline LoggingConfiguration
- remove dead code and unused exports in Cardano.Config.Logging

Checklist
---------

### DIFF
--- a/cardano-config/src/Cardano/Config/Logging.hs
+++ b/cardano-config/src/Cardano/Config/Logging.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE CPP                 #-}
+{-# LANGUAGE LambdaCase          #-}
 {-# LANGUAGE NamedFieldPuns      #-}
 {-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE Rank2Types          #-}
@@ -9,12 +10,7 @@
 
 module Cardano.Config.Logging
   ( LoggingLayer (..)
-  , LoggingFlag (..)
-  , LoggingConfiguration (..)
   , createLoggingFeature
-  , createLoggingFeatureCLI
-  , loggingCardanoFeatureInit
-  , loggingCLIConfiguration
   -- re-exports
   , Trace
   , Configuration
@@ -36,6 +32,7 @@ import           Cardano.BM.Backend.Aggregation (plugin)
 import           Cardano.BM.Backend.EKGView (plugin)
 import           Cardano.BM.Backend.Monitoring (plugin)
 import           Cardano.BM.Backend.TraceForwarder (plugin)
+import           Cardano.BM.Backend.Switchboard (Switchboard)
 import qualified Cardano.BM.Backend.Switchboard as Switchboard
 import           Cardano.BM.Configuration (Configuration)
 import qualified Cardano.BM.Configuration as Config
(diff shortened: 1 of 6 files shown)


cardano-config/src/Cardano/Config/TraceConfig.hs
```diff
@@ -0,0 +1,163 @@
...
+data TraceConfig'
+
+  -- | Intended to mirror Configuration with TurnOnLogging: False.
+  = TracingDisabled
+
+  -- | Detailed tracing options.  Except for verbosity, eaach option controls
+  --   a particular tracer.
+  | TraceOptions
+  -- Common part.
+  { traceVerbosity :: !TracingVerbosity
```
erikd, line 77: That is a partial function. Production quality Haskell code should avoid using partial functions and more importantly should avoid adding them. Do we not still have a `-Wpartial-field` warnings enabled?

deepfire: @erikd, it's never exported, see the comment.  The partiality is strictly cointained to the module.

cc @dcoutts, as he approved this pattern previously.

erikd: We should have ( and I thought we did have) `'-Wpartial-fields` enabled in all the cabal files. Why has this warning not triggerred? Are we not using `-Werror` ?

If we really are going to use this, then we first need to make sure `-Wpartial-fields` and `-Werror` are working as intended and than disable the partial fields error in this module using `-{-# OPTIONS_GHC -Wno-partial-field #-}`

erikd:
```
λ > import Cardano.Config.TraceConfig 
*Cardano.Config.CommonCLI Cardano.Config.TraceConfig
λ > :t traceAcceptPolicy
traceAcceptPolicy
  :: Cardano.Config.TraceConfig.TraceConfig' -> Bool
*Cardano.Config.CommonCLI Cardano.Config.TraceConfig
λ > :i Cardano.Config.TraceConfig.TraceConfig'

<interactive>:1:1: error:
    Not in scope: ‘Cardano.Config.TraceConfig.TraceConfig'’
*Cardano.Config.CommonCLI Cardano.Config.TraceConfig
λ > 
```
Ok, that works as advertised and the comments should warn people not to do stupid.

cardano-config/src/Cardano/Config/TraceConfig.hs
```diff
@@ -0,0 +1,180 @@
...
+
+import           Data.Aeson
+import           Data.Aeson.Types (Parser)
+
+import           Cardano.BM.Data.Tracer (TracingVerbosity (..))
+
+import           Cardano.Config.Orphanage ()
+
+
+newtype TraceConfig = TraceConfig TraceConfig'
```
Jimbo4350, line 63: I think you can remove `newtype TraceConfig = TraceConfig TraceConfig'`, remove it's `Show` instance and redefine `type TraceId = TraceConfig' -> Bool` as `type TraceId = TraceConfig -> Bool`.

cardano-config/src/Cardano/Config/Logging.hs
```diff
@@ -115,61 +99,22 @@ data LoggingLayer = LoggingLayer
...
-
-
-    -- we construct the cardano feature
-    let cardanoFeature = CardanoFeature
-                            { featureName = "LoggingMonitoringFeature"
-                            , featureStart = liftIO . void $ pure loggingLayer
-                            , featureShutdown = liftIO $ cleanUpLogging loggingLayer
-                            }
+    -> ExceptT ConfigError IO Configuration
+loggingCLIConfiguration = maybe emptyConfig readConfig
```
Jimbo4350, line 107: :+1:

cardano-config/src/Cardano/Config/TraceConfig.hs
deepfire: @Jimbo4350, that `newtype` is necessary to prevent the partiality from escaping from the module -- please, see the explanation in the added comments.

So no, I don't think we can remove it and make the dependent changes you propose.

deepfire:
bors merge

PR #819, opened 2020-04-17
Extract modules from cardano-node to create package cardano-cli

intricate:


cardano-cli/cardano-cli.cabal
```diff
@@ -0,0 +1,123 @@
+name:                  cardano-cli
+-- TODO @intricate: Should this match the `cardano-node` version?
+version:               0.0.1
```
intricate, line 3: @erikd @Jimbo4350

erikd: I think all packages in this repo should have the same version number. @dcoutts ?

intricate: Set it to the same version as `cardano-node`.

cardano-cli/NOTICE
```diff
@@ -0,0 +1,14 @@
+Copyright 2019 Input Output (Hong Kong) Ltd.
```
Jimbo4350, line 1: Should be 2020?

intricate: Right

PR #844, opened 2020-04-24
CAD-830:  [RFC] flip some tracers for ~a good luck~ performance

deepfire:
Note, this is more about raising a discussion -- because the reason why this is an improvement is quite puzzling.  But if we also want a 7% perf increase in default config, right now -- hey, why not..

Context: https://jira.iohk.io/browse/CAD-792?focusedCommentId=15892&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15892 supplies an intriguing story, whereby flipping (on -> off, off -> on) some tracers in the default configuration leads to measurable performance gain.

This picks the creme du crope of the tracers to flip, and -- if we're to believe a (hopefully) thorough benchmark that operates on a (very!) noisy subject -- delivers ~7% more blocks per second.

UPDATE:  the benchmark was reran with extended run length (15 instead of 3 epochs), and the variance went down significantly -- but the measured improvement is still there, at 7%.

NOTE: reading the above JIRA comment in CAD-792 can be really beneficial for getting up to speed with the weirdness..

Checklist
---------


coot:
I don't understand why turning tracers on makes positive impact on performance, that's counter intuitive, and I'd think that it's a performance problem of the logging system. The tracers that were turned on are quite verbose (e.g. logging every `chain-sync` message).

coot:
What is the difference between turning a tracer off and on?

| Branch | of best | wall, s | σ/μ | total CPU, s | σ/μ | total alloc, GB | σ/μ | copied, GB | σ/μ | max live, MB | σ/μ |
|-|-|-|-|-|-|-|-|-|-|-|-|
| dat-disable-all-tracers | 1.13 | 126.3 | 0.15 | 275.5 | 0.13 | 293.3 | 0.21 | 3.65 | 0.05 | 361.6 | 0.04 |

The hint here is that even though the memory footprint was the smallest (max live column), total allocation size and bytes copied by the GC are much higher (so the performance is worse). Maybe it's worth to run profiling build and see what are these allocations.  You could try `profiling-detail=all-functions` in the logging and `none` for other libraries, [profiling-detail](https://www.haskell.org/cabal/users-guide/nix-local-build.html#cfg-field-profiling-detail)

deepfire:
> What is the difference between turning a tracer off and on?

The difference is in the family of configuration options called `Trace*`.

deepfire:
> I don't understand why turning tracers on makes positive impact on performance, that's counter intuitive, and I'd think that it's a performance problem of the logging system.

What evidence, do you think, supports that?

deepfire:
> Maybe it's worth to run profiling build and see what are these allocations.

I already did some form of that space profiling, and monitoring wasn't in the list.

coot:
@deepfire inline answers follows:

> The difference is in the family of configuration options called Trace*.

I am not chasing configuration, but rather what changes it implies in the running code.

> What evidence, do you think, supports that?

It'd like to understand why switching off tracers causes GC to work harder (which is might be the cause of worse performance).

erikd:
@deepfire Could it be that turning on tracers means those tracer objects get used and garbage collected while if the tracer is disabled they are held on to and not GC-ed until much latter?

erikd:
This also suggests that we have a space leak: https://twitter.com/InputOutputHK/status/1256302392267739136

deepfire:
> I am not chasing configuration, but rather what changes it implies in the running code.

> @deepfire Could it be that turning on tracers means those tracer objects get used and garbage collected while if the tracer is disabled they are held on to and not GC-ed until much latter?

Basically, "disabling a tracer" means that the `ouroboros-network` receives a `nullTracer`.

So to explain what this means, one has to explain what a `Tracer m a` is -- it is essentially a typed callback that receives values of certain type, that it can trace in a given context.

So, a `nullTracer` is just a callback that does nothing: https://github.com/input-output-hk/iohk-monitoring-framework/blob/master/contra-tracer/src/Control/Tracer.lhs#L133

So, this is precisely the effect of disabling a tracer, f.e. see https://github.com/input-output-hk/cardano-node/blob/master/cardano-node/src/Cardano/Tracing/Tracers.hs#L262-L268 , where `tracerOnOff` decides what is passed to `ouroboros-network`: https://github.com/input-output-hk/cardano-node/blob/master/cardano-node/src/Cardano/Tracing/Tracers.hs#L321-L323

erikd:
The memory usage graphs posted to twitter *absolutely does* suggest a space leak.

deepfire:
> The memory usage graphs posted to twitter absolutely does suggest a space leak.

@erikd, sure, but what is the relevance here?

erikd:
This is months old and has conflicts. Closing.

PR #865, opened 2020-04-30
config: Add ToJSON instances required for NetworkTopology

erikd:
With round trip tests for the JSON.

### DIFF
--- a/cardano-config/src/Cardano/Config/Types.hs
+++ b/cardano-config/src/Cardano/Config/Types.hs
@@ -31,7 +31,7 @@ module Cardano.Config.Types
     , parseNodeConfigurationFP
     ) where
 
-import           Prelude (read, show)
+import           Prelude (show)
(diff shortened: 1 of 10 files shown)


cardano-config/src/Cardano/Config/Types.hs
```diff
@@ -308,15 +308,28 @@ instance Condense NodeAddress where
...
-newtype NodeHostAddress = NodeHostAddress { getAddress :: Maybe IP.IP }
-                          deriving (Eq, Ord, Show)
+instance ToJSON NodeAddress where
+  toJSON na =
+    object
+      [ "addr" .= maybe "null" toJSON (naHostAddress na)
+      , "port" .= (fromIntegral (naPort na) :: Int)
+      ]
+
+newtype NodeHostAddress
```
Jimbo4350, line 321: :+1:

PR #880, opened 2020-05-04
Add functionality for querying the current protocol parameters from a Shelley node

intricate:


cardano-api/src/Cardano/Api/LocalStateQuery.hs
```diff
@@ -0,0 +1,188 @@
...
+import           Cardano.Config.Types (SocketPath (..))
+
+import qualified Codec.CBOR.Term as CBOR
+
+import           Control.Monad.Class.MonadSTM.Strict
+                   (MonadSTM, StrictTMVar, atomically, newEmptyTMVarM, tryPutTMVar, takeTMVar)
+import           Control.Monad.Trans.Except.Extra (newExceptT)
+import           Control.Tracer (Tracer)
+
+import           Data.ByteString.Lazy (ByteString)
```
erikd, line 29: I think `Cardano.Prelude` has a `LByteString` which means you would not have to hide it and import it here.

cardano-api/src/Cardano/Api/LocalStateQuery.hs
```diff
@@ -0,0 +1,188 @@
...
+  => StrictTMVar IO (Either AcquireFailure result)
+  -> Trace IO Text
+  -> Protocol blk (BlockProtocol blk)
+  -> SocketPath
+  -> (Point blk, Query blk result)
+  -> IO ()
+queryNodeLocalState resultVar trce ptcl (SocketFile socketPath) pointAndQuery = do
+    logInfo trce $ "queryNodeLocalState: Connecting to node via " <> textShow socketPath
+    let ProtocolInfo{pInfoConfig} = protocolInfo ptcl
+    NtC.withIOManager $ \iocp -> do
```
erikd, line 117: Had to scroll up to figure out what `NtC` is. Might be better to use `NodeToClient`. Yes, its more characters but far more readable.

cardano-api/src/Cardano/Api/LocalStateQuery.hs
```diff
@@ -0,0 +1,188 @@
...
+  => Proxy blk
+  -> StrictTMVar IO (Either AcquireFailure result)
+  -> Trace IO Text
+  -> TopLevelConfig blk
+  -> (Point blk, Query blk result)
+  -> NtC.Versions
+      NtC.NodeToClientVersion
+      NtC.DictVersion
+      (NtC.LocalConnectionId
+        -> OuroborosApplication 'InitiatorApp ByteString IO (Either AcquireFailure result) Void)
```
erikd, line 150: :facepalm: APIs

intricate: :man_shrugging:

cardano-api/src/Cardano/Api/LocalStateQuery.hs
```diff
@@ -0,0 +1,190 @@
...
+      NodeToClient.connectTo
+        (NodeToClient.localSnocket iocp socketPath)
+        NetworkConnectTracers
+          { nctMuxTracer = muxTracer
+          , nctHandshakeTracer = handshakeTracer
+          }
+        (localInitiatorNetworkApplication proxy resultVar trce pInfoConfig pointAndQuery)
+        socketPath
+  where
+    -- TODO: it should be passed as an argument
```
erikd, line 129: I think this stuff is probably still in flux, so this is fine!

intricate: I think I actually copied this comment along with some of your code from the `TxSubmit` module :smile:

cardano-api/src/Cardano/Api/LocalStateQuery.hs
```diff
@@ -0,0 +1,190 @@
...
+import           Cardano.BM.Trace (Trace, appendName, logInfo)
+
+import           Cardano.Config.Protocol ()
+import           Cardano.Config.Types (SocketPath (..))
+
+import qualified Codec.CBOR.Term as CBOR
+
+import           Control.Monad.Class.MonadSTM.Strict
+                   (MonadSTM, StrictTMVar, atomically, newEmptyTMVarM, tryPutTMVar, takeTMVar)
+import           Control.Monad.Trans.Except.Extra (newExceptT)
```
Jimbo4350, line 26: Please see the imports section of the style guide and make the necessary corrections.

PR #882, opened 2020-05-05
Much work on the CLI

erikd:


cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -181,6 +182,12 @@ newtype PrivKeyFile
   = PrivKeyFile FilePath
   deriving (Eq, Show)
 
+newtype TxBodyFile = TxBodyFile FilePath
+  deriving (Eq, Show)
```
Jimbo4350, line 188: @erikd should these two have `deriving newtype` ?

cardano-cli/test/cli/address-describe/run
```diff
@@ -0,0 +1,60 @@
+#!/bin/sh -u
```
Jimbo4350, line 1: :100:

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
erikd: Yeah probably, but not a big issue. Can be fixed in the next PR set.

cardano-api/test/Test/Cardano/Api/Gen.hs
```diff
@@ -142,3 +133,55 @@ genTxUnsignedShelley =
...
+-- -------------------------------------------------------------------------------------------------
+
+genByteString :: Range Int -> Gen ByteString
+genByteString lenRange =
+  BS.pack <$> Gen.list lenRange (Gen.element [ '\x00' .. '\xff' ])
+
+-- Generates a fake TxId by applying the right hashing function to a random ByteString.
+genFakeTxId :: Gen TxId
+genFakeTxId =
+  TxId . {- Crypto. -} hashRaw identity <$> genByteString (Range.linear 10 50)
```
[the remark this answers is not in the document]
erikd: No that was intended as a helper for when `hashRaw` bubbles up from `base`.

cardano-api/src/Cardano/Api/Convert.hs
```diff
@@ -0,0 +1,96 @@
+module Cardano.Api.Convert
+  ( addressFromHex
+  , addressToHex
+  , parseTxIn
+  , parseTxOut
+  , renderTxIn
+  , renderTxOut
```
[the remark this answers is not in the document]
erikd: Agree, but will leave that for later. Tests will have to be moved as well.

PR #894, opened 2020-05-07
cli: Implement 'shelley transaction sign' command

erikd:


cardano-cli/src/Cardano/CLI/Key.hs
```diff
@@ -82,8 +82,8 @@ prettyPublicKey vk =
 -- TODO:  we need to support password-protected secrets.
 -- | Read signing key from a file.  Throw an error if the file can't be read or
 -- fails to deserialise.
-readSigningKey :: CardanoEra -> SigningKeyFile -> ExceptT CliError IO SigningKey
-readSigningKey era (SigningKeyFile fp) = do
+readEraSigningKey :: CardanoEra -> SigningKeyFile -> ExceptT CliError IO SigningKey
+readEraSigningKey era (SigningKeyFile fp) = do
```
intricate, line 86: Appreciate the rename :smile:

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -641,6 +644,16 @@ pColdSigningKeyFile =
      <> Opt.help "Filepath of the cold signing key."
      )
 
+pSomeSigningKeyFiles :: Parser [SigningKeyFile]
+pSomeSigningKeyFiles =
+  some $
```
Jimbo4350, line 649: :+1:

PR #898, opened 2020-05-08
Minor CLI improvements

erikd:


cardano-cli/test/cardano-cli-test.hs
```diff
@@ -8,7 +8,9 @@ import           Control.Monad (foldM, forM, when)
 import           Data.Maybe (isNothing)
 import qualified Data.List as List
 import qualified Data.Text as Text
-import           Data.Text.ANSI (brightRed, green)
+import qualified Data.Text.IO
+ as Text
```
intricate, line 12: Was this a mistake?

```suggestion
import qualified Data.Text.IO as Text
```

erikd: Yes, was a goof. Will fix.

PR #899, opened 2020-05-08
Update cabal project file flags and fix warnings

Jimbo4350:
Issue
-----------

Checklist
---------

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -6,11 +6,17 @@ packages:
     cardano-config
     cardano-node
 
+package cardano-api
+  ghc-options: -Werror -Wall -Wredundant-constraints -Wincomplete-record-updates -Wincomplete-uni-patterns -Wpartial-fields
+
+package cardano-cli
+  ghc-options: -Werror -Wall -Wredundant-constraints -Wincomplete-record-updates -Wincomplete-uni-patterns -Wpartial-fields
+
 package cardano-config
-  ghc-options: -Wall -fwarn-redundant-constraints
+  ghc-options: -Werror -Wall -Wredundant-constraints -Wincomplete-record-updates -Wincomplete-uni-patterns -Wpartial-fields
 
 package cardano-node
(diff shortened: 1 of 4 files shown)


cabal.project
```diff
@@ -6,11 +6,17 @@ packages:
     cardano-config
     cardano-node
 
+package cardano-api
+  ghc-options: -Wall -Wredundant-constraints -Wincomplete-record-updates -Wincomplete-uni-patterns -Wpartial-fields -Werror
```
erikd, line 10: Probabl best to move the `-Werror` to the front so its obvious.

Jimbo4350:
bors merge

cabal.project
```diff
@@ -6,11 +6,17 @@ packages:
...
+package cardano-cli
+  ghc-options: -Werror -Wall -Wredundant-constraints -Wincomplete-record-updates -Wincomplete-uni-patterns -Wpartial-fields
+
 package cardano-config
-  ghc-options: -Wall -fwarn-redundant-constraints
+  ghc-options: -Werror -Wall -Wredundant-constraints -Wincomplete-record-updates -Wincomplete-uni-patterns -Wpartial-fields
 
 package cardano-node
-  ghc-options: -Wall -fwarn-redundant-constraints
+  ghc-options: -Werror -Wall -Wredundant-constraints -Wincomplete-record-updates -Wincomplete-uni-patterns -Wpartial-fields
```
mrBliss, line 19: So we removed it from the stack.yaml file (https://github.com/input-output-hk/cardano-node/pull/879/) but added it to the cabal.project file? That's rather inconsistent.

IMO, `-Werror` shouldn't be enabled by default, as it is very annoying during development. It *should* be enabled on CI, so that warnings don't make it into master. For example, see https://github.com/input-output-hk/ouroboros-network/blob/21c6eda6aaddd8729562999ccc27f7b363e6139b/nix/haskell.nix#L42.

PR #917, opened 2020-05-10
cli: Read local node's socket path from an environment variable

erikd:


erikd:
There will be something in the docs

PR #922, opened 2020-05-10
Add cli commands to create registration and retirement stake pool certificates

Jimbo4350:
Issue
-----------

Checklist
---------


cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -112,6 +117,31 @@ data PoolCmd
...
+      [VerificationKeyFile]
+      -- ^ Pool owner verification staking key(s).
+      [ShelleyStakePoolRelay]
+      -- ^ Stake pool relays.
+      OutputFile
+  | PoolRetirmentCert
+      VerificationKeyFile
+      -- ^ Stake pool verification key.
+      Shelley.EpochNo
+      -- ^ Epoch in which to retire the stake pool. --TODO: Double check this
```
Jimbo4350, line 142: Need to double check this.

Jimbo4350:
bors merge

PR #931, opened 2020-05-11
Add command to calculate the minimum fee for a transaction

intricate:


cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -192,6 +203,18 @@ data GenesisCmd
 -- Shelley CLI flag/option data types
 --
 
+newtype ProtocolParamsFile
+  = ProtocolParamsFile FilePath
+  deriving (Show, Eq)
```
erikd, line 208: Of you do `deriving newtype` the default `Show` instance will be much nicer.

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -961,6 +1005,24 @@ pTxFile fdir =
       <> Opt.help (show fdir ++ " filepath of the Tx.")
       )
 
+pNumTxIns :: Parser NumTxIns
+pNumTxIns =
+  NumTxIns <$>
+    Opt.option Opt.auto
+      (  Opt.long "num-tx-ins"
```
erikd, line 1012: Maybe "tx-in-count" instead?

intricate: Sure

PR #933, opened 2020-05-11
Add Shelley tip query command

intricate:


erikd:
This command needs a `--network-magic` parameter for non-mainnet.

intricate:
> This command needs a `--network-magic` parameter for non-mainnet.

The command does accept a `--network-magic` parameter:

```  | QueryTip Network```

erikd:
Choosing the network has been sorted out.

PR #937, opened 2020-05-11
CAD-900: Node traces processed txs num.

denisshevchenko:
Issue
-----------

- Node traces the number of processed transactions.

- related to https://github.com/input-output-hk/cardano-benchmarking/pull/51

Checklist
---------

### DIFF
--- a/cardano-node/src/Cardano/Node/Run.hs
+++ b/cardano-node/src/Cardano/Node/Run.hs
@@ -117,7 +117,11 @@ runNode loggingLayer npm@NodeCLI{protocolFiles} = do
         Left err -> (putTextLn $ renderProtocolInstantiationError err) >> exitFailure
         Right (SomeConsensusProtocol p) -> pure $ SomeConsensusProtocol p
 
-    tracers <- mkTracers (ncTraceConfig nc) trace
+    txsProcessedCounter :: MVar Integer <- newMVar 0
+
+    tracers <- mkTracers (ncTraceConfig nc) trace txsProcessedCounter
+
+    txsProcessedThread <- Async.async $ traceTxsProcessed (appendName "metrics" trace) txsProcessedCounter 0
 
 #ifdef UNIX
     let viewmode = ncViewMode nc
@@ -132,6 +136,7 @@ runNode loggingLayer npm@NodeCLI{protocolFiles} = do
 
         handleSimpleNode p trace tracers npm (const $ pure ())
         Async.uninterruptibleCancel upTimeThread
+        Async.uninterruptibleCancel txsProcessedThread
 
       LiveView   -> do
 #ifdef UNIX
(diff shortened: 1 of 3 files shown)


cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -117,7 +117,11 @@ runNode loggingLayer npm@NodeCLI{protocolFiles} = do
         Left err -> (putTextLn $ renderProtocolInstantiationError err) >> exitFailure
         Right (SomeConsensusProtocol p) -> pure $ SomeConsensusProtocol p
 
-    tracers <- mkTracers (ncTraceConfig nc) trace
+    txsProcessedCounter :: MVar Integer <- newMVar 0
```
CodiePP, line 120: in the near future, this can become a data structure that contains more state variables.

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -117,7 +117,11 @@ runNode loggingLayer npm@NodeCLI{protocolFiles} = do
...
+
+    tracers <- mkTracers (ncTraceConfig nc) trace txsProcessedCounter
+
+    txsProcessedThread <- Async.async $ traceTxsProcessed (appendName "metrics" trace) txsProcessedCounter 0
```
CodiePP, line 124: is the name space: "cardano.node.metrics" ?

cardano-node/src/Cardano/Node/Run.hs
denisshevchenko: Exactly.

cardano-node/src/Cardano/Node/Run.hs
denisshevchenko: Yes.

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -407,15 +408,19 @@ mkTracers traceConf tracer = do
...
         traceNamedObject tr' (meta, logValue1)
         traceNamedObject tr' (meta, logValue2)
-        traceNamedObject tr' (meta, logValue3)
+ 
+        case mempoolEvent of
+          TraceMempoolRemoveTxs txs _ ->
+            -- TraceMempoolRemoveTxs are previously valid transactions that are no longer valid because of
+            -- changes in the ledger state. These transactions are already removed from the mempool,
+            -- so we can treat them as completely processed.
+            modifyMVar_ txsProcessedCounter $ \counter -> return $ counter + fromIntegral (length txs)
```
mrBliss, line 422: You're storing a thunk in the `MVar` that holds onto `txs`. As the `+` is lazy, you're accumulating them. This means that you're holding on to *all* the transactions that are removed.

The only thing forcing the thunk is `when (txsProcessed /= txsNum)`, so you're only holding onto the transactions that were removed in a single second. So not a huge space leak, but many small ones can still make the ship sink.

So: force that `counter` before storing it! Alternatively, would `atomicModifyIORef'` suffice here? And why an `Integer`? What about `Word64` or `Word`?

denisshevchenko: > And why an `Integer`?

Because we want to trace that value (from `MVar`) as `LogValue` with `PureI` (which takes `Integer`).

Anyway, @mrBliss, this is "not-finished" solution. I mean, we need some additional information 
to trace (for example, the number of forged blocks), so we'll store a structure in this `MVar`, not a single value. Moreover, it will probably be `TVar` instead of `MVar`.

mrBliss: > > And why an `Integer`?
> 
> Because we want to trace that value (from `MVar`) as `LogValue` with `PureI` (which takes `Integer`).

Alright.

> 
> Anyway, @mrBliss, this is "not-finished" solution. I mean, we need some additional information to trace (for example, the number of forged blocks, please see CAD-803). So we'll store a structure in this `MVar`, not a single value. Moreover, it will probably be `TVar` instead of `MVar`. This PR is solving one single bug.

I don't understand why you then merge this "not-finished" solution. You just added a small space leak (reset every second, fortunately) :man_shrugging:

erikd: @denisshevchenko We must fix this as a high priority task.

denisshevchenko: > I don't understand why you then merge this "not-finished" solution.

Because it solves a bug. An old bug, both in TUI and in RTView service.

PR #940, opened 2020-05-11
Add CODEOWNERS

erikd:
The idea is to disallow merging of PRs without approval of CODEOWNERS.


erikd:
Had a an approval from Duncan in Slack.

PR #943, opened 2020-05-12
CAD-803: Trace forged blocks. Strict evaluation.

denisshevchenko:
We keep counters of transactions processed by this node by observing events of `TraceMempoolRemoveTxs`, and the number of forged blocks by this node from observing events of `TraceForgedBlock`.

These counters are then traced under the named context `cardano.node.metrics` and displayed in the *TUI* (tbd in another PR) and *RTview*.

-----

_This PR fixes the bug with laziness (please see https://github.com/input-output-hk/cardano-node/pull/937#discussion_r423216013)._

### DIFF
--- a/cardano-node/src/Cardano/Node/Run.hs
+++ b/cardano-node/src/Cardano/Node/Run.hs
@@ -28,6 +28,7 @@ import           Control.Tracer
 import qualified Data.ByteString.Char8 as BSC
 import           Data.Either (partitionEithers)
 import           Data.Functor.Contravariant (contramap)
+import           Data.IORef (IORef, newIORef)
 import qualified Data.List as List (lookup, null)
 import           Data.Proxy (Proxy (..))
 import           Data.Semigroup ((<>))
@@ -122,11 +123,9 @@ runNode loggingLayer npm@NodeCLI{protocolFiles} = do
         Left err -> putTextLn (renderProtocolInstantiationError err) >> exitFailure
         Right (SomeConsensusProtocol p) -> pure $ SomeConsensusProtocol p
 
-    txsProcessedCounter :: MVar Integer <- newMVar 0
+    bcCounters :: IORef BlockchainCounters <- newIORef initialBlockchainCounters
 
-    tracers <- mkTracers (ncTraceConfig nc) trace txsProcessedCounter
-
-    txsProcessedThread <- Async.async $ traceTxsProcessed (appendName "metrics" trace) txsProcessedCounter 0
+    tracers <- mkTracers (ncTraceConfig nc) trace bcCounters
 
 #ifdef UNIX
     let viewmode = ncViewMode nc
@@ -141,7 +140,6 @@ runNode loggingLayer npm@NodeCLI{protocolFiles} = do
 
         handleSimpleNode p trace tracers npm (const $ pure ())
         Async.uninterruptibleCancel upTimeThread
-        Async.uninterruptibleCancel txsProcessedThread
 
       LiveView   -> do
 #ifdef UNIX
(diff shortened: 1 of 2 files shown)


denisshevchenko:
@mrBliss > It is strange that you're storing both counters in a single TVar. They seem to be independent from each other, so I would expect two separate TVars.

Yes, it's true. I could work with two separate `MVar`s. But in the nearest future we add a few more counters here (slots missed, forks created, etc.), in this case we'll have a lot of `MVar`s. So I thought 
that 1 single `TVar` with all the counters is a better idea.

denisshevchenko:
Please do not merge it for now, this PR will be simplified soon.

erikd:
> Please do not merge it for now, this PR will be simplified soon.

Switching this to a Draft.

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -413,13 +422,18 @@ mkTracers traceConf tracer txsProcessedCounter = do
...
+          TraceMempoolRemoveTxs txs _ -> do
             -- TraceMempoolRemoveTxs are previously valid transactions that are no longer valid because of
             -- changes in the ledger state. These transactions are already removed from the mempool,
             -- so we can treat them as completely processed.
-            modifyMVar_ txsProcessedCounter $ \counter -> return $ counter + fromIntegral (length txs)
+            newCounter <- atomicModifyIORef' bcCounters $ \counters ->
+              let newCounter = bcTxsProcessedNum counters + fromIntegral (length txs)
+              in (counters { bcTxsProcessedNum = newCounter }, newCounter)
+            meta' <- mkLOMeta Notice Public
+            traceNamedObject tr' (meta', LogValue "txsProcessedNum" (PureI $ fromIntegral newCounter))
```
njd42, line 436: What is the semantics`bcTxsProcessedNum` - you are counting them only when removed - this needs to be carefully explained so that end users understand this

denisshevchenko: Yes, I explained it in the comments:

-- TraceMempoolRemoveTxs are previously valid transactions that are no longer valid because of
-- changes in the ledger state. These transactions are already removed from the mempool,
-- so we can treat them as completely processed.

When txs were added in the mempool - they are prepared to processing, and only when they are removed from the mempool, it means that they are completely processed.

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -117,11 +118,9 @@ runNode loggingLayer npm@NodeCLI{protocolFiles} = do
         Left err -> (putTextLn $ renderProtocolInstantiationError err) >> exitFailure
         Right (SomeConsensusProtocol p) -> pure $ SomeConsensusProtocol p
 
-    txsProcessedCounter :: MVar Integer <- newMVar 0
+    bcCounters :: IORef BlockchainCounters <- newIORef (BlockchainCounters 0 0)
```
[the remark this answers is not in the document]
denisshevchenko: Done.

denisshevchenko:
@njd42 I made changes you asked for.

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -381,6 +385,21 @@ mkTracers traceConf tracer bcCounters = do
                   _ -> pure ()
               _ -> pure ()
 
+    notifyForkIsCreated :: Tracer IO (WithSeverity (ChainDB.TraceEvent blk))
+    notifyForkIsCreated =
+        Tracer $ \(WithSeverity _ ev') ->
+          case ev' of
+              (ChainDB.TraceAddBlockEvent ev) -> case ev of
```
erikd, line 392: Un-needed parens.

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -381,6 +385,21 @@ mkTracers traceConf tracer bcCounters = do
...
+                  ChainDB.SwitchedToAFork _ _ _ -> do
```
erikd, line 393: If you use `ChainDB.SwitchedToAFork {}` there will no error message when the number of fields changes.

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -481,6 +501,33 @@ mkTracers traceConf tracer bcCounters = do
           -- The rest of the constructors.
           _ -> pure ()
 
+        notifySlotsMissedIfNeeded = Tracer $ \case
+          Consensus.TraceNodeIsLeader _ -> do
```
erikd, line 505: `Consensus.TraceNodeIsLeader {}`

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -481,6 +501,33 @@ mkTracers traceConf tracer bcCounters = do
...
+            newCounter <- atomicModifyIORef' bcCounters $ \counters ->
+              let nc = bcNodeIsLeaderNum counters + 1
+              in (counters { bcNodeIsLeaderNum = nc }, nc)
+            meta <- mkLOMeta Notice Public
+            traceNamedObject (appendName "metrics" tracer)
+                             (meta, LogValue "nodeIsLeaderNum" (PureI $ fromIntegral newCounter))
+          Consensus.TraceNodeNotLeader _ -> do
```
erikd, line 512: Same/

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -413,18 +446,27 @@ mkTracers traceConf tracer txsProcessedCounter = do
...
+      TraceMempoolRemoveTxs txs _ -> do
+        -- TraceMempoolRemoveTxs are previously valid transactions that are no longer valid because of
+        -- changes in the ledger state. These transactions are already removed from the mempool,
+        -- so we can treat them as completely processed.
+        newCounter <- atomicModifyIORef' bcCounters $ \counters ->
+          let nc = bcTxsProcessedNum counters + fromIntegral (length txs)
+          in (counters { bcTxsProcessedNum = nc }, nc)
+        meta <- mkLOMeta Notice Public
+        traceNamedObject (appendName "metrics" tracer)
+                         (meta, LogValue "txsProcessedNum" (PureI $ fromIntegral newCounter))
```
mrBliss, line 462: This piece of code is repeated in many places. Best to make a function out of it.

CodiePP:
all issues were addressed.
can you please recheck @njd42 and @erikd ?

erikd:
bors merge

PR #950, opened 2020-05-12
Add shelley update protocol parameters cli command

Jimbo4350:
- Adds a command `create-update-proposal` for the shelley era
- Adds an additional argument to `build-raw` for the inclusion of update proposals.

### DIFF
--- a/cardano-api/src/Cardano/Api.hs
+++ b/cardano-api/src/Cardano/Api.hs
@@ -39,7 +39,7 @@ module Cardano.Api
   , TxIx
   , Lovelace (..)
   , SlotNo (..)
-
+  , Update (..)
   , ShelleyTxBody
 
   , getTxSignedBody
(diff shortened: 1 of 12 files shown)


cardano-api/src/Cardano/Api.hs
```diff
@@ -382,6 +417,10 @@ buildShelleyTransaction txins txouts ttl fee certs = do
    certDiscrim (ShelleyGenesisDelegationCertificate sGenDelegCert) = sGenDelegCert
    certDiscrim (ShelleyMIRCertificate mirCert) = mirCert
 
+   updateDiscrim :: Maybe Update ->  Shelley.StrictMaybe ShelleyUpdate
```
erikd, line 420: Slightly odd name for that. `toStrictMaybe` instead?

In fact, isn't that a lib function somewhere?

Jimbo4350: `maybeToStrictMaybe :: Maybe a -> StrictMaybe a` lives in the ledger specs library but my function is `a -> b`. I can rename it.

Jimbo4350:
bors merge

PR #952, opened 2020-05-12
Add new cli tests

ArturWieczorek:
Add new cli test
Fix address-key-hash test


cardano-cli/test/cli/address-key-hash/run
```diff
@@ -12,12 +12,13 @@ setup_data_dir "${testname}"
 
 error=0
 
+assert_file_exists "${DATA_DIR}/address.vkey"
```
ArturWieczorek, line 15: This test was marked as passed when I ran it - **however** it was failing in reality - because there was a file inside test dir with name `data` and it should be a file with name `address.vkey`  inside data dir: `data\address.vkey`

cardano-cli/test/cli/genesis-create/run
```diff
@@ -78,4 +78,32 @@ if test "${keyhash_delegate_count}" != "${gendlg_count}" ; then
...
+
+    assert_string 'Genesis signing key' "${TEST_DIR}/genesis-keys/genesis${i}.skey"
+    assert_string 'Genesis verification key' "${TEST_DIR}/genesis-keys/genesis${i}.vkey"
+
+    # Check delegate keys:
+    assert_file_exists "${TEST_DIR}/delegate-keys/delegate${i}.skey"
+    assert_file_exists "${TEST_DIR}/delegate-keys/delegate${i}.vkey"
+    assert_file_exists "${TEST_DIR}/delegate-keys/delegate-opcert${i}.counter"
+
+    assert_string 'Node operator signing key' "${TEST_DIR}/delegate-keys/delegate${i}.skey"
```
ArturWieczorek, line 97: I guess there is no need to check every `i` file for titles/headers, should be enough for 1 of each kind

cardano-cli/test/cli/genesis-key-gen-delegate/run
```diff
@@ -0,0 +1,35 @@
...
+assert_file_exists "${TEST_DIR}/key-gen.vkey"
+assert_file_exists "${TEST_DIR}/key-gen.skey"
+assert_file_exists "${TEST_DIR}/op-cert.counter"
+
+# Check type:
+assert_string 'Node operator verification key' "${TEST_DIR}/key-gen.vkey"
+assert_string 'Node operator signing key' "${TEST_DIR}/key-gen.skey"
+assert_string 'Node operational certificate issue counter' "${TEST_DIR}/op-cert.counter"
+
+# Check title:
```
ArturWieczorek, line 31: This one (and probably some below too) looks like clone of this test `node-key-gen/run` - so not sure if it is needed - **however** - the titles for those files are different, so we can check for those differences if we want to

cardano-cli/test/cli/genesis-initial-addr/run
```diff
@@ -0,0 +1,24 @@
...
+setup_data_dir "${testname}"
+assert_file_exists "${DATA_DIR}/utxo.vkey"
+
+${CARDANO_CLI} shelley genesis key-hash \
+    --verification-key-file "${DATA_DIR}/utxo.vkey" \
+    > "${TEST_DIR}/initial-utxo-addr"
+
+assert_file_exists "${TEST_DIR}/initial-utxo-addr"
+assert_line_count 1 "${TEST_DIR}/initial-utxo-addr"
+assert_string "34257772315e97141af2d6f6ddd097d9a8923c5828434a143ab29e2d4e27f406" "${TEST_DIR}/initial-utxo-addr"
```
ArturWieczorek, line 22: We probably should add this to all tests - if  for example `--verification-key-file "${DATA_DIR}/utxo.vkey" ` will have a mistake in the filename then `"${TEST_DIR}/initial-utxo-addr"` will be created and it will pass because it will contain a one line... but it will be following message: `verification key read error: TextView IO exception at: test/cli/genesis-initial-addr/data/utxo.vkeyk Error: test/cli/genesis-initial-addr/data/utxo.vkeyk: openBinaryFile: does not exist (No such file or directory)` so I think we should also check the content and do it for other tests.

cardano-cli/test/cli/address-key-hash/run
```diff
@@ -12,12 +12,15 @@ setup_data_dir "${testname}"
...
+
 ${CARDANO_CLI} shelley address key-hash \
     --verification-key-file "${DATA_DIR}/address.vkey" \
     > "${TEST_DIR}/vkey-hash"
 
 assert_file_exists "${TEST_DIR}/vkey-hash"
-
 assert_line_count 1 "${TEST_DIR}/vkey-hash"
+assert_string "f8d336c2e57d6f675bbe48c88584d8f4ee199b76676526c3b9fadb1d1a7b7ebb" "${TEST_DIR}/vkey-hash"
```
ArturWieczorek, line 23: If there is error in the file name argument this will still pass, file with error message will be created that's why I added this check.

cardano-cli/test/cli/address-key-hash/run
erikd: Thanks for catching that. I don;t think we need an assert that input data exists but we should fix it whenever we find it to be incorrect. Now that ists fixed we should remove the assertion that it exists.

cardano-cli/test/cli/genesis-create/run
erikd: i am fine with that being tested in a for loop. I would just as the you follow the somewhat idiosyncratic indenation style already present in these  (2 space indent, `done` indented the same as the body and `do` in the same line as the `for` with a semicolon.

cardano-cli/test/cli/genesis-key-gen-delegate/run
erikd: Worth removing that comment and keeping the test :D.

Tests are there to make sure we know when we break something. We should keep this.

cardano-cli/test/cli/genesis-initial-addr/run
erikd: There is something I need to fix with the way errors are reported, so that may not be necessary.

Bsically error messages should always go the `stderr` and not `stdout`.

ArturWieczorek:
I noticed that for some recently updated/added commands we don't use titles in generated by them files, is that okay or they should be updated @erikd ? (Other files had titles)

**ex 1:**
```
type: TxSignedShelley
title: Free form text
cbor-hex:
18b183a4...
```
**ex 2:**
```
type: TxUnsignedShelley
title: Free form text
cbor-hex:
 18b3a400....
```

cardano-cli/test/cli/genesis-key-gen-delegate/run
erikd: The title is free form test. The user can edit and modify it. We should not be checking it.

erikd: I have pushed a fix that removes that title check. I will merge this once CI is happy.

PR #965, opened 2020-05-13
CAD-971 check required tracers in LiveView

CodiePP:
improved user experience when using "LiveView"

because the "TracerConfig" cannot be altered after it has been parsed from the configuration file, the node's startup process is only checking whether the required tracers are turned on. The missing tracer configurations are printed to the console and the startup is waiting for the user to acknowledge before continuing. If all required tracers are turned on then no interruption happens.

### DIFF
--- a/cardano-node/src/Cardano/Node/Run.hs
+++ b/cardano-node/src/Cardano/Node/Run.hs
@@ -17,22 +17,21 @@
 
 module Cardano.Node.Run
   ( runNode
-  , ViewMode(..)
   )
 where
 
-import           Cardano.Prelude hiding (ByteString, atomically, take, unlines, trace)
-import           Prelude (String, error, unlines)
+import           Cardano.Prelude hiding (ByteString, atomically, take, trace)
+import           Prelude (error)
 
 import qualified Control.Concurrent.Async as Async
 import           Control.Tracer
 import qualified Data.ByteString.Char8 as BSC
 import           Data.Either (partitionEithers)
 import           Data.Functor.Contravariant (contramap)
-import qualified Data.List as List
+import qualified Data.List as List (lookup, null)
 import           Data.Proxy (Proxy (..))
 import           Data.Semigroup ((<>))
-import           Data.Text (Text, breakOn, pack, take)
+import           Data.Text (Text, breakOn, pack, take, unlines)
 import           GHC.Clock (getMonotonicTimeNSec)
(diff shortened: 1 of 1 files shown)


cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -141,6 +142,17 @@ runNode loggingLayer npm@NodeCLI{protocolFiles} = do
       LiveView   -> do
 #ifdef UNIX
         let c = llConfiguration loggingLayer
+        -- check required tracers are turned on
+        let reqtrs = [("TraceBlockFetchDecisions",traceBlockFetchDecisions), ("TraceChainDb",traceChainDB), ("TraceForge",traceForge), ("TraceMempool",traceMempool)]
+            trsinactive = filter (\(_,f) -> not $ traceEnabled (ncTraceConfig nc) f) reqtrs
+        when (length trsinactive > 0) $ do
```
erikd, line 148: Probably does not matter in this case but `length` forces the evaluation of the list in order to calculate the length.

A better solution would be:
```
     unless (List.null trsinactive) $ do
```
because that just inspects the head of the list to determine if it is empty or not.

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -315,7 +327,7 @@ handleSimpleNode p trace nodeTracers npm onKernel = do
          traceWith tracer $
            "System started at " <> show (nodeStartTime (Proxy @blk) cfg)
 
-         traceWith tracer $ unlines
+         traceWith tracer $ Prelude.unlines
```
erikd, line 330: Would it not be better to use `Text.unlines` which specifles exactly what data type the we are operating on.

Since `Text` and `ByteString` export many of the same function names, it is best to always uses those functions qualified, ie as `Text.unlines` or `BS.unlines` or `LBS.unlines`.

cardano-node/src/Cardano/Node/Run.hs
CodiePP: thanks, much better! :+1:

cardano-node/src/Cardano/Node/Run.hs
CodiePP: yes, that 'unlines' was causing warnings and was 'bit awkward. Changing the type of the tracer to 'Text' and using Text.unlines is a very good solution. Thanks!

PR #968, opened 2020-05-14
config: Fix NetworkTopology breakage

erikd:
@karknu I was anot able to get to the bottom of what broke with my previous change so i have changed it back to having the `Maybe` inside the newtype. If nothing else I at least added some tests.

### DIFF
--- a/cardano-config/src/Cardano/Config/Parsers.hs
+++ b/cardano-config/src/Cardano/Config/Parsers.hs
@@ -232,15 +232,15 @@ parseNodeId desc =
     )
 
 parseNodeAddress :: Parser NodeAddress
-parseNodeAddress = NodeAddress <$> optional parseHostAddr <*> parsePort
+parseNodeAddress = NodeAddress <$> parseHostAddr <*> parsePort
 
 parseHostAddr :: Parser NodeHostAddress
 parseHostAddr =
-    option (NodeHostAddress <$> auto) (
+    option (eitherReader parseNodeHostAddress) (
           long "host-addr"
        <> metavar "HOST-NAME"
        <> help "Optionally limit node to one ipv6 or ipv4 address"
-       <> value (NodeHostAddress "127.0.0.1")
+       <> value (NodeHostAddress Nothing)
     )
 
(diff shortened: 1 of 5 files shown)


cardano-config/src/Cardano/Config/Parsers.hs
```diff
@@ -232,15 +232,15 @@ parseNodeId desc =
...
 
 parseHostAddr :: Parser NodeHostAddress
 parseHostAddr =
-    option (NodeHostAddress <$> auto) (
+    option (eitherReader parseNodeHostAddress) (
           long "host-addr"
        <> metavar "HOST-NAME"
        <> help "Optionally limit node to one ipv6 or ipv4 address"
-       <> value (NodeHostAddress "127.0.0.1")
+       -- <> value Nothing
```
karknu, line 243: Should be ` <> (value $ NodeHostAddress Nothing)`
Without any specif address specified `cardano-node` will use any IPv4 and/or IPv6 address.

erikd: Fixed!

cardano-config/test/Test/Cardano/Config/Json.hs
```diff
@@ -18,13 +20,17 @@ prop_roundtrip_NodeAddress_JSON =
...
+    Hedgehog.cover 1 "  has address" $ isJust (unNodeHostAddress $ naHostAddress na)
+    Hedgehog.cover 1 "  no address" $ isNothing (unNodeHostAddress $ naHostAddress na)
 
 prop_roundtrip_NodeHostAddress_JSON :: Property
 prop_roundtrip_NodeHostAddress_JSON =
   Hedgehog.property $ do
     nha <- Hedgehog.forAll genNodeHostAddress
     Hedgehog.tripping nha toJSON fromJSON
     Hedgehog.tripping nha encode decode
+    Hedgehog.cover 1 "  has address" $ isJust (unNodeHostAddress nha)
```
Jimbo4350, line 32: :+1:

Issue #770, opened 2020-04-07
High CPU Usage on MacOS


kevinhammond:
Users report full CPU usage on MacOS even though the node has finished syncing.  We would expect the CPU usage to be near zero.

It is possible that this is not a bug in the node, but just the wallet making repeated queries.
[logs-2020-04-07T224954.0732Z.zip](https://github.com/input-output-hk/cardano-node/files/4447382/logs-2020-04-07T224954.0732Z.zip)

kevinhammond:
Testing on another Mac does not replicate the issue (i.e. the CPU usage drops to zero once synced).  Here are the corresponding logs.  In total, the process has used 4:11:27.65 CPU since starting (which is high), memory is stable at 773MB.  It has received 6.21GB of data and sent 12.0MB.  It has written 2.81GB to disk and read 5.34GB.

[logs-2020-04-08T091633.0206Z.zip](https://github.com/input-output-hk/cardano-node/files/4449498/logs-2020-04-08T091633.0206Z.zip)

bergr01:
The two issues I have observed on Mac so far are:

TUI (LiveView) is reporting CPU usage of 70% of above when in fact the cardano-node process seems to be using under 1% of the CPU
Similarly, the Network In field of the LiveView TUI is sometimes showing some ridiculous value (but in the heading, i.e. what the maximum value was observed - not the current value) - e.g. 

Network In:                         17929164000000000.0 KB/s

disassembler:
This is resolved in https://github.com/input-output-hk/cardano-node/pull/986

tatyanavych:
Both @nikola.glumac and @darko-mijic have tested this, and confirmed to be resolved.

kevinhammond:
If it is resolved, we can close it.

PR #969, opened 2020-05-14
bump haskell.nix

angerman:
This is so that input-output-hk/haskell.nix#604 is included (macOS 200% cpu utilisation fix) /cc @karknu


karknu:
This fixes #770

PR #977, opened 2020-05-15
Doc : provides  supply parameter when regenerating genesis.json before starting nodes

Proxiweb:
If the genesis.json is regenerate without --supply parameter, initialFunds will be set to 0 and the subsequent utxo querying step will return 0 instead of 500000000

### DIFF
--- a/doc/shelley-genesis.md
+++ b/doc/shelley-genesis.md
@@ -955,7 +955,7 @@ be 30 seconds into the future, and then we can start our nodes. This assumes
 you did not do any manual tweaking of the generated `genesis.json`, as it will
 be overwritten.
 ```
-$ cardano-cli shelley genesis create --genesis-dir example/
+$ cardano-cli shelley genesis create --genesis-dir example/ --supply 1000000000
 ```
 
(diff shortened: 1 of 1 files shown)


doc/shelley-genesis.md
```diff
@@ -955,7 +955,7 @@ be 30 seconds into the future, and then we can start our nodes. This assumes
 you did not do any manual tweaking of the generated `genesis.json`, as it will
 be overwritten.
 ```
-$ cardano-cli shelley genesis create --genesis-dir example/
+$ cardano-cli shelley genesis create --genesis-dir example/ --supply 1000000000
```
erikd, line 958: This example does not set an initial supply, but one just a little later does.

Rather than adding a supply amount here, I think it would be better to add a note and explaining that it will be set in a later coommand.

I have a patch in a WIP branch that adds the extra comment.

Proxiweb: The confusion came for me from de manually / semi-automagically parts : I started with the manually and skipped automagically, so I didn't type the later automagical command. le mixed manual / automagical is a bit confusing. Maybe giving the later command in both format would be better.

PR #983, opened 2020-05-16
Doc : add missing supply parameter  

Proxiweb:
The previous pull request was on the wrong line. Missing supply parameter puts funds to 0 in the genesis.json so querying utxos doesn't give good results


erikd:
This issue was addressed in https://github.com/input-output-hk/cardano-node/pull/985/commits/2e8be1b86d3f9878409fc7de45976346dbf95d10 .

Proxiweb:
@erikd this PR was not about the same line, the first one was on the wrong line. This issue is still relevant.

PR #993, opened 2020-05-18
CAD-977 AdoptBlock with list of tx ids

CodiePP:
tracing the list of transaction ids as a list and not a text. this aids analysis.

### DIFF
--- a/cardano-config/src/Cardano/TracingOrphanInstances/Consensus.hs
+++ b/cardano-config/src/Cardano/TracingOrphanInstances/Consensus.hs
@@ -755,7 +755,7 @@ instance ( Condense (HeaderHash blk)
       [ "kind" .= String "TraceAdoptedBlock"
       , "slot" .= toJSON (unSlotNo slotNo)
       , "block hash" .=  (condense $ blockHash blk)
-      , "tx ids" .= (show $ map txId txs)
+      , "tx ids" .= toJSON (map (show . txId) txs)
       ]
   toObject _verb (TraceAdoptedBlock slotNo blk _txs) =
     mkObject


cardano-config/src/Cardano/TracingOrphanInstances/Consensus.hs
```diff
@@ -755,7 +755,7 @@ instance ( Condense (HeaderHash blk)
       [ "kind" .= String "TraceAdoptedBlock"
       , "slot" .= toJSON (unSlotNo slotNo)
       , "block hash" .=  (condense $ blockHash blk)
-      , "tx ids" .= (show $ map txId txs)
+      , "tx ids" .= toJSON (map (show . txId) txs)
```
erikd, line 758: `txId` is a raw ByteString isn't it? If so, it should be Base16 encoded.

CodiePP: these tx ids are not raw bytestring but base16/hex encoded:
```
{"at":"2020-05-15T19:38:20.32Z","env":"1.11.0:66f0e","ns":["cardano.node.Forge"],"data":{"kind":"TraceAdoptedBlock","slot":341,"tx ids":"[txid:d1315d5c,txid:e1aa6320,txid:d795aa9b,txid:5f089341,txid:38de8a9a,txid:9803848d,txid:bfd47d95,txid:31484cca,txid:fa99e623,txid:0448dc15,txid:2faa15da,txid:dc1cc3d7,txid:0f0525c2,txid:bedfec7a,txid:732c9acb,txid:203fd829,txid:1b7d1fa7,txid:a483f777,txid:153c97d1,txid
```

PR #1007, opened 2020-05-20
CAD-973: Trace peers from node kernel. Additional fields.

denisshevchenko:
1. Add new fields in the TUI, for additional tracers (missed slots, created forks, leadership, and number of blocks forged by this node).
2. Extract information from NodeKernel is moved from LiveViewBK, because it's not TUI-specific code.
3. Trace information about connected peers (it will be received by RTView service).

### DIFF
--- a/cardano-node/src/Cardano/Node/TUI/LiveView.hs
+++ b/cardano-node/src/Cardano/Node/TUI/LiveView.hs
@@ -22,7 +22,8 @@ module Cardano.Node.TUI.LiveView
     , captureCounters
     , liveViewPostSetup
     , setNodeThread
-    , setNodeKernel
+    , storePeersInLiveView
+    , storeMempoolDataInLiveView
     ) where
 
 import           Cardano.Prelude hiding (modifyMVar_, newMVar, on, readMVar, show)
@@ -35,9 +36,6 @@ import           Control.Concurrent.MVar.Strict (MVar, modifyMVar_, newMVar, rea
 import           Control.DeepSeq (rwhnf)
 import           Control.Monad (forever, void)
 import           Control.Monad.IO.Class (liftIO)
-import qualified Control.Monad.Class.MonadSTM.Strict as STM
-import qualified Data.Map.Strict as Map
-import qualified Data.Set as Set
 import           Data.Text (Text)
 import qualified Data.Text as Text
 import           Data.Time.Calendar (Day (..))
@@ -77,23 +75,12 @@ import           Cardano.BM.Data.Observable
 import           Cardano.BM.Data.Severity
 import           Cardano.BM.Data.SubTrace
 import           Cardano.BM.Trace
+import           Cardano.Tracing.Peer (Peer (..), ppPeer)
 
 import           Cardano.Config.Types
 import           Cardano.Config.GitRev (gitRev)
-import           Cardano.Slotting.Slot (unSlotNo)
-import qualified Ouroboros.Network.AnchoredFragment as Net
-import qualified Ouroboros.Network.Block as Net
-import           Ouroboros.Network.NodeToClient (LocalConnectionId)
-import           Ouroboros.Network.NodeToNode (RemoteConnectionId)
-import           Ouroboros.Consensus.Block (GetHeader(..))
-import           Ouroboros.Consensus.Ledger.Extended (ExtLedgerState)
-import           Ouroboros.Consensus.Node (NodeKernel(..), remoteAddress)
-import           Ouroboros.Consensus.Mempool.API (MempoolCapacityBytes (..), getCapacity)
 import           Ouroboros.Consensus.NodeId (CoreNodeId (..), NodeId (..))
-import           Ouroboros.Consensus.Storage.ChainDB.API (getCurrentLedger)
 
-import qualified Ouroboros.Network.BlockFetch.ClientState as Net
-import qualified Ouroboros.Network.BlockFetch.ClientRegistry as Net
 import           Paths_cardano_node (version)
 
 import           Text.Printf (printf)
@@ -144,6 +131,9 @@ data LiveViewState blk a = LiveViewState
     , lvsBlockNum             :: !Word64
     , lvsChainDensity         :: !Double
     , lvsBlocksMinted         :: !Word64
+    , lvsLeaderNum            :: !Word64
+    , lvsSlotsMissedNum       :: !Word64
+    , lvsCreatedForksNum      :: !Word64
     , lvsTransactions         :: !Word64
     , lvsPeersConnected       :: !Word64
     , lvsMempool              :: !Word64
@@ -185,30 +175,9 @@ data LiveViewState blk a = LiveViewState
     , lvsMetricsThread        :: !LiveViewThread
     , lvsNodeThread           :: !LiveViewThread
 
-    , lvsNodeKernel           :: !(SMaybe (LVNodeKernel blk))
-    , lvsPeersLastTime        :: !Word64
-    , lvsPeers                :: [LVPeer blk]
+    , lvsPeers                :: [Peer blk]
     , lvsColorTheme           :: !ColorTheme
     } deriving (Generic, NFData, NoUnexpectedThunks)
-
-data SMaybe a
-  = SNothing
-  | SJust !a
-  deriving (Foldable, Functor, Generic, NFData, NoUnexpectedThunks, Traversable)
-
-fromSMaybe :: a -> SMaybe a -> a
-fromSMaybe x SNothing = x
-fromSMaybe _ (SJust x) = x
-
-data LVNodeKernel blk = LVNodeKernel
-  { getNodeKernel :: !(NodeKernel IO RemoteConnectionId LocalConnectionId blk) }
-  deriving (Generic)
-
-instance NoUnexpectedThunks (LVNodeKernel blk) where
-    whnfNoUnexpectedThunks _ _ = pure NoUnexpectedThunks
-
-instance NFData (LVNodeKernel blk) where
-    rnf _ = ()
 
 -- | Type wrapper to simplify derivations.
 newtype LiveViewThread = LiveViewThread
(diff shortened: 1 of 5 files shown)


cabal.project
```diff
@@ -417,11 +417,12 @@ source-repository-package
...
-  hedgehog >= 1.0,
-  bimap >= 0.4.0,
-  brick >= 0.47,
-  libsystemd-journal >= 1.4.4
+    ip < 1.5
+  , hedgehog >= 1.0
+  , bimap >= 0.4.0
+  , brick >= 0.47
+  , libsystemd-journal >= 1.4.4
+  , systemd >= 2.3.0
```
erikd, line 425: Does this branch need to be rebased? I see to remember doing exactly this just recently.

cardano-node/src/Cardano/Node/TUI/LiveView.hs
```diff
@@ -118,7 +105,7 @@ instance NoUnexpectedThunks a => NoUnexpectedThunks (LiveViewBackend blk a) wher
     return result
 
 type LiveViewMVar blk a = MVar (LiveViewState blk a)
-newtype LiveViewBackend blk a = LiveViewBackend { getbe :: LiveViewMVar blk a }
+data LiveViewBackend blk a = LiveViewBackend { getbe :: LiveViewMVar blk a }
```
erikd, line 108: Why?

cardano-node/src/Cardano/Node/TUI/LiveView.hs
```diff
@@ -414,6 +383,33 @@ instance IsEffectuator (LiveViewBackend blk) Text where
...
+
+                                return $ lvs { lvsSlotsMissedNum = fromIntegral missedSlotsNum }
+
+                    LogValue "forksCreatedNum" (PureI createdForksNum) ->
+                        modifyMVar_ (getbe lvbe) $ \lvs -> do
+
+                                checkForUnexpectedThunks ["forksCreatedNum LiveViewBackend"] lvs
+
+                                return $ lvs { lvsCreatedForksNum = fromIntegral createdForksNum }
+
```
erikd, line 412: This chunk of 3 lines is repeated over and over again. Maybe it should be a function.

cardano-node/src/Cardano/Node/TUI/LiveView.hs
```diff
@@ -615,24 +585,14 @@ setNodeThread lvbe nodeThr =
...
-              , lvsMempoolCapacityBytes = fromIntegral mempoolCapacity
-              }
+storePeersInLiveView :: NFData a => [LVPeer blk] -> LiveViewBackend blk a -> IO ()
+storePeersInLiveView peers lvbe = modifyMVar_ (getbe lvbe) $ \lvs -> return $ lvs { lvsPeers = peers }
+
+storeMempoolDataInLiveView :: NFData a => Word64 -> Word64 -> LiveViewBackend blk a -> IO ()
+storeMempoolDataInLiveView capacity capacityBytes lvbe = modifyMVar_ (getbe lvbe) $ \lvs ->
+  return $ lvs { lvsMempoolCapacity = capacity
+               , lvsMempoolCapacityBytes = capacityBytes
+               }
```
erikd, line 595: This reads less well than:
```
storeMempoolDataInLiveView capacity capacityBytes lvbe =
  modifyMVar_ (getbe lvbe) $ \lvs ->
    return $ lvs { lvsMempoolCapacity = capacity
                 , lvsMempoolCapacityBytes = capacityBytes
                 } 
```
Also `return` is going to be replaced with `pure` and `pure` is a better name for what it does. Maybe we should start using `pure` instead.

cardano-node/src/Cardano/Node/TUI/LiveView.hs
```diff
@@ -185,31 +175,10 @@ data LiveViewState blk a = LiveViewState
     , lvsMetricsThread        :: !LiveViewThread
     , lvsNodeThread           :: !LiveViewThread
 
-    , lvsNodeKernel           :: !(SMaybe (LVNodeKernel blk))
-    , lvsPeersLastTime        :: !Word64
     , lvsPeers                :: [LVPeer blk]
     , lvsColorTheme           :: !ColorTheme
     } deriving (Generic, NFData, NoUnexpectedThunks)
```
mrBliss, line 180: How large is this record (like C's `sizeof`, not including the things it points to)? Each time you modify it, you make a copy of the *entire record*. Depending on how often it is updated, this might be putting extra pressure on the GC.

You could group related values into sub records (e.g, if all the disk usage fields are read together, group them in a single record), then the number of bytes to copy shrinks. For example, grouping 10 `Word64` in one record will shrink the size of `LiveViewState` by 9 * 8 bytes.

cardano-node/src/Cardano/Node/TUI/LiveView.hs
denisshevchenko: I've experimented with additional fields, just forgot to return `newtype`. Fixed.

cardano-node/src/Cardano/Node/TUI/LiveView.hs
denisshevchenko: Done.

cabal.project
denisshevchenko: Done.

cardano-node/src/Cardano/Node/TUI/LiveView.hs
denisshevchenko: > You could group related values into sub records..

Agree. It will be done in another PR.

cardano-node/src/Cardano/Tracing/Peer.hs
```diff
@@ -0,0 +1,209 @@
...
+{-# LANGUAGE FlexibleInstances     #-}
+{-# LANGUAGE LambdaCase            #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE NamedFieldPuns        #-}
+{-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE ScopedTypeVariables   #-}
+{-# LANGUAGE TupleSections         #-}
+{-# LANGUAGE UndecidableInstances  #-}
+
+{-# OPTIONS_GHC -Wno-orphans  #-}
```
erikd, line 15: This should not be needed. All the instances in this file seem to be for types defined in this file.

cardano-node/src/Cardano/Tracing/Peer.hs
```diff
@@ -0,0 +1,209 @@
...
+import           Cardano.Prelude hiding (atomically)
+import           Prelude (String)
+
+import qualified Control.Monad.Class.MonadSTM.Strict as STM
+
+import           Data.Aeson (ToJSON(..), toJSON, Value (..), (.=))
+import           Data.IORef (IORef, readIORef, writeIORef)
+import qualified Data.Map.Strict as Map
+import qualified Data.Set as Set
+import           Data.Text (Text, pack)
```
erikd, line 36: Please only import `Text` unqualifed. and use everything else qualified.

cardano-node/src/Cardano/Tracing/Peer.hs
```diff
@@ -0,0 +1,209 @@
...
+{-# LANGUAGE DeriveTraversable     #-}
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE FlexibleInstances     #-}
+{-# LANGUAGE LambdaCase            #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE NamedFieldPuns        #-}
+{-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE ScopedTypeVariables   #-}
+{-# LANGUAGE TupleSections         #-}
+{-# LANGUAGE UndecidableInstances  #-}
```
erikd, line 13: Only 6 of these language pragmas are actually needed. Please remove the unneeded ones.

cardano-node/src/Cardano/Tracing/Peer.hs
```diff
@@ -0,0 +1,209 @@
...
+data SMaybe a
+  = SNothing
+  | SJust !a
+  deriving (Foldable, Functor, Generic, NFData, NoUnexpectedThunks, Traversable)
+
+fromSMaybe :: a -> SMaybe a -> a
+fromSMaybe x SNothing = x
+fromSMaybe _ (SJust x) = x
+
+data LVNodeKernel blk = LVNodeKernel
```
erikd, line 112: `newtype`

cardano-node/src/Cardano/Tracing/Peer.hs
```diff
@@ -0,0 +1,209 @@
...
+instance NoUnexpectedThunks (LVNodeKernel blk) where
+    whnfNoUnexpectedThunks _ _ = pure NoUnexpectedThunks
+
+instance NFData (LVNodeKernel blk) where
+    rnf _ = ()
+
+data NodeKernelData blk = NodeKernelData
+  { nkdMempoolCapacity      :: !Word64
+  , nkdMempoolCapacityBytes :: !Word64
+  , nkdKernel               :: !(SMaybe (LVNodeKernel blk))
```
erikd, line 125: Redundant parens.

cardano-node/src/Cardano/Tracing/Peer.hs
```diff
@@ -0,0 +1,209 @@
...
+
+-- | Instances for convertin [Peer blk] to Object.
+
+instance ToObject [Peer blk] where
+  toObject MinimalVerbosity _ = emptyObject
+  toObject _ [] = emptyObject
+  toObject verb xs = mkObject
+    [ "kind"  .= String "NodeKernelPeers"
+    , "peers" .= toJSON
+      (foldl' (\acc x -> toObject verb x : acc) [] xs) ]
```
erikd, line 199: Here, the closing sqaure bracket is at the end of the line, but 10 lines below, it is on a new line.

Consistency please. Personally I prefer the closing square bracket on a new line like below. Its easier to read.

cardano-node/src/Cardano/Tracing/Peer.hs
```diff
@@ -0,0 +1,209 @@
...
+    peerStates <- fmap tuple3pop <$> (   STM.atomically
+                                       . (>>= traverse readFetchClientState)
+                                       . Net.readFetchClientsStateVars
+                                       . getFetchClientRegistry $ kernel
+                                     )
+    candidates <- STM.atomically . getCandidates . getNodeCandidates $ kernel
+
+    pure $ Map.elems . flip Map.mapMaybeWithKey candidates $
+      \cid af -> Map.lookup cid peerStates <&>
+      \(status, inflight) -> Peer cid af status inflight
```
erikd, line 177: Surely there is a simpler more readable way of doing this.

cardano-node/src/Cardano/Tracing/Peer.hs
```diff
@@ -0,0 +1,209 @@
...
+    rnf _ = ()
+
+data NodeKernelData blk = NodeKernelData
+  { nkdMempoolCapacity      :: !Word64
+  , nkdMempoolCapacityBytes :: !Word64
+  , nkdKernel               :: !(SMaybe (LVNodeKernel blk))
+  }
+
+initialNodeKernelData :: NodeKernelData blk
+initialNodeKernelData = NodeKernelData 0 0 (SNothing)
```
erikd, line 129: Redundant parens.

denisshevchenko:
@erikd Thanks for suggestions! Fixed.

cardano-node/src/Cardano/Tracing/Peer.hs
```diff
@@ -0,0 +1,203 @@
+{-# LANGUAGE ConstraintKinds       #-}
```
erikd, line 1: You do not need `ConstraintKinds` in this code.

cardano-node/src/Cardano/Tracing/Peer.hs
```diff
@@ -0,0 +1,203 @@
...
+import           Cardano.Prelude hiding (atomically)
+import           Prelude (String)
+
+import qualified Control.Monad.Class.MonadSTM.Strict as STM
+
+import           Data.Aeson (ToJSON(..), toJSON, Value (..), (.=))
+import           Data.IORef (IORef, readIORef, writeIORef)
+import qualified Data.Map.Strict as Map
+import qualified Data.Set as Set
+import qualified Data.Text as T
```
erikd, line 27: Better to have `import qualified Data.Text as Text`.

cardano-node/src/Cardano/Tracing/Peer.hs
denisshevchenko: Fixed.

cardano-node/src/Cardano/Tracing/Peer.hs
denisshevchenko: Fixed.

PR #1016, opened 2020-05-22
cli: Add 'shelley query ledger-state' command

erikd:


cardano-api/src/Cardano/Api/LocalStateQuery.hs
```diff
@@ -149,6 +150,24 @@ queryStakeDistributionFromLocalState network socketPath point = do
       socketPath
       pointAndQuery
 
+queryLocalLedgerState
+  :: blk ~ ShelleyBlock TPraosStandardCrypto
+  => Network
+  -> SocketPath
+  -> Point blk
+  -> ExceptT LocalStateQueryError IO LByteString
```
erikd, line 158: Should the return type here be `LByteString` or something else?

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -155,27 +192,27 @@ runQueryStakeDistribution network mOutFile = do
   let ptclClientInfo = pClientInfoCodecConfig . protocolClientInfo $ mkNodeClientProtocolTPraos
   tip <- liftIO $ withIOManager $ \iomgr ->
     getLocalTip iomgr ptclClientInfo network sockPath
-  stakeDistr <- firstExceptT NodeLocalStateQueryError $
+  stakeDist <- firstExceptT NodeLocalStateQueryError $
```
erikd, line 195: These changes were required to avoid name shadowing warnings.

cardano-api/src/Cardano/Api/LocalStateQuery.hs
intricate: Why can't we get a `LedgerState` here?

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE ScopedTypeVariables #-}
```
intricate, line 2: Is this being used? If not:

```suggestion
```

cardano-api/src/Cardano/Api/LocalStateQuery.hs
intricate: Okay, I just read the documentation here:

https://github.com/input-output-hk/ouroboros-network/blob/9d9754c9ddcfff82b27c371a545aa4680d86d996/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs#L320-L333

Should we attempt the `decodeFull` here and return the `LedgerState` on success?

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
erikd: Yes, use it in `Left (lstate :: Ledger ....)` somewhere below.

erikd:
```
  case decodeFull lbs of
    Right (lstate :: Ledger.LedgerState TPraosStandardCrypto) ->
```

cardano-api/src/Cardano/Api/LocalStateQuery.hs
erikd: We need to be able to return the decoded `Ledger.LedgerState TPraosStandardCrypto` if we can and if that fails return the `LByteString`. I could do that with an `Either` but :shrug: .

erikd: Changed the type of that so its:
```
ExceptT LocalStateQueryError IO (Either LByteString (Ledger.LedgerState TPraosStandardCrypto))
```
which I think is a minor improvement.

erikd: > Should we attempt the decodeFull here and return the LedgerState on success?

If decide fails we want to do a generic CBOR decode.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
erikd: That was removed when I moved the decode logic to the `LocalStateQuery` module.

cardano-api/src/Cardano/Api/LocalStateQuery.hs
```diff
@@ -149,6 +151,29 @@ queryStakeDistributionFromLocalState network socketPath point = do
       socketPath
       pointAndQuery
 
+queryLocalLedgerState
+  :: blk ~ ShelleyBlock TPraosStandardCrypto
+  => Network
+  -> SocketPath
+  -> Point blk
+  -> ExceptT LocalStateQueryError IO (Either LByteString (Ledger.LedgerState TPraosStandardCrypto))
```
Jimbo4350, line 159: Shouldn't we move that `LByteString` error into `LocalStateQueryError`?

erikd: No, these really are two different things. If CBOR decoding is successful, the return the `LedgerState`. If decoding fails, return the ByteString.

Jimbo4350: Understood now: https://github.com/input-output-hk/cardano-node/pull/1016/files/ba755abfd4b9f4f467629aba4a577dcd59574e08#diff-b548d89cce82358cb320307aed52a300R135

PR #1022, opened 2020-05-22
Update BadInputsUTxO & ValueNotConserved logging errors

Jimbo4350:


cardano-config/src/Cardano/TracingOrphanInstances/Shelley.hs
```diff
@@ -215,7 +218,8 @@ instance Crypto c => ToObject (PredicateFailure (UTXOW c)) where
 instance Crypto c => ToObject (PredicateFailure (UTXO c)) where
   toObject _verb (BadInputsUTxO badInputs) =
     mkObject [ "kind" .= String "BadInputsUTxO"
-             , "badInputs" .= badInputs
+             , "error" .= renderBadInputsUTxOErr badInputs
+
```
intricate, line 222:
```suggestion
```

cardano-config/src/Cardano/TracingOrphanInstances/Shelley.hs
```diff
@@ -241,9 +245,25 @@ instance Crypto c => ToObject (PredicateFailure (UTXO c)) where
...
+  | Set.null txIns = String "There are no transaction inputs in this transaction."
+  | otherwise = String $ "These transaction inputs do not exist in the UTxO set: " <> unwrapTxIns txIns
+  where
+    unwrapTxIns :: Set (TxIn c) -> Text
+    unwrapTxIns badTxins = textShow . Set.toList $ Set.map (\(TxIn txId' index) -> (txId', index)) badTxins
+
+renderValueNotConservedErr :: Coin -> Coin -> Value
+renderValueNotConservedErr consumed produced
+  | consumed > produced = String "This transaction has consumed more ADA than it has produced."
+  | consumed < produced = String "This transaction has produced more ADA than it has consumed."
```
intricate, line 264: I think we should use the term "Lovelace" here:

```suggestion
  | consumed > produced = String "This transaction has consumed more Lovelace than it has produced."
  | consumed < produced = String "This transaction has produced more Lovelace than it has consumed."
```

PR #1069, opened 2020-05-26
Add CLI command to get Pool Id

### LINKED ISSUE #1018: [FR] - add cli command to return the pool-id given the verification key
After creating and registering a stake pool, we need to know its pool-id.

kantp:
This adds a command cardano-cli shelley node pool-id, that takes the cold
verification key, and produces the stake pool id, by hashing it.

Closes #1018


kantp:
Moved the command to the `stake-pool` subcommand, specifically `cardano-cli shelley stake-pool id`

PR #1073, opened 2020-05-27
Hedgehog testing of pioneers exercise 1

Jimbo4350:
Tests the building of payment and stake addresses as described in [pioneers exercise 1](https://github.com/input-output-hk/cardano-tutorials/blob/master/pioneers-testnet/pioneers-exercise-1.md). The test failures leverage `optparse-applicative` errors.

### DIFF
--- a/cardano-cli/cardano-cli.cabal
+++ b/cardano-cli/cardano-cli.cabal
@@ -131,7 +131,6 @@ executable cardano-cli
                      , optparse-applicative
                      , text
                      , transformers-except
-
   default-extensions:  NoImplicitPrelude
 
 test-suite cardano-cli-test
@@ -151,6 +150,34 @@ test-suite cardano-cli-test
                       , text
                       , text-ansi
                       , unix
+
+  default-language:     Haskell2010
+  default-extensions:   NoImplicitPrelude
+
+  ghc-options:          -Wall
+                        -Wincomplete-record-updates
+                        -Wincomplete-uni-patterns
+                        -Wredundant-constraints
+                        -Wpartial-fields
+                        -Wcompat
+                        -threaded -rtsopts -with-rtsopts=-N -with-rtsopts=-T
+
+test-suite cardano-cli-pioneers
+  hs-source-dirs:       test
+  main-is:              cardano-cli-pioneers.hs
+  type:                 exitcode-stdio-1.0
+
+  build-depends:        base
+                      , cardano-cli
+                      , cardano-prelude
+                      , directory
+                      , hedgehog
+                      , optparse-applicative
+                      , text
+                      , transformers-except
+
+  other-modules:        Test.OptParse
+                        Test.Pioneers.Exercise1
 
   default-language:     Haskell2010
   default-extensions:   NoImplicitPrelude
--- a/cardano-cli/app/cardano-cli.hs
+++ b/cardano-cli/app/cardano-cli.hs
@@ -4,10 +4,9 @@ import           Cardano.Prelude hiding (option)
 
 import           Control.Monad.Trans.Except.Exit (orDie)
 import qualified Options.Applicative as Opt
-import           Options.Applicative (ParserInfo, ParserPrefs, showHelpOnEmpty)
 
-import           Cardano.CLI.Parsers (parseClientCommand)
-import           Cardano.CLI.Run (ClientCommand, renderClientCommandError, runClientCommand)
+import           Cardano.CLI.Parsers (opts, pref)
+import           Cardano.CLI.Run (renderClientCommandError, runClientCommand)
 import           Cardano.Config.TopHandler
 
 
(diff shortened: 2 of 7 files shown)


cardano-cli/test/Test/Scenarios/OptParse.hs
```diff
@@ -0,0 +1,87 @@
...
+import qualified Hedgehog as H
+import           Hedgehog.Internal.Property (failWith)
+
+-- | Purely evalutate the cardano-cli parser.
+-- e.g evalCardanoCLIParser ["shelley"] would be equivalent to cabal exec cardano-cli shelley
+-- without running underlying IO.
+evalCardanoCLIParser :: [String] -> Opt.ParserResult ClientCommand
+evalCardanoCLIParser args = Opt.execParserPure pref opts args
+
+-- | This takes a `ParserResult`, which is pure, and executes it.
```
intricate, line 32:
```suggestion
-- | This takes a 'ParserResult', which is pure, and executes it.
```

cardano-cli/test/Test/Scenarios/ShelleyAddressBuild.hs
```diff
@@ -0,0 +1,60 @@
...
+
+    -- Build shelley address
+    execCardanoCLIParser
+      []
+      "Failure: prop_shelleyAddressBuild.build_address"
+        $ evalCardanoCLIParser [ "shelley", "address", "build"
+                               , "--payment-verification-key-file", addrVerKey
+                               ]
+
+    liftIO $ fileCleanup [addrVerKey, addrSignKey]
```
intricate, line 52: Should we still attempt a file clean up in the failure case?

n.b. I see that `execCardanoCLIParser` will clean up in the failure case, but this is in reference to the `H.assert`s above.

cardano-cli/test/Test/Scenarios/ShelleyAddressBuild.hs
```diff
@@ -0,0 +1,60 @@
...
+
+
+-- | This property checks the scenario where we would like to
+-- build a shelley address.
+-- 1. We generate a key pair
+-- 2. We check for the existence of the key pair
+-- 3. We use the generated verification key to build a shelley address.
+prop_shelleyAddressBuild :: Property
+prop_shelleyAddressBuild =
+  H.withTests 1 . H.property $ do
```
intricate, line 23: Why don't we run this more than once? Seems like it'd work fine.

cardano-cli/test/cardano-cli-test.hs
```diff
@@ -22,8 +22,10 @@ import           System.Process (rawSystem)
 import           System.IO (BufferMode (..))
 import qualified System.IO as IO
 
+
```
intricate, line 25: A bit of `diff` noise in this file :smile: 

```suggestion
```

cardano-cli/test/cardano-cli-test.hs
```diff
@@ -22,8 +22,10 @@ import           System.Process (rawSystem)
...
 main :: IO ()
 main = do
+
```
intricate, line 28:
```suggestion
```

cardano-cli/cardano-cli.cabal
```diff
@@ -162,3 +161,34 @@ test-suite cardano-cli-test
...
                         -Wcompat
                         -threaded -rtsopts -with-rtsopts=-N -with-rtsopts=-T
+
+test-suite cardano-cli-scenarios
+  hs-source-dirs:       test
+  main-is:              cardano-cli-scenarios.hs
+  type:                 exitcode-stdio-1.0
+
+  if os(windows)
+    buildable: False
```
intricate, line 171: Is it true that this isn't buildable on Windows? I don't think we're doing anything *nix specific here, right?

cardano-cli/test/Test/Pioneers/Exercise1/BuildShelleyPaymentAddress.hs
```diff
@@ -0,0 +1,56 @@
...
+
+
+-- | This property checks the scenario where we would like to
+-- build a shelley address.
+-- 1. We generate a key pair
+-- 2. We check for the existence of the key pair
+-- 3. We use the generated verification key to build a shelley address.
+prop_buildShelleyPaymentAddress :: Property
+prop_buildShelleyPaymentAddress =
+  H.withTests 1 . H.property $ do
```
erikd, line 22: Apart from `H.withTests` and `H.property` what are we using from HH?

Could this not just be a function in `IO Bool` or similar? Ie returns `True` on success.

Jimbo4350: I was planning to use `Group` to group the exercises as we go along and `hedgehog` helps with the error rendering. If you are opposed to using `hedgehog` for these I can remove its usage.

erikd: I was not so much opposed to HH as wondering why. Nice error rendering actually makes sense.

So it you want it for the error rendering you should pull out a function like:

```
propertyOnce =  H.withTests 1 . H.property
```
and then use that everywhere.

Jimbo4350:
bors merge

PR #1076, opened 2020-05-27
CAD-1085: Port in TUI, bug fix.

denisshevchenko:
PR fixes a small bug with `Port` in TUI: no more `Just`.

### DIFF
--- a/cardano-node/src/Cardano/Node/TUI/LiveView.hs
+++ b/cardano-node/src/Cardano/Node/TUI/LiveView.hs
@@ -570,7 +570,10 @@ liveViewPostSetup lvbe ncli nc = do
     nodeId =
       case nodeMode ncli of
         MockProtocolMode -> namenum (ncNodeId nc)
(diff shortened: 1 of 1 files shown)


cardano-node/src/Cardano/Node/TUI/LiveView.hs
```diff
@@ -570,7 +570,10 @@ liveViewPostSetup lvbe ncli nc = do
     nodeId =
       case nodeMode ncli of
         MockProtocolMode -> namenum (ncNodeId nc)
-        RealProtocolMode -> Text.pack $ "Port: " <> show (naPort <$> nodeAddr ncli)
+        RealProtocolMode -> Text.pack $ "Port: " <>
+          case naPort <$> nodeAddr ncli of
+            Nothing -> "-"
```
intricate, line 575: Is this what we want to display in the `Nothing` case? Or is there another means by which we can identify the node?

denisshevchenko: > Or is there another means by which we can identify the node?

I don't know it.

PR #1091, opened 2020-05-29
CAD-1132 trace block size in AdoptedBlock event

CodiePP:
added the block's size in the traced event AdoptedBlock.
this simplifies analysis of benchmarking results.
the size is available from the block's header. Thanks @mrBliss  for pointing this out!

### DIFF
--- a/cardano-config/src/Cardano/TracingOrphanInstances/Consensus.hs
+++ b/cardano-config/src/Cardano/TracingOrphanInstances/Consensus.hs
@@ -22,14 +22,14 @@ import           Cardano.TracingOrphanInstances.Common
 import           Cardano.TracingOrphanInstances.Network (showTip, showPoint)
 
 import           Ouroboros.Consensus.Block
-                   (BlockProtocol, Header, headerPoint,
+                   (BlockProtocol, Header, getHeader, headerPoint,
                     RealPoint, realPointSlot, realPointHash)
 import           Ouroboros.Consensus.MiniProtocol.BlockFetch.Server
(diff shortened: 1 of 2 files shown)


cardano-config/src/Cardano/TracingOrphanInstances/Consensus.hs
```diff
@@ -745,6 +747,7 @@ instance ToObject MempoolSize where
 instance ( Condense (HeaderHash blk)
          , HasTxId tx
          , LedgerSupportsProtocol blk
```
mrBliss, line 749: Now this constraint is redundant, it is implied by `RunNode`.
```suggestion
```

cardano-config/src/Cardano/TracingOrphanInstances/Consensus.hs
```diff
@@ -218,6 +219,7 @@ showT = pack . show
...
```
mrBliss, line 221: Now this constraint is redundant, it is implied by `RunNode`.
```suggestion
```

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -244,8 +244,8 @@ initialBlockchainCounters = BlockchainCounters 0 0 0 0 0
 mkTracers
   :: forall peer localPeer blk.
      ( LedgerSupportsProtocol blk
```
mrBliss, line 246:
```suggestion
```
Now this constraint is redundant, it is implied by `RunNode`.

PR #1101, opened 2020-05-30
Refactor Cardano.Tracing.Tracers

Jimbo4350:
- First PR towards making `Cardano.Tracing.Tracers` more easily understandable & explicit.
- All constraints are now explicit,  many functions from within `where` clauses are moved to the top level & given type signatures.
-  `TraceConfig'` is simplified and partial function free.

### DIFF
--- a/cardano-config/src/Cardano/TracingOrphanInstances/Network.hs
+++ b/cardano-config/src/Cardano/TracingOrphanInstances/Network.hs
@@ -35,6 +35,12 @@ import           Ouroboros.Network.NodeToNode
                    (WithAddr(..), ErrorPolicyTrace(..), TraceSendRecv (..))
 import           Ouroboros.Network.Protocol.BlockFetch.Type
                    (BlockFetch, Message(..))
+import           Ouroboros.Network.Protocol.ChainSync.Type (ChainSync)
+import qualified Ouroboros.Network.Protocol.ChainSync.Type as ChainSync
+import           Ouroboros.Network.Protocol.LocalStateQuery.Type (LocalStateQuery)
+import qualified Ouroboros.Network.Protocol.LocalStateQuery.Type as LocalStateQuery
+import           Ouroboros.Network.Protocol.LocalTxSubmission.Type (LocalTxSubmission)
+import qualified Ouroboros.Network.Protocol.LocalTxSubmission.Type as LocalTxSub
 import           Ouroboros.Network.Protocol.TxSubmission.Type
                    (Message (..), TxSubmission)
 import           Ouroboros.Network.Snocket (LocalAddress (..))
@@ -49,8 +55,6 @@ import           Ouroboros.Network.TxSubmission.Outbound
 -- We do need some consensus imports to provide useful trace messages for some
 -- network protocols
 import           Ouroboros.Consensus.Util.Condense (Condense, condense)
-import           Ouroboros.Consensus.Ledger.SupportsMempool (GenTx, HasTxId, HasTxs(..),
-                   TxId, txId)
 
 
 showTip :: Condense (HeaderHash blk)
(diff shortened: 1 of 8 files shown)


cardano-config/src/Cardano/TracingOrphanInstances/Network.hs
```diff
@@ -382,22 +386,10 @@ instance (Show peer)
...
-         , Condense (TxId (GenTx blk))
-         , HasHeader blk
-         , HasTxs blk
-         , HasTxId (GenTx blk)
-         )
+instance ( Condense (HeaderHash blk))
       => ToObject (AnyMessage (BlockFetch blk)) where
-  toObject _verb (AnyMessage (MsgBlock blk)) =
-    mkObject
-      [ "kind" .= String "MsgBlock"
```
Jimbo4350, line 394: I know information was removed here however, the `HasHeader blk` constraint also applies for serialised blocks. It does not make sense to have a `HasHeader` constraint for a serialized block. I'm not entirely sure how we can get around this.

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -273,417 +278,479 @@ mkTracers traceConf tracer bcCounters = do
...
+      traceI :: Integral a => Text -> a -> IO ()
+      traceI msg i = traceNamedObject tr' (meta, LogValue msg (PureI (fromIntegral i)))
+
+  traceD "density"     (fromRational density)
+  traceI "slotNum"     slots
+  traceI "blockNum"    blocks
+  traceI "slotInEpoch" (slots `rem` epochSlots)
+  traceI "epoch"       (slots `div` epochSlots)
+ where
+   epochSlots :: Word64 = 21600  -- TODO
```
mrBliss, line 370: Please rebase after https://github.com/input-output-hk/cardano-node/pull/1108 is merged.

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -146,14 +148,19 @@ runNode loggingLayer npm@NodeCLI{protocolFiles} = do
...
                      , ("TraceChainDb",traceChainDB)
                      , ("TraceForge",traceForge)
                      , ("TraceMempool",traceMempool)
                      ]
-            trsinactive = filter (\(_,f) -> not $ traceEnabled (ncTraceConfig nc) f) reqtrs
-        unless (List.null trsinactive) $ do
+
+        trsinactive <- case ncTraceConfig nc of
+                         TracingOff -> return []
+                         TracingOn traceConf -> return $ map (\(tName,f) -> (tName, f traceConf)) reqtrs
```
CodiePP, line 161: previously, the list only included those tracers which are turned off.

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -102,17 +100,21 @@ runNode
...
-                           MinimalVerbosity -> "minimal"
-                           MaximalVerbosity -> "maximal"
+    case ncTraceConfig nc of
+      TracingOff -> return ()
+      TracingOn traceConf ->
+        case traceVerbosity traceConf of
+          NormalVerbosity -> traceWith tracer $ "tracing verbosity = normal verbosity "
+          MinimalVerbosity -> traceWith tracer $ "tracing verbosity = minimal verbosity "
+          MaximalVerbosity -> traceWith tracer $ "tracing verbosity = maximal verbosity "
+
```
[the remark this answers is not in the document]
Jimbo4350: Agreed that is more concise. Let me circle back in a follow up PR as I have an idea on how to improve on the above.

cardano-node/src/Cardano/Node/Run.hs
Jimbo4350: My most recent change should now function as before. I'd like to also circle back and propose some further changes here.

Issue #1063, opened 2020-05-25
[BUG] - pool margin value not shown correctly in ledger-state


bergr01:
cardano-cli / cardano-node version 1.12.0 on master.

Registered a new stake pool on FnF network - specified non-zero margin (e.g. 0.75) when creating stake pool registration certificate. Followed the rest of the steps, submitted the transaction - pool record was created in ledger-state output (seen via cardano-cli shelley query ledger-state --testnet-magic 42 command). 

Expected to see poolMargin field with the same value as was specified during pool creation - instead, poolMargin is listed as 0 for all stake pools.

onyxstakepool:
@dcoutts The precision epsilon of the rational approximation was set to 1 in the parser. 
Thus for most x in [0, 1] the best rational approximation of x in the interval [x - 1, x + 1] is 0/1.

This is fixed by setting epsilon equal to 0 and thus enforcing a machine precision rational approximation. 

```diff
diff --git a/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs b/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
index 2caff9dd..fd4302a7 100644
--- a/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
+++ b/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
@@ -1112,7 +1112,7 @@ pPoolCost =
 
 pPoolMargin :: Parser ShelleyStakePoolMargin
 pPoolMargin =
-  (\dbl -> maybeOrFail . Shelley.mkUnitInterval $ approxRational (dbl :: Double) 1) <$>
+  (\dbl -> maybeOrFail . Shelley.mkUnitInterval $ approxRational (dbl :: Double) 0) <$>
     Opt.option Opt.auto
       (  Opt.long "pool-margin"
       <> Opt.metavar "DOUBLE"
```

onyxstakepool:
@dcoutts I have tested the fix with the pool margin set to the irrational Euler constant gamma = 0.57721 56649 01532 86060... The parsed pool margin approximates the Euler constant up to 15 digits as expected for double precision numbers.

PR #1102, opened 2020-05-31
Pool margin fix

onyxstakepool:
Proposed fix for issue #1063


erikd:
Pull this into #1107 so it can go thru CI.

erikd:
No its own IOHK PR so it will go through CI #1110 .

PR #1135, opened 2020-06-03
SRE-100 musl64 and macos release builds

jbgi:
And other improvements along the way.


disassembler:
bors try

jbgi:
binaries:
 - linux: https://hydra.iohk.io/build/2870638
 - macos: https://hydra.iohk.io/build/2877117
 - win64: https://hydra.iohk.io/build/2870639

PR #1137, opened 2020-06-03
shelley_qa: new network

disassembler:


erikd:
Checked out this branch an got:
```
trace: Using IOHK default nixpkgs
error: attribute 'qa' in selection path 'scripts.qa.node' not found
```

erikd:
Needs to be `nix-build -A scripts.shelley_qa.node -o ....`

bors r+

PR #1142, opened 2020-06-03
cli: Tweak the cli test genesis.spec.json file

erikd:
Makes it more realistic.


erikd:
Sorry, this is wrong.

PR #1192, opened 2020-06-05
Restore configurability of traces

deepfire:
Restore configurability that a number of traces had lost:

- ChainDB
- DnsSubscription
- BlockFetchDecision
- ForgeState
- Forge
- Mempool

Add configurability for the BlockchainTime trace (off by default).

### DIFF
--- a/cardano-config/src/Cardano/Config/TraceConfig.hs
+++ b/cardano-config/src/Cardano/Config/TraceConfig.hs
@@ -30,6 +30,7 @@ data TraceSelection
 
   -- Per-trace toggles, alpha-sorted.
   , traceAcceptPolicy :: !Bool
+  , traceBlockchainTime :: !Bool
   , traceBlockFetchClient :: !Bool
   , traceBlockFetchDecisions :: !Bool
   , traceBlockFetchProtocol :: !Bool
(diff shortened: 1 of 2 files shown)


deepfire:
> the function `tracerOnOff` has hardcoded "MinimalVerbosity" and is not taking into account the verbosity level as set in the configuration (line 908 of Tracers.hs)
> would it be possible to also pass in the TracingVerbosity to this function?

Yes, that deviation of the original, obvious `tracerOnOff` semantic is very unwelcome.

I think that needs to be fixed, but it's probably out of scope for this PR.

cc @Jimbo4350

Jimbo4350:
> > the function `tracerOnOff` has hardcoded "MinimalVerbosity" and is not taking into account the verbosity level as set in the configuration (line 908 of Tracers.hs)
> > would it be possible to also pass in the TracingVerbosity to this function?
> 
> Yes, that deviation of the original, obvious `tracerOnOff` semantic (which used to match its name) is very confusing and counterproductive.
> 
> I think that needs to be fixed, but it's probably out of scope for this PR.
> 
> cc @Jimbo4350

The current `tracerOnOff` mechanism will remain as is for now. The recent refactor has made things significantly clearer and we have more refactoring planned for this module.

PR #1193, opened 2020-06-05
CAD-1192: Restore trace verbosity configuration

Jimbo4350:
The trace verbosity was hard-coded in a large recent refactor, this PR removes the hardcoded value and restores the ability to change the tracing verbosity via the config yaml file.

### DIFF
--- a/cardano-node/src/Cardano/Tracing/Tracers.hs
+++ b/cardano-node/src/Cardano/Tracing/Tracers.hs
@@ -290,23 +290,24 @@ mkTracers protoInfo traceConf' tracer bcCounters = do
 
   case traceConf' of
     TracingOn traceConf ->
-      pure Tracers
-        { chainDBTracer = tracerOnOff' (traceChainDB traceConf) $
-            annotateSeverity . teeTraceChainTip traceConf' bcCounters elidedChainDB $ appendName "ChainDB" tracer
-        , consensusTracers
-            = mkConsensusTracers' protoInfo elidedFetchDecision blockForgeOutcomeExtractor forgeTracers tracer (TracingOn traceConf) bcCounters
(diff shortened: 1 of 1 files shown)


Jimbo4350:
Waiting on #1192 to merge

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -901,12 +908,17 @@ traceCounter logValueName aCounter tracer = do
 
 tracerOnOff
   :: Transformable Text IO a
-  => Bool -> LoggerName -> Trace IO Text -> Tracer IO a
-tracerOnOff False _ _ = nullTracer
-tracerOnOff True name trcer = annotateSeverity
-                                $ toLogObject' MinimalVerbosity
+  => Bool -> TracingVerbosity -> LoggerName -> Trace IO Text -> Tracer IO a
```
CodiePP, line 911: :+1:

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -902,10 +908,10 @@ traceCounter logValueName aCounter tracer = do
 
 tracerOnOff
```
deepfire, line 909: Maybe the function should get another name, as it clearly no longer does exactly what it name implies?

Jimbo4350: Feel free to suggest alternative names :slightly_smiling_face:

CodiePP: that's only delaying this PR

Jimbo4350:
bors merge

PR #1203, opened 2020-06-08
Fix punc

wi-ski:


erikd:
Please squash those commits down into a single commit so the addition/deletion of the `ff-genesis.json` will disappear.

doc/shelley-genesis.md
```diff
@@ -57,7 +57,7 @@ should be created offline, kept offline and only the verification keys shared.
 
 For a demo we're going to put all the keys together in once place
 
-To start with we will set up our template directory:
+To start, with we will set up our template directory:
```
mrBliss, line 60: This doesn't improve things, IMO, "with we will set up ..."  is just wrong.

I'm not convinced this needs changing, but I would go for something like "We will start with setting up ...".

intricate: or this could be formatted as:

```suggestion
To start with, we will set up our template directory:
```

intricate:
Superseded by #1770

PR #1209, opened 2020-06-08
Add test for pioneers exercise 2

Jimbo4350:
- Improve error messages for address and tx out cli parsers

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
+++ b/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
@@ -949,7 +949,7 @@ pTxIn =
 
 pTxOut :: Parser TxOut
 pTxOut =
-  Opt.option (Opt.maybeReader (parseTxOut . Text.pack))
+  Opt.option (Opt.eitherReader (parseTxOut . Text.pack))
     (  Opt.long "tx-out"
     <> Opt.metavar "TX-OUT"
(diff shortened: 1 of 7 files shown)


cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -949,7 +949,7 @@ pTxIn =
 
 pTxOut :: Parser TxOut
 pTxOut =
-  Opt.option (Opt.maybeReader (parseTxOut . Text.pack))
+  Opt.option (Opt.eitherReader (\str -> parseTxOut $ Text.pack str))
```
erikd, line 952: You should be able to do `Opt.option (Opt.eitherReader (parseTxOut . Text.pack))`

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -1042,15 +1042,12 @@ pQueryFilter = pAddresses <|> pure NoFilter
 
 pFilterByHexEncodedAddress :: Parser Address
 pFilterByHexEncodedAddress =
-  Opt.option maybeHexEncodedAddressReader
+  Opt.option (Opt.eitherReader (\str -> first show . addressFromHex $ Text.pack str))
```
erikd, line 1045: Same.

Jimbo4350:
bors merge

PR #1258, opened 2020-06-14
Fix output of cardano-node --version command

AndrewWestberg:
cardano-node is outputting cardano-cli in its version. Likely a copy/pate error.


erikd:
Closing in favor of https://github.com/input-output-hk/cardano-node/pull/1263 .

PR #1277, opened 2020-06-15
Make genesis parser backwards compatible

kderme:
Uses https://github.com/input-output-hk/cardano-ledger-specs/pull/1553 which allows field `minPoolCost` to not exist


kderme:
not needed after https://github.com/input-output-hk/cardano-node/pull/1308

PR #1290, opened 2020-06-16
SRE-118 multi-node cluster test framework for QA.

jbgi:
The user-supplied test script is run on an edge node and can use `cardano-cli` and `cardano-node` in `PATH`, as well as the following environment variable :
- `CARDANO_NODE_SOCKET_PATH`
- `CONFIG_DIR` (containing genesis and all genesis and core nodes keys)
- `TOPOLOGY` (path to topology file with all other nodes of the cluster);

To run the test:
`nix build -f default.nix nixosTests.qaCluster`
(test to be duplicated with different arguments if needed).

### DIFF
--- a/cardano-cli/test/cli/core/common
+++ b/cardano-cli/test/cli/core/common
@@ -1,4 +1,4 @@
-#!/bin/bash -eu
+#!/usr/bin/env bash -eu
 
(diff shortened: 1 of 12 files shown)


jbgi:
replaced by #2026.

PR #1300, opened 2020-06-18
supervisord-cluster: initial setup

disassembler:
Adds a supervisord controlled local cluster that can be used standalone from cardano-node or as a script ran by daedalus. Also runs a basic webserver to serve metadata for the registered pools.

TODO:

- [x] change supervisord to run in background
- [x] build the tx, sign and submit to fund pool owners with pledge and register pools
- [x] integrate with devops shell so if network=local `start-cluster` command starts the cluster. Also set `CARDANO_NODE_SOCKET_PATH` to bft1.socket in that shell
- [x] map across bft and pool nodes to auto-generate supervisord services that work for any value of `bftNum` or `poolNum`.

Stretch Goals:
* start at `d=1` and do an update proposal that switches to `d=0.1`
* MIR certificate for ITN pubkey of wallet embedded in initialFunds to preload rewards
* register bft1 on-chain as a relay for both pools.
* add to daedalus and extend service config to run smash, postgres and db-sync

Usage:
```
nix-build -A cluster && ./result/bin/start-cluster
```


disassembler:
Usage:

nix/supervisord-cluster/default.nix
```diff
@@ -0,0 +1,283 @@
...
+      prefix = [];
+      modules = import ../nixos/module-list.nix ++ [ systemdCompat extra ];
+      args = { inherit pkgs; };
+    };
+  in pkgs.writeScript "cardano-node" ''
+    #!${pkgs.stdenv.shell}
+    ${eval.config.services.cardano-node.script}
+  '';
+  topologyFile = selfPort: {
+    Producers = map (p :
```
johnalotoski, line 38: p:

nix/supervisord-cluster/default.nix
```diff
@@ -0,0 +1,283 @@
...
+      {
+        addr = "127.0.0.1";
+        port = p;
+        valency = 1;
+      }) (lib.filter (p: p!= selfPort) (lib.genList (i: basePort + i + 1) (numBft + numPools)));
```
johnalotoski, line 43: p !=

PR #1317, opened 2020-06-22
Simplifies the Docker image to enable more freedom at runtime and better correlation with the underlying interface

rhyslbw:
- Removes the conveniences that came at the cost of direct correlation with the underlying interface
- The entrypoint is now `cardano-node` when NETWORK is not set
- docker run inputput/cardano-node [...ARGS] = cardano-node [...ARGS]
- :warning:  BREAKING CHANGE (however will fail early with a familiar error message)

### DIFF
--- a/nix/docker.nix
+++ b/nix/docker.nix
@@ -5,22 +5,22 @@
 #
 #   docker load -i $(nix-build -A dockerImage --no-out-link)
 #
-# To launch with provided mainnet configuration and persist state in a docker volume run:
+# To launch with pre-loaded configuration, using the NETWORK env.
+# An example using a docker volume to persist state:
 #
-#   docker run -v /data -e NETWORK=mainnet inputoutput/cardano-node:<TAG>
+#   docker run -v /data -e NETWORK=mainnet inputoutput/cardano-node
 #
-# To launch with provided testnet configuration without persisting state run:
+# Provide a complete command otherwise:
 #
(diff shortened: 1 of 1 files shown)


nix/docker.nix
```diff
@@ -5,22 +5,22 @@
...
-# To launch with provided testnet configuration without persisting state run:
+# Provide a complete command otherwise:
 #
-#   docker run -e NETWORK=testnet inputoutput/cardano-node:<TAG>
-#
-# To launch with custom config, mount a dir containing configuration.yaml, genesis.json, and topology.json
-#
-#   docker run -v $PATH_TO/configuration:/configuration \
-#     inputoutput/cardano-node:<TAG>
+#   docker run -v $PWD/configuration/defaults/byron-mainnet:/configuration \
```
rhyslbw, line 15: @kevinhammond example use-case

nix/docker.nix
```diff
@@ -5,22 +5,22 @@
...
 #
-# To launch with provided mainnet configuration and persist state in a docker volume run:
+# To launch with pre-loaded configuration, using the NETWORK env.
+# An example using a docker volume to persist state:
 #
-#   docker run -v /data -e NETWORK=mainnet inputoutput/cardano-node:<TAG>
+#   docker run -v /data -e NETWORK=mainnet inputoutput/cardano-node
 #
-# To launch with provided testnet configuration without persisting state run:
+# Provide a complete command otherwise:
```
rhyslbw, line 13: @kevinhammond example run command

PR #1349, opened 2020-06-26
Shelley Testnet Fork

disassembler:
Commits needed for shelley-testnet (CI only, DO NOT MERGE under ANY circumstances)


mrBliss:
> DO NOT MERGE under ANY circumstances

I just turned it into a draft to make it even more obvious.

PR #1378, opened 2020-07-01
Add golden test for stake pool metadata hash command

intricate:


cardano-cli/test/Test/Metadata.hs
```diff
@@ -0,0 +1,56 @@
...
+
+      -- Check that the stake pool metadata hash file content is correct.
+      actualStakePoolMetadataHash <- liftIO $ readFile outputStakePoolMetadataHashFp
+      expectedStakePoolMetadataHash === actualStakePoolMetadataHash
+
+      liftIO $ fileCleanup allFiles
+
+      H.success
+  where
+    exampleStakePoolMetadata :: Text
```
Jimbo4350, line 45: Let's convert these to golden files just for consistency

PR #1402, opened 2020-07-07
supervisord cluster: update for 1.15

disassembler:


nix/supervisord-cluster/clusterlib.py
```diff
@@ -63,14 +72,31 @@ def sendTx(self, txins=[], txouts=[], certificates=[], signing_keys=[], change_a
...
+        build_args = [ "--ttl", str(ttl), "--fee", str(fee), "--out-file", "tx.body" ] + txin_args + txout_args + cert_args
+
+        if proposal:
+            build_args.extend(["--update-proposal-file", proposal])
+
+        self.CLI(["cardano-cli", "shelley", "transaction", "build-raw"] + build_args )
+
+        if proposal:
+            build_args.extend(["--update-proposal-file", proposal])
+
```
johnalotoski, line 99: Extending build_args twice with `if proposal`?  Here and on L92-93?

PR #1410, opened 2020-07-08
Fix Cardano CLI tests on MacOS:

newhoggy:
* Use date in a cross-platform way to work on MacOS (OpenBSD) and Linux
    * Use xargs to strip whitespace from either side of the line count so
      that line count comparisions work on MacOS
    * Invoke shellcheck differently to avoid use of 'type', which triggers
      a lint error on MacOS

### DIFF
--- a/cardano-cli/test/cli/core/common
+++ b/cardano-cli/test/cli/core/common
@@ -65,7 +65,7 @@ assert_string () {
 }
 
 assert_line_count () {
-  line_count=$(wc -l < "$2")
+  line_count=$(wc -l < "$2" | xargs)
   if test "${line_count}" != "$1" ; then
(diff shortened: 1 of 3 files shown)


newhoggy:
The following commands show that the output of `TZ=UTC date +"%Y-%m-%dT%H:%M:%S+00:00"` is equivalent to `gdate` behaviour:

```
$ TZ=UTC gdate +"%Y-%m-%dT%H:%M:%S+00:00"; TZ=UTC gdate --iso-8601=seconds; TZ=UTC date +"%Y-%m-%dT%H:%M:%S+00:00"
2020-07-08T00:24:55+00:00
2020-07-08T00:24:55+00:00
2020-07-08T00:24:55+00:00
```

PR #1412, opened 2020-07-08
Update bech32 dependency to a version where tests succeed without needing bech32 to already be installed

newhoggy:
Previously `cabal test all` failed for me due to `bech32` not being installed.

See https://github.com/input-output-hk/bech32/pull/22

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -117,8 +117,8 @@ package cardano-ledger
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/bech32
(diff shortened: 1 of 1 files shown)


newhoggy:
Example failure:

```
  test/AppSpec.hs:57:14:
  1) App, bech32 command-line, any bech32 string can be decoded to hex
       uncaught exception: IOException of type NoSuchThing
       bech32: readCreateProcess: runInteractiveProcess: exec: does not exist (No such file or directory)
       (after 1 test and 12 shrinks)
         MinString "aaaaaaaaa"
```

PR #1413, opened 2020-07-08
Disable tests in bech32 package

newhoggy:
To avoid unnecessary tests and also avoid invoking tests which in an earlier version of bech32 invokes the bech32 executable that may not exist and cause test failure.


Jimbo4350:
bors merge

PR #1417, opened 2020-07-08
Make clusterlib more pythonic

mkoura:
More idiomatic Python that follows pep8. Some code refactoring towards single responsibility principle.

The `nix-build -A clusterTests.updateProposal` tested on f800f466227976f4d357fd99c69e9715b95d99bc.

### DIFF
--- a/nix/supervisord-cluster/clusterlib.py
+++ b/nix/supervisord-cluster/clusterlib.py
@@ -1,5 +1,4 @@
 """Wrapper for node-cli."""
-
 import functools
 import json
 import subprocess
@@ -12,7 +11,7 @@ class CLIError(Exception):
 
 
 class ClusterLib:
-    """Cluster Lib"""
+    """Cluster Lib."""
 
     def __init__(self, network_magic, state_dir):
         self.network_magic = network_magic
@@ -105,7 +104,8 @@ class ClusterLib:
                 str(txbody_file),
             ]
         )
-        return int(stdout.decode().split(" ")[1])
+        fee, __ = stdout.decode().split(" ")
+        return int(fee)
 
     def get_tx_fee(
         self, txins=None, txouts=None, certificates=None, signing_keys=None, proposal_file=None,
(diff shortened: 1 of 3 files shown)


nix/supervisord-cluster/clusterlib.py
```diff
@@ -1,167 +1,369 @@
...
-            self.CLI(["cardano-cli", "shelley", "transaction", "submit", "--testnet-magic", str(self.networkMagic), "--tx-file", "tx.signed"])
-        except CLIError as e:
-            print("Failed to submit transaction!")
-            print(f"txins: {txins} txouts: {txouts} signing keys: {signing_keys}")
-            raise e
-
-    def getAddress(self, payment=None, stake=None):
-        args = []
-        if not payment and stake:
-            raise CLIError("Must set payment, stake or both")
```
intricate, line 119: Seems like this was a logic bug. 

`not payment and stake` would have only been `True` in the event that `bool(payment) == False` and `bool(stake) == True`. Therefore this doesn't catch the case where neither `payment` nor `stake` have been provided.

nix/supervisord-cluster/clusterlib.py
```diff
@@ -1,167 +1,369 @@
...
+                str(self.network_magic),
+                "--tx-file",
+                str(tx_file),
+            ]
+        )
+
+    def get_address(self, payment=None, stake=None):
+        cli_args = []
+        if not (payment or stake):
+            raise CLIError("Must set payment, stake or both.")
```
intricate, line 257: I think the condition should actually just be `not payment`.

The `shelley address build` command _must_ be provided with a `payment-verification-key-file`, while the `stake-verification-key-file` is optional.

intricate: Also, since this method only utilizes the `shelley address build` command, it might be more accurate to call this method `get_payment_address`.

Unless there are plans to extend it to utilize the `shelley stake-address build` command in the future, @disassembler? e.g. in the case where _only_ a `stake-verification-key-file` is provided?

mkoura: Thanks for info! I changed it accordingly so that `stake` is now used alongside `payment` when specified.

nix/supervisord-cluster/clusterlib.py
```diff
@@ -348,7 +348,11 @@ def send_tx_genesis(
                 f"txins: {txins} txouts: {txouts} signing keys: {signing_keys}\n{err}"
             )
 
-    def submit_update_proposal(self, cli_args, epoch=1):
+    def get_current_epoch_no(self):
+        tip = self.get_tip()
+        return int(tip["slotNo"] / self.epoch_length)
```
intricate, line 353: Note that this is not necessarily the _current_ epoch number. The tip represents the last block that was successfully applied to the ledger, so this method actually returns the epoch number of the last adopted block. This method may return an incorrect result as there could very well be empty slots following the last adopted block.

intricate: I think that, for the time being, I would leave this as it was prior to `e65a66df7d6ad575adc5499ad0c030397a4bf1e0`. We currently don't expose a CLI utility for retrieving the current slot number.

mkoura: Thanks for explanation, I reverted the change.

PR #1424, opened 2020-07-09
Add more hedghog tests

Jimbo4350:
- Depends on #1423 
- Add extended payment key golden test
- Add `SerialiseAsCBOR` roundtrip tests for `TxBody`, `Tx` and `Witness` types
- Add Bech32 roundtrip tests

### DIFF
--- a/cardano-api/test/Test/Cardano/Api/Typed/Gen.hs
+++ b/cardano-api/test/Test/Cardano/Api/Typed/Gen.hs
@@ -1,10 +1,16 @@
 module Test.Cardano.Api.Typed.Gen
   ( genAddressByron
   , genAddressShelley
+  , genByronKeyWitness
   , genOperationalCertificate
   , genOperationalCertificateIssueCounter
+  , genShelleyWitness
   , genSigningKey
   , genStakeAddress
+  , genTxByron
+  , genTxShelley
+  , genTxBodyByron
+  , genTxBodyShelley
   , genVerificationKey
   ) where
 
@@ -14,6 +20,8 @@ import           Cardano.Prelude
 
 import           Control.Monad.Fail (fail)
 
+import qualified Cardano.Crypto.Hash as Crypto
+import           Cardano.Slotting.Slot (SlotNo(..))
 import           Ouroboros.Network.Magic (NetworkMagic(..))
 
 import           Hedgehog (Gen)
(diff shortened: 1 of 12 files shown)


cardano-api/test/Test/Cardano/Api/Typed/Gen.hs
```diff
@@ -105,28 +118,110 @@ genStakeCredential = do
...
+    <*> genTxBodyShelley
+ where
+   genWitnessList :: Gen [Witness Shelley]
+   genWitnessList = do
+     bsWits <- Gen.list (Range.constant 0 10) genShelleyBootstrapWitness
+     keyWits <- Gen.list (Range.constant 0 10) genShelleyKeyWitness
+     return $ bsWits ++ keyWits
+
+genTxExtraContent :: Gen TxExtraContent
+genTxExtraContent = return txExtraContentEmpty
```
newhoggy, line 181: Is empty intended for this generator?

cardano-api/test/Test/Cardano/Api/Typed/Gen.hs
```diff
@@ -105,28 +118,110 @@ genStakeCredential = do
...
+genShelleyTxOut =
+  TxOut <$> genAddressShelley <*> genLovelace
+
+genShelleyHash :: Gen (Crypto.Hash Crypto.Blake2b_256 ())
+genShelleyHash = return $ Crypto.hash ()
+
+genSlotNo :: Gen SlotNo
+genSlotNo = SlotNo <$> Gen.word64 Range.constantBounded
+
+-- TODO: Should probably have a naive generator that generates no inputs, no outputs etc
```
newhoggy, line 144: Is this TODO saying we should have an additional generator?

Jimbo4350: Yep, so we may want to have a naive generator that generates no transactions inputs (which would be an invalid transaction) for other testing purposes. It's okay to not have right now.

cardano-api/test/Test/Cardano/Api/Typed/Gen.hs
Jimbo4350: Yes this is intentional.

Jimbo4350:
bors merge

PR #1436, opened 2020-07-10
GITHUB-1086:  shut down the switchboard at exit

deepfire:
1. Expose `LoggingLayer` shutdown functionality, that was removed during `cardano-shell` elimination.  This allows proper termination of the switchboard thread.
2. Shutdown the `LoggingLayer` at exit.


rvl:
1. This bug also existed while cardano-shell was being used, so I don't know how removal of cardano-shell is related.
2. Looks plausible. We had the same issue in cardano-wallet. The app would hang if the switchboard were used after exit.

PR #1445, opened 2020-07-13
Eliminate dependency on lens

mrBliss:
Even though lens is already pulled in because of the dependency on it in
`cardano-ledger-specs`, simple uses of lenses like this can be catered by
`microlens`.


Jimbo4350:
bors merge

Jimbo4350:
I'd prefer to remove all lens usage but I'll see what others think about that.

PR #1468, opened 2020-07-15
update ouroboros-network

karknu:


mrBliss:
See https://github.com/input-output-hk/cardano-node/pull/1465

karknu:
> See #1465

Great, that saves me the trouble.

Issue #1420, opened 2020-07-09
[BUG] - A relay only downloads one block at the time in Blockfetch deadline mode.


karknu:
*Internal*

**Describe the feature you'd like**
The following scenario has been observed on the Shelley testnet.

Node A, IOHK and C are close, node B is far away.
1. A header for a new block, H_n from node A arrives at IOHK’s relay
2. It is for a new slot so IOHK’s relay decides to download it
3. A header for a new block H_n’ from node B arrives at IOHK’s relay, it is
   for the same slot but with a lower vrf value.
   IOHK’s relay knows that it will switch to it but it can’t start the download
   because it is limited by `bfcMaxConcurrencyDeadline`.
4. Block H_n arrives, and is adopted by IOHK’s relay so it announces it
   to all its peers, even if it knows it would like to switch ASAP.
5. Node C learns of H_n from IOHK and ask for it.
6. Block H_n arrives at Node C and is adopted.
7. Node C produces a new block
8. Block H_n’ arrives at IOHK’s relay. It switches and announces it to
   its peers.

Make `bfcMaxConcurrencyDeadline` and `bfcMaxConcurrencyBulkSync` configurable.
`bfcMaxConcurrencyDeadline` of one will work fine for normal clients (Daedalus) but should be set to two or more for pool relays.

**Describe alternatives you've considered**
One could increase the hardcoded bfcMaxConcurrencyDeadline  in ouroboros-network, but it that would mean that it is active for all nodes.

karknu:
Depends on https://github.com/input-output-hk/ouroboros-network/pull/2363

PR #1469, opened 2020-07-15
Blockfetch Config

karknu:
Implements #1420 

Should be rebased when #1465 is merged.


cardano-config/src/Cardano/Config/Types.hs
```diff
@@ -261,6 +263,17 @@ instance HasKESMetricsData ByronBlock where
 
 instance HasKESMetricsData (SimpleBlock a b) where
 
+newtype MaxConcerrencyBulkSync = MaxConcerrencyBulkSync
```
mrBliss, line 266: concurrency, not concerrency

PR #1482, opened 2020-07-16
devops-shell: add cluster-hfc scripts

disassembler:
Adds support for local testing of hard forks

* Refactors python helper code into a library (cardanolib-py)
* Separate out cluster logic and CLI wrapper classes (CLI wrapper is used by cluster)
* Renames keys to shelley to be specific for era

### DIFF
--- a/nix/supervisord-cluster/tests/base.py
+++ b/nix/supervisord-cluster/tests/base.py
@@ -1,13 +1,14 @@
-"""Usage: cluster-test.py [--network-magic=MAGIC] [--state-dir=DIR]
+"""Usage: cluster-test.py [--network-magic=INT] [--state-dir=DIR] [--num-delegs=INT]
 
 Options:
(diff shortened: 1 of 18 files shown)


nix/supervisord-cluster/tests/base.py
```diff
@@ -1,13 +1,14 @@
-"""Usage: cluster-test.py [--network-magic=MAGIC] [--state-dir=DIR]
+"""Usage: cluster-test.py [--network-magic=INT] [--state-dir=DIR] [--num-delegs=INT]
 
 Options:
     -n, --network-magic <magic>  network magic [default: 42]
```
johnalotoski, line 4: <magic> --> <int> to match L1 and L5?

disassembler: yeah, the change doesn't really do anything, just changed it to the type of it instead of some other description.

PR #1488, opened 2020-07-17
Fix Condense instance of OneEraHash

mrBliss:
`OneEraHash` is the `HeaderHash` type of `CardanoBlock`.

Now instead of

    "\214\157\227\245\SI\135CSI\238\241\185\196\137\FS;@\ETBB\208\201>\128\145\206\162\185\207\ETB\245t\180"<32b>

it should print the base16 encoding.


mrBliss:
bors merge

PR #1498, opened 2020-07-19
Update install.md

moomanchoo:
"cabal --version" in the text font looks like a single dash which doesn't work. Better to use -V which does and appears the same.


doc/getting-started/install.md
```diff
@@ -81,7 +81,7 @@ Update cabal
 
 Above instructions install Cabal version `3.2.0.0`. You can check the version by typing
 
-   cabal --version
+   cabal -V
```
intricate, line 84:
```suggestion
    cabal --version
```

PR #1502, opened 2020-07-20
Update install.md

iggirex:
listed version of cardano-node to DL no longer works with rest of instructions. Specifically, when trying `cardano-node run <options>`, getting errors about invalid blocks and the program just hangs

isItRunning


newhoggy:
Thanks for the PR.  The docs have already been updated to not refer to a specific tag.  Closing.

PR #1508, opened 2020-07-20
DOCS - removed nav components

robcohen:
* removed nav components, instead user simply hits "Go back to Cardano Documentation"


robcohen:
bors merge

PR #1519, opened 2020-07-22
Convert genesis tests to hedgehog

newhoggy:
Tests converted to hedgehog tests:

* `genesis-initial-txin`
* `genesis-key-gen-delegate`
* `genesis-key-gen-genesis`
* `genesis-key-gen-utxo`
* `genesis-key-hash`


cardano-cli/test/Test/CLI/Shelley/Golden/Genesis/InitialTxin.hs
```diff
@@ -0,0 +1,34 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Test.CLI.Shelley.Golden.Genesis.InitialTxin
```
intricate, line 4: I think we should go with:

```suggestion
module Test.CLI.Shelley.Golden.Genesis.InitialTxIn
```

cardano-cli/test/Test/CLI/Shelley/Golden/Genesis/InitialTxin.hs
```diff
@@ -0,0 +1,34 @@
...
+        utxoHashFile = tempDir <> "/utxo_hash"
+        goldenUtxoHashFile = "test/Test/golden/shelley/keys/genesis_utxo_hashes/utxo_hash"
+
+    utxoHash <-liftIO $ OP.execCardanoCLI
+        [ "shelley","genesis","initial-txin"
+        , "--testnet-magic", "16"
+        , "--verification-key-file", verificationKeyFile
+        ]
+
+    liftIO $ IO.writeFile utxoHashFile utxoHash
```
intricate, line 30: We don't need to write out this file, do we? If we plan to use it, we should remember to include it for clean up in the call to `OP.equivalence`.

cardano-cli/test/Test/CLI/Shelley/Golden/Genesis/InitialTxin.hs
```diff
@@ -0,0 +1,34 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeApplications #-}
```
intricate, line 2: Are we using `TypeApplications` here? Same question for the other test modules introduced in this PR.

cardano-cli/test/Test/CLI/Shelley/Golden/Genesis/InitialTxin.hs
```diff
@@ -0,0 +1,34 @@
...
+
+module Test.CLI.Shelley.Golden.Genesis.InitialTxin
+  ( golden_shelleyGenesisInitialTxin
+  ) where
+
+import Cardano.Prelude hiding (to)
```
intricate, line 8: I think we can remove this `hiding` clause here and also in the other test modules introduced in this PR.

cardano-cli/test/Test/OptParse.hs
```diff
@@ -195,8 +195,8 @@ assertFilesExist allFiles@(file:rest) = do
   else liftIO (fileCleanup allFiles) >> failWithCustom callStack Nothing (file <> " has not been successfully created.")
 
 -- | Assert the file contains the given number of occurences of the given string
-assertFileOccurences :: Int -> String -> FilePath -> H.PropertyT IO ()
-assertFileOccurences n s fp = do
+assertFileOccurences :: HasCallStack => Int -> String -> FilePath -> H.PropertyT IO ()
+assertFileOccurences n s fp = withFrozenCallStack $ do
```
intricate, line 199: :+1:

cardano-cli/test/Test/CLI/Shelley/Golden/Genesis/InitialTxin.hs
newhoggy: We don't need to clean up because it's written to the workspace which will be deleted.

We don't need to write it, but if it might be useful for producing a file we can copy to the golden file.

cardano-cli/test/Test/CLI/Shelley/Golden/Genesis/InitialTxin.hs
newhoggy: Fixed!

cardano-cli/test/Test/CLI/Shelley/Golden/Genesis/InitialTxin.hs
newhoggy: Fixed!

newhoggy:
bors merge

PR #1538, opened 2020-07-24
Convert stake address tests to hedgehog

newhoggy:


cardano-cli/test/Test/CLI/Shelley/Golden/StakeAddress/RegistrationCertificate.hs
```diff
@@ -0,0 +1,27 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module Test.CLI.Shelley.Golden.StakeAddress.RegistrationCertificate
+  ( golden_shelleyStakeAddressRegistrationCertificate
+  ) where
+
+import Cardano.Prelude hiding (to)
```
erikd, line 7: Wat? I do not see a clashing or alternative defintion.

newhoggy:
> Just remove any un-needed import `hiding` bits and this is good to go.

Done!

cardano-cli/test/Test/CLI/Shelley/Golden/Address/Info.hs
```diff
@@ -4,7 +4,7 @@ module Test.CLI.Shelley.Golden.Address.Info
   ( golden_shelleyAddressInfo
   ) where
 
-import Cardano.Prelude hiding (to)
+import Cardano.Prelude
```
erikd, line 7: And that should be quashed into the previous commit that added them :D

newhoggy:
bors merge

PR #1542, opened 2020-07-24
Minor dependency trimming

mrBliss:
I was actually hoping to remove all dependencies on the ledger testsuites, which bring in `lens`, `goblins`, the byron spec ledger (totally unused by `cardano-node`), but the ledger testsuites are used by the tests of the node.

I'm getting rid of `lens` in another way here: https://github.com/input-output-hk/cardano-ledger-specs/pull/1705

So this PR barely does any dependency trimming, but it would have been a shame to throw away this minor cleanup.


mrBliss:
bors merge

PR #1552, opened 2020-07-24
fixes LiveView memory display

CodiePP:
minimal change to reenable display of memory on Linux and Mac OSX in LiveView/TUI


mark-stopka:
I was worried this did not fix the issue properly as the bar was most of the time at full, however now I see there was some memory usage spike and it fell back to less than 100% of the maximum, so I can confirm that this works when cherry-picked into 1.18.0 release...

image

intricate:
Thanks @mark-stopka

bergr01:
@intricate @CodiePP Are those "checkForUnexpectedThunks ["Meresident LiveViewBackend"] lvs" correct or do they need adjusting too?

PR #1576, opened 2020-07-28
Stylish haskell applied

newhoggy:


cardano-cli/src/Cardano/CLI/Byron/Commands.hs
```diff
@@ -1,4 +1,4 @@
-{-# LANGUAGE GeneralisedNewtypeDeriving #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
```
erikd, line 1: Is this commit still needed?

cardano-cli/test/Test/OptParse.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE FlexibleContexts #-}
```
erikd, line 1: What change made this needed?

.stylish-haskell.yaml
```diff
@@ -15,41 +15,13 @@ steps:
   #     # true.
   #     add_language_pragma: true
 
-  # Format record definitions. This is disabled by default.
```
erikd: This should probably be squashed down into one of the previous commits.

cardano-cli/src/Cardano/CLI/Byron/Commands.hs
newhoggy: I've removed the pragma changes to see what happens.

cardano-cli/test/Test/OptParse.hs
newhoggy: I've removed the pragma changes to see what happens in CI.

.stylish-haskell.yaml
newhoggy: Squashed!

cardano-cli/test/Test/OptParse.hs
newhoggy: Looks to be working.

cardano-cli/src/Cardano/CLI/Byron/Commands.hs
newhoggy: Looks to be working.

newhoggy:
bors merge

PR #1580, opened 2020-07-28
Add legacy key generation documentation

SebastienGllmt:
I can't find any command for generating legacy keys and there are no tests for this, but empirically this works. I also documented the format for the cborHex in the legacy payment key case since it's really not obvious (you would think it's just the `xprv | chaincode`)


robcohen:
bors merge

PR #1581, opened 2020-07-28
doc: Docs aligned with mainnet

CarlosLopezDeLara:
replace config files and options in examples to work with mainnet.


robcohen:
bors merge

PR #1592, opened 2020-07-29
Tidy up remaining tests

newhoggy:
It's recommended to review this PR with this link https://github.com/input-output-hk/cardano-node/pull/1592/files?w=1

### Summary
* Convert all remaining tests to use workspaces so that they don't need to manually handle cleanup, which is error prone.  Possible errors prevented include, accidentally creating stray files that get committed, failing to clean up files, sharing files between concurrently run tests.
* Remove `fps` (files to cleanup) argument from `equivalence` and `checkTextEnvelopeFormat` functions because the argument was no longer used.
* New `readFile` function in `OptParse` module, the use of which does not require `liftIO` and handles exceptions in a way that produces useful test output.
* Use `void . execCardanoCLI` in place of `execCardanoCLIParser allFiles . evalCardanoCLIParser []` for brevity.
* Includes https://github.com/input-output-hk/cardano-node/pull/1589 which adds `cardano-cli` as `build-tools` of cli tests to ensure that the CLI executable is always built before running tests.
* Remove unused functions `evalCardanoCLIParser`, `execCardanoCLIParser` and `fileCleanup`

### DIFF
--- a/cardano-cli/test/Test/CLI/Shelley/Golden/Node/IssueOpCert.hs
+++ b/cardano-cli/test/Test/CLI/Shelley/Golden/Node/IssueOpCert.hs
@@ -5,7 +5,6 @@ module Test.CLI.Shelley.Golden.Node.IssueOpCert
   ) where
 
 import           Cardano.Prelude
-
 import           Hedgehog (Property)
 
 import qualified System.Directory as IO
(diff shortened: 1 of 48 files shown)


cardano-cli/test/Test/CLI/Shelley/Golden/Node/IssueOpCert.hs
```diff
@@ -14,33 +13,32 @@ import qualified Test.OptParse as OP
...
+golden_shelleyNodeIssueOpCert = OP.propertyOnce . OP.moduleWorkspace "tmp" $ \tempDir -> do
+  hotKesVerificationKeyFile <- OP.noteInputFile "test/Test/golden/shelley/keys/kes_keys/verification_key"
+  coldSigningKeyFile <- OP.noteInputFile "test/Test/golden/shelley/keys/genesis_delegate_keys/signing_key"
+  originalOperationalCertificateIssueCounterFile <- OP.noteInputFile "test/Test/golden/shelley/keys/genesis_delegate_keys/operational_certificate_counter"
+  operationalCertificateIssueCounterFile <- OP.noteTempFile tempDir "delegate-op-cert.counter"
+  operationalCertFile <- OP.noteTempFile tempDir "operational.cert"
+
+  void . liftIO $ IO.copyFile originalOperationalCertificateIssueCounterFile operationalCertificateIssueCounterFile
+
+  -- We could generate the required keys here, but then if the ket generation fails this
```
Jimbo4350, line 25: Small typo "ket" should be "kes"

cardano-cli/test/Test/CLI/Shelley/Golden/StakeAddress/Build.hs
```diff
@@ -14,17 +13,16 @@ import qualified Test.OptParse as OP
...
+      [ "shelley","stake-address","build"
+      , "--mainnet"
+      , "--staking-verification-key-file", verificationKeyFile
+      ]
 
-    void . liftIO $ IO.writeFile rewardAddressFile rewardAddress
+  void . liftIO $ IO.writeFile rewardAddressFile rewardAddress
 
-    rewardAddress === "stake1uxqmgfzls3vn7c7qlu3fdycz2nmh5p5sl2w7t7tfetp8evqacghf3\n"
+  rewardAddress === "stake1uxqmgfzls3vn7c7qlu3fdycz2nmh5p5sl2w7t7tfetp8evqacghf3\n"
```
Jimbo4350, line 28: This should be a golden file

cardano-cli/test/Test/CLI/Shelley/Golden/Node/IssueOpCert.hs
intricate:
```suggestion
  -- We could generate the required keys here, but then if the KES generation fails this
```

cardano-cli/test/Test/CLI/Shelley/Golden/StakeAddress/DeregistrationCertificate.hs
```diff
@@ -5,25 +5,23 @@ module Test.CLI.Shelley.Golden.StakeAddress.DeregistrationCertificate
...
+  verificationKeyFile <- OP.noteInputFile "test/Test/golden/shelley/keys/stake_keys/verification_key"
+  deregistrationCertFile <- OP.noteTempFile tempDir "deregistrationCertFile"
 
-    void $ OP.execCardanoCLI
-        [ "shelley","stake-address","deregistration-certificate"
-        , "--staking-verification-key-file", verificationKeyFile
-        , "--out-file", deregistrationCertFile
-        ]
+  void $ OP.execCardanoCLI
+      [ "shelley","stake-address","deregistration-certificate"
```
Jimbo4350, line 20: Isn't SH supposed to enforce a 2 space indent?

cardano-cli/test/Test/CLI/Shelley/Golden/TextEnvelope/Certificates/OperationalCertificate.hs
```diff
@@ -4,71 +4,60 @@ module Test.CLI.Shelley.Golden.TextEnvelope.Certificates.OperationalCertificate
...
-import           Cardano.Prelude
-
 import           Cardano.Api.Typed (AsType (..), HasTextEnvelope (..))
-
+import           Cardano.Prelude
 import           Hedgehog (Property)
-import qualified Hedgehog as H
 
-import           Test.OptParse as OP
+import qualified Test.OptParse as OP
```
Jimbo4350, line 11: IMO the qualified import of `Test.OptParse` takes away from the readability of these tests. If we were in a situation where we imported several modules or there were name clashes I'd say fair enough. I don't think this is the case here though. @intricate @erikd thoughts?

cardano-cli/test/Test/OptParse.hs
```diff
@@ -112,59 +92,14 @@ execCardanoCLI arguments = do
       ]
     IO.ExitSuccess -> return stdout
 
--- | This takes a 'ParserResult', which is pure, and executes it.
-execCardanoCLIParser
```
Jimbo4350, line 116: :wave:

cardano-cli/test/Test/CLI/Shelley/Golden/StakeAddress/DeregistrationCertificate.hs
newhoggy: Seems like SH isn't enforcing this.

cardano-cli/test/Test/CLI/Shelley/Golden/TextEnvelope/Certificates/OperationalCertificate.hs
newhoggy: Good coding styles usually say to use an import list or use qualified imports so that it's clear where the function comes from.

I feel that excessive use of import lists clashes with the strict flags we use that ban unused imports.

With import lists, the experience for me has been that I've had to add and remove functions to and from this lists often repeatedly.  For example if I wanted to add some debugging, I import the function I need, and then when I'm done, I need to remove them again.

A different story happens with refactoring, that doesn't concern the use of our strict flags.

When I copy files from another file to use as a template, I need to fiddle with import lists, whereas with qualified imports it's less so.  Qualified imports results in more refactor friendly code for some kinds of refactoring.

cardano-cli/test/Test/CLI/Shelley/Golden/Node/IssueOpCert.hs
newhoggy: Fixed!

cardano-cli/test/Test/CLI/Shelley/Golden/StakeAddress/Build.hs
newhoggy: Fixed!

cardano-cli/test/Test/CLI/Shelley/Golden/TextEnvelope/Certificates/OperationalCertificate.hs
newhoggy: But, I agree it does look a little more readable without the qualified imports 😂

Jimbo4350: All very fair points!

PR #1593, opened 2020-07-29
fix start-cluster and stop-cluster scripts

mkoura:
* the `start-cluster` is missing `--shelley-mode` and is failing with `cardano-cli: HardForkEncoderFutureEra (SingleEraInfo {singleEraName = "Shelley"})`
* the scripts are missing shebang and therefore are not detected by "exec" as scripts (e.g. when running from python scripts)


Jimbo4350:
Spoke with @mkoura who agreed to close this as the issue was fixed in another PR.

PR #1594, opened 2020-07-29
Remove types and parsers module from cardano-config

Jimbo4350:
- Split the types and parsers modules into `cardano-cli` and `cardano-node`


cardano-cli/src/Cardano/CLI/Byron/Parsers.hs
```diff
@@ -721,3 +719,47 @@ cliParseTxId :: String -> TxId
...
+parseSigningKeyFile opt desc = SigningKeyFile <$> parseFilePath opt desc
+
+
+parseGenesisFile :: String -> Parser GenesisFile
+parseGenesisFile opt =
+  GenesisFile <$> parseFilePath opt "Genesis JSON file."
+
+parseFlag' :: a -> a -> String -> String -> Parser a
+parseFlag' def active optname desc =
+  flag def active $ long optname <> help desc
```
newhoggy, line 765: What's the significance of the prime suffix?

Jimbo4350: :shrug:....I didn't implement this function, but I'm removing it because its only used in one place.

cardano-node/src/Cardano/Node/Types.hs
```diff
@@ -45,13 +59,100 @@ import           Ouroboros.Network.Block (MaxSlotNo (..))
...
+      Nothing -> panic $ "Parsing of IP failed: " <> ipStr
+  parseJSON Null = pure $ NodeHostAddress Nothing
+  parseJSON invalid = panic $ "Parsing of IP failed due to type mismatch. "
+                            <> "Encountered: " <> (Text.pack $ show invalid) <> "\n"
+
+instance ToJSON NodeHostAddress where
+  toJSON mha =
+    case unNodeHostAddress mha of
+      Just ip -> String (Text.pack $ show ip)
+      Nothing -> Null
```
intricate, line 154: Note that all of this is duplicated in `Cardano.CLI.Types`. Is that intentional?

Jimbo4350: Yep at this point it is. I've just realized some of these types are tied to "dead" cli commands.

Jimbo4350: Removed along with the cli commands that were using them

PR #1604, opened 2020-07-30
Add a jsonFromMetaData function

erikd:


cardano-api/src/Cardano/Api/MetaData.hs
```diff
@@ -0,0 +1,159 @@
...
+
+import           Cardano.Api.Typed as Api
+
+import qualified Data.Aeson as Aeson
+import qualified Data.Attoparsec.Text as Atto
+import qualified Data.ByteString.Char8 as BS
+import qualified Data.ByteString.Base16 as Base16
+import qualified Data.HashMap.Strict as HashMap
+import qualified Data.List as List
+import qualified Data.Map.Strict as Map
```
newhoggy, line 19: Any particular reason we use strict containers?

erikd: We use strict pretty much thoughout the whole of the cardano code base.

We tend to use strict data and lazy functions as the default.

cardano-api/test/Test/Cardano/Api/MetaData.hs
```diff
@@ -0,0 +1,83 @@
...
+-- -----------------------------------------------------------------------------
+
+-- Generate 'TxMetadata' that will round trip correctly.
+-- The only valid value that is known to not round trip correctly is a list of
+-- lists where all the sublists are of length 2.
+-- The reason this does not round trip is because a list of pairs aliases with
+-- a map.
+genMetaData :: Gen TxMetadata
+genMetaData = do
+  idxs <- List.nub <$> Gen.list (Range.linear 1 5) (Gen.word64 $ Range.linear 0 999999999999)
```
newhoggy, line 39: What's the significance of `999999999999` and why don't we test the entire `Word64` range?

newhoggy: Is it invalid for the list to be empty?  What should happen in this case?

cardano-api/test/Test/Cardano/Api/MetaData.hs
```diff
@@ -0,0 +1,83 @@
...
+              , (2, TxMetaBytes <$> genByteString)
+              , (1, genFlatTxMetadataValue)
+              ]
+  mapM (\k -> (k,) <$> genTxMetadataValue) (List.sort $ List.nub keys)
+
+genByteString :: Gen ByteString
+genByteString = BS.pack <$> Gen.list (Range.linear 1 32) Gen.latin1
+
+genText :: Gen Text
+genText = Text.pack <$> Gen.list (Range.linear 1 32) Gen.alphaNum
```
newhoggy, line 77: I presume we don't support empty `Text` & `ByteString`.  Are these impossible states?

cardano-api/test/Test/Cardano/Api/MetaData.hs
```diff
@@ -0,0 +1,83 @@
...
+
+genPairList :: Range Int -> Gen [(TxMetadataValue, TxMetadataValue)]
+genPairList len = do
+  keys <- Gen.list len $
+            Gen.frequency
+              [ (3, TxMetaText <$> genText)
+              , (2, TxMetaBytes <$> genByteString)
+              , (1, genFlatTxMetadataValue)
+              ]
+  mapM (\k -> (k,) <$> genTxMetadataValue) (List.sort $ List.nub keys)
```
newhoggy, line 71: Non-duplicate keys seem sensible enough.  What's the reason they need to be sorted and how is it guaranteed by the input?

newhoggy: `genTxMetadataValue` and `genPairList` are mutually recursive generators.

I've never done this before.  Does hedgehog have protection against this blowing up into an infinite data structure?

newhoggy: > Non-duplicate keys seem sensible enough. What's the reason they need to be sorted and how is it guaranteed by the input?

I suspect round tripping mentioned in the property. has something to do with it. 😁

cardano-api/src/Cardano/Api/MetaData.hs
```diff
@@ -0,0 +1,159 @@
...
+            _otherwise -> Nothing
+        _otherwise -> Nothing
+
+
+jsonToMetadataString :: Text -> Either MetaDataJsonConversionError TxMetadataValue
+jsonToMetadataString txt
+    -- If the text is encoded in hex, we convert it to a byte string.
+  | BS.take 2 utf8 == "0x"
+  , let (raw, trailing) = Base16.decode (BS.drop 2 utf8)
+  , BS.null trailing
```
newhoggy, line 148: I didn't know `let` was permitted in pattern guards.  I like it.

cardano-api/src/Cardano/Api/MetaData.hs
```diff
@@ -0,0 +1,159 @@
...
+  deriving (Eq, Ord, Show)
+
+jsonFromMetadata :: Api.TxMetadata -> Aeson.Value
+jsonFromMetadata (Api.TxMetadata (MetaData meta)) =
+    Aeson.Object $ HashMap.fromList (map convertPair $ Map.toList meta)
+  where
+    convertPair :: (Word64, MetaDatum) -> (Text, Aeson.Value)
+    convertPair (w, d) = (Text.pack (show w), toAeson d)
+
+    toAeson :: MetaDatum ->  Aeson.Value
```
newhoggy, line 47: Double space after arrow.

cardano-api/test/Test/Cardano/Api/MetaData.hs
erikd: Generating `Word64` in hedgehog is a pain in the neck, mainly because `Gen.enumBounded` fails whenever the generated `Word64` is greater than `maxBound :: Int64`.

cardano-api/test/Test/Cardano/Api/MetaData.hs
erikd: They are not impossible, but likely to result in problems.

cardano-api/test/Test/Cardano/Api/MetaData.hs
erikd: The need to be sorted because they go via a `HashMap` and that messes up the ordering.

cardano-api/src/Cardano/Api/MetaData.hs
erikd: Fixed.

cardano-api/test/Test/Cardano/Api/MetaData.hs
erikd: Just tested it, they are both fine with `0` as the lower bound, so I kept it.

cardano-api/test/Test/Cardano/Api/MetaData.hs
erikd: Hmm, switched that to `Gen.word64 Range.constantBounded`

erikd:
bors merge

PR #1605, opened 2020-07-30
Remove cardano-config 

Jimbo4350:


Jimbo4350:
Likely failing because CI needs to be upgraded to use `cabal` 3.2. Sub-library use in `cardano-node` appears to be causing the failures seen in CI here.

PR #1607, opened 2020-07-30
Update minimum required cabal version to cabal 3.0

Jimbo4350:


mrBliss:
If we go for 3.2.0.0, we could start listing multiple subdirs in the `subdir` field of `source-repository-package` (when the Nix integration supports it).

Jimbo4350:
Subsumed by #1615. CI needs to be upgraded to use cabal 3.2

PR #1614, opened 2020-07-30
config: Update cardano mainnet configs

erikd:
Grabbed the mainnet config files from https://hydra.iohk.io/build/3627080/download/1/index.html

Tested using:
```
/nix/store/5n4y7algf5jri6l176qy1h26qsl429f6-cardano-node-exe-cardano-node-1.18.0/bin/cardano-node run \
    --config configuration/cardano/mainnet-config.json \
    --database-path state-node-mainnet/db-mainnet \
    --topology configuration/cardano/mainnet-topology.json \
    --host-addr 127.0.0.1 \
    --port 3001 \
    --socket-path state-node-mainnet/node.socket
```
and the node continues syncing.


mrBliss:
Should the `mainnet-genesis.json` in the root of the repo be removed?

erikd:
> Should the mainnet-genesis.json in the root of the repo be removed?

`db-sync` used to rely on that file being there. I am concerned about what might break if that file is removed.

erikd:
bors merge

PR #1615, opened 2020-07-30
Upgrade cabal file version

newhoggy:
Testing to see what happens in CI

### DIFF
--- a/cardano-api/cardano-api.cabal
+++ b/cardano-api/cardano-api.cabal
@@ -1,3 +1,5 @@
+cabal-version: 2.4
+
 name:                   cardano-api
 version:                1.18.0
 description:            The cardano api
@@ -8,7 +10,6 @@ license-files:
   LICENSE
   NOTICE
 build-type:             Simple
-cabal-version:          >= 1.10
 extra-source-files:     README.md, ChangeLog.md
 
 library


cardano-api/cardano-api.cabal
```diff
@@ -1,3 +1,5 @@
+cabal-version: 2.4
```
erikd, line 1: Probably should be `3.0` because we are using `3.0` features (ie the `cabal.project` file).

newhoggy: I tried `3.0` but it failed the build.  For some reason `2.4` is as high as we can go.

erikd: Ok, I would be fine leaving it like that for now. The required nix fix is probably large and hairy.

newhoggy:
Thanks!

newhoggy:
bors merge

mrBliss:
I was looking into this again in #1670 and discovered the following:

In `cabal-install-3.2.0.0`, the `cabal-version` field [is parsed as follows](https://github.com/haskell/cabal/blob/329ebb1ad4c132e26b99e128450f9dafa1144133/Cabal/Distribution/PackageDescription/Parsec.hs#L106-L110):

```haskell
[..]
    -- if we get too new version, fail right away
    case ver of
        Just v | v > mkVersion [3,0] -> parseFailure zeroPos
            "Unsupported cabal-version. See https://github.com/haskell/cabal/issues/4899."
        _ -> pure ()
[..]
```

So even version 3.2.0.0 explicitly does not support a version >= 3.0 :roll_eyes:. Although, I discovered that this version number corresponds to the "cabal spec version", which, I suppose, not necessarily has to match the "Cabal/cabal-install version".

This was fixed in 3.4.0.0:

```haskell
        -- if we get too new version, fail right away
        Just v -> case cabalSpecFromVersionDigits (versionNumbers v) of
            Just csv -> return (Just csv)
            Nothing  -> parseFatalFailure zeroPos $
                "Unsupported cabal-version " ++ prettyShow v ++ ". See https://github.com/haskell/cabal/issues/4899."
```

I suppose we'll have to wait until 3.4.0.0 is officially released and wide-spread before we can increase this version number.

PR #1617, opened 2020-07-31
Spawn cardano-node in a test for 10 seconds.

newhoggy:
### Summary

* New test that spawns `cardano-node` for 10 seconds then interrupts it as a proof of concept of process handling
* Support code for process handling in tests
* Upgrade all cabal files to cabal file version `2.4`
* Add `build-tool-depends: cardano-node:cardano-node` to `test` component of  `cardano-node.cabal` to ensure the executable is built before tests because tests depend on it.
* Add `/cardano-node/logs` to `.gitignore`
* Some configuration files committed under `chairman` directory.  These will be removed in future and generated as part of tests.


cardano-node/test/Test/Cardano/Node/Chairman.hs
```diff
@@ -0,0 +1,41 @@
...
+import           Hedgehog (Property, discover)
+
+import qualified Hedgehog as H
+import qualified System.Process as IO
+import qualified Test.Common.Base as H
+import qualified Test.Common.Process as H
+
+prop_spawnOneNode :: Property
+prop_spawnOneNode = H.propertyOnce $ do
+  (_mIn, _mOut, _mErr, hProcess) <- H.createProcess $
```
Jimbo4350, line 18: Nice! Not sure what direction you are going in here but I would create a function that accepts a record type parameterizing the database path, socket path etc. Not necessarily in this PR though.

newhoggy: Sounds good.  I'll look into that.

cardano-node/test/Test/Common/Process.hs
```diff
@@ -0,0 +1,36 @@
...
+  => ProcessHandle
+  -> H.PropertyT IO ()
+interruptProcessGroupOf hProcess = GHC.withFrozenCallStack $ do
+  H.evalM . liftIO $ IO.interruptProcessGroupOf hProcess
+
+waitForProcess :: HasCallStack
+  => ProcessHandle
+  -> H.PropertyT IO ExitCode
+waitForProcess hProcess = GHC.withFrozenCallStack $ do
+  H.evalM . liftIO $ IO.waitForProcess hProcess
```
Jimbo4350, line 36: Maybe `waitForProcess` is indefinitely waiting?

newhoggy: I thought of that possibility, but there's no visibility in CI, so how can I confirm?

newhoggy: I'm also wondering if there is something spinning now on the build server.  Usually there should be a timeout on builds if there is no progress.

Jimbo4350: You will need to get in touch with devops for assistance.

cardano-node/src/Cardano/Node/Types.hs
```diff
@@ -545,7 +545,7 @@ parseNodeConfigurationFP :: ConfigYamlFilePath -> IO NodeConfiguration
 parseNodeConfigurationFP (ConfigYamlFilePath fp) = do
     nc <- decodeFileThrow fp
     -- Make all the files be relative to the location of the config file.
-    pure $ adjustFilePaths (takeDirectory fp </>) nc
+    pure $ adjustFilePaths (\p -> takeDirectory fp <> "/" <> p) nc
```
newhoggy, line 548: Instead of constructing this kind of string on Windows (not the backslash):

```
source/configuration/defaults/simpleview\genesis/genesis.json
```

newhoggy:
bors merge

cardano-node/moo.txt
```diff
@@ -0,0 +1 @@
+received shutdown request 2
```
Jimbo4350, line 1: @newhoggy Left over file?

chairman/configuration/defaults/simpleview/config-0.yaml
```diff
@@ -0,0 +1,212 @@
+# global filter; messages must have at least this severity to pass:
```
Jimbo4350, line 1: @newhoggy We should put these in the `configuration` folder. Previously we had configurations all over the place. It just makes it easier in the events where we want to add a tracer for example, they all exist in `configuration`.

chairman/configuration/defaults/simpleview/genesis/delegate-keys.005.key
```diff
@@ -0,0 +1 @@
+X�����C��%��Y����G�c�b~�L�n�<.���J�xH�Q��V���钮�C�Ӷ�Q�&j%�" 7��F7ҥ�a����6�[��/=cK41(i���@�W����f��w�����
```
Jimbo4350, line 1: @newhoggy  The genesis files should be generated on the fly so these need to be removed.

newhoggy: Yep, I'm working on the file generation at the moment.

cardano-node/moo.txt
newhoggy: Will remove.  Thanks!

PR #1619, opened 2020-07-31
Begin removing usages of `Condense HeaderHash` from the project

### LINKED ISSUE #570: Inaccurate ToObject instance for `Point blk`
https://github.com/input-output-hk/cardano-node/blob/f1a5b69d205599ad99c07f19ed9d6cf50575564a/cardano-node/src/Cardano/Tracing/ToObjectOrphans.hs#L597-L602

This particular definition outputs a JSON object whose `"kind"` key is mapped to a value `"Tip"`. This seems incorrect as a `Point blk` is not necessarily the tip.

intricate:
This PR also happens to close #570.

### DIFF
--- a/cardano-node/src/Cardano/Tracing/OrphanInstances/Consensus.hs
+++ b/cardano-node/src/Cardano/Tracing/OrphanInstances/Consensus.hs
@@ -5,6 +5,7 @@
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE UndecidableInstances #-}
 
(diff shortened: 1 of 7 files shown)


cardano-node/src/Cardano/Tracing/OrphanInstances/Consensus.hs
```diff
@@ -47,15 +51,20 @@ import           Ouroboros.Consensus.Util.Condense
...
 
 import qualified Ouroboros.Consensus.Storage.ChainDB as ChainDB
 import qualified Ouroboros.Consensus.Storage.LedgerDB.OnDisk as LedgerDB
 
 
 
+instance ConvertRawHash (Header blk) where
+  toShortRawHash = toShortRawHash
+  fromShortRawHash = fromShortRawHash
+  hashSize = hashSize
```
intricate, line 66: Perhaps this should be in `ouroboros-consensus`.

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
```diff
@@ -538,12 +507,17 @@ instance (Show txid, Show tx)
...
-  toObject verb p =
-    mkObject [ "kind" .= String "Tip" --TODO: why is this a Tip not a Point?
-             , "tip" .= showPoint verb p ]
+  toObject _verb GenesisPoint =
+    mkObject
+      [ "kind" .= String "GenesisPoint" ]
+  toObject verb (BlockPoint slot h) =
+    mkObject
+      [ "kind" .= String "BlockPoint"
+      , "slot" .= toJSON (unSlotNo slot)
```
intricate, line 518: I _could_ use `ToObject SlotNo` here, but it feels odd to represent it as an object when its really just a sort of wrapper for a `Number`.

PR #1625, opened 2020-08-03
Default hlint file

newhoggy:
This file disables all the `hlint` rules in the project that might cause the lint to fail.

To test if `hlint` works in CI, delete the `.hlint.yaml` file and the build should fail.

This PR was created so that the SRE team can create a commit on top to enable `hlint` in CI.  See https://jira.iohk.io/browse/SRE-131

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -1,0 +1,82 @@
+# HLint configuration file
+# https://github.com/ndmitchell/hlint
+##########################
+
+# This file contains a template configuration file, which is typically
+# placed as .hlint.yaml in the root of your project
+
+
+# Warnings currently triggered by your code
+- ignore: {name: "Unused LANGUAGE pragma"}
+- ignore: {name: "Use bimap"}
+- ignore: {name: "Move brackets to avoid $"}
+- ignore: {name: "Redundant bracket"}
+- ignore: {name: "Redundant flip"}
+- ignore: {name: "Redundant where"}
+- ignore: {name: "Eta reduce"}
+- ignore: {name: "Use camelCase"}
+- ignore: {name: "Use newtype instead of data"}
+- ignore: {name: "Redundant $"}
+- ignore: {name: "Avoid lambda"}
+- ignore: {name: "Reduce duplication"}
+- ignore: {name: "Redundant as"}
+- ignore: {name: "Parse error: lexical error in string/character literal at character '\\n'"}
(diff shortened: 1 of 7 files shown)


jbgi:
there is still some rules that cause the lint to fail:
https://hydra.iohk.io/build/3746935/nixlog/1

newhoggy:
What version of `hlint` is being used?

newhoggy:
I'd have to have any additional rules be excluded in the `hlint` config to enable the build to pass, then my team would be able to gradually look at re-enabling rules on a case-by-case basis.

jbgi:
it's using htlint 2.2.10, I can look into using a newer version (the one provided by haskell.nix hackage snapshot)

jbgi:
mmm. tried to use hlint through haskell.nix, but got
`error: The Haskell package set does not contain the package: cpphs (build dependency)`

newhoggy:
I can confirm the `hlint` version difference is responsible for most of the warnings.

jbgi:
updated to use hlint 3.1.6.

now the only warnings are:
```
cardano-node/test/Test/Common/Base.hs:75:19-114: Warning: Use maybe
Found:
  fromMaybe
    "UnknownModule"
    (fmap
       (GHC.srcLocModule . snd) (listToMaybe (getCallStack callStack)))
Perhaps:
  maybe
    "UnknownModule" (GHC.srcLocModule . snd)
    (listToMaybe (getCallStack callStack))

cardano-node/test/Test/Common/Process.hs:37:75-91: Warning: Use unwords
Found:
  L.intercalate " "
Perhaps:
  unwords
```

newhoggy:
Nice!

newhoggy:
I pushed some warning fixes on top.  Hopefully it builds now.

newhoggy:
bors merge

newhoggy:
Looks like there's a conflict in `release.nix`:

```
<<<<<<< HEAD
  filterJobsCross = mapAttrsRecursiveCond (a: !(isDerivation a)) (path: value:
    if (any (p: take (length p) path == p) noCrossBuild) then null
    else value
  );
=======
  filterJobsCross = filterAttrs (n: _: n != "dockerImage" && n != "shell" && n != "cluster" && n != "hlint");
>>>>>>> Add hlint ci check
```

newhoggy:
@jbgi Thanks so much for your help!

PR #1629, opened 2020-08-03
Simplify & improve shelley cli errors

Jimbo4350:


cardano-cli/src/Cardano/CLI/Shelley/Run/Address.hs
```diff
@@ -24,18 +24,18 @@ import           Cardano.CLI.Shelley.Run.Address.Info (ShelleyAddressInfoError,
 import           Cardano.CLI.Types
 
 data ShelleyAddressCmdError
-  = ShelleyAddressCmdAddressInfoError !ShelleyAddressInfoError
-  | ShelleyAddressCmdReadFileError !(FileError TextEnvelopeError)
-  | ShelleyAddressCmdWriteFileError !(FileError ())
+  = AddressInfoError !ShelleyAddressInfoError
+  | ReadFileError !(FileError TextEnvelopeError)
+  | WriteFileError !(FileError ())
```
intricate, line 29: I don't think naming our errors like this is a good idea. I would prefer to keep all of the constructor names prefixed as before. 

If we do it this way, we now have multiple constructors for different types with the same name: `ReadFileError`, `WriteFileError`, etc.

erikd: I am pretty firmly with Luke on this. `ReadFileError` is incredibly generic whereas `ShelleyAddressCmdReadFileError` is incredibly specific. Yes, it is more characters, but it will be read at least 100x more often than written.

Jimbo4350: I initially wrote these error constructors to be very verbose because we were relying on the error rendering functions in the module. Therefore they needed to be verbose and specific so we could write sensible errors in the error rendering function. 

However, I changed my mind because:
- `FilerError ()`/`FileError TextEnvelopeError` will render an error message indicating what went wrong and where it went wrong. This eliminates the need for the verbose and specific error constructors in each module barring those that do not involve some kind of IO error.
- We don't export any of the error constructors.

cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
```diff
@@ -98,6 +132,17 @@ data KeyCmd
...
 
+renderKeyCmd :: KeyCmd -> Text
+renderKeyCmd cmd =
+  case cmd of
+    KeyGetVerificationKey _ _ -> "key verification-key"
+    KeyNonExtendedKey  _ _ -> "key non-extended-key"
+    KeyConvertByronKey _ _ _ _ -> "key convert-byron-key"
+    KeyConvertByronGenesisVKey _ _ -> "key convert-byron-genesis-key"
+    KeyConvertITNStakeKey _ _ -> "key convert-itn-key"
+    KeyConvertITNExtendedToStakeKey _ _ -> "key convert-itn-extended-key"
```
newhoggy, line 143: Words on LHS and RHS don't match?

Jimbo4350: Sorry I don't understand what you mean? I've mapped the `KeyCmd` type to what is display in the cli executable.

Jimbo4350:
Closing this, too much time has gone by.

PR #1632, opened 2020-08-04
Shutdown cleanup

newhoggy:
Refactoring PR built on top of https://github.com/input-output-hk/cardano-node/pull/1631


cardano-node/src/Cardano/Node/Handlers/Shutdown.hs
```diff
@@ -136,20 +121,30 @@ withShutdownHandling
...
-
-   mkInternalShutdown :: IO ShutdownFDs
-   mkInternalShutdown = do
-     (r, w) <- IO.createPipeFd
-     pure $ InternalShutdown (ShutdownListener $ Fd r) (ShutdownDoorbell $ Fd w)
+  sfds <- case cli of
+    NodeCLI{shutdownIPC = Just fd} -> pure $ ExternalShutdown (ShutdownListener fd)
+    NodeCLI{shutdownOnSlotSynced = MaxSlotNo{}} -> do
+      (r, w) <- IO.createPipeFd
+      pure $ InternalShutdown (ShutdownListener $ Fd r) (ShutdownDoorbell $ Fd w)
```
newhoggy, line 128: After this refactoring, it became obvious that the `ShutdownDoorbell` constructed here is ignored below at `_w` and so is unused.  Did we forget to use it or is the `ShutdownDoorbell` redundant?

Jimbo4350: I believe @deepfire created this module

deepfire: @newhoggy, not ignored -- the `ShutdownFDs` are passed to `action`, where the write side of the doorbell can be used (i.e. closed).

PR #1639, opened 2020-08-04
Refine type signature of withShutdownHandler

newhoggy:
`withShutdownHandler` was previously taking `ShutdownFDs` argument, which contains more information than it needs to work.  Refining the type down to take `Maybe ShutdownListener` instead.


newhoggy:
bors merge

PR #1644, opened 2020-08-06
Minimal testnet

newhoggy:


scripts/lite/shelley-testnet.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
```
erikd, line 1: Have you run the `shellcheck` program over this?

newhoggy: Thanks for the suggestion.

I have applied `shellcheck` recommendations.

newhoggy:
bors merge

PR #1649, opened 2020-08-06
Replace epoch by chunk in ChainDB.OpenedImmDB trace message

kderme:
I see logs when i run on mainnet like

`[kostas-u:cardano.node.ChainDB:Info:5] [2020-08-06 10:54:59.55 UTC] Opened imm db with immutable tip at (Point 5080840, "643179a075814eff3bafe38967474b8281640ffd77135b8db3b8293756f8c24a") and epoch 235
`

which is wrong since there is no epoch 235 yet

The immutable db has changed and no longer creates a file per epoch but per, what it calls, a chunk.


mrBliss:
Maybe make the title of this PR a bit more specific (was logging broken? how?), e.g., "Replace epoch by chunk in ChainDB.OpenedImmDB trace message".

PR #1657, opened 2020-08-10
Run cluster exception safe

newhoggy:
Tested by modifying the `main` function like this:

```
import           Cardano.Prelude

import           Hedgehog.Main (defaultMain)

import qualified Control.Concurrent as IO
import qualified Control.Exception as IO
import qualified System.IO as IO
import qualified Test.Cardano.Node.Chairman
import qualified Test.Cardano.Node.Json

main :: IO ()
main =
  flip IO.finally (IO.putStrLn "Done" >> IO.threadDelay 10000000) $
    defaultMain
      [ Test.Cardano.Node.Chairman.tests
      , Test.Cardano.Node.Json.tests
      ]
```

Then adding `void H.failure` to the end of the test.

The failure causes the test to fail.  This triggers the actions registered on `ResourceT` to be called, which closes all the process handles and terminates the process.

The `main` function would run the test, then print `Done`.  The test process remains waiting for another 10 seconds.  During this time, I can observe that all the node processes have exited the moment `Done` is printed.


newhoggy:
In practise the previous iteration works just as well, but probably because the garbage collector kicks in quickly to terminate the process.

The change I guess makes the cleanup a guarantee rather than relying on something as temperamental as a garbage collector.

An added advantage is we no longer need to remember to clean up as clean up happens automatically at the end of scope.

cardano-node/test/Test/Common/Base.hs
```diff
@@ -15,6 +15,8 @@ module Test.Common.Base
 
 import           Control.Monad
 import           Control.Monad.IO.Class (liftIO)
+import           Control.Monad.Morph
+import           Control.Monad.Trans.Resource
```
erikd, line 19: Everything that is imported from outside the current repo should used either qualified or explicit imports.

newhoggy: Chose to use explicit imports.

cardano-node/test/Test/Cardano/Node/Chairman.hs
```diff
@@ -8,43 +9,80 @@ module Test.Cardano.Node.Chairman
...
+          , "--config", tempDir <> "/config-" <> si <> ".yaml"
+          , "--signing-key", tempDir <> "/genesis/delegate-keys.00" <> si <> ".key"
+          , "--delegation-certificate", tempDir <> "/genesis/delegation-cert.00" <> si <> ".json"
+          , "--shutdown-ipc", "0"
+          ]
+        ) <&>
+        ( \cp -> cp
+          { IO.std_in = IO.CreatePipe
+          , IO.std_out = IO.UseHandle hNodeStdout
+          , IO.std_err = IO.UseHandle hNodeStderr
```
newhoggy, line 78: This will redirect `stdout` and `stderr` output to files.

Subsequently, will need to do something more sophisticated to handle test assertions based on the output.

newhoggy:
bors merge

PR #1660, opened 2020-08-10
Implement ToJSON/FromJSON instances of MultiSigScript

Jimbo4350:
In an effort to keep things simple, the multisig syntax will use JSON for now unless we need . This is not integrated into cardano-cli as yet. The format is as follows:

Require `all` signatures:
```
{ "all": [ "2f3d4cf10d0471a1db9f2d2907de867968c27bca6272f062cd1c2413", "f856c0c5839bab22673747d53f1ae9eed84afafb085f086e8e988614" ] }
```
Require `any` signature:
```
{ "any": [ "2f3d4cf10d0471a1db9f2d2907de867968c27bca6272f062cd1c2413", "f856c0c5839bab22673747d53f1ae9eed84afafb085f086e8e988614" ] }
```
Require `atLeast` signatures:
e.g 2 of 4:
```
{
    "atLeast": {
        "paymentKeyHashes": [
            "2f3d4cf10d0471a1db9f2d2907de867968c27bca6272f062cd1c2413",
            "f856c0c5839bab22673747d53f1ae9eed84afafb085f086e8e988614",
            "b275b08c999097247f7c17e77007c7010cd19f20cc086ad99d398538",
            "686024aecb5884d73a11b9ae4e63931112ba737e878d74638b78513a"
        ],
        "required": 2,
        "total": 4
    }
}
```

### DIFF
--- a/cardano-api/src/Cardano/Api/Typed.hs
+++ b/cardano-api/src/Cardano/Api/Typed.hs
@@ -37,7 +37,6 @@ module Cardano.Api.Typed (
     Shelley,
     HasTypeProxy(..),
     AsType(..),
-
     -- * Cryptographic key interface
     -- $keys
     Key,
@@ -307,24 +306,23 @@ module Cardano.Api.Typed (
     toShelleyNetwork,
   ) where
 
-
 import           Prelude
 
 import           Data.Aeson.Encode.Pretty (encodePretty')
 import           Data.Bifunctor (first)
+import qualified Data.HashMap.Strict as HMS
 import           Data.Kind (Constraint)
 import           Data.List as List
 import qualified Data.List.NonEmpty as NonEmpty
 import           Data.Maybe
 import           Data.Proxy (Proxy (..))
-import           Data.Typeable (Typeable)
-import           Data.Void (Void)
-import           Data.Word
---import           Data.Either
 import           Data.String (IsString (fromString))
 import           Data.Text (Text)
 import qualified Data.Text as Text
 import qualified Data.Text.Encoding as Text
+import           Data.Typeable (Typeable)
+import           Data.Void (Void)
+import           Data.Word
 import           Numeric.Natural
 
 import           Data.IP (IPv4, IPv6)
@@ -344,6 +342,7 @@ import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
 import qualified Data.Sequence.Strict as Seq
 import qualified Data.Set as Set
+import           Data.Vector (Vector)
 import qualified Data.Vector as Vector
 
 import qualified Codec.Binary.Bech32 as Bech32
@@ -357,10 +356,9 @@ import           Control.Monad.Trans.Except (ExceptT (..))
 import           Control.Monad.Trans.Except.Extra
 import           Control.Tracer (nullTracer)
 
-import           Data.Aeson (FromJSON (..), ToJSON (..), (.:))
+import           Data.Aeson (FromJSON (..), ToJSON (..), Value (..), object, (.:), (.=))
 import qualified Data.Aeson as Aeson
 import qualified Data.Aeson.Types as Aeson
-
 
 --
 -- Common types, consensus, network
@@ -465,8 +463,6 @@ import           Ouroboros.Network.Protocol.ChainSync.Client as ChainSync
 import           Ouroboros.Network.Protocol.LocalStateQuery.Client as StateQuery
 import           Ouroboros.Network.Protocol.LocalStateQuery.Type (AcquireFailure)
 import           Ouroboros.Network.Protocol.LocalTxSubmission.Client as TxSubmission
-
-
 
 
 -- ----------------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


cardano-api/src/Cardano/Api/Typed.hs
```diff
@@ -1565,6 +1565,100 @@ data MultiSigScript = RequireSignature (Hash PaymentKey)
...
+      (_, Just mss, _) -> return mss
+      (_, _, Just mss) -> return mss
+      (_, _, _) -> fail "Please specify only one type of JSON multisig object."
+
+gatherSignatures :: Vector Value -> [MultiSigScript]
+gatherSignatures anyList = do
+  let mPaymentKeyHashes = Vector.map (\val -> case filterValue val of
+                                                Just hashPayKey ->
+                                                  Just $ RequireSignature hashPayKey
+                                                Nothing -> Nothing
```
newhoggy, line 1606: Probably use `fmap` here

cardano-api/test/Test/Cardano/Api/Typed/MultiSigScript.hs
```diff
@@ -0,0 +1,38 @@
...
+import           Test.Cardano.Api.Typed.Gen
+import           Test.Cardano.Api.Typed.Orphans ()
+
+
+prop_generateMofNcorrectly :: Property
+prop_generateMofNcorrectly = H.property $ do
+  RequireMOf req sigs <- H.forAll genMofNRequiredSig
+  case length sigs >= req of
+    False -> failWith Nothing $ "genMofNRequiredSig: Number of required \
+                                \signatures exceedes number of available key hashes. \
```
newhoggy, line 25: Typo: "exceedes"

cardano-api/src/Cardano/Api/Typed.hs
```diff
@@ -1565,6 +1565,100 @@ data MultiSigScript = RequireSignature (Hash PaymentKey)
...
+
+    -- Parse "mOfn" multisig objects
+    mofN <- return $ filterMofN obj
+
+    case (allReqHashes, anyReqHashes, mofN) of
+      (Nothing, Nothing, Nothing) -> fail "Script does not contain \"all\", \"any\" \
+                                          \or \"mOfn\" JSON objects."
+      (Just mss, _, _) -> return mss
+      (_, Just mss, _) -> return mss
+      (_, _, Just mss) -> return mss
```
newhoggy, line 1598: Hmm, based on the compile error below, the above probably should have `Nothing` in place of `_`.

newhoggy: Could write it this way perhaps.

```
case catMaybes [allReqHashes, anyReqHashes, mofN] of
  [mss] -> return mss
  [] -> fail "Script does not contain \"all\", \"any\" or \"mOfn\" JSON objects."
  _ -> fail "Please specify only one type of JSON multisig object."
```

cardano-api/src/Cardano/Api/Typed.hs
```diff
@@ -1565,6 +1565,103 @@ data MultiSigScript = RequireSignature (Hash PaymentKey)
...
+   allPotentialScripts obj' = catMaybes . map (parseMaybe scriptParser . Text.unpack)
+                                $ HMS.keys obj'
+
+   onlyMofN :: ParsedScript -> Maybe (Int, String, Int)
+   onlyMofN (All _) = Nothing
+   onlyMofN (Any _) = Nothing
+   onlyMofN (MofN mon) = Just mon
+
+
+data ParsedScript = All String | Any String | MofN (Int, String, Int)
```
newhoggy, line 1647: I think the `String` in `MofN (Int, String, Int)` is redundant because it always has the same value.  Can delete without loss of functionality.

Jimbo4350: Yep that's a good point.

cardano-api/src/Cardano/Api/Typed.hs
```diff
@@ -1565,6 +1564,103 @@ data MultiSigScript = RequireSignature (Hash PaymentKey)
...
+  toJSON (RequireAnyOf reqSigs) = object [ "any" .= map toJSON reqSigs ]
+  toJSON (RequireAllOf reqSigs) = object [ "all" .= map toJSON reqSigs ]
+  toJSON (RequireMOf reqNum reqSigs) =
+    case reqNum > length reqSigs of
+      False -> object [convertMofN .= map toJSON reqSigs]
+      True -> error $ "mofn: Required number of signatures \
+                      \exceed number of payment key hashes provided. m: "
+                      ++ show reqNum ++ " n: " ++ show (length reqSigs)
+   where
+     convertMofN = Text.pack $ (show reqNum ++ "of" ++ show (length reqSigs))
```
intricate, line 1578: To stick with the `camelCase` convention for JSON keys, I would use `Of`:

```suggestion
     convertMofN = Text.pack $ (show reqNum ++ "Of" ++ show (length reqSigs))
```

So now we'd have keys like `3Of5` and `2Of6`

cardano-api/src/Cardano/Api/Typed.hs
```diff
@@ -1565,6 +1564,103 @@ data MultiSigScript = RequireSignature (Hash PaymentKey)
...
+                       _ -> return Nothing
+
+    -- Parse "any" multisig objects
+    anyReqHashes <- case HMS.lookup "any" obj of
+                       Just (Array vecAnyReqHashes) ->
+                         return . Just . RequireAnyOf $ gatherSignatures vecAnyReqHashes
+                       _ -> return Nothing
+
+    -- Parse "mofn" multisig objects
+    mofN <- return $ filterMofN obj
```
intricate, line 1595: It seems inconvenient to have the client parse `m` and `n` from the JSON key. I think it'd be easier for the typical client if we stick with a static known key, such as `mOfN`, as opposed to dynamic keys, such as `3Of5` and `2Of6`.

I wonder if something like this would be more appropriate:

```json
{
  "mOfN": {
    "m": 2,
    "n": 6,
    "keyHashes": ["examplepaymentkeyhash", "examplepaymentkeyhash", "examplepaymentkeyhash", "examplepaymentkeyhash", "examplepaymentkeyhash", "examplepaymentkeyhash"]
  }
}
```

intricate: Opinions? @erikd @newhoggy @dcoutts

cardano-api/src/Cardano/Api/Typed.hs
```diff
@@ -1565,6 +1564,103 @@ data MultiSigScript = RequireSignature (Hash PaymentKey)
...
+      (_, _, _) -> fail "Please specify only one type of JSON multisig object."
+
+gatherSignatures :: Vector Value -> [MultiSigScript]
+gatherSignatures anyList =
+  let mPaymentKeyHashes = Vector.map (\val -> RequireSignature <$> filterValue val) anyList
+  in catMaybes $ Vector.toList mPaymentKeyHashes
+ where
+  filterValue :: Value -> Maybe (Hash PaymentKey)
+  filterValue (String hpk) = Just $ convertToHash hpk
+  filterValue _ = Nothing
```
intricate, line 1612: Aren't scripts supposed to be able to contain other nested scripts and not just `RequireSignature`s?

cardano-api/src/Cardano/Api/Typed.hs
newhoggy: The above suggestion would be easier to parse from something like `jq`.

cardano-api/src/Cardano/Api/Typed.hs
Jimbo4350: Discussed on a video call. The consensus is to leave the scripts as simple as possible and adjust if user requirements change.

cardano-api/src/Cardano/Api/Typed.hs
Jimbo4350: Changed to:
```
{
    "atLeast": {
        "paymentKeyHashes": [
            "2f3d4cf10d0471a1db9f2d2907de867968c27bca6272f062cd1c2413",
            "f856c0c5839bab22673747d53f1ae9eed84afafb085f086e8e988614",
            "b275b08c999097247f7c17e77007c7010cd19f20cc086ad99d398538",
            "686024aecb5884d73a11b9ae4e63931112ba737e878d74638b78513a"
        ],
        "required": 2,
        "total": 4
    }
}
```

cardano-api/test/Golden/MultiSig/all
```diff
@@ -0,0 +1,12 @@
...
+        "e09d36c79dec9bd1b3d9e152247701cd0bb860b5ebfd1de8abb6735a",
+        "a687dcc24e00dd3caafbeb5e68f97ca8ef269cb6fe971345eb951756",
+        "0bd1d702b2e6188fe0857a6dc7ffb0675229bab58c86638ffa87ed6d",
+        "dd0044a26cf7d4491ecea720fda11afb59d5725b53afa605fdf695e6",
+        "cf223afe150cc8e89f11edaacbbd55b011ba44fbedef66fbd37d8c9d",
+        "372643e7ef4b41fd2649ada30a89d35cb90b7c14cb5de252e6ce6cb7",
+        "aa453dc184c5037d60e3fbbadb023f4a41bac112f249b76be9bb37ad",
+        "6b732c60c267bab894854d6dd57a04a94e603fcc4c36274c9ed75952"
+    ]
+}
```
intricate, line 12: We should make sure that there are newlines at EOF for each of these golden files

cardano-api/src/Cardano/Api/Typed.hs
```diff
@@ -1565,6 +1561,110 @@ data MultiSigScript = RequireSignature (Hash PaymentKey)
...
+                                 \does not equal total. " ++ " total: " ++ show total
+                                 ++ "Number of key hashes: " ++ show (length keys)
+
+  | length keys < required = error $ "required exceeds the number of payment key \
+                                   \hashes. Number of keys: " ++ show (length keys)
+                                   ++ " required: " ++ show required
+
+  | required == 1 = error $ "required is equal to one, you should use the \"any\" multisig script"
+
+  | required == total = error "required = total, you should use the \"all\" multisig script"
```
intricate, line 1582:
```suggestion
  | required == total = error "required is equal to the total, you should use the \"all\" multisig script"
```

or

```suggestion
  | required == total = error "required == total, you should use the \"all\" multisig script"
```

cardano-api/test/Test/Cardano/Api/Examples.hs
```diff
@@ -29,6 +37,62 @@ import           Shelley.Spec.Ledger.PParams (PParams' (..), emptyPParams)
...
+                   $ convertToHash "b275b08c999097247f7c17e77007c7010cd19f20cc086ad99d398538"
+               , RequireSignature
+                   $ convertToHash "686024aecb5884d73a11b9ae4e63931112ba737e878d74638b78513a"
+               ]
+
+convertToHash :: Text -> Api.Hash Api.PaymentKey
+convertToHash txt =
+  case Api.deserialiseFromRawBytesHex (Api.AsHash Api.AsPaymentKey) $ Text.encodeUtf8 txt of
+    Just payKeyHash -> payKeyHash
+    Nothing -> error $ "Test.Cardano.Api.Examples.Error deserialising payment key hash: "
```
intricate, line 93: Did you mean:

```suggestion
    Nothing -> error $ "Test.Cardano.Api.Examples: Error deserialising payment key hash: "
```

cardano-api/src/Cardano/Api/Typed.hs
```diff
@@ -1565,6 +1561,110 @@ data MultiSigScript = RequireSignature (Hash PaymentKey)
                     | RequireMOf Int [MultiSigScript]
   deriving (Eq, Show)
 
+instance ToJSON MultiSigScript where
```
intricate, line 1564: Just a note that maybe we should move this `MultiSigScript` stuff into its own module at some point.

cardano-api/src/Cardano/Api/Typed.hs
```diff
@@ -1565,6 +1561,110 @@ data MultiSigScript = RequireSignature (Hash PaymentKey)
...
+mofn val =
+  MofN $ Aeson.withObject "mofn"
+           (\o -> do mofnObj <- o .: "atLeast"
+                     required <- mofnObj .: "required"
+                     total <- mofnObj .: "total"
+                     keyHashes <- case HMS.lookup "paymentKeyHashes" mofnObj of
+                                    Just (Array keyhashes) -> return $ gatherSignatures keyhashes
+                                    _ -> return []
+                     fromJSONmOfnChecks keyHashes required total)
+           val
```
intricate, line 1637: You don't have to style it exactly like this but:

```suggestion
mofn = MofN $ Aeson.withObject "mofn" $ \o -> do 
  mofnObj <- o .: "atLeast"
  required <- mofnObj .: "required"
  total <- mofnObj .: "total"
  keyHashes <- case HMS.lookup "paymentKeyHashes" mofnObj of
    Just (Array keyhashes) -> return $ gatherSignatures keyhashes
    _ -> return []
  fromJSONmOfnChecks keyHashes required total
```

cardano-api/src/Cardano/Api/Typed.hs
```diff
@@ -1565,6 +1561,110 @@ data MultiSigScript = RequireSignature (Hash PaymentKey)
...
+-- Parse "mofn" multisig objects
+mofn :: Value -> ParseMultiSigScript
+mofn val =
+  MofN $ Aeson.withObject "mofn"
+           (\o -> do mofnObj <- o .: "atLeast"
+                     required <- mofnObj .: "required"
+                     total <- mofnObj .: "total"
+                     keyHashes <- case HMS.lookup "paymentKeyHashes" mofnObj of
+                                    Just (Array keyhashes) -> return $ gatherSignatures keyhashes
+                                    _ -> return []
```
intricate, line 1635: You might be able to use `(.:?)` and `(.!=)` here.

intricate, line 1635: You might be able to use `(.:?)` and `(.!=)` here.

cardano-api/test/Golden/MultiSig/all
Jimbo4350: Why? The golden test will fail.

cardano-api/src/Cardano/Api/Typed.hs
Jimbo4350: I agree, but I'm waiting on @dcoutts to decide on the best way to break up the whole `Typed.hs` module

cardano-api/src/Cardano/Api/Typed.hs
```diff
@@ -1565,6 +1561,110 @@ data MultiSigScript = RequireSignature (Hash PaymentKey)
...
+                     fromJSONmOfnChecks keyHashes required total)
+           val
+
+
+
+fromJSONmOfnChecks :: [MultiSigScript] -> Int -> Int -> Aeson.Parser MultiSigScript
+fromJSONmOfnChecks keys required total
+  | required == 1 = fail $ "required is equal to one, you should use the \"any\" multisig script"
```
intricate, line 1643: What happens if `required <= 0`?

cardano-api/src/Cardano/Api/Typed.hs
```diff
@@ -1565,6 +1561,110 @@ data MultiSigScript = RequireSignature (Hash PaymentKey)
...
+  | required == total = fail "required = total, you should use the \"all\" multisig script"
+  | length keys < required = fail $ "required exceeds the number of payment key hashes. \
+                                    \Number of keys: " ++ show (length keys)
+                                  ++ " required: " ++ show required
+  | length keys == total = return $ RequireMOf required keys
+  | otherwise = fail $ "Cardano.Api.Typed.fromJSONmOfnChecks failure occured. required: "
+                     ++ show required ++ " total: " ++ show total
+                     ++ " keys: " ++ show keys
```
intricate, line 1651: Should we also check whether `required <= 0` as well as whether `required > total`?

Jimbo4350: I think we should. I'll add in a check for that.

Jimbo4350:
bors merge

cardano-api/src/Cardano/Api/Typed.hs
```diff
@@ -1565,6 +1561,114 @@ data MultiSigScript = RequireSignature (Hash PaymentKey)
...
+toJSONmOfnChecks keys required total
+  | length keys /= total = error $ "The number of payment key hashes submitted \
+                                 \does not equal total. " ++ " total: " ++ show total
+                                 ++ "Number of key hashes: " ++ show (length keys)
+
+  | length keys < required = error $ "required exceeds the number of payment key \
+                                   \hashes. Number of keys: " ++ show (length keys)
+                                   ++ " required: " ++ show required
+
+  | required <= 0 = error $ "setting required to 0 is nonsensiscal. Please refer to the docs."
```
intricate, line 1580: Nitpicking about spelling as usual :smile:

```suggestion
  | required <= 0 = error $ "setting required to 0 is nonsensical. Please refer to the docs."
```

An alternative error message might be: `The required number of payment key hashes cannot be less than or equal to 0`.

Jimbo4350: :+1: Will fix in a follow up PR

PR #1663, opened 2020-08-11
Remove unused language pragmas

newhoggy:
Removed unused language pragmas and enable `hlint` rule for this.

Surprisingly we depend on `NamedFieldPuns` to suppress warnings.  See below.

### DIFF
--- a/cardano-api/src/Cardano/Api/TextView.hs
+++ b/cardano-api/src/Cardano/Api/TextView.hs
@@ -56,8 +56,8 @@ data TextView = TextView
   } deriving (Eq, Show)
 
 instance ToJSON TextView where
-  toJSON (TextView (TextViewType tvType) (TextViewDescription desc) rawCBOR) =
-    object [ "type" .= Text.decodeUtf8 tvType
+  toJSON (TextView (TextViewType tvType') (TextViewDescription desc) rawCBOR) =
+    object [ "type" .= Text.decodeUtf8 tvType'
            , "description" .= Text.decodeUtf8 desc
            , "cborHex" .= (Text.decodeUtf8 $ Base16.encode rawCBOR)
            ]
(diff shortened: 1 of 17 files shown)


cardano-api/src/Cardano/Api/TextView.hs
```diff
@@ -57,8 +56,8 @@ data TextView = TextView
   } deriving (Eq, Show)
 
 instance ToJSON TextView where
-  toJSON (TextView (TextViewType tvType) (TextViewDescription desc) rawCBOR) =
-    object [ "type" .= Text.decodeUtf8 tvType
```
newhoggy, line 61: The above actually produces a compiler warning in the absence of `NamedFieldPuns`.  Seems like `NamedFieldPuns` suppresses shadow variable warnings.  If we value the shadow variable warning, it's a good reason to avoid this extension - especially since we actually don't use the substance of this extension.

cardano-cli/src/Cardano/CLI/Shelley/Run/Pool.hs
```diff
@@ -158,8 +156,8 @@ runStakePoolRetirementCert (VerificationKeyFile sPvkeyFp) retireEpoch (OutputFil
       . newExceptT
       $ readFileTextEnvelope (AsVerificationKey AsStakePoolKey) sPvkeyFp
 
-    let stakePoolId = verificationKeyHash stakePoolVerKey
-        retireCert = makeStakePoolRetirementCertificate stakePoolId retireEpoch
+    let stakePoolId' = verificationKeyHash stakePoolVerKey
+        retireCert = makeStakePoolRetirementCertificate stakePoolId' retireEpoch
```
intricate, line 160: Did `stakePoolId` have to be renamed here?

newhoggy: Yep.  This was the error:

```
src/Cardano/CLI/Shelley/Run/Pool.hs:159:9: error: [-Wname-shadowing, -Werror=name-shadowing]
    This binding for ‘stakePoolId’ shadows the existing binding
      imported from ‘Cardano.Api.Typed’ at src/Cardano/CLI/Shelley/Run/Pool.hs:19:1-34
    |
159 |     let stakePoolId = verificationKeyHash stakePoolVerKey
    |         ^^^^^^^^^^^
```

newhoggy:
bors merge

PR #1665, opened 2020-08-11
WIP: Lazy show tvar in tests

newhoggy:
This is a demo to demonstrate a fun and potentially controversial idea and gauge the reaction.

Here is a new `newTVar` function which can be used to create a TVar and annotate its initial and final value at the point the test completes (whether success or failure).

It works by using lazy IO.  Internally it uses `unsafeInterleaveIO` to ensure the final value is not read until it is forced by `hedgehog` annotation machinery, which runs after the test completes.

Example output:
```
  ✗ prop_lazyAnnotations failed at test/Test/Cardano/Node/Chairman.hs:27:3
    after 1 test.
  
       ┏━━ test/Test/Cardano/Node/Chairman.hs ━━━
    22 ┃ prop_lazyAnnotations :: Property
    23 ┃ prop_lazyAnnotations = H.propertyOnce $ do
    24 ┃   tv <- H.newTVar @Int 0
       ┃   │ 0 -> 1
    25 ┃   void . H.evalM . liftIO . STM.atomically $ STM.writeTVar tv 1
    26 ┃   _ <- H.lazyShowTVar tv
       ┃   │ 1
    27 ┃   H.failure
       ┃   ^^^^^^^^^
```

Note that the annotation on line `24` reports the `TVar` having transitioned from its initial value of `0` to the value of `1` by the end of the test.

### DIFF
--- a/cardano-node/cardano-node.cabal
+++ b/cardano-node/cardano-node.cabal
@@ -291,6 +291,7 @@ test-suite cardano-node-test
                       , resourcet
                       , shelley-spec-ledger
                       , shelley-spec-ledger-test
+                      , stm
                       , temporary
                       , time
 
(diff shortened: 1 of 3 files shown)


newhoggy:
Looks like this won't fly for "interesting" reasons: https://mail.haskell.org/pipermail/haskell-cafe/2012-January/098453.html

I'll investigate further later when I have time.  Closing for now.

PR #1666, opened 2020-08-11
Allow to build for non-systemd systems

### LINKED ISSUE #1200: [BUG] - Build System: flag -systemd is not being respected by cabal-install
**Exernal**

**Summary**
While [there is clearly an attempt to remove `systemd` dependency in `cardano-node`](https://github.com/input-output-hk/cardano-node/blob/aba9ab8d935fea5fa291002f9ec102b621cb7cfb/cardano-node/cardano-node.cabal#L18), when trying to use it, `cabal` fails to resolve dependencies, see error below...

```
cabal: Could not resolve dependencies:
[__0] trying: lobemo-scribe-systemd-0.1.0.0 (user goal)
[__1] next goal: libsystemd-journal (dependency of lobemo-scribe-systemd)
[__1] rejecting: libsystemd-journal-1.4.4 (conflict: pkg-config package
libsystemd==209 || >209, not found in the pkg-config database)
[__1] rejecting: libsystemd-journal-1.4.3, libsystemd-journal-1.4.2,
libsystemd-journal-1.4.1, libsystemd-journal-1.4.0, libsystemd-journal-1.3.4,
libsystemd-journal-1.3.3, libsystemd-journal-1.3.1, libsystemd-journal-1.3.0,
libsystemd-journal-1.2.0, libsystemd-journal-1.1.0, libsystemd-journal-1.0.0
(constraint from project config TODO requires >=1.4.4)
[__1] fail (backjumping, conflict set: libsystemd-journal,
lobemo-scribe-systemd)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: libsystemd-journal,
lobemo-scribe-systemd
```

I am building on CentOS, using `cabal-install-3.2.0.0` using following commands:

```
RUN cabal install cardano-cli \
      --installdir=/usr/local/bin \
      --install-method=copy \
      -f-systemd && \
    cabal install cardano-node \
      --install-method=copy \
      --installdir=/usr/local/bin \
      -f-systemd
```

I have also tried `--flags="-systemd"` notation with same results [as per relevant `cabal` documentation](https://www.haskell.org/cabal/users-guide/nix-local-build.html#cfg-field-flags). You can use following **Dockerfile** to reproduce the issue:

```
FROM perlur/centos-haskell-ghcup AS builder

ENV CABAL_VERSION="3.2.0.0"
ENV GHC_VERSION="8.6.5"

ENV CARDANO_NODE_VERSION="1.13.0"

RUN yum update -y && \
    yum install -y openssl-devel zlib-devel
RUN ghcup install-cabal ${CABAL_VERSION}
RUN ghcup install ${GHC_VERSION} && \
         ghcup set ${GHC_VERSION}

WORKDIR /usr/src
RUN git clone --recurse-submodules https://github.com/input-output-hk/cardano-node
WORKDIR cardano-node
RUN git fetch --tags
RUN git checkout ${CARDANO_NODE_VERSION}


# Ugly workaround for vty-5.25.1
RUN ln -s /usr/lib64/libtinfo.so.6 /usr/lib64/libtinfo.so
RUN cabal update
RUN cabal install cardano-cli \
      --installdir=/usr/local/bin \
      --install-method=copy \
      -f-systemd && \
    cabal install cardano-node \
      --install-method=copy \
      --installdir=/usr/local/bin \
      -f-systemd

FROM perlur/centos-base

MAINTAINER "Mark Stopka <mark.stopka@perlur.cloud>"

ENV SERVICE_NAME "cardano-node"

RUN yum update -y && \
    yum clean all && \
    dnf clean all && \
    rm -rf /var/cache/yum && \
    rm -rf /var/cache/dnf

EXPOSE 3000

COPY --from=builder /usr/local/bin/cardano-node /usr/local/bin/
COPY --from=builder /usr/local/bin/cardano-cli /usr/local/bin/
COPY usr/local/lib/* /usr/local/lib/
COPY usr/local/bin/* /usr/local/bin/
```

**Expected behavior**
`cabal-install` will not require `systemd-devel` package to resolve dependencies and build.

- OS: CentOS
- Version: 8
- Node version: 1.13.0

- If applicable, add screenshots, config files and/or logs to help explain the problem.

**Additional context**

hasufell:
This will disable systemd integration for all cabal users,
but fixes #1200.

There are two other ways to fix this:

* provide a hackage overlay, which contains lobemo-scribe-systemd
  only or all 3rdparty packages and then add that overlay to
  cabal.project (downside: stack doesn't seem to be able to use
  hackage overlays)
* provide 2 cabal.project files and let user choose via e.g.:
  --project-file=cabal.project.nosystemd
  (downside: maintaining yet another file)

----

Edit: third option could be to document for non-systemd users how to edit cabal.project, but I guess that's a bit awkward?

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -279,13 +279,6 @@ source-repository-package
   location: https://github.com/input-output-hk/iohk-monitoring-framework
   tag: efa4b5ecd7f0a13124616b12679cd42517cd905a
   --sha256: 0h1h5ifl5d7dl3y6fym9pjd6rfrbh5rbyxs0xw5las503pibv2bf
-  subdir:   plugins/scribe-systemd
-
-source-repository-package
-  type: git
-  location: https://github.com/input-output-hk/iohk-monitoring-framework
-  tag: efa4b5ecd7f0a13124616b12679cd42517cd905a
-  --sha256: 0h1h5ifl5d7dl3y6fym9pjd6rfrbh5rbyxs0xw5las503pibv2bf
   subdir:   plugins/backend-aggregation
 
 source-repository-package
@@ -424,3 +417,7 @@ constraints:
 
 package comonad
   flags: -test-doctests
+
+-- force systemd off until there's a solution to
+-- https://github.com/haskell/cabal/issues/5444
+flags: -systemd


hasufell:
That's certainly an option. What about providing a hackage overlay? I find that an interesting idea in general and would be eager to look into it. But it will add more steps to release cycle.

rvl:
@hasufell Why do you need to disable systemd anyway?
You can just install the systemd development package for your Linux distribution and then the cabal build will succeed.
If you have religious objections to systemd, there is always `cabal.project.local`.

hasufell:
>  You can just install the systemd development package for your Linux distribution and then the cabal build will succeed.

1. not all distros have systemd: https://pkgs.alpinelinux.org/packages?name=systemd&branch=edge
2. on some distros you cannot have eudev and systemd installed at the same time, because they are both udev providers: https://gitlab.exherbo.org/exherbo/arbor/-/blob/aec8e96ed61483f92995680ad0e3b414934fee0a/packages/sys-apps/eudev/eudev.exlib#L34

So installing systemd is not an option for me.

disassembler:
> That doesn't work, as discussed in this PR, because of #1200 and [haskell/cabal#5444](https://github.com/haskell/cabal/issues/5444)

This seems to me like an issue that should be fixed upstream.

hasufell:
> This seems to me like an issue that should be fixed upstream.

Yes, there already is an issue open, but there is no clear timeline, so it needs a workaround.

disassembler:
I don't see a reason to address this. The work-around is you modify the cabal.proejct file (as this patch does, and ideally when cabal is fixed, just add to cabal.project.local), but this patch would break our nix builds and cabal builds for anyone that wants systemd (which there are a number of pool operators I've been helping this weekend trying to get systemd services setup on non-nixos).

hasufell:
@disassembler as @erikd pointed out, this could be done by a shell script. I don't believe users will figure this out on their own. Is there any reason to not accept a patch for that?

PR #1667, opened 2020-08-11
Update dependencies

mrBliss:
* Big refactoring of ForgeState(Info), KESInfo, ..., in consensus.

* No more mock protocols.

* Bonus: no more dependencies on `lens`.


README.rst
```diff
@@ -100,8 +100,6 @@ The ``--config`` flag points to a ``.yaml`` file that is responsible to configur
 `configuration.yaml <https://github.com/input-output-hk/cardano-node/blob/master/configuration/defaults/byron-mainnet/configuration.yaml>`_.
 Some of the more important settings are as follows:
 
-* ``NodeId: 0``  -- Used in mock protocols only to differentiate nodes.
```
mrBliss, line 103: NodeId was only used for mock protocols, I removed it wherever I could.

(We have way too many scripts, probably horribly broken).

cardano-node/src/Cardano/Node/Configuration/Logging.hs
```diff
@@ -166,8 +166,8 @@ createLoggingLayer ver nodecli@NodeCLI{configFile} = do
      -> Trace IO Text
      -> IO ()
    loggingPreInit nodeConfig logConfig switchBoard trace = do
-     Config.getEKGport logConfig >>= \p ->
-       when (p > 0) $
+     Config.getEKGBindAddr logConfig >>= \mbEndpoint ->
+       when (isJust mbEndpoint) $
```
mrBliss, line 170: @denisshevchenko This changed in https://github.com/input-output-hk/iohk-monitoring-framework/pull/570, is this correct?

cardano-node/src/Cardano/Node/Protocol/Types.hs
```diff
@@ -40,34 +37,21 @@ instance FromJSON Protocol where
...
       "Byron" -> pure ByronProtocol
       "Shelley" -> pure ShelleyProtocol
       "Cardano" -> pure CardanoProtocol
 
       -- The old names
-      "BFT" -> pure (MockProtocol MockBFT)
-    --"MockPBFT" -- same as new name
-      "Praos" -> pure (MockProtocol MockPraos)
       "RealPBFT" -> pure ByronProtocol
       "TPraos" -> pure ShelleyProtocol
```
mrBliss, line 46: We should replace `RealPBFT` and `Praos` in all the configuration files with `Byron` and `Shelley`. (Better yet, remove all outdated configuration files).

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -400,49 +399,6 @@ handleSimpleNode p trace nodeTracers npm onKernel = do
 
          when validateDB $ traceWith tracer "Performing DB validation"
 
-  --TODO: there's still lots of duplication here, when only minor things are
-  -- different
```
mrBliss, line 404: I'm happy this is gone.

cardano-node/src/Cardano/Node/TUI/Run.hs
```diff
@@ -52,16 +52,10 @@ liveViewPostSetup lvbe ncli nc = do
...
-
-        _ -> Text.pack $ "Port: " <>
-          case naPort <$> nodeAddr ncli of
-            Nothing -> "-"
-            Just port -> show port
-
+        Text.pack $ "Port: " <>
+        case naPort <$> nodeAddr ncli of
+          Nothing -> "-"
+          Just port -> show port
```
mrBliss, line 58: As the TODo above it mentions, `nodeId` is the wrong name for this.

cardano-node/src/Cardano/Tracing/Metrics.hs
```diff
@@ -68,9 +78,31 @@ instance HasKESMetricsData (ShelleyBlock c) where
...
+      -- Multiple eras can have 'KESMetricsData', e.g., Shelley and Shelley +
+      -- native assets. We don't know which era we're in, so we have to make a
+      -- choice: we choose the last era's 'KESMetricsData'.
+      --
+      -- TODO this might need revision when we have updated everything for the
+      -- first hard fork after Shelley, but before it actually happened. At
+      -- that moment, we'll trace the KESMetricsData of the upcoming era,
+      -- instead of the KESMetricsData of the current Shelley era.
+      combineKESMetrics :: [KESMetricsData] -> KESMetricsData
+      combineKESMetrics = lastDef NoKESMetricsData
```
mrBliss, line 108: This will need some attention when we do the next hard fork.

cardano-node/src/Cardano/Node/TUI/EventHandler.hs
```diff
@@ -355,7 +355,7 @@ initLiveViewState = do
     return $ LiveViewState
                 { lvsScreen                 = MainView
                 , lvsRelease                = "Release not set yet"
-                , lvsProtocol               = MockProtocol MockPBFT -- Needs a real value. Will be overwritten later.
+                , lvsProtocol               = CardanoProtocol -- Needs a real value. Will be overwritten later.
```
intricate, line 358: I wonder if this comment is still valid. @denisshevchenko any idea?

cardano-node/src/Cardano/Tracing/OrphanInstances/Consensus.hs
```diff
@@ -242,22 +240,18 @@ instance ( tx ~ GenTx blk
...
-  trTransformer = trStructuredText
-
 instance ( tx ~ GenTx blk
          , ConvertRawHash blk
          , HasTxId tx
          , LedgerSupportsProtocol blk
-         , Show (TxId tx))
+         , Show (TxId tx)
+         , Show (ForgeStateUpdateError blk)
+         , Show (CannotForge blk))
```
intricate, line 254: I'll revise the usage of `Show` in these tracing instances in a follow-up PR.

cardano-node/src/Cardano/Tracing/Metrics.hs
mrBliss: After thinking a little about this: when we do the next hard fork, we might as well use the same KES key for the current and the next era, in which case this doesn't matter much.

mrBliss:
One of the reasons CI fails is that there are warnings when compiling [`iohk-monitoring-lib-iohk-monitoring-0.1.10.1-x86_64-w64-mingw32`](https://hydra.iohk.io/build/3793069/nixlog/38), I suppose because of some CPP? @CodiePP Can you look into this? Either `-Werror` is not enabled on that repo or it doesn't build `w64-mingw32`, or *both*. I will leave out the dependency bump of `iohk-monitoring` until it is fixed.

mrBliss:
The other reason CI [failed](https://hydra.iohk.io/build/3793033/nixlog/48) was because of a flaky test, which will be fixed by #1669.

mrBliss:
bors merge

cardano-node/src/Cardano/Node/Configuration/Logging.hs
denisshevchenko: @mrBliss Yes, it is correct: https://github.com/input-output-hk/iohk-monitoring-framework/pull/570/files#diff-a4a2e19aead2886583c9bee71a24fbacR300

cardano-node/src/Cardano/Node/Protocol/Types.hs
mark-stopka: How would you then differentiate between e.q. Praos and Praos with some featured backported from Ouroboros Chronos? It would still be "Shelley" era, but with different version of Ouroboros consensus...

mrBliss: The block type, e.g., `ShelleyBlock`, determines the protocol, e.g., `TPraos`. Switching to a new protocol like Ouroboros Chronos means hard forking to a new block type.

PR #1668, opened 2020-08-12
Reorganize cardano-cli tests

erikd:


cardano-cli/test/Test/Cli/ITN.hs
```diff
@@ -1,6 +1,6 @@
 {-# LANGUAGE OverloadedStrings #-}
 
-module Test.ITN
+module Test.Cli.ITN
```
intricate, line 3: Should we stick with `Test.CLI.*` for better consistency with the rest of the project? i.e. `Cardano.CLI.*`

Although, I actually do prefer `*.Cli.*` overall.

erikd: I think we should change the others to be like this :grinning: .

newhoggy: I also happen to prefer `Cli` 🤷

cardano-cli/cardano-cli.cabal
```diff
@@ -174,56 +174,97 @@ test-suite cardano-cli-test
...
+                      , lifted-base
+                      , optparse-applicative
+                      , process
+                      , temporary
+                      , text
+                      , time
+                      , transformers-except
+                      , unordered-containers
+
+  other-modules:        Test.Golden.Shelley
```
intricate, line 223: I wonder if these should be `Test.Cli.Golden.*`.

erikd: I specifically chopped the `Cli` part out because the test exe is `cardano-cli-golden`.

The other test in that package is `cardano-cli-test` .

erikd:
bors merge

PR #1671, opened 2020-08-12
cli: Improve stake-address-info query output

erikd:


cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -387,19 +386,18 @@ data DelegationsAndRewards
...
-      delegAndRwdToJson acc k (d, r) =
-        HMS.insert
-          (toKey k)
-          (Aeson.object ["delegation" .= d, "rewardAccountBalance" .= r])
-          acc
+        :: (Ledger.Credential 'Ledger.Staking TPraosStandardCrypto, (Maybe (Hash StakePoolKey), Coin))
+        -> Aeson.Value
+      delegAndRwdToJson (k, (d, r)) =
+        Aeson.object
+          [ "address" .= toKey k
```
intricate, line 397: I would rename this `toKey` function since the resulting value is not being used as a key in a map anymore. 

Alternatively, since `toKey` isn't actually being reused, factor it out entirely and do `serialiseAddress (StakeAddress (toShelleyNetwork nw) k)`.

erikd: Renamed to `renderAddress` .

erikd:
bors merge

PR #1680, opened 2020-08-13
New isPortOpen function

newhoggy:
This function is for testing, and tests are included to check it actually works before use.

The `listenOn` function is defined to facilitate the testing of the test function.

### DIFF
--- a/cardano-node/test/Test/Common/Network.hs
+++ b/cardano-node/test/Test/Common/Network.hs
@@ -1,0 +1,51 @@
+module Test.Common.Network
+  ( isPortOpen
+  , listenOn
+  ) where
+
+import           Control.Monad
+import           Control.Monad.Fail
+import           Control.Monad.IO.Class (MonadIO, liftIO)
+import           Data.Bool
+import           Data.Either
+import           Data.Eq
+import           Data.Function
+import           Data.Functor
+import           Data.Int
+import           Data.Maybe
+import           Foreign.C.Error (Errno (..), eCONNREFUSED)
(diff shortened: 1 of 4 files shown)


newhoggy:
> The code looks fine, but the commit title should mention that this function is needed for and implemented in the test. Maybe:
> 
> > tests: Implement new isPortOpen function for testing

Fixed!

cardano-node/test/Test/Common/Network.hs
```diff
@@ -0,0 +1,49 @@
...
+
+-- | Check if it is possible to connect to a socket address
+canConnect :: SockAddr -> IO Bool
+canConnect sockAddr = IO.bracket (IO.socket AF_INET Stream 6) IO.close' $ \sock -> do
+  result <- IO.try $ IO.connect sock sockAddr
+  case result of
+    Right () -> return True
+    Left e -> if (Errno <$> IO.ioe_errno e) == Just eCONNREFUSED
+      then return False
+      else IO.throwIO e
```
rvl, line 39: If you want the tests to also work on windows, then be aware that the network exceptions are different.
We have similar code in cardano-wallet:
https://github.com/input-output-hk/cardano-wallet/blob/5c1240cc668993fcb7135f244f2b1cc9c96a3238/lib/core/src/Cardano/Wallet/Network/Ports.hs#L104

newhoggy: Your suggestion worked a charm.  Thanks!

newhoggy:
bors merge

PR #1682, opened 2020-08-13
api: Fix JSON representation of a MetaDatum ByteString value

erikd:


cardano-api/src/Cardano/Api/MetaData.hs
```diff
@@ -120,14 +129,23 @@ jsonToMetadataValue  av =
...
-        Api.TxMetaMap . List.sortOn fst <$> traverse convertPair (HashMap.toList kvs)
+        if HashMap.keys kvs /= ["hex"]
+          then Api.TxMetaMap . List.sortOn fst <$> traverse convertPair (HashMap.toList kvs)
+          else
+            case HashMap.elems kvs of
+              [Aeson.String bs] ->
+                case Base16.decode (Text.encodeUtf8 bs) of
+                  (bs2, "") -> Right $ TxMetaBytes bs2
+                  (_, bsf) -> Left $ ConversionErrBadBytes bsf
+              what -> Left $ ConversionErrExpected "[Aeson.String x]" (show what)
```
newhoggy, line 140: Probably should show `kvs` here instead of `what`, because `what` is a list of values and `kvs` is a list of key/values.

When we get an error, we need to keys to figure out why `"hex"` is not the only key.

cardano-api/src/Cardano/Api/MetaData.hs
```diff
@@ -120,14 +129,23 @@ jsonToMetadataValue  av =
...
   where
     convertPair :: (Text, Aeson.Value)
                 -> Either MetaDataJsonConversionError (TxMetadataValue, TxMetadataValue)
     convertPair (k, v) =
       (,) <$> jsonToMetadataValue (Aeson.String k) <*> jsonToMetadataValue v
 
-    -- If the list of values is a list of pairs, then return it as such.
+    -- If the list of values is a list of pairs, then return it as such, otherwise
+    -- return Nothing.
```
intricate, line 148: :+1:

cardano-api/src/Cardano/Api/MetaData.hs
erikd: Fixed.

cardano-api/src/Cardano/Api/MetaData.hs
```diff
@@ -120,14 +129,23 @@ jsonToMetadataValue  av =
...
-        Api.TxMetaMap . List.sortOn fst <$> traverse convertPair (HashMap.toList kvs)
+        if HashMap.keys kvs /= ["hex"]
+          then Api.TxMetaMap . List.sortOn fst <$> traverse convertPair (HashMap.toList kvs)
+          else
+            case HashMap.elems kvs of
+              [Aeson.String bs] ->
+                case Base16.decode (Text.encodeUtf8 bs) of
+                  (bs2, "") -> Right $ TxMetaBytes bs2
+                  (_, bsf) -> Left $ ConversionErrBadBytes bsf
+              _otherwise -> Left $ ConversionErrExpected "[Aeson.String x]" (show kvs)
```
intricate, line 140: :+1:

cardano-api/src/Cardano/Api/MetaData.hs
```diff
@@ -70,9 +73,12 @@ jsonFromMetadataValue md =
         Map _ -> Nothing
         List _ -> Nothing
         I _ -> Nothing
-        B bs -> Just ("0x" <> Text.decodeUtf8 (Base16.encode bs), jsonFromMetadataValue b)
+        B _ -> Nothing
```
intricate, line 76: Ah, this is tricky because the spec actually denotes that all of these metadata value types would be valid keys. 

That being said, it's probably the right choice to disallow these metadata value types for the JSON representation of tx metadata since it becomes a bit complicated. However, I'm not sure if it's the right decision to silently ignore them as we're doing here.

I think we should address this and also better document it in a follow up PR.

intricate: Right now, we don't seem to use `jsonFromMetadata` outside of the tests but, if we did, this could be problematic as data could be silently lost when serializing tx metadata as JSON.

erikd: Hmm, maybe I need better comments here.

erikd: Better comments added.

erikd:
bors merge

cardano-api/src/Cardano/Api/MetaData.hs
intricate: Oops, disregard my comments. I misunderstood what was going on here.

Regardless, better docs are good :smiley:

PR #1685, opened 2020-08-13
Fix invalid string literal

newhoggy:


cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -182,7 +185,7 @@ checkLiveViewrequiredTracers traceConfig = do
   if all (== True) reqTracers
   then pure ()
   else do putTextLn "for full functional 'LiveView', please turn on the following \
-                    \tracers in the configuration file: TraceBlockFetchDecisions, \\
+                    \tracers in the configuration file: TraceBlockFetchDecisions, \
```
newhoggy, line 188: This is the substantive fix.

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -73,18 +74,20 @@ import qualified Ouroboros.Consensus.Storage.ChainDB as ChainDB
...
 
+import           Cardano.Node.Configuration.Socket (SocketOrSocketInfo (..),
+                     gatherConfiguredSockets)
+import           Cardano.Node.Configuration.Topology
+import           Cardano.Node.Handlers.Shutdown
 import           Cardano.Node.Protocol (SomeConsensusProtocol (..), mkConsensusProtocol,
                      renderProtocolInstantiationError)
-import           Cardano.Node.Handlers.Shutdown
-import           Cardano.Node.Configuration.Socket (SocketOrSocketInfo (..), gatherConfiguredSockets)
-import           Cardano.Node.Configuration.Topology
```
newhoggy, line 80: Our coding style currently allows for `100` characters per line, but we're committing code past this.  Should we extend the line limit to `120`

Jimbo4350: I vote to leave the width at 100 chars

mrBliss: As a sucker for code style/formatting, I suggest letting CI enforce the use of `stylish-haskell`, like we do in consensus (https://github.com/input-output-hk/ouroboros-network/blob/7454eaca85e3ffc298e973767316ab2eac2bbe7b/.buildkite/pipeline.yml#L12). This repository is touched by too many different people who are not aware of the style conventions (I'm guilty myself!). Although I think not everybody will agree with making `stylish-haskell` mandatory :grin:.

intricate: I agree with @mrBliss

cardano-node/src/Cardano/Node/Run.hs
newhoggy: I moved the above function to its own module to avoid `CPP` language extension interference.

newhoggy:
bors merge

PR #1692, opened 2020-08-13
Remove redundant description of Output2

backuitist:


backuitist:
@intricate just wondering about IOHK policy w.r.t to external contributions, do you require some sort of CLA?

kevinhammond:
Thank you for your contribution!  There is no need for a CLA, but we assume you are happy with the code being distributed under our standard Apache licence.

PR #1704, opened 2020-08-16
WIP Less CPP

newhoggy:
Trying to reduce the amount of CPP in the code.

### DIFF
--- a/cardano-node/src/Cardano/Node/Platform/Posix.hs
+++ b/cardano-node/src/Cardano/Node/Platform/Posix.hs
@@ -1,0 +1,3 @@
+module Cardano.Node.Platform.Posix
+  (
+  ) where
--- a/cardano-node/src/Cardano/Node/Run.hs
+++ b/cardano-node/src/Cardano/Node/Run.hs
@@ -1,25 +1,30 @@
 {-# LANGUAGE BangPatterns #-}
-{-# LANGUAGE CPP #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
 
-#if !defined(mingw32_HOST_OS)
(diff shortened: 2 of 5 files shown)


cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -126,37 +110,36 @@ runNode loggingLayer npm@NodeCLI{protocolFiles} = do
         Async.uninterruptibleCancel upTimeThread
         Async.uninterruptibleCancel peersThread
 
-      LiveView -> do
-#ifdef UNIX
-        let c = llConfiguration loggingLayer
+      LiveView -> if isPosix
```
mrBliss, line 113: Note:
```haskell
    let viewmode = if isPosix then ncViewMode nc else SimpleView
```
So when `viewmode == LiveView`, it must be that `isPosix` is `True`.

cardano-node/src/Cardano/Node/Platform/Posix.hs
```diff
@@ -0,0 +1,3 @@
+module Cardano.Node.Platform.Posix
```
mrBliss, line 1: Maybe make compilation of this module conditional in the `.cabal` file? Same for `Window.hs`. Because as soon as you add platform-specific code/dependencies, it will cause compilation issues on the other platform.

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -29,62 +34,45 @@ import           Data.Version (showVersion)
 import           GHC.Clock (getMonotonicTimeNSec)
 import           Network.HostName (getHostName)
 import           Network.Socket (AddrInfo, Socket)
-import           System.Directory (canonicalizePath, makeAbsolute)
-import           System.Environment (lookupEnv)
-
-import           Cardano.BM.Data.Aggregated (Measurable (..))
-import           Paths_cardano_node (version)
-#ifdef UNIX
```
mrBliss, line 37: I'm assuming that all of these dependencies also work on non-POSIX platforms.

newhoggy:
Closed.  My try again at some future point.

PR #1709, opened 2020-08-17
Simplify validation

newhoggy:


cardano-api/src/Cardano/Api/MetaData.hs
```diff
@@ -78,44 +78,35 @@ validateTxMetadata
...
-      :: Either TxMetadataValidationError TxMetadata
-      -> MetaDatum
-      -> Either TxMetadataValidationError TxMetadata
-    foldFn lastResult metaDatum =
-      case lastResult of
-        Left err -> Left err
-        Right _ -> validate metaDatum
-
-    validate :: MetaDatum -> Either TxMetadataValidationError TxMetadata
+    validate :: MetaDatum -> Either TxMetadataValidationError ()
```
newhoggy, line 83: `Right txMd` was returned for every occurrence of `Right`, so no information is conveyed by the payload of `Right`.

I changed the type signature to return `Either TxMetadataValidationError ()` to reflect this.

cardano-api/src/Cardano/Api/MetaData.hs
```diff
@@ -78,44 +78,35 @@ validateTxMetadata
   :: TxMetadata
   -> Either TxMetadataValidationError TxMetadata
 validateTxMetadata txMd@(TxMetadata (MetaData mdMap)) =
-    Map.foldl' foldFn (Right txMd) mdMap
+    txMd <$ traverse_ validate (Map.elems mdMap)
```
newhoggy, line 81: Previously `txMd` was threaded through the implementation of the function.  I modified the code to use `txMd <$` to indicate that if validation succeeds, the value returned is always `txMd`.  Written this way it's clearer that `validateTxMetadata` will not succeed with any other value.

newhoggy: On the RHS of `<$`, we are in `Either` monad, so `traverse_ validate` means to run `validate` on every element in the list, and abort on the first failure.

cardano-api/src/Cardano/Api/MetaData.hs
```diff
@@ -78,44 +78,25 @@ validateTxMetadata
...
-            (_, Left err) -> Left err
-            _ -> validate (Map mdPairs)
-
-        List [] -> Right txMd
-        List (md:mds) ->
-          case validate md of
-            Right _ -> validate (List mds)
-            Left err -> Left err
+        Map mdPairs -> for_ mdPairs $ \(k, v) -> validate k >> validate v
+        List mds -> for_ mds validate
```
newhoggy, line 87: Use `for_` which is the same as `flip traverse_`, we can write this more intuitively.

Logically we are looping over the list or map and validating each element or key value and aborting on the first failure.

PR #1696, opened 2020-08-14
Update dependencies

mrBliss:
Highlights:
* https://github.com/input-output-hk/cardano-ledger-specs/pull/1784
* https://github.com/input-output-hk/cardano-ledger-specs/pull/1785
* https://github.com/input-output-hk/cardano-ledger-specs/pull/1779
* https://github.com/input-output-hk/cardano-ledger-specs/pull/1785
* https://github.com/input-output-hk/cardano-ledger-specs/pull/1742
* https://github.com/input-output-hk/ouroboros-network/pull/2512

PR #1711, opened 2020-08-17
Use a released version of cborg 

mrBliss:
The dependency on cborg was pinned to an unreleased revision to bring in
<https://github.com/well-typed/cborg/pull/223>. That PR has been included in
`cborg-0.2.3.0`, so we can remove the dependency pin in favour of bumping the
lower version bound on the dependency.

Instead of using 0.2.3 as the lower version bound, bump it to >= 0.2.4, as this
version is compatible with the last version of the `primitive` package, which
will soon be needed by `cardano-ledger-specs`, see
<https://github.com/input-output-hk/cardano-ledger-specs/pull/1785>.

Another advantage of using a released version is that the package will now be
stored in the global cabal store and shared between projects (although that will
also be the case for source dependencies in Cabal 3.4.0.0).

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -46,9 +46,6 @@ package ouroboros-consensus
 package ouroboros-consensus-cardano
   tests: False
 
-package ouroboros-consensus-cardano
-  tests: False
-
 package ouroboros-network
   tests: False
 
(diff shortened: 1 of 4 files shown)


mrBliss:
The commits from this PR are also included in https://github.com/input-output-hk/cardano-node/pull/1711, as they're needed to fix a dependency issue (`primitive >= 0.7.1.0`). So if #1711 gets merged first, this PR can be closed.

nc6:
No longer needed, since #1696 was merged

mrBliss:
Ughh, I mentioned the wrong PR in my last comment :facepalm:. Thanks for cleaning this up.

PR #1712, opened 2020-08-17
When checking if can connect to socket, consider WSAECONNRESET as false

newhoggy:
To fix this flaky test problem: https://hydra.iohk.io/build/3848276/nixlog/1


newhoggy:
bors merge

PR #1733, opened 2020-08-20
Enable more hlint rules

newhoggy:
Enable the following default `hlint` rules:

* Move brackets to avoid $
* Redundant bracket
* Redundant flip
* Use camelCase
* Use newtype instead of data
* Redundant $
* Reduce duplication
* Use first
* Use record patterns
* Use :
* Use const

In some cases, I've selectively re-enabled the rule on a file-by-file bases by adding `HLINT ignore` pragmas to the file.  I did this in cases where it seemed stylistically sensible to relax those rules in those circumstances.  For example we break the camelCase rule in function names of property tests.

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -7,18 +7,7 @@
 
 
 # Warnings currently triggered by your code
-- ignore: {name: "Move brackets to avoid $"}
-- ignore: {name: "Redundant bracket"}
-- ignore: {name: "Redundant flip"}
-- ignore: {name: "Use camelCase"}
-- ignore: {name: "Use newtype instead of data"}
-- ignore: {name: "Redundant $"}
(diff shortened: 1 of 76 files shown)


erikd:
Probably worth squashing those down to a single commit, and listing the current commit messages as points.

mrBliss:
Is HLint enforced by CI? If not, I'm afraid you'll have to clean up HLint warnings created by others from time to time.

newhoggy:
> Is HLint enforced by CI? If not, I'm afraid you'll have to clean up HLint warnings created by others from time to time.

HLint is already enforced by CI.

erikd (approved):

newhoggy:
bors merge

PR #1735, opened 2020-08-20
Make the multisig script syntax recursive

Jimbo4350:


doc/reference/multisig.md
```diff
@@ -0,0 +1,63 @@
...
+        "622be5fab3b5c3f371a50a535e4d3349c942a98cecee93b24e2fd11d"
+    ]
+}
+```
+
+## atLeast
+
+The "atLeast" key indicates that in order to spend this tx output, we require "atLeast" 2 corresponding signatures from the list of payment key hashes. The "required" key indicates the minimum number of signatures we need to spend the tx output.
+
+```
```
newhoggy, line 51: Change to ````json` and also for above examples.

cardano-api/src/Cardano/Api/Typed.hs
```diff
@@ -1561,111 +1562,129 @@ data MultiSigScript = RequireSignature (Hash PaymentKey)
...
+                                     ++ " required: " ++ show required
+    | length keys == total = RequireMOf required keys
+    | otherwise = error $ "Cardano.Api.Typed.fromJSONmOfnChecks failure occured. required: "
+                        ++ show required ++ " total: " ++ show total
+                        ++ " keys: " ++ show keys
+
+  convertToHash :: Text -> Hash PaymentKey
+  convertToHash txt = case deserialiseFromRawBytesHex (AsHash AsPaymentKey) $ Text.encodeUtf8 txt of
+                        Just payKeyHash -> payKeyHash
+                        Nothing -> error $ "Error deserialising payment key hash: " <> Text.unpack txt
```
newhoggy, line 1687: Maybe we should consider using `error` less because it means users of the API will be unable to handle this error programmatically (except maybe in `IO` by catching an exception).

A fix would be returning `Either Text (Hash PaymentKey)` instead.

erikd (approved):

doc/reference/multisig.md
Jimbo4350: Good shout! I wasn't aware of this :+1:

cardano-api/src/Cardano/Api/Typed.hs
Jimbo4350: In this instance I think its ok because it's at the top level close to the JSON parsing.

PR #1744, opened 2020-08-20
fix pledge in test cluster

disassembler:


nix/supervisord-cluster/default.nix
```diff
@@ -336,7 +336,7 @@ let
     ${pkgs.python3Packages.supervisor}/bin/supervisord --config ${supervisorConfig} $@
     while [ ! -S $CARDANO_NODE_SOCKET_PATH ]; do echo "Waiting 5 seconds for bft node to start"; sleep 5; done
     echo "Transfering genesis funds to pool owners, register pools and delegations"
-    cardano-cli shelley transaction submit \
+    cardano-cli shelley transaction submit --shelley-mode \
```
newhoggy, line 339: Does our test cluster run in Shelley only mode?

dorin100: @disassembler can you please help us to merge this PR? The scripts need this in order to provide accurate results. Thank you.

PR #1753, opened 2020-08-24
Fix instructions for installing nix

newhoggy:
As per https://github.com/input-output-hk/cardano-node/pull/1752

### DIFF
--- a/doc/getting-started/building-the-node-using-nix.md
+++ b/doc/getting-started/building-the-node-using-nix.md
@@ -10,7 +10,7 @@ than installing on Ubuntu/Debian.
 The [Nix Package Manager][nix] can be installed on most Linux distributions by downloading and
 running the installation script:
 ```
-curl https://nixos.org/nix/install > install-nix.sh
+curl -L https://nixos.org/nix/install > install-nix.sh
 ./install-nix.sh
 ```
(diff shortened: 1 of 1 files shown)


doc/getting-started/building-the-node-using-nix.md
```diff
@@ -10,7 +10,7 @@ than installing on Ubuntu/Debian.
 The [Nix Package Manager][nix] can be installed on most Linux distributions by downloading and
 running the installation script:
 ```
-curl https://nixos.org/nix/install > install-nix.sh
+curl -L https://nixos.org/nix/install > install-nix.sh
```
erikd, line 13: We should uses the long version `--location` of the argument just for documentation sake.

erikd (changes requested):

doc/getting-started/building-the-node-using-nix.md
newhoggy: This is how `nix` documents the install step.  It's probably best to follow their lead on this.

newhoggy: image

intricate:
Closing as duplicate of #1752

PR #1755, opened 2020-08-24
Remove changelog from extra-source-files.  This is necessary for 'cabal install' to work

newhoggy:


mrBliss (commented):
If there is no `ChangeLog.md` and the tool itself is not meant for public use, I would remove the `extra-source-files:     ChangeLog.md` line from the cabal file instead of adding an empty `ChangeLog.md`.

newhoggy:
Removed as suggested.

mrBliss (approved):
Thanks for taking care of this!

newhoggy:
Thanks!

newhoggy:
bors merge

PR #1773, opened 2020-08-27
A version of shelley testnet lite script based on mkfiles.sh

newhoggy:


Jimbo4350 (approved):
LGTM!

cardano-node-chairman/cardano-node-chairman.cabal
```diff
@@ -53,9 +54,12 @@ test-suite tests
                       , aeson
                       , async
                       , cardano-node-chairman
+                      , containers
                       , directory
                       , filepath
                       , hedgehog
+                      , lens
+                      , lens-aeson
```
mrBliss, line 62: Would it be possible to just use [`microlens`](http://hackage.haskell.org/package/microlens) instead of `lens`? I recently managed to remove the dependency on `lens` from all downstream repos.

(Not that I hate lenses, I just don't want the whole dependency tree)

newhoggy: At the moment, I'm just using the fastest path to getting this to work.  I can explore other options once I've completed the translation of the shell script to Haskell.

newhoggy: I'm also going to move this to a new PR and merge only the shell script.

newhoggy:
bors merge

PR #1784, opened 2020-08-31
Add command for creating genesis key delegation certificates

intricate:
Add a command for creating genesis key delegation certificates.


cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
```diff
@@ -303,3 +312,21 @@ newtype TxFile
...
+  | VerificationKeyFilePath !VerificationKeyFile
+  -- ^ A path to a verification key file.
+  -- Note that this file hasn't been validated at all (whether it exists,
+  -- contains a key of the correct type, etc.)
+
+deriving instance (Show (VerificationKey keyrole), Show (Hash keyrole))
+  => Show (VerificationKeyOrHashOrFile keyrole)
+
+deriving instance (Eq (VerificationKey keyrole), Eq (Hash keyrole))
+  => Eq (VerificationKeyOrHashOrFile keyrole)
```
newhoggy, line 332: Is this the same as `deriving (Eq, Show)`?

newhoggy (approved):

cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
intricate: Yeah, but I couldn't do that directly since I needed the constraints: `Eq (VerificationKey keyrole), Eq (Hash keyrole)`.

PR #1786, opened 2020-09-01
Add TraceBlockchainTime to config files

newhoggy:
`TraceBlockchainTime` is a configuration that was added back in June, but was not mentioned in documentation or in config files found within the project.

This setting must be set to `True` in order to see the "until genesis start time at" when the start time is set to the future.

This PR makes those additions.

Additionally, the `Byron` chairman test now asserts that the text "until genesis start time at" appears in the log file when the start time is in the future.

### DIFF
--- a/cardano-node-chairman/src/Chairman/IO/File.hs
+++ b/cardano-node-chairman/src/Chairman/IO/File.hs
@@ -1,0 +1,15 @@
+module Chairman.IO.File
+  ( fileContains
+  ) where
+
+import           Data.Bool
+import           Data.Functor
+import           Data.String (String)
+import           System.IO (FilePath, IO)
+
+import qualified Data.List as L
+import qualified System.IO as IO
+
+-- | Determine if the given string is found in the given file.
+fileContains :: String -> FilePath -> IO Bool
+fileContains text path = (text `L.isInfixOf`) <$> IO.readFile path
--- a/cardano-node-chairman/test/Test/Cardano/Node/Chairman/Byron.hs
+++ b/cardano-node-chairman/test/Test/Cardano/Node/Chairman/Byron.hs
@@ -16,12 +16,14 @@ import           Data.Int
 import           Data.Maybe
 import           Data.Ord
 import           Data.Semigroup
+import           Data.String (String)
 import           GHC.Num
 import           Hedgehog (Property, discover)
 import           System.IO (IO)
 import           Text.Show
 
 import qualified Chairman.Base as H
+import qualified Chairman.IO.File as IO
 import qualified Chairman.IO.Network.Socket as IO
 import qualified Chairman.IO.Network.Sprocket as IO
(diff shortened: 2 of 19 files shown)


cardano-node-chairman/src/Chairman/IO/File.hs
```diff
@@ -0,0 +1,16 @@
...
+
+import           Data.Bool
+import           Data.Functor
+import           Data.String (String)
+import           System.IO (FilePath, IO)
+
+import qualified Data.List as L
+import qualified System.IO as IO
+
+-- | Determin if the given string is found in the given file.
```
mrBliss, line 13:
```suggestion
-- | Determine if the given string is found in the given file.
```

mrBliss (commented):
Thank you for adding this! I have wondered a few times why it was/went missing but never investigated.

cardano-node-chairman/src/Chairman/IO/File.hs
newhoggy: Fixed.  Thanks!

cardano-node-chairman/test/Test/Cardano/Node/Chairman/Byron.hs
```diff
@@ -88,7 +95,9 @@ prop_spawnByronCluster = H.propertyOnce . H.workspace "chairman" $ \tempAbsPath
     H.createDirectoryIfMissing $ tempBaseAbsPath <> "/" <> socketDir
 
     H.copyFile (baseConfig <> "/topology-node-" <> si <> ".json") (tempAbsPath <> "/topology-node-" <> si <> ".json")
-    H.copyFile (baseConfig <> "/config-" <> si <> ".yaml") (tempAbsPath <> "/config-" <> si <> ".yaml")
+    H.writeFile (tempAbsPath <> "/config-" <> si <> ".yaml") . L.unlines . fmap rewriteConfiguration . L.lines =<<
```
Jimbo4350, line 98: Wouldn't this potentially turn off all logging and only leave on: `TraceBlockchainTime: True`

Jimbo4350 (commented):
LGTM but just one comment before I approve

cardano-node-chairman/test/Test/Cardano/Node/Chairman/Byron.hs
newhoggy: All the other logging would be as per the original configuration files: `configuration/chairman/defaults/simpleview/config-*.yaml`

The original configuration file has a few trace options turned, and those lines are preserved.

I will use this rewrite mechanism to turn on any additional logging as I need it.

Jimbo4350: Ah I misread the code.

Jimbo4350 (approved):

newhoggy:
bors merge

PR #1790, opened 2020-09-01
Add CLI functionality for reading Bech32 and hex signing keys

intricate:
At the moment, `cardano-cli` commands only accept signing keys as "text envelope" formatted files.

This PR allows signing keys to be specified as files containing other data formats such as Bech32 and hex.

### DIFF
--- a/cardano-api/src/Cardano/Api/Typed.hs
+++ b/cardano-api/src/Cardano/Api/Typed.hs
@@ -184,6 +184,7 @@ module Cardano.Api.Typed (
     SerialiseAsBech32,
     serialiseToBech32,
     deserialiseFromBech32,
+    deserialiseAnyOfFromBech32,
     Bech32DecodeError(..),
 
     -- ** Addresses
@@ -2863,6 +2864,47 @@ deserialiseFromBech32 asType bech32Str = do
 
     return value
 
+deserialiseAnyOfFromBech32
+  :: forall b.
+     [FromSomeType SerialiseAsBech32 b]
+  -> Text
+  -> Either Bech32DecodeError b
+deserialiseAnyOfFromBech32 types bech32Str = do
+    (prefix, dataPart) <- Bech32.decodeLenient bech32Str
+                            ?!. Bech32DecodingError
+
+    let actualPrefix = Bech32.humanReadablePartToText prefix
+
+    FromSomeType actualType fromType <-
+      findForPrefix actualPrefix
+        ?! Bech32UnexpectedPrefix actualPrefix permittedPrefixes
+
+    payload <- Bech32.dataPartToBytes dataPart
+                 ?! Bech32DataPartToBytesError (Bech32.dataPartToText dataPart)
+
+    value <- deserialiseFromRawBytes actualType payload
+               ?! Bech32DeserialiseFromBytesError payload
+
+    let expectedPrefix = bech32PrefixFor value
+    guard (actualPrefix == expectedPrefix)
+      ?! Bech32WrongPrefix actualPrefix expectedPrefix
+
+    return (fromType value)
+  where
+    findForPrefix
+      :: Text
+      -> Maybe (FromSomeType SerialiseAsBech32 b)
+    findForPrefix prefix =
+      find
+        (\(FromSomeType t _) -> prefix `elem` bech32PrefixesPermitted t)
+        types
+
+    permittedPrefixes :: [Text]
+    permittedPrefixes =
+      concat
+        [ bech32PrefixesPermitted ttoken
+        | FromSomeType ttoken _f <- types
+        ]
 
 -- | Bech32 decoding error.
 data Bech32DecodeError =
--- a/cardano-cli/src/Cardano/CLI/Shelley/Key.hs
+++ b/cardano-cli/src/Cardano/CLI/Shelley/Key.hs
@@ -1,0 +1,161 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+-- | Shelley CLI option data types and functions for cryptographic keys.
+module Cardano.CLI.Shelley.Key
+  ( SigningKeyDecodeError (..)
+  , readSigningKeyFile
+  , readSigningKeyFileAnyOf
+  ) where
+
+import           Cardano.Prelude
+
+import           Control.Monad.Trans.Except.Extra (handleIOExceptT)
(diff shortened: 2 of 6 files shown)


cardano-api/src/Cardano/Api/Typed.hs
```diff
@@ -2862,6 +2863,49 @@ deserialiseFromBech32 asType bech32Str = do
...
+  :: [FromSomeType SerialiseAsBech32 b]
+  -> Text
+  -> Either Bech32DecodeError b
+deserialiseAnyOfFromBech32 types bech32Str = go [] types
+  where
+    go
+      :: [Bech32DecodeError]
+      -> [FromSomeType SerialiseAsBech32 b]
+      -> Either Bech32DecodeError b
+    go errs (FromSomeType ttoken fromType:ts) =
```
newhoggy, line 2876: Seems like `errs` accumulates data, but ultimately always gets thrown away?

intricate: It isn't always thrown away, see the `go errs []` case.

intricate: I've made some changes to this function, so marking this thread as "resolved".

cardano-api/src/Cardano/Api/Typed.hs
```diff
@@ -2862,6 +2863,47 @@ deserialiseFromBech32 asType bech32Str = do
...
+      ?! Bech32WrongPrefix actualPrefix expectedPrefix
+
+    return (fromType value)
+  where
+    findForPrefix
+      :: Text
+      -> Maybe (FromSomeType SerialiseAsBech32 b)
+    findForPrefix prefix =
+      find
+        (\(FromSomeType t _) -> prefix `elem` bech32PrefixesPermitted t)
```
Jimbo4350, line 2898: Where is `bech32PrefixesPermitted` coming from?

intricate: `bech32PrefixesPermitted` is a function of the `SerialiseAsBech32` typeclass

cardano-cli/src/Cardano/CLI/Shelley/Key.hs
```diff
@@ -0,0 +1,161 @@
...
+  deriving (Eq, Show)
+
+-- | Render an error message for a 'SigningKeyDecodeError'.
+renderSigningKeyDecodeError :: SigningKeyDecodeError -> Text
+renderSigningKeyDecodeError err =
+  case err of
+    SigningKeyTextEnvelopeError textEnvErr ->
+      Text.pack (displayError textEnvErr)
+    SigningKeyBech32DecodeError decodeErr ->
+      Text.pack (displayError decodeErr)
```
newhoggy, line 43: The display output doesn't have any information about which of the above two constructor paths happened.  Is the error still sensible?

cardano-cli/src/Cardano/CLI/Shelley/Key.hs
```diff
@@ -0,0 +1,162 @@
...
+    SigningKeyInvalidError -> "Invalid signing key."
+
+instance Error SigningKeyDecodeError where
+  displayError = Text.unpack . renderSigningKeyDecodeError
+
+-- | Read a signing key from a file.
+--
+-- The contents of the file can either be Bech32-encoded, hex-encoded, or in
+-- the text envelope format.
+readSigningKeyFile
```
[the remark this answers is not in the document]
intricate: I think it'd probably be less user-friendly to have the user specify the encoding/formatting of their signing key file when we can already make that determination from the input.

cardano-cli/src/Cardano/CLI/Shelley/Key.hs
intricate: The result of `displayError` for those errors (`TextEnvelopeError` and `Bech32DecodeError`) is pretty detailed. 

However, I could prefix it with something like `"Error reading text envelope signing key: "` or `"Error reading Bech32 signing key: "` or something.

cardano-cli/src/Cardano/CLI/Shelley/Key.hs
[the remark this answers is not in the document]
intricate: I think it'd probably be less user-friendly to have the user specify the encoding/formatting of their signing key file when we can already make that determination from the input.

erikd (approved):

cardano-cli/src/Cardano/CLI/Shelley/Key.hs
```diff
@@ -0,0 +1,161 @@
...
+renderSigningKeyDecodeError :: SigningKeyDecodeError -> Text
+renderSigningKeyDecodeError err =
+  case err of
+    SigningKeyTextEnvelopeError textEnvErr ->
+      Text.pack (displayError textEnvErr)
+    SigningKeyBech32DecodeError decodeErr ->
+      Text.pack (displayError decodeErr)
+    SigningKeyInvalidError -> "Invalid signing key."
+
+instance Error SigningKeyDecodeError where
```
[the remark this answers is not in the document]
intricate: `Error` is a typeclass defined in `Cardano.Api.Typed` that defines a function `displayError :: e -> String`.

In this PR, you'll see that I added a constructor to the `ShelleyKeyCmdError` error type: `ShelleyKeyCmdReadSigningKeyFileError !(FileError SigningKeyDecodeError)`. 

There is no `renderFileError` function, but there is an `Error` instance of `FileError e` with the constraint `Error e`. This is why I defined an `Error` instance for `SigningKeyDecodeError`, such that we could render `FileError SigningKeyDecodeError` values with `displayError`.

However, I could write a `renderFileError` function, but we already consistently use `displayError` for rendering `FileErrors` throughout `cardano-cli`.

cardano-cli/src/Cardano/CLI/Shelley/Key.hs
```diff
@@ -0,0 +1,161 @@
...
+-- the text envelope format.
+readSigningKeyFile
+  :: forall keyrole.
+     ( HasTextEnvelope (SigningKey keyrole)
+     , SerialiseAsBech32 (SigningKey keyrole)
+     )
+  => AsType keyrole
+  -> SigningKeyFile
+  -> IO (Either (FileError SigningKeyDecodeError) (SigningKey keyrole))
+readSigningKeyFile asType (SigningKeyFile fp) =
```
intricate, line 61: I wonder if it'd make more sense to attempt to detect the encoding of the file and then try deserializing it in the appropriate way.

PR #1799, opened 2020-09-03
Copy chairman app to new cabal project

newhoggy:
This allows for the changes to be made to the new chairman app without breaking the old chairman tests.

### DIFF
--- a/cardano-node-chairman/app/Cardano/Chairman.hs
+++ b/cardano-node-chairman/app/Cardano/Chairman.hs
@@ -1,0 +1,530 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE GADTs #-}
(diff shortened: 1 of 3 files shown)


cardano-node-chairman/app/Main.hs
```diff
@@ -0,0 +1,161 @@
+{-# LANGUAGE NamedFieldPuns #-}
```
erikd, line 1: Should rename that to maybe `chariman.hs`. The leading lower case shows that it its a Haskell main and not a module.

erikd (approved):
Just think we need a rename but once that is done all good.

cardano-node-chairman/app/Main.hs
newhoggy: Should I use `cardano-node-chairman.hs` for now, since that's the name of the exe?

newhoggy:
bors merge

PR #1801, opened 2020-09-03
Improve cardano-cli shelley errors

Jimbo4350:
Shelley `cardano-cli` commands now fail in this format: `Shelley command failed: $failedcommand Error: $error`

Example
```
Shelley command failed: node issue-op-cert  Error: node-bft1/kes.vkey: node-bft1/kes.vkey: openBinaryFile: does not exist (No such file or directory)
```

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Run.hs
+++ b/cardano-cli/src/Cardano/CLI/Run.hs
@@ -39,18 +39,20 @@ data ClientCommand =
 
 data ClientCommandErrors
   = ByronClientError ByronClientCmdError
-  | ShelleyClientError ShelleyClientCmdError
+  | ShelleyClientError ShelleyCommand ShelleyClientCmdError
   deriving Show
   --TODO: We should include an AgnosticClientError
 
 runClientCommand :: ClientCommand -> ExceptT ClientCommandErrors IO ()
-runClientCommand (ByronCommand   c) = firstExceptT ByronClientError $ runByronClientCommand c
-runClientCommand (ShelleyCommand c) = firstExceptT ShelleyClientError $ runShelleyClientCommand c
-runClientCommand DisplayVersion     = runDisplayVersion
+runClientCommand (ByronCommand c) = firstExceptT ByronClientError $ runByronClientCommand c
+runClientCommand (ShelleyCommand c) = firstExceptT (ShelleyClientError c) $ runShelleyClientCommand c
+runClientCommand DisplayVersion = runDisplayVersion
 
 renderClientCommandError :: ClientCommandErrors -> Text
-renderClientCommandError (ByronClientError err) = renderByronClientCmdError err
-renderClientCommandError (ShelleyClientError err) = renderShelleyClientCmdError err
+renderClientCommandError (ByronClientError err) =
+  renderByronClientCmdError err
+renderClientCommandError (ShelleyClientError cmd err) =
+  runIdentity $ renderShelleyClientCmdError cmd err
 
 runDisplayVersion :: ExceptT ClientCommandErrors IO ()
 runDisplayVersion = do
--- a/cardano-cli/src/Cardano/CLI/Run.hs
+++ b/cardano-cli/src/Cardano/CLI/Run.hs
@@ -52,7 +52,7 @@ renderClientCommandError :: ClientCommandErrors -> Text
 renderClientCommandError (ByronClientError err) =
   renderByronClientCmdError err
 renderClientCommandError (ShelleyClientError cmd err) =
-  runIdentity $ renderShelleyClientCmdError cmd err
+  renderShelleyClientCmdError cmd err
 
 runDisplayVersion :: ExceptT ClientCommandErrors IO ()
 runDisplayVersion = do
--- a/cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
+++ b/cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
@@ -16,6 +16,7 @@ module Cardano.CLI.Shelley.Commands
   , GovernanceCmd (..)
   , GenesisCmd (..)
   , TextViewCmd (..)
+  , renderShelleyCommand
 
     -- * CLI flag types
   , AddressKeyType (..)
@@ -74,6 +75,19 @@ data ShelleyCommand
   | TextViewCmd     TextViewCmd
   deriving (Eq, Show)
 
+renderShelleyCommand :: ShelleyCommand -> Text
+renderShelleyCommand sc =
+  case sc of
+    AddressCmd cmd -> renderAddressCmd cmd
+    StakeAddressCmd cmd -> renderStakeAddressCmd cmd
+    KeyCmd cmd -> renderKeyCmd cmd
+    TransactionCmd cmd -> renderTransactionCmd cmd
+    NodeCmd cmd -> renderNodeCmd cmd
+    PoolCmd cmd -> renderPoolCmd cmd
+    QueryCmd cmd -> renderQueryCmd cmd
+    GovernanceCmd cmd -> renderGovernanceCmd cmd
+    GenesisCmd cmd -> renderGenesisCmd cmd
+    TextViewCmd cmd -> renderTextViewCmd cmd
 
 data AddressCmd
   = AddressKeyGen AddressKeyType VerificationKeyFile SigningKeyFile
@@ -84,6 +98,17 @@ data AddressCmd
   | AddressInfo Text (Maybe OutputFile)
   deriving (Eq, Show)
 
+
+renderAddressCmd :: AddressCmd -> Text
+renderAddressCmd cmd =
+  case cmd of
+    AddressKeyGen {} -> "address key-gen"
+    AddressKeyHash {} -> "address key-hash"
+    AddressBuild {} -> "address build"
+    AddressBuildMultiSig {} -> "address build-multisig"
+    AddressInfo {} -> "address info"
+    AddressBuildScript {} -> "address build-script"
+
 data StakeAddressCmd
   = StakeAddressKeyGen VerificationKeyFile SigningKeyFile
   | StakeAddressKeyHash VerificationKeyFile (Maybe OutputFile)
@@ -92,6 +117,16 @@ data StakeAddressCmd
   | StakeKeyDelegationCert VerificationKeyFile StakePoolVerificationKeyHashOrFile OutputFile
   | StakeKeyDeRegistrationCert VerificationKeyFile OutputFile
   deriving (Eq, Show)
+
+renderStakeAddressCmd :: StakeAddressCmd -> Text
+renderStakeAddressCmd cmd =
+  case cmd of
+    StakeAddressKeyGen {} -> "stake-address key-gen"
+    StakeAddressKeyHash {} -> "stake-address key-hash"
+    StakeAddressBuild {} -> "stake-address build"
+    StakeKeyRegistrationCert {} -> "stake-address registration-certificate"
+    StakeKeyDelegationCert {} -> "stake-address delegation-certificate"
+    StakeKeyDeRegistrationCert {} -> "stake-address deregistration-certificate"
 
 data KeyCmd
   = KeyGetVerificationKey SigningKeyFile VerificationKeyFile
@@ -102,6 +137,17 @@ data KeyCmd
   | KeyConvertITNExtendedToStakeKey SomeKeyFile OutputFile
   | KeyConvertITNBip32ToStakeKey SomeKeyFile OutputFile
   deriving (Eq, Show)
+
+renderKeyCmd :: KeyCmd -> Text
+renderKeyCmd cmd =
+  case cmd of
+    KeyGetVerificationKey {} -> "key verification-key"
+    KeyNonExtendedKey {} -> "key non-extended-key"
+    KeyConvertByronKey {} -> "key convert-byron-key"
+    KeyConvertByronGenesisVKey {} -> "key convert-byron-genesis-key"
+    KeyConvertITNStakeKey {} -> "key convert-itn-key"
+    KeyConvertITNExtendedToStakeKey {} -> "key convert-itn-extended-key"
+    KeyConvertITNBip32ToStakeKey {} -> "key convert-itn-bip32-key"
 
 data TransactionCmd
   = TxBuildRaw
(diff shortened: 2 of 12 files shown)


cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
```diff
@@ -84,6 +98,17 @@ data AddressCmd
...
+
+renderAddressCmd :: AddressCmd -> Text
+renderAddressCmd cmd =
+  case cmd of
+    (AddressKeyGen _ _ _) -> "address key-gen"
+    (AddressKeyHash _ _) -> "address key-hash"
+    (AddressBuild _ _ _ _) -> "address build"
+    (AddressBuildMultiSig) -> "address build-multisig"
+    (AddressInfo _ _) -> "address info"
+    (AddressBuildScript _ _ _) -> "address build-script"
```
intricate, line 110: I'm hesitant of doing it this way when the strings used in `Cardano.CLI.Shelley.Parsers` could diverge from these.

intricate: cc @erikd @newhoggy

newhoggy: They already diverge somewhat already don't they?  Do you expect the above strings to be embedded in the program description?

intricate: It seems like these strings are supposed to be the same as the subcommand strings, e.g. `cardano-cli shelley address key-gen`. If we were to make a change to one of those in `Cardano.CLI.Shelley.Parsers`, then we'd need to remember to also make that change here. Maybe it's not a big problem, but I thought it'd be worth mentioning.

newhoggy: I had another look and it seems the text is meant to resemble the command used in the CLI not the program description.  And that both should resemble the constructor name, so it's probably fine?

erikd: `(AddressKeyGen _ _ _) -> ` will break as soon as parameter is added or removed.

`AddressKeyGen {}` wil not break.

cardano-cli/src/Cardano/CLI/Shelley/Run.hs
```diff
@@ -36,19 +36,36 @@ data ShelleyClientCmdError
...
-    ShelleyCmdGovernanceError govCmdErr -> renderShelleyGovernanceError govCmdErr
-    ShelleyCmdNodeError nodeCmdErr -> renderShelleyNodeCmdError nodeCmdErr
-    ShelleyCmdPoolError poolCmdErr -> renderShelleyPoolCmdError poolCmdErr
-    ShelleyCmdStakeAddressError stakeAddrCmdErr -> renderShelleyStakeAddressCmdError stakeAddrCmdErr
-    ShelleyCmdTextViewError txtViewErr -> renderShelleyTextViewFileError txtViewErr
-    ShelleyCmdTransactionError txErr -> renderShelleyTxCmdError txErr
-    ShelleyCmdQueryError queryErr -> renderShelleyQueryCmdError queryErr
-    ShelleyCmdKeyError keyErr -> renderShelleyKeyCmdError keyErr
+-- Identity monad is used to avoid boilerplate
+renderShelleyClientCmdError :: ShelleyCommand -> ShelleyClientCmdError -> Identity Text
```
erikd, line 40: Why the identitiy monad?

cardano-cli/src/Cardano/CLI/Shelley/Run.hs
```diff
@@ -36,19 +36,36 @@ data ShelleyClientCmdError
...
+                  return $ renderShelleyTxCmdError txErr
+                ShelleyCmdQueryError queryErr ->
+                  return $ renderShelleyQueryCmdError queryErr
+                ShelleyCmdKeyError keyErr ->
+                  return $ renderShelleyKeyCmdError keyErr
+
+  return $ "Shelley command failed: "
+         <> renderShelleyCommand cmd
+         <> "  Error: "
+         <> cmdError
```
erikd, line 67: `mapend` is great for two components. As soo as its more, `mconcat` makes more sense (although with laziness probably makes close to zero difference).

cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
newhoggy: I somewhat agree.  Which is better depends on whether you want it to break.  Probably `hlint` suggested using `{}`, if we know better, we can suppress with a `HLINT` pragma.  There are examples in many Haskell files.

erikd (changes requested):
A couple of minor changes needed.

cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
erikd: Hlint is correct in this case :D

cardano-cli/src/Cardano/CLI/Shelley/Run.hs
newhoggy: Is this primarily a performance concern?

cardano-cli/src/Cardano/CLI/Shelley/Run.hs
Jimbo4350: I used the `Identity` monad to avoid this boilerplate:
```
case err  of 
  ShelleyCmdADdressError addrCmdErr -> "Shelley command failed:" <> renderShelleyCommand cmd
                                                                                                               <> "  Error: "
                                                                                                               <> renderShelleyAddressCmdError addrCmdErr
  ShelleyCmdGenesisError genesisCmdErr -> "Shelley command failed:" <> renderShelleyCommand cmd
                                                                                                               <> "  Error: "
                                                                                                               <> renderShelleyGenesisCmdError addrCmdErr
```

cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
```diff
@@ -211,6 +305,19 @@ data GenesisCmd
...
+  case cmd of
+    GenesisCreate {} -> "genesis create"
+    GenesisKeyGenGenesis {} -> "genesis key-gen-genesis"
+    GenesisKeyGenDelegate {} -> "genesis key-gen-delegate"
+    GenesisKeyGenUTxO {} -> "genesis key-gen-utxo"
+    GenesisCmdKeyHash {} -> "genesis key-hash"
+    GenesisVerKey {} -> "genesis get-ver-key"
+    GenesisTxIn {} -> "genesis initial-txin"
+    GenesisAddr {} -> "genesis initial-addr"
+    GenesisHashFile {} -> "genesis hash"
```
newhoggy, line 319: Actually, I'm curious how in this case constructor name and the RHS text don't match.

Jimbo4350: We can change the `Strings` to match the constructor names but the `Strings` are meant to match the `cardano-cli` commands

cardano-cli/src/Cardano/CLI/Shelley/Run.hs
```diff
@@ -36,19 +36,36 @@ data ShelleyClientCmdError
...
-    ShelleyCmdGenesisError genesisCmdErr -> renderShelleyGenesisCmdError genesisCmdErr
-    ShelleyCmdGovernanceError govCmdErr -> renderShelleyGovernanceError govCmdErr
-    ShelleyCmdNodeError nodeCmdErr -> renderShelleyNodeCmdError nodeCmdErr
-    ShelleyCmdPoolError poolCmdErr -> renderShelleyPoolCmdError poolCmdErr
-    ShelleyCmdStakeAddressError stakeAddrCmdErr -> renderShelleyStakeAddressCmdError stakeAddrCmdErr
-    ShelleyCmdTextViewError txtViewErr -> renderShelleyTextViewFileError txtViewErr
-    ShelleyCmdTransactionError txErr -> renderShelleyTxCmdError txErr
-    ShelleyCmdQueryError queryErr -> renderShelleyQueryCmdError queryErr
-    ShelleyCmdKeyError keyErr -> renderShelleyKeyCmdError keyErr
+-- Identity monad is used to avoid boilerplate
```
newhoggy, line 39: Can you elaborate?  Is the use of the `Identity` monad for the benefit of the caller or to improve the implementation of the function?

cardano-cli/src/Cardano/CLI/Shelley/Run.hs
newhoggy, line 67: If it's for a better way to implement the `renderShelleyClientCmdError` function, the shocking thing is you don't need a monad to use the `do` notation.  For example this works with no monad in sight:

```
moo :: Int -> String
moo n = do
  x <- case n of
    1 -> "1"
    2 -> "2"
  return (x <> "!")
```

cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
newhoggy: I was thinking to change the constructor names to match the strings.

newhoggy:
Do you know the origins of this duplication in the string?  "Error: node-bft1/kes.vkey: node-bft1/kes.vkey"

Is there a reason we want this duplication?

cardano-cli/src/Cardano/CLI/Shelley/Run.hs
Jimbo4350: Isn't `String` = `[Char]` so we would be in the list monad?

cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
erikd: Sorry, I do not understand, what is the question?

cardano-cli/src/Cardano/CLI/Shelley/Run.hs
erikd: If you remove the `Indentity` monad, the function becomes pure.

cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
Jimbo4350: > I was thinking to change the constructor names to match the strings, although not necessarily in this PR.
Yeah I don't mind doing this in a follow up PR.

cardano-cli/src/Cardano/CLI/Shelley/Run.hs
erikd: What is wrong with
```
mconcat [ "Shelley command failed: ". renderShelleyCommand cmd, "  Error: ",  cmdError ]
``` 
?

cardano-cli/src/Cardano/CLI/Run.hs
```diff
@@ -52,7 +52,7 @@ renderClientCommandError :: ClientCommandErrors -> Text
 renderClientCommandError (ByronClientError err) =
   renderByronClientCmdError err
 renderClientCommandError (ShelleyClientError cmd err) =
-  runIdentity $ renderShelleyClientCmdError cmd err
+  renderShelleyClientCmdError cmd err
```
erikd, line 55: \o/

cardano-cli/src/Cardano/CLI/Shelley/Run.hs
```diff
@@ -36,35 +36,37 @@ data ShelleyClientCmdError
...
+    ShelleyCmdKeyError keyErr ->
+       renderError cmd renderShelleyKeyCmdError keyErr
+ where
+   renderError :: ShelleyCommand -> (a -> Text) -> a -> Text
+   renderError shelleyCmd renderer shelCliCmdErr =
+      mconcat [ "Shelley command failed: "
+              , renderShelleyCommand shelleyCmd
+              , "  Error: "
+              , renderer shelCliCmdErr
+              ]
```
erikd, line 69: Still a **LOT** better, but how about putting the `case err of .....` in a `where` statement and then having the function body as:
```
  mconat [ .... ]
```

cardano-cli/src/Cardano/CLI/Shelley/Run.hs
newhoggy: Nothing wrong.  Just more of a curiosity.

newhoggy: > Isn't `String` = `[Char]` so we would be in the list monad? I've also re-pushed some changes to get rid of the `Identity` monad. Would probably be easier to look at the diff and see my rational for using `Identity`

I made a typo in the origin code where I used `<-` and `return`.  If you remove all the monad things, `do` still works.  I don't actually write code non-monadic code with a `do` block, and not suggesting it.  😁

cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
Jimbo4350: @newhoggy at this point we're only concerned with the name of the command, not necessarily it's arguments so `{}` is fine for now.

Jimbo4350:
> Do you know the origins of the duplication in the string? "Error: node-bft1/kes.vkey: node-bft1/kes.vkey"
> 
> Is there a reason we want this duplication?

It's from:
```
instance Error e => Error (FileError e) where
  displayError (FileIOError path ioe) =
    path ++ ": " ++ displayException ioe
  displayError (FileError path e) =
    path ++ ": " ++ displayError e
```

newhoggy:
> > Do you know the origins of the duplication in the string? "Error: node-bft1/kes.vkey: node-bft1/kes.vkey"
> > Is there a reason we want this duplication?
> 
> It's from the typed api:
> 
> ```
> instance Error e => Error (FileError e) where
>   displayError (FileIOError path ioe) =
>     path ++ ": " ++ displayException ioe
>   displayError (FileError path e) =
>     path ++ ": " ++ displayError e
> ```

Thanks!

newhoggy (approved):

cardano-cli/src/Cardano/CLI/Shelley/Run.hs
Jimbo4350: https://github.com/input-output-hk/cardano-node/pull/1801#discussion_r484230581

cardano-cli/src/Cardano/CLI/Shelley/Run.hs
erikd: Nope I was wrong.

erikd (approved):

cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
newhoggy: > Sorry, I do not understand, what is the question?

For example in the code:

```
GenesisHashFile {} -> "genesis hash"
```

The words on the LHS are not the same as the RHS, but it seemed like in most cases it is and perhaps would be good if it consistently was the case.

PR #1803, opened 2020-09-03
CAD-1765 stack on 1.19.1

CodiePP:
stack compiles Release 1.19.1 again

```
$ stack exec cardano-node -- --version
cardano-node 1.19.1 - linux-x86_64 - ghc-8.6
git rev 497afd7dedc5d5b9bdcdb0e3cac6a50bd9f7dd54
```

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,4 @@
-resolver: https://raw.githubusercontent.com/input-output-hk/cardano-prelude/71ea865408f2e03e6d6832359423546699730849/snapshot.yaml
+resolver: https://raw.githubusercontent.com/input-output-hk/cardano-prelude/0c5b0a6619fadf22f4d62a12154e181a6d035c1c/snapshot.yaml
 compiler: ghc-8.6.5
 
 packages:
@@ -12,29 +12,31 @@ ghc-options:
   cardano-config:       -Wall -fwarn-redundant-constraints
   cardano-node:         -Wall -fwarn-redundant-constraints
 
-allow-newer: true
-
 extra-deps:
   - QuickCheck-2.12.6.1
-  - quickcheck-instances-0.3.19
+  - Unique-0.4.7.6
+  - aeson-1.4.7.1
+  - assoc-1.0.1
   - base58-bytestring-0.1.0
   - base64-0.4.2
-  - bech32-1.0.2
-  - brick-0.47
-  - binary-0.8.7.0
+  - bech32-1.1.0
+  - bifunctors-5.5.7
   - bimap-0.4.0
+  - binary-0.8.7.0
+  - brick-0.47
   - canonical-json-0.6.0.0
   - cborg-0.2.4.0
   - clock-0.8
   - config-ini-0.2.4.0
   - containers-0.5.11.0
   - data-clist-0.1.2.2
+  - directory-1.3.6.1
+  - esqueleto-3.2.2
   - generic-monoid-0.1.0.0
+  - generics-sop-0.5.1.0
   - ghc-byteorder-4.11.0.0.10
   - gray-code-0.3.1
   - hedgehog-1.0.2
-  - hedgehog-corpus-0.2.0
-  - hedgehog-quickcheck-0.1.1
   - hspec-2.7.0
   - hspec-core-2.7.0
   - hspec-discover-2.7.0
@@ -43,33 +45,51 @@ extra-deps:
   - micro-recursion-schemes-5.0.2.2
   - moo-1.2
   - network-3.1.1.1
+  - persistent-2.10.5.1
+  - persistent-postgresql-2.10.1.2
+  - persistent-template-2.8.2.3
+  - pipes-safe-2.3.2
+  - primitive-0.7.1.0
+  - process-1.6.8.2
+  - protolude-0.3.0
   - quiet-0.2
+  - quickcheck-instances-0.3.19
+  - serialise-0.2.3.0
   - snap-core-1.0.4.1
   - snap-server-1.1.1.1
+  - sop-core-0.5.0.1
   - statistics-linreg-0.3
   - streaming-binary-0.3.0.1
   - systemd-2.3.0
   - tasty-hedgehog-1.0.0.2
   - text-1.2.4.0
-  - text-ansi-0.1.0
   - text-zipper-0.10.1
+  - th-abstraction-0.3.2.0
+  - th-lift-0.8.1
   - th-lift-instances-0.1.14
+  - these-1.1
+  - time-1.8.0.4
+  - time-compat-1.9.3
   - time-units-1.0.0
   - transformers-except-0.1.1
-  - Unique-0.4.7.6
-  - word-wrap-0.4.1
+  - typerep-map-0.3.3.0
+  - unix-2.7.2.2
+  - webdriver-0.9.0.1
   - websockets-0.12.6.1
+  - Win32-2.8.5.0
+  - word-wrap-0.4.1
 
   - git: https://github.com/input-output-hk/cardano-ledger-specs
     commit: 74188f62ff743c690bdf287f7fa18eaa2ee354d4
     subdirs:
       # small-steps
       - semantics/executable-spec
+      - semantics/small-steps-test
       # cs-ledger
       - byron/ledger/executable-spec
       - byron/ledger/impl
-      - byron/crypto
       - byron/ledger/impl/test
+      - byron/crypto
       - byron/crypto/test
       # cs-blockchain
       - byron/chain/executable-spec
@@ -92,8 +112,8 @@ extra-deps:
       - binary
       - binary/test
       - cardano-crypto-class
+      - cardano-crypto-praos
       - slotting
-
 
     # Needed for `cardano-ledger-specs`
   - git: https://github.com/input-output-hk/goblins
@@ -116,8 +136,6 @@ extra-deps:
   - prometheus-2.1.2
   - libsystemd-journal-1.4.4
   - katip-0.8.4.0
-
-    #Ouroboros-network dependencies
 
   - git: https://github.com/input-output-hk/ouroboros-network
     commit: f0eb6e439e7c0121476ded5e88d2f638e8aa36ac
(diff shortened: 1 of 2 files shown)


rvl:
Thanks @CodiePP.

I'm getting some error from stack 2.1.3.

```
rodney@blue:~/iohk/cardano-node % stack build --fast
Aeson exception:
Error in $.packages[0].completed: failed to parse field 'packages': failed to parse field 'completed': Could not parse a UnresolvedPackageLocationImmutable from: Object (fromList [("subdir",String "cborg"),("name",String "cborg"),("version",String "0.2.2.1"),("git",String "https://github.com/well-typed/cborg"),("pantry-tree",Object (fromList [("size",Number 2711.0),("sha256",String "46906962e840d9d9c7a37c15fac8d5b5a7b44d172d037b46ba5256b524ce8b6f")])),("commit",String "42a83192749774268337258f4f94c97584b80ca6")])
```

stack.yaml
```diff
@@ -43,33 +46,57 @@ extra-deps:
...
+  - word-wrap-0.4.1
+
+  - git: https://github.com/erebe/systemd
+    commit: dba4b563dac99ff52328db916d99d928bc4a80fb
+
+  - git: https://github.com/CodiePP/cborg.git
+    commit: f6a2b7cf5ff4f515accaf026c95b4992d9f771bf
+    subdirs:
+        - cborg
+        - serialise
```
mrBliss, line 87: Please just use `cborg-0.2.4.0` and `serialise-0.2.3.0`. https://github.com/CodiePP/cborg/commit/f6a2b7cf5ff4f515accaf026c95b4992d9f771bf should not be necessary, see https://github.com/well-typed/cborg/blob/cborg-0.2.4.0/cborg/cborg.cabal

stack.yaml
```diff
@@ -43,33 +46,57 @@ extra-deps:
...
 
   - git: https://github.com/input-output-hk/cardano-ledger-specs
-    commit: 74188f62ff743c690bdf287f7fa18eaa2ee354d4
+    commit: 08f22528915bd2e125f20387dd171ba0f4e09742
```
mrBliss, line 90: This commit is not on master, just in an unmerged PR. If you close that PR, rebase it, or it is rebased/squashed when merged, that commit will eventually disappear, breaking the build.

KtorZ:
@rvl remove your stack.yaml.lock and try again :+1:

CodiePP:
in https://github.com/input-output-hk/cardano-ledger-specs/pull/1835 @nc6 and @mrBliss proposed to pin "protolude" to a newer version. It works, thanks.

stack.yaml
CodiePP: right. this is not necessary anymore.

stack.yaml
CodiePP: this is gone now too.

mrBliss: I mean, you can you use the version from Hackage now, no need for a specific revision.

stack.yaml
```diff
@@ -43,33 +46,58 @@ extra-deps:
...
-  - Unique-0.4.7.6
-  - word-wrap-0.4.1
+  - typerep-map-0.3.3.0
+  - unix-2.7.2.2
+  - webdriver-0.9.0.1
   - websockets-0.12.6.1
+  - Win32-2.8.5.0
+  - word-wrap-0.4.1
+
+  - git: https://github.com/erebe/systemd
```
rvl, line 81: Why does this need to be a git rev and not a hackage release? A yaml comment would be helpful.

stack.yaml
```diff
@@ -43,33 +46,58 @@ extra-deps:
...
+    commit: dba4b563dac99ff52328db916d99d928bc4a80fb
+
+  - git: https://github.com/well-typed/cborg.git
```
rvl, line 84: I think someone else also mentioned - better to use `cborg-0.2.4.0` from hackage.

rvl (approved):
Builds for me (after removing the local file `stack.yaml.lock`).

stack.yaml
```diff
@@ -16,25 +16,28 @@ allow-newer: true
```
mrBliss, line 16: BTW, this `allow-newer: true` changes all dependency pins into lower bounds, which *is not* the same thing.

CodiePP: :+1: 
this is not needed anymore; removed.

stack.yaml
mrBliss: Indeed, I did, but GitHub no longer shows it.

CodiePP: yes, not needed; removed.

stack.yaml
CodiePP: yes, not needed; removed.

mrBliss:
LGTM, although I haven't tried it, as I don't use Stack.

CodiePP (commented):
> LGTM, although I haven't tried it, as I don't use Stack.

give it a try :-)

hasufell:
```
Error: While constructing the build plan, the following exceptions were encountered:

In the dependencies for serialise-0.2.1.0:
    primitive-0.7.1.0 from stack configuration does not match >=0.5 && <0.7  (latest matching version is 0.6.4.0)
needed due to cardano-api-1.19.1 -> serialise-0.2.1.0

Some different approaches to resolving this:

  * Set 'allow-newer: true' in /root/.stack/config.yaml to ignore all version constraints and build anyway.

  * Recommended action: try adding the following to your extra-deps in /root/cardano-node/stack.yaml:

- primitive-0.6.4.0@sha256:5b6a2c3cc70a35aabd4565fcb9bb1dd78fe2814a36e62428a9a1aae8c32441a1,2079

Plan construction failed.
```

```
# stack --version
Version 2.3.3, Git revision cb44d51bed48b723a5deb08c3348c0b3ccfc437e x86_64 hpack-0.33.0
# git status
On branch cad-1765-update-stack-to-1.19.1
Your branch is up to date with 'origin/cad-1765-update-stack-to-1.19.1'.

nothing to commit, working tree clean
```

rvl:
@hasufell Why are you building as the `root` user? Which OS are you using? Why do you use the `--system-ghc` flag?
Try removing the `.stack-work` directories, `stack.yaml.lock`, and `~/.stack`, then start again.

mrBliss:
> ```
> # stack --system-ghc build --flag 'cardano-node:-systemd' --flag 'cardano-config:-systemd'
> Error: While constructing the build plan, the following exceptions were encountered:
> 
> In the dependencies for serialise-0.2.1.0:
>     primitive-0.7.1.0 from stack configuration does not match >=0.5 && <0.7  (latest matching version is 0.6.4.0)
> needed due to cardano-api-1.19.1 -> serialise-0.2.1.0
> 
> Some different approaches to resolving this:
> 
>   * Set 'allow-newer: true' in /root/.stack/config.yaml to ignore all version constraints and build anyway.
> 
>   * Recommended action: try adding the following to your extra-deps in /root/cardano-node/stack.yaml:
> 
> - primitive-0.6.4.0@sha256:5b6a2c3cc70a35aabd4565fcb9bb1dd78fe2814a36e62428a9a1aae8c32441a1,2079
> 
> Plan construction failed.
> ```

Neither solution is correct, the [ledger requires `primitive >= 0.7.1.0`](https://github.com/input-output-hk/cardano-ledger-specs/blob/7265f655856525b0a6d44b8d79f19f6afe42e748/shelley/chain-and-ledger/executable-spec/shelley-spec-ledger.cabal#L114) because it uses a function added in that function. `allow-newer: true` is looking for trouble as you're ignoring upper version bounds.

I think the following will solve it: use `serialise-0.2.3.0`, which has the following bounds for `primitive`: `>= 0.5 && <0.8`.

CodiePP:
added `serialise-0.2.3.0`
also builds with stack 2.3.3

hasufell:
@rvl 

> Why are you building as the root user?

I like to live on the edge... nah, I'm in a docker container.

> Which OS are you using?

Alpine.

> Why do you use the --system-ghc flag?

Stack doesn't provide a GHC for 8.6.5 for alpine, so I'm using ghcup to install the right GHC.

@mrBliss 

> I think the following will solve it: use serialise-0.2.3.0, which has the following bounds for primitive: >= 0.5 && <0.8.

I retried with the latest commit of the PR and the resolution problem is gone. But I still have the path-io issue, even after cleaning up stack root and stack workdir:

```
path-io                          > <command line>: can't load .so/.DLL for: /root/.stack/snapshots/x86_64-linux/3ec8f3be85a3b1b894f5e464d2713829670e33ce73aa73c3a6693f7e58686e15/8.6.5/lib/x86_64-linux-ghc-8.6.5/libHSunix-compat-0.5.1-rv6R0hBPRL6Jl74DCAgLB-ghc8.6.5.so (Error relocating /root/.stack/snapshots/x86_64-linux/3ec8f3be85a3b1b894f5e464d2713829670e33ce73aa73c3a6693f7e58686e15/8.6.5/lib/x86_64-linux-ghc-8.6.5/libHSunix-compat-0.5.1-rv6R0hBPRL6Jl74DCAgLB-ghc8.6.5.so: major: symbol not found)
```

Maybe this issue is alpine specific?

rvl:
The latest version of this branch builds for me.

@hasufell The symbol that the linker can't find comes from glibc ... which is missing on Alpine. I'm sorry but in my opinion this is not a suitable OS to use for Haskell development.

hasufell:
> The symbol that the linker can't find comes from glibc ... which is missing on Alpine. I'm sorry but in my opinion this is not a suitable OS to use for Haskell development.

I don't think so. If it's because of missing glibc, then this is musl specific, not alpine specific. And if you want properly static binaries, you need musl, even with nix.

In addition, this linking (or whatever) issue cannot be reproduced with cabal. So I'm tempted to believe it's a stack bug.

KtorZ:
@hasufell to build in a Docker container with stack, have a look at: https://github.com/input-output-hk/cardano-addresses/blob/master/Dockerfile

Multi-stage docker builds are very handy here, you can use a debian or ubuntu image to build, and then use alpine with glibc patched to run.

hasufell:
I tracked down the linking bug with stack on alpine to `unix-compat-0.5.1`. Bumping it to **0.5.2** fixes the build.

CodiePP:
this should be moved to some other place for stack users

hasufell:
> this should be moved to some other place for stack users

What other place? Can you give more information why this ticket was closed?

PR #1804, opened 2020-09-04
Run chairman in chairman tests

newhoggy:
* Moved option parsing to its own module to keep the main module clean.
* Invoke the chairman executable from the chairman tests.
* New `waitSecondsForProcess` function which waits for a process to exit by a deadline.
* New `cat` function to print contents of a file.
* New `procChairman` for invoking `cardano-node-chairman` process.
* Replace magic constant `42` with `testnetMagic` constant

### DIFF
--- a/cardano-node-chairman/app/Cardano/Chairman/Options.hs
+++ b/cardano-node-chairman/app/Cardano/Chairman/Options.hs
@@ -1,0 +1,138 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Cardano.Chairman.Options
+  ( mkNodeClientProtocol
+  , ChairmanArgs(..)
+  , parseChairmanArgs
+  , opts
+  ) where
+
+import           Cardano.Api.Protocol.Byron
+import           Cardano.Api.Protocol.Cardano
+import           Cardano.Api.Protocol.Shelley
+import           Cardano.Api.Protocol.Types
+import           Cardano.Api.Typed (NetworkMagic (..))
+import           Cardano.Chain.Slotting (EpochSlots (..))
+import           Cardano.Node.Protocol.Types (Protocol (..))
+import           Cardano.Node.Types
+import           Cardano.Prelude hiding (option)
+import           Control.Applicative (some)
+import           Control.Monad.Class.MonadTime (DiffTime)
+import           Options.Applicative
+import           Ouroboros.Consensus.BlockchainTime (SlotLength, slotLengthFromSec)
+import           Ouroboros.Consensus.Cardano (SecurityParam (..))
+import           Ouroboros.Network.Block (BlockNo)
+
+import qualified Options.Applicative as Opt
+
+--TODO: replace this with the new stuff from Cardano.Api.Protocol
+mkNodeClientProtocol :: Protocol -> SomeNodeClientProtocol
+mkNodeClientProtocol protocol =
+  case protocol of
+    ByronProtocol ->
+      mkSomeNodeClientProtocolByron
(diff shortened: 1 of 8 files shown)


cardano-node-chairman/test/Test/Cardano/Node/Chairman/Shelley.hs
```diff
@@ -57,6 +57,9 @@ import qualified System.Process as IO
 
 {- HLINT ignore "Redundant <&>" -}
 {- HLINT ignore "Redundant flip" -}
+{- HLINT ignore "Reduce duplication" -}
+
+{-# ANN module ("HLint: ignore Reduce duplication" :: Text) #-}
```
newhoggy, line 62: Not sure why I need to all use the old style pragma to suppress this `hlint` warning.

cardano-node-chairman/app/Cardano/Chairman/Options.hs
```diff
@@ -0,0 +1,138 @@
...
+    <> completer (bashCompleter "file")
+    <> metavar "FILEPATH"
+    )
+
+parseRunningTime :: Parser DiffTime
+parseRunningTime =
+  option ((fromIntegral :: Int -> DiffTime) <$> auto)
+    (  long "timeout"
+    <> short 't'
+    <> metavar "Time"
```
Jimbo4350, line 83: This should probable be `"SECONDS"`

cardano-node-chairman/app/Cardano/Chairman/Options.hs
```diff
@@ -0,0 +1,138 @@
...
+      <> Opt.metavar "INT"
+      <> Opt.help "The testnet network magic number"
+      )
+
+parseProgress :: Parser BlockNo
+parseProgress =
+  option ((fromIntegral :: Int -> BlockNo) <$> auto)
+    (  long "require-progress"
+    <> short 'p'
+    <> metavar "Blocks"
```
Jimbo4350, line 118: This should probably be `"INT"`

cardano-node-chairman/app/cardano-node-chairman.hs
```diff
@@ -49,113 +35,3 @@ main = do
     caSocketPaths
     someNodeClientProtocol
     caNetworkMagic
-
```
Jimbo4350, line 52: :+1:

cardano-node-chairman/src/Chairman/Process.hs
```diff
@@ -128,6 +132,22 @@ waitForProcess :: HasCallStack
 waitForProcess hProcess = GHC.withFrozenCallStack $ do
   H.evalM . liftIO $ catch (fmap Just (IO.waitForProcess hProcess)) $ \(_ :: AsyncCancelled) -> return Nothing
 
+waitSecondsForProcess :: HasCallStack
+  => Int
```
Jimbo4350, line 136: Would it make sense to put the initial `waitSecondsForProcess` in a `where` clause here? I'm not sure what the rational is for keeping them separate.

Jimbo4350 (approved):
LGTM! Just a couple of minor comments.

newhoggy:
bors merge

newhoggy:
Oops, I missed the comments.  Will fix.

cardano-node-chairman/src/Chairman/Process.hs
newhoggy: In this case I put the `IO` related functions in a separate module from the `Integration` functions so I don't muddle them up and I can give them the same name because they essentially do the same thing.

All the `IO` and `Integration` functions of the same name should do the same thing, the only difference being that the `Integration` functions additionally add annotations, have callstack handling and ensure any exceptions thrown are seen by `hedgehog`.

PR #1806, opened 2020-09-04
Force evaluation in assertIO

newhoggy:
Trying to fix this problem:

https://hydra.iohk.io/build/4032130/nixlog/16

My suspicion is `doesSocketExist` throws an exception lazily when the returned `Bool` is evaluated which would allow the exception to evade `hedgehog`.  Forcing the return value to be evaluated should ensure the exception is thrown within `evalIO` which will allow `hedgehog` to catch it and print the error.


Jimbo4350 (approved):

newhoggy:
bors merge

PR #1807, opened 2020-09-04
Handle IOException in isPortOpen

newhoggy:


cardano-node-chairman/src/Chairman/IO/Network/Socket.hs
```diff
@@ -34,7 +36,7 @@ import qualified UnliftIO.Exception as IO
 
 -- | Check if a TCP port is open
 isPortOpen :: Int -> IO Bool
-isPortOpen port = do
+isPortOpen port = handle (\(_ :: IOException) -> return False) $ do
```
mrBliss, line 39: Now you're catching in two places: here using `catch` and in `canConnect`  using `try`. Could you accomplish the same by returning `False` in the `Left e` case of `canConnect`?

newhoggy: Is `IOException` definitely the only exception that we need to handle?

I think your suggestion is sensible if this is the case.  Otherwise the type signature might end up being complicated by multiple exception types.

mrBliss: I think you're catching in both cases `IOException`. Here because of the annotation and in `canConnect` because you're calling `IO.ioe_errno e`

newhoggy: This other PR failed because to build in CI because of an issue that this PR fixes:  https://github.com/input-output-hk/cardano-node/pull/1806

I'd like to merge this PR fix the flaky build and work on address your comments in a new PR on Monday if that's alright.

newhoggy:
Thanks!  I'll give your suggestions a go on Monday!

newhoggy:
bors merge

PR #1811, opened 2020-09-06
Update register_stakepool.md

PeterMika:
Running Cardano 1.19.0 the command instructions "Get Pool ID" was generating output in format bech32 which the succeeding command to verify you PoolID does not use. 

Setting output format to --output-format "hex" resolved this issue.

Available options:
  --verification-key-file FILE
                           Output filepath of the verification key.
  --output-format STRING   Optional output format. Accepted output formats are
                           "hex" and "bech32" (default is "bech32").

### DIFF
--- a/doc/stake-pool-operations/register_stakepool.md
+++ b/doc/stake-pool-operations/register_stakepool.md
@@ -179,7 +179,7 @@ All amounts in Lovelace
 
 Get Pool ID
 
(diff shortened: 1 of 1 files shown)


Jimbo4350:
Thanks, this PR was subsumed by: https://github.com/input-output-hk/cardano-node/pull/2442

PR #1812, opened 2020-09-07
Simplify error handling

newhoggy:
Simplify exception handling

### DIFF
--- a/cardano-node-chairman/src/Chairman/IO/Network/Socket.hs
+++ b/cardano-node-chairman/src/Chairman/IO/Network/Socket.hs
@@ -1,12 +1,6 @@
-{-# LANGUAGE CPP #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
 
-{-|
-Module      : Chairman.IO.Network.Socket
-Description : Network functions
-
-OS specific functions are found in submodules.
--}
 module Chairman.IO.Network.Socket
   ( doesSocketExist
   , isPortOpen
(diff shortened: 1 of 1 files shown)


cardano-node-chairman/src/Chairman/IO/Network/Socket.hs
```diff
@@ -17,42 +11,32 @@ module Chairman.IO.Network.Socket
...
 -- | Check if a TCP port is open
 isPortOpen :: Int -> IO Bool
-isPortOpen port = handle (\(_ :: IOException) -> return False) $ do
+isPortOpen port = do
   socketAddressInfos <- IO.getAddrInfo Nothing (Just "127.0.0.1") (Just (show port))
   case socketAddressInfos of
-    socketAddressInfo:_ -> canConnect (IO.addrAddress socketAddressInfo)
+    socketAddressInfo:_ ->
+      handle (return . const @Bool @IOException False) $
+        canConnect (IO.addrAddress socketAddressInfo) $> True
```
newhoggy, line 33: Moved the exception handling to around `canConnect`.

I'm intentionally excluding `getAddrInfo` because if that fails, I want the exception to propagate and be reported by tests.  This is because a failure at that point means it's impossible to test if the port is open and any information associated with the exception would be useful in diagnosing the problem.

cardano-node-chairman/src/Chairman/IO/Network/Socket.hs
```diff
@@ -17,42 +11,32 @@ module Chairman.IO.Network.Socket
...
+canConnect :: SockAddr -> IO ()
 canConnect sockAddr = IO.bracket (IO.socket AF_INET Stream 6) IO.close' $ \sock -> do
-  result <- IO.try $ IO.connect sock sockAddr
-  case result of
-    Right () -> return True
-    Left e
-      | (Errno <$> IO.ioe_errno e) == Just eCONNREFUSED -> return False
-      | "WSAECONNREFUSED" `L.isInfixOf` show e -> return False
-      | "WSAECONNRESET" `L.isInfixOf` show e -> return False
-      | otherwise -> IO.throwIO e
```
newhoggy, line 55: Exception has been moved to the caller of `canConnect`.  This means exception handling can happen in one place.

Moreover, if an exception is thrown at all, we'll report that the port was not open.  This is to avoid tests being flaky, which has been the case when some other unhandled variant of `IOException` was thrown.

erikd (approved):

newhoggy:
bors merge

mrBliss:
:+1:

PR #1817, opened 2020-09-07
Update README.rst

untopor:
minor formatting fix


Jimbo4350:
Thanks @untopor this was included in #1959

PR #1820, opened 2020-09-08
Validate shelley genesis before starting node

### LINKED ISSUE #1149: cardano-node is allowed to be started with a genesis setttings combination that prevents rewards being paid out
We ran into this the other day with a non-FnF network, that had epoch length set to a short value (30 minutes - 900 of 2s slots), and had a securityParmeter set to a rather high value. While the network was able to be started, with BFT and eventually Praos nodes and pool registration and funds transfers and delegations/pledging, there were no rewards sighted in the ledger state even after a dozen epochs. 

After discussing this with IOHK developer we were referred to some documentation related to epoch stages and certain expectations by Ourobouros in regards to the k and f parameters - eventually we lowered our securityParameter setting value and reset the network and got rewards being paid out correctly - however it would be better if extra validation could be added on the genesis settings involved, to alert the operator / network manager about the incorrectly configured genesis.

### LINKED ISSUE #1478: [FR] - Check the validity of the Shelley genesis config
**Internal**

In the last few months, many people have run into problems caused by an incorrect genesis config for their Shelley net.

A function with some basic sanity checks of the Shelley genesis config was introduced in [`cardano-ledger-specs`](https://github.com/input-output-hk/cardano-ledger-specs/pull/1644).

This function is [now](https://github.com/input-output-hk/ouroboros-network/pull/2423) used in the consensus layer to `assert` that the Shelley genesis config is valid. Note that this check is only done when assertions are enabled, which is not the case for real nodes.

`cardano-node` should use the `validateGenesis` function from https://github.com/input-output-hk/cardano-ledger-specs/pull/1644/ to check the validity of the Shelley genesis config file and report a proper error/exception in case of failure.

Jimbo4350:
Closes #1149
Closes #1478

### DIFF
--- a/cardano-node/src/Cardano/Node/Protocol/Shelley.hs
+++ b/cardano-node/src/Cardano/Node/Protocol/Shelley.hs
@@ -39,8 +39,10 @@ import           Ouroboros.Consensus.Cardano.ShelleyHFC
 
 import           Ouroboros.Consensus.Shelley.Node (Nonce (..), ShelleyGenesis,
                      TPraosLeaderCredentials (..))
-import           Ouroboros.Consensus.Shelley.Protocol (TPraosCanBeLeader (..), StandardShelley)
+import           Ouroboros.Consensus.Shelley.Protocol (StandardShelley, TPraosCanBeLeader (..))
 
+import           Shelley.Spec.Ledger.Genesis (ValidationErr (..), describeValidationErr,
+                     validateGenesis)
 import           Shelley.Spec.Ledger.Keys (coerceKeyRole)
 import           Shelley.Spec.Ledger.PParams (ProtVer (..))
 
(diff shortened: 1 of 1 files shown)


cardano-node/src/Cardano/Node/Protocol/Shelley.hs
```diff
@@ -188,6 +190,7 @@ data ShelleyProtocolInstantiationError =
        GenesisReadError !FilePath !IOException
      | GenesisHashMismatch !GenesisHash !GenesisHash -- actual, expected
      | GenesisDecodeError !FilePath !String
+     | GenesisValidationFailure [ValidationErr]
```
intricate, line 193:
```suggestion
     | GenesisValidationFailure ![ValidationErr]
```

intricate:
Related: https://github.com/input-output-hk/cardano-node/issues/1478

cardano-node/src/Cardano/Node/Protocol/Shelley.hs
```diff
@@ -213,7 +216,7 @@ renderShelleyProtocolInstantiationError pie =
     GenesisDecodeError fp err ->
         "There was an error parsing the genesis file: "
      <> toS fp <> " Error: " <> T.pack (show err)
-
+    GenesisValidationFailure vErrs -> T.unlines $ map describeValidationErr vErrs
```
intricate, line 219: All of the cases here are separated by a newline. I think we should continue doing that for consistency:

```suggestion
    GenesisValidationFailure vErrs -> T.unlines $ map describeValidationErr vErrs
    
```

intricate (approved):

PR #1823, opened 2020-09-08
Call validateTxMetadata on transaction metadata parsed from JSON

intricate:
At this time, `jsonToMetadata` already performs the appropriate validation on the transaction metadata during its JSON parsing. However, it still might make sense to call `validateTxMetadata` just in case more validation rules that aren't covered by `jsonToMetadata` are added in the future.


rvl (approved):
Good idea

rvl:
Sorry my review wasn't sufficient.

Bors-ng just requires single review approval (`bors.toml` sets `required_approvals=1`).
This GitHub repo has master branch protection with codeowner approvals enabled.

When bors-ng tried to push the merge commit to github it was rejected.

Rather than handling the error and sending a comment to the PR, bors-ng does the erlang thing and crashes.

The error message is visible by checking the :x: status from bors in the merge commit.

newhoggy (approved):

PR #1824, opened 2020-09-08
Extend deadline in Shelley-only chairman test

newhoggy:
Extend the deadline a bit in case there is something else happening that might slow down the node and cause it to miss the deadline.


erikd (approved):

newhoggy:
bors merge

Issue #1451, opened 2020-07-14
[BUG] - CLI Parser is traversing same option twice


rdlrt:
**Internal/Exernal**
*External* 

**Summary**

Relevant code [here](https://github.com/input-output-hk/cardano-node/blob/8c72fc00cb75b7e7a2c9423ba50c639f6236c88b/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs#L1686-L1719)

CLI Parser is currently traversing options from different protocols (shelley|byron|cardano) , but repeating parameters "epoch-slots" and "security-param" twice in output - once for Byron, and once for Cardano.

**Steps to reproduce**
1. Run `cardano-cli shelley query tip`
2. You see output contains same parameters (`--epoch-slots` and `--security-param`) twice - once for `byron-mode` and once for `cardano-mode`.

**Expected behavior**
The parameters should not be repeated for a given command.

**System info (please complete the following information):**
- OS: *all*
- Version *all*
- Node version: 1.15.1

kevinhammond:
Thank you.  We will deal with this as we review the codebase

vix-io:
hi @Jimbo4350 is this resolved already?

intricate:
@vix-io: This isn't resolved yet.

rdlrt:
Still the case on 1.26.0

Jimbo4350:
Closing this. If this is still relevant please re-open.

PR #1827, opened 2020-09-09
Fix cli protocol parser

Jimbo4350:
#1451

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
+++ b/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
@@ -1932,16 +1932,16 @@ pProtocol =
         (SecurityParam defaultSecurityParam))
   where
     pByron :: Parser Protocol
-    pByron = ByronProtocol <$> pEpochSlots <*> pSecurityParam
+    pByron = ByronProtocol <$> pEpochSlotsByron <*> pSecurityParam
 
     pShelley :: Parser Protocol
     pShelley = pure ShelleyProtocol
 
     pCardano :: Parser Protocol
-    pCardano = CardanoProtocol <$> pEpochSlots <*> pSecurityParam
+    pCardano = CardanoProtocol <$> pEpochSlotsCardano <*> pSecurityParam
 
-    pEpochSlots :: Parser EpochSlots
-    pEpochSlots =
+    pEpochSlotsByron :: Parser EpochSlots
+    pEpochSlotsByron =
       EpochSlots <$>
         Opt.option Opt.auto
           (  Opt.long "epoch-slots"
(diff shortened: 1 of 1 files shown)


newhoggy:
Seems to not fix the problem for me in that `--epoch-slots` is still duplicated:

```
➜  cardano-node git:(jordan/fix-cli-protocol-parser) cabal build cardano-cli
Up to date
➜  cardano-node git:(jordan/fix-cli-protocol-parser) cabal exec -- cardano-cli shelley query tip
Usage: cardano-cli shelley query tip [--shelley-mode | --byron-mode
                                       [--epoch-slots NATURAL]
                                       [--security-param NATURAL] |
                                       --cardano-mode --epoch-slots NATURAL
                                       [--security-param NATURAL]]
                                     [--out-file FILE]
  Get the node's current tip (slot no, hash, block no)

Available options:
  --shelley-mode           For talking to a node running in Shelley-only mode.
  --byron-mode             For talking to a node running in Byron-only mode.
  --epoch-slots NATURAL    The number of slots per epoch for the Byron
                           era. (default: 21600)
  --security-param NATURAL The security parameter. (default: 2160)
  --cardano-mode           For talking to a node running in full Cardano mode
                           (default).
  --epoch-slots NATURAL    The number of slots per epoch for the Cardano era.
  --security-param NATURAL The security parameter. (default: 2160)
```

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -1951,6 +1951,15 @@ pProtocol =
           <> Opt.showDefault
           )
 
+    pEpochSlotsCardano :: Parser EpochSlots
+    pEpochSlotsCardano =
+      EpochSlots <$>
+        Opt.option Opt.auto
+          (  Opt.long "epoch-slots"
+          <> Opt.metavar "NATURAL"
+          <> Opt.help "The number of slots per epoch for the Cardano era."
```
intricate, line 1960: There isn't such a thing as the "Cardano" era, but either Byron or Shelley. The `EpochSlots` value only applies to the Byron era, so I think that this was correct the way it was before (using `pEpochSlots`).

intricate (changes requested):
It doesn't seem that this addresses https://github.com/input-output-hk/cardano-node/issues/1451.

Also, I think that these particular changes (renaming `pEpochSlots` and adding `pEpochSlotsCardano`) should be reverted. See https://github.com/input-output-hk/cardano-node/pull/1827#discussion_r486004244.

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
Jimbo4350: So the conclusion would be to revert the changes and close the issue?

Jimbo4350:
Closing, will revisit in the future

PR #1828, opened 2020-09-09
No close socket

newhoggy:


cardano-node-chairman/src/Chairman/IO/Network/Socket.hs
```diff
@@ -65,8 +66,11 @@ allocateRandomPorts n = do
   -- Create n sockets with randomly bound ports, grab the port numbers and close those ports
   addr:_ <- IO.getAddrInfo (Just hints) (Just "127.0.0.1") (Just "0")
   socks <- forM [1..n] $ \_ -> IO.socket (IO.addrFamily addr) (IO.addrSocketType addr) (IO.addrProtocol addr)
-  forM_ socks $ \sock -> IO.bind sock (IO.addrAddress addr)
+  forM_ socks $ \sock -> do
+    IO.setSocketOption sock IO.ReuseAddr 1
+    IO.setSocketOption sock IO.NoDelay 1
```
karknu, line 71: Since you shouldn't send any data on these sockets NoDelay option isn't needed.

PR #1829, opened 2020-09-09
Remove cardano-config

Jimbo4350:


newhoggy (approved):

rvl (approved):

newhoggy:
There's an error in CI:

```
<no location info>: error:
    <command line>: can't load .so/.DLL for: libHScardano-api-1.19.1-9lA2untDJmALfxneTVdNlH.so (Error loading shared library libHScardano-api-1.19.1-9lA2untDJmALfxneTVdNlH.so: No such file or directory)
```

rvl:
Hmm strange that this error appears just because the code was moved. It seems to be due to template haskell on the cross-compiled musl build.

Jimbo4350:
> Hmm strange that this error appears just because the code was moved. It seems to be due to template haskell on the cross-compiled musl build.

I've run into this before but I have no idea how to fix it. How did you figure this out?

rvl:
@Jimbo4350 Sorry, after another look at this, I still don't know what is wrong. All that is different from before is that the `Git.Rev` module is now in the same package where it's used, and there are now two copies of it (under `Cardano` and `Cardano.CLI`). @angerman might have some ideas.

Jimbo4350:
Closing as we currently cannot figure out why CI is failing

PR #1830, opened 2020-09-09
Update dependency on ouroboros-network

mrBliss:
WARNING: this breaks binary compatibility of Byron and Shelley ledger
snapshots. The ledger will have to be replayed from genesis!

* `Protocol(Client)Info` no longer needs the `SecurityParam`.
* New `TraceForgeEvent` messages were added. The constructors have been sorted
  to match the order in which they're defined in
  `Ouroboros.Consensus.Node.Tracers`.

### DIFF
--- a/cardano-node/src/Cardano/Tracing/OrphanInstances/Consensus.hs
+++ b/cardano-node/src/Cardano/Tracing/OrphanInstances/Consensus.hs
@@ -166,19 +166,22 @@ instance HasSeverityAnnotation () where
 
 instance HasPrivacyAnnotation (TraceForgeEvent blk)
 instance HasSeverityAnnotation (TraceForgeEvent blk) where
-  getSeverityAnnotation TraceForgedBlock {}            = Info
   getSeverityAnnotation TraceStartLeadershipCheck {}   = Info
-  getSeverityAnnotation TraceNodeNotLeader {}          = Info
-  getSeverityAnnotation TraceNodeCannotForge {}        = Error
-  getSeverityAnnotation TraceNodeIsLeader {}           = Info
-  getSeverityAnnotation TraceForgeStateUpdateError {}  = Error
-  getSeverityAnnotation TraceNoLedgerState {}          = Error
-  getSeverityAnnotation TraceNoLedgerView {}           = Error
-  getSeverityAnnotation TraceBlockFromFuture {}        = Error
   getSeverityAnnotation TraceSlotIsImmutable {}        = Error
-  getSeverityAnnotation TraceAdoptedBlock {}           = Info
+  getSeverityAnnotation TraceBlockFromFuture {}        = Error
+  getSeverityAnnotation TraceBlockContext {}           = Debug
+  getSeverityAnnotation TraceNoLedgerState {}          = Error
+  getSeverityAnnotation TraceLedgerState {}            = Debug
+  getSeverityAnnotation TraceNoLedgerView {}           = Error
+  getSeverityAnnotation TraceLedgerView {}             = Debug
+  getSeverityAnnotation TraceForgeStateUpdateError {}  = Error
+  getSeverityAnnotation TraceNodeCannotForge {}        = Error
+  getSeverityAnnotation TraceNodeNotLeader {}          = Info
+  getSeverityAnnotation TraceNodeIsLeader {}           = Info
+  getSeverityAnnotation TraceForgedBlock {}            = Info
   getSeverityAnnotation TraceDidntAdoptBlock {}        = Error
   getSeverityAnnotation TraceForgedInvalidBlock {}     = Error
+  getSeverityAnnotation TraceAdoptedBlock {}           = Info
 
 
 instance HasPrivacyAnnotation (TraceLocalTxSubmissionServerEvent blk)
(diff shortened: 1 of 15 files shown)


cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -502,19 +502,22 @@ teeForge ::
 teeForge ft nodeKern tverb tr = Tracer $ \ev@(WithSeverity sev event) -> do
   flip traceWith ev $ fanning $ \(WithSeverity _ e) ->
     case e of
-      Consensus.TraceNodeCannotForge {} -> teeForge' (ftTraceNodeCannotForge ft)
-      Consensus.TraceForgeStateUpdateError{} -> teeForge' (ftTraceForgeStateUpdateError ft)
-      Consensus.TraceForgedBlock{} -> teeForge' (ftForged ft)
       Consensus.TraceStartLeadershipCheck{} -> teeForge' (ftForgeAboutToLead ft)
+      Consensus.TraceSlotIsImmutable{} -> teeForge' (ftTraceSlotIsImmutable ft)
+      Consensus.TraceBlockFromFuture{} -> teeForge' (ftTraceBlockFromFuture ft)
+      Consensus.TraceBlockContext{} -> panic "no point in tracing this"
```
mrBliss, line 508: @denisshevchenko I left 6 `panic`s in this file because IMO they're no point in forwarding these (debug) messages. What should we do?

CodiePP: panic is probably not the right thing to enter production.
do you want to suppress this particular messages instead?

cardano-node/src/Cardano/Tracing/OrphanInstances/Consensus.hs
```diff
@@ -837,64 +852,49 @@ instance ( tx ~ GenTx blk
          , ToObject (CannotForge blk)
          , ToObject (ForgeStateUpdateError blk))
       => ToObject (TraceForgeEvent blk) where
-  toObject MaximalVerbosity (TraceAdoptedBlock slotNo blk txs) =
+  toObject _verb (TraceStartLeadershipCheck slotNo) =
     mkObject
-      [ "kind" .= String "TraceAdoptedBlock"
```
CodiePP, line 842: "TraceAdoptedBlock" is an important event for metrics in benchmarking. The reordering in this file makes it difficult to assess if this is still traced.

cardano-node/src/Cardano/Tracing/Tracers.hs
deepfire: @mrBliss, this case ought to return a tracer -- so if we want to ignore something, the current solution is by returning a `nullTracer`.

cardano-node/src/Cardano/Tracing/OrphanInstances/Consensus.hs
mrBliss: I did this for a good reason, to stick to the same order as in which they're defined: https://github.com/input-output-hk/ouroboros-network/blob/7cda58405b6ee0d335b11e88e5c9989c7a3a6e03/ouroboros-consensus/src/Ouroboros/Consensus/Node/Tracers.hs#L143-L310

IMO, by maintaining the same order, it is easier to spot (by eye) which events are already handled and which are missing (the compiler tells you this as well, of course).

cardano-node/src/Cardano/Tracing/Tracers.hs
mrBliss: Yeah, the `panic` is of course not meant to be merged, it's only to signal that this still needs to be addressed :slightly_smiling_face: 

@deepfire Thanks for the answer!

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -541,33 +544,38 @@ teeForge' tr =
...
-        Consensus.TraceForgedBlock slot _ _ _ ->
-          LogValue "forgedSlotLast" $ PureI $ fromIntegral $ unSlotNo slot
         Consensus.TraceStartLeadershipCheck slot ->
           LogValue "aboutToLeadSlotLast" $ PureI $ fromIntegral $ unSlotNo slot
+        Consensus.TraceSlotIsImmutable slot _tipPoint _tipBlkNo ->
+          LogValue "slotIsImmutable" $ PureI $ fromIntegral $ unSlotNo slot
+        Consensus.TraceBlockFromFuture slot _slotNo ->
+          LogValue "blockFromFuture" $ PureI $ fromIntegral $ unSlotNo slot
+        Consensus.TraceBlockContext slot _tipBlkNo _tipPoint ->
+          LogValue "blockContext" $ PureI $ fromIntegral $ unSlotNo slot
```
mrBliss, line 554: @deepfire What about this one (and `"ledgerState"`)? As you know, they're only for debug purposes. I don't know how these `LogValue`s are used, so up to you.

mrBliss:
Ready to be merged, IMO

cardano-node/src/Cardano/Tracing/Tracers.hs
erikd: I would really prefer not to have a `panic` there, expecially not with such a poor error message.

Using `nullTracer` seems like the sensible solution.

erikd (changes requested):
Please switch from `panic` to `nullTracer` as suggested by the logging team.

erikd (approved):

mrBliss:
bors merge

cardano-node/src/Cardano/Tracing/Tracers.hs
deepfire: I think this should just remain configurable -- if it's not traced then it's configuration-by-editing at debug time.

PR #1846, opened 2020-09-15
Increase test code re-use by moving test support code into separate common library

newhoggy:
Increase test code re-use by moving test support code into separate common `hedgehog-extras` library.

* Move generic test code into the new `hedgehog-extras` library with a module structure consistent with its genericity.
* Both `cardano-node-chairman` and `cardano-cli` modified to use the new library
* Avoid using `System.IO` libraries directly in `cardano-cli` because exceptions cause test annotations to show making tests difficult.  Instead use equivalent functions from `hedgehog-extras`
* Re-implement `execCardanoCLI` in terms of `execFlex` which has the advantage of not depending on `cabal exec` which occasionally causes problems in terms of pollution of `stdout`

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -6,6 +6,7 @@ packages:
     cardano-config
     cardano-node
     cardano-node-chairman
+    hedgehog-extras
 
 package cardano-api
   ghc-options: -Werror -Wall -Wredundant-constraints -Wincomplete-record-updates -Wincomplete-uni-patterns -Wpartial-fields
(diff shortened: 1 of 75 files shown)


Jimbo4350 (approved):
LGTM!

newhoggy:
bors merge

newhoggy:
bors merge

newhoggy:
bors merge

PR #1847, opened 2020-09-15
Add simple mainnet node script

Jimbo4350:


edsko:
Seems to work just fine, but perhaps it'd be better to use a db in a specific directory rather than a new temp directory, created each time the script is invoked? The old mainnet script stored it in `db/byron` or some such.

edsko (approved):

newhoggy (approved):
Looks good.  Is there anything that needs to be added to `.gitignore`?

Jimbo4350:
> Looks good. Is there anything that needs to be added to `.gitignore`?

Ah yes. Just added `/mainnet` to the `.gitignore`

PR #1850, opened 2020-09-15
Implement partial options monoid configuration technique in cardano-node

Jimbo4350:
POI: https://medium.com/@jonathangfischoff/the-partial-options-monoid-pattern-31914a71fc67

This allows us to combine configuration options from the command line and from the config yaml file in a seamless way. The default configuration will connect to mainnet. I.E running `cabal exec cardano-node run` in the top level `cardano-node` dir will connect the node to mainnet.

### DIFF
--- a/cardano-node/src/Cardano/Node/Configuration/POM.hs
+++ b/cardano-node/src/Cardano/Node/Configuration/POM.hs
@@ -65,7 +65,7 @@ data NodeConfiguration
        , ncLoggingSwitch  :: !Bool
        , ncLogMetrics     :: !Bool
        , ncTraceConfig    :: !TraceOptions
-       } deriving Show
+       } deriving (Eq, Show)
 
 
 data PartialNodeConfiguration
--- a/cardano-node/src/Cardano/Node/Configuration/POM.hs
+++ b/cardano-node/src/Cardano/Node/Configuration/POM.hs
@@ -1,0 +1,315 @@
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE NamedFieldPuns #-}
+
+{-# OPTIONS_GHC -Wno-noncanonical-monoid-instances #-}
+
+module Cardano.Node.Configuration.POM
+  ( NodeConfiguration (..)
+  , PartialNodeConfiguration(..)
+  , defaultPartialNodeConfiguration
+  , lastOption
+  , makeNodeConfiguration
+  , parseNodeConfigurationFP
+  , pncProtocol
+  , ncProtocol
+  )
+where
+
+import           Cardano.Prelude
+import           Prelude (String)
+
+import           Control.Monad (fail)
+import           Data.Aeson
+import           Data.Semigroup (Semigroup (..))
+import           Data.Yaml (decodeFileThrow)
+import           Generic.Data (gmappend)
+import           Generic.Data.Orphans ()
+import           Options.Applicative
+import           System.FilePath (takeDirectory, (</>))
+import           System.Posix.Types (Fd (..))
+
+import qualified Cardano.Chain.Update as Byron
+import           Cardano.Crypto (RequiresNetworkMagic (..))
+import           Cardano.Node.Protocol.Types (Protocol (..))
+import           Cardano.Node.Types
+import           Cardano.Tracing.Config
+import           Ouroboros.Network.Block (MaxSlotNo (..))
+
+data NodeConfiguration
+  = NodeConfiguration
+      {  ncNodeAddr        :: !(Maybe NodeAddress)
+          -- | Filepath of the configuration yaml file. This file determines
+          -- all the configuration settings required for the cardano node
+          -- (logging, tracing, protocol, slot length etc)
+       , ncConfigFile      :: !ConfigYamlFilePath
+       , ncTopologyFile    :: !TopologyFile
+       , ncDatabaseFile    :: !DbFile
+       , ncProtocolFiles   :: !ProtocolFilepaths
+       , ncValidateDB      :: !Bool
+       , ncShutdownIPC     :: !(Maybe Fd)
+       , ncShutdownOnSlotSynced :: !MaxSlotNo
+
+        -- Protocol-specific parameters:
+       , ncProtocolConfig :: !NodeProtocolConfiguration
+
+         -- Node parameters, not protocol-specific:
+       , ncSocketPath     :: !(Maybe SocketPath)
+
+         -- BlockFetch configuration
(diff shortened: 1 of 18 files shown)


cardano-node/src/Cardano/Node/Configuration/POM.hs
```diff
@@ -0,0 +1,328 @@
...
+
+instance AdjustFilePaths PartialNodeConfiguration where
+  adjustFilePaths f x@PartialNodeConfiguration {
+                        pncProtocolConfig,
+                        pncSocketPath
+                      } =
+    x {
+      pncProtocolConfig = adjustFilePaths f pncProtocolConfig,
+      pncSocketPath     = adjustFilePaths f pncSocketPath
+    }
```
erikd, line 111: is that really better than:
```
  adjustFilePaths f x =
    x {
      pncProtocolConfig = adjustFilePaths f (pncProtocolConfig x),
      pncSocketPath     = adjustFilePaths f (pncSocketPath x)
    }
```

cardano-node/src/Cardano/Node/Configuration/POM.hs
```diff
@@ -0,0 +1,328 @@
...
+      , pncShutdownOnSlotSynced = pncShutdownOnSlotSynced x <> pncShutdownOnSlotSynced y
+      , pncProtocolConfig = pncProtocolConfig x <> pncProtocolConfig y
+      , pncSocketPath = pncSocketPath x <> pncSocketPath y
+      , pncMaxConcurrencyBulkSync = pncMaxConcurrencyBulkSync x <> pncMaxConcurrencyBulkSync y
+      , pncMaxConcurrencyDeadline = pncMaxConcurrencyDeadline x <> pncMaxConcurrencyDeadline y
+      , pncViewMode = pncViewMode x <> pncViewMode y
+      , pncLoggingSwitch = pncLoggingSwitch x <> pncLoggingSwitch y
+      , pncLogMetrics = pncLogMetrics x <> pncLogMetrics y
+      , pncTraceConfig = pncTraceConfig x <> pncTraceConfig y
+      }
```
erikd, line 139: Is there anything there that could not be derived?

newhoggy, line 139: Shouldn't `(<>)` and `mappend` do the same thing?

cardano-node/src/Cardano/Node/Configuration/POM.hs
```diff
@@ -0,0 +1,328 @@
...
+        npcTestShelleyHardForkAtEpoch   <- v .:? "TestShelleyHardForkAtEpoch"
+        npcTestShelleyHardForkAtVersion <- v .:? "TestShelleyHardForkAtVersion"
+        npcShelleyHardForkNotBeforeEpoch <- v .:? "ShelleyHardForkNotBeforeEpoch"
+        pure NodeHardForkProtocolConfiguration {
+               npcTestShelleyHardForkAtEpoch,
+               npcTestShelleyHardForkAtVersion,
+               npcShelleyHardForkNotBeforeEpoch
+             }
+
+-- Default configuration is mainnet
```
newhoggy, line 257: Probably upgrade to haddock documentation.

cardano-node/src/Cardano/Node/Configuration/POM.hs
```diff
@@ -0,0 +1,328 @@
...
+            Last . Just . NodeProtocolConfigurationByron <$> parseByronProtocol v
+
+          ShelleyProtocol ->
+            Last . Just . NodeProtocolConfigurationShelley <$> parseShelleyProtocol v
+
+          CardanoProtocol ->
+            Last . Just  <$> (NodeProtocolConfigurationCardano <$> parseByronProtocol v
+                                                               <*> parseShelleyProtocol v
+                                                               <*> parseHardForkProtocol v)
+      pure mempty {
```
erikd, line 175: Why `mempty` there? Would it not be better to use the actual struct name?

cardano-node/src/Cardano/Node/Configuration/POM.hs
```diff
@@ -0,0 +1,328 @@
...
+    { pncConfigFile = Last . Just $ ConfigYamlFilePath "configuration/cardano/mainnet-config.json"
+    , pncDatabaseFile = Last . Just $ DbFile "mainnet/db/"
+    , pncLoggingSwitch = Last $ Just True
+    , pncSocketPath = Last $ Just "mainnet/socket/nodesocket"
+    , pncTopologyFile = Last . Just $ TopologyFile "configuration/cardano/mainnet-topology.json"
+    , pncViewMode = Last $ Just SimpleView
+    }
+
+lastOption :: Parser a -> Parser (Last a)
+lastOption parser = fmap Last $ optional parser
```
erikd, line 270: `last <$> optional parser` 
or point free style:
```
lastOption = fmap Last . optional
```

newhoggy:
A quick test to sanity check the partial options monoid might be useful.

cardano-node/src/Cardano/Node/Parsers.hs
```diff
@@ -46,33 +47,33 @@ nodeRunParser = do
...
-    , shutdownOnSlotSynced
-    }
+  nodeConfigFp <- lastOption parseConfigFile
+
+  validate <- lastOption parseValidateDB
+  shutdownIPC <- lastOption parseShutdownIPC
+
+  shutdownOnSlotSynced <- lastOption parseShutdownOnSlotSynced
+
+  pure $ mempty
```
erikd, line 60: `mempty` again?

cardano-node/src/Cardano/Node/Configuration/POM.hs
Jimbo4350: This was a copy/paste. I can change it.

cardano-node/src/Cardano/Node/Configuration/POM.hs
Jimbo4350: Yep, I implemented the `Monoid` instance to use the notation `mempty { ... }` to avoid writing `mempty` for each record didn't have a value. I've removed the instance because being explicit makes it clearer what I'm doing.

Jimbo4350: Remove the instance for the following reason: https://github.com/input-output-hk/cardano-node/pull/1850#discussion_r490073930

cardano-node/src/Cardano/Node/Configuration/POM.hs
Jimbo4350: I agree, it makes it clearer what I'm doing.

erikd (approved):
I will leave it up to you whether you want to add some tests.

However, if it break, you own both pieces :D

Jimbo4350:
> A quick test to sanity check the partial options monoid might be useful.

Sanity test added! :smile:

PR #1872, opened 2020-09-18
CAD-1866: Prepared TraceForwarderBK in defaut configs.

denisshevchenko:
1. Now the default configs of the node already contain metric forwarding to `TraceForwarderBK`
2. The `TraceForwarderBK` can be commented out in the default backends section (this makes it easy for someone to turn on `RTView`).

_Please note that this PR doesn't touch any code, only configuration files_.

### DIFF
--- a/configuration/chairman/byron-shelley/configuration.yaml
+++ b/configuration/chairman/byron-shelley/configuration.yaml
@@ -79,11 +79,13 @@ TracingVerbosity: NormalVerbosity
 # configuration below. The logging backend is called Katip.
 setupBackends:
   - KatipBK
+# - TraceForwarderBK
 
 # This specifies the default backends that trace output is sent to if it
 # is not specifically configured to be sent to other backends.
 defaultBackends:
   - KatipBK
+# - TraceForwarderBK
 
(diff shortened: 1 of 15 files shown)


configuration/chairman/byron-shelley/configuration.yaml
```diff
@@ -260,30 +262,31 @@ options:
   # go to the default backend, only to the explicitly listed backends.
   mapBackends:
     cardano.node.BlockFetchDecision.peers:
-      - TraceForwarderBK
+      # - TraceForwarderBK
```
CodiePP, line 265: you can keep it here; it won't hurt. Once the backend is activated in "setupBackends", then it will become active. If the backend is not loaded, then this configuration does not have an effect.

CodiePP: the benefit for the user: only one change will turn on the "TraceForwarderBK" and thus the routing to _RTview_

denisshevchenko:
Strictly speaking, after the latest changes in RTView, all these changes in default configuration files are not mandatory anymore. This is because RTView, during interactive configuration dialog, will show to user _particular changes_ that should be made in the node's configuration file. It's much easier for the user.

PR #1878, opened 2020-09-20
Correct cardano-cli location and 'version' arg

LintCroft:
Built version 1.9.3
cardano-cli location appears to be in 'cardano-node-<TAGGED VERSION>/x/cardano-cli/build/cardano-cli/' now.
'--version' arg doesn't exist, must use 'version'


intricate:
Just a note that both `cardano-cli --version` and `cardano-cli version` should work fine.

LintCroft:
> Just a note that both `cardano-cli --version` and `cardano-cli version` should work fine.
> 
> Edit: Wait, version `1.9.3`? The CLI has undergone significant changes since then.

I observe:
```
$ cardano-cli --version
Invalid option '--version'

Did you mean this?
    version
```

Good point about the version, I checkout out that without confirming which was latest. I will run through this again with 1.20.0 and update this PR.

LintCroft:
Documentation is correct for latest version - 1.20.0. Closing PR.

PR #1884, opened 2020-09-22
Windows builds

newhoggy:


.github/workflows/haskell.yml
```diff
@@ -0,0 +1,172 @@
...
+      fail-fast: false
+      matrix:
+        ghc: ["8.6.5"]
+        os: [ubuntu-latest, windows-latest]
+
+    steps:
+    - uses: actions/checkout@v1
+
+    - name: Install pkgconfiglite
+      if: matrix.os == 'windows-latest'
```
coot, line 23: Ach, that's the trick to run a step on a given platform, I was missing it somewhere else :)

.github/workflows/haskell.yml
```diff
@@ -0,0 +1,172 @@
...
+        cat >> cabal.project <<EOF
+        package cardano-crypto-praos
+          flags: -external-libsodium-vrf
+        EOF
+
+    - name: Cabal update
+      run: cabal update
+
+    - name: Select build directory
+      run: echo "::set-env name=CABAL_BUILDDIR::dist"
```
coot, line 97: Will this pick the current directory?  On windows I had problems with length of paths names, so I always build somewhere very close to the root, e.g. `C://d`.

newhoggy: So far this has been working.  Shortening the `dist-newstyle` to `dist` was just enough to squeeze in under the path name limit.

PR #1894, opened 2020-09-25
Use tasty to drive chairman

newhoggy:
This makes it possible to select which tests you want to run.  For example:

```
$ cabal test cardano-node-chairman --enable-tests --test-show-details=direct --test-options '--pattern Network'
Running 1 test suites...
Test suite tests: RUNNING...
test/Spec.hs
  Spec
    Network
      isPortOpen False: OK
          ✓ isPortOpen False passed 1 test.
      isPortOpen True:  OK
          ✓ isPortOpen True passed 1 test.
```

Being able to select the test you want to run will be important for bringing up testnets without also runnings the tests and vice versa.


erikd (approved):

newhoggy:
bors merge

PR #1896, opened 2020-09-25
Remove stack.yaml.  This is no longer supported

newhoggy:


deepfire:
@newhoggy, this also needs to remove https://github.com/input-output-hk/cardano-node/blob/master/.buildkite/pipeline.yml#L7-L10

erikd (approved):

newhoggy:
bors merge

PR #1897, opened 2020-09-27
Run a chairman process per node

newhoggy:


erikd (approved):

newhoggy:
bors merge

newhoggy:
bors merge

cardano-node-chairman/test/Test/Cardano/Node/Chairman/ByronShelley.hs
```diff
@@ -652,6 +653,13 @@ prop_chairman = H.propertyOnce . H.workspace "chairman" $ \tempAbsPath -> do
         )
       )
 
+    return hProcess
+
+  -- Check for chairman success
+  forM_ (L.zip allNodes hChairmanProcesses) $ \(node, hProcess) -> do
```
coot, line 659: I think this is not right.  There should be a single `chairman` process connected to `allNodes`, not a `chairman` per node.

coot: Or is this going to be fixed in a subsequent PR?

PR #1904, opened 2020-09-30
Update the mainnet config to add ShelleyHardForkNotBeforeEpoch setting

erikd:
See the commit comment for details.

### DIFF
--- a/configuration/cardano/mainnet-config.json
+++ b/configuration/cardano/mainnet-config.json
@@ -11,6 +11,7 @@
   "RequiresNetworkMagic": "RequiresNoMagic",
   "ShelleyGenesisFile": "mainnet-shelley-genesis.json",
   "ShelleyGenesisHash": "1a3be38bcbb7911969283716ad7aa550250226b76a61fc51cc9a9a35d9276d81",
+  "ShelleyHardForkNotBeforeEpoch": 208,
   "TraceBlockFetchClient": false,
   "TraceBlockFetchDecisions": false,
   "TraceBlockFetchProtocol": false,


configuration/cardano/mainnet-config.json
```diff
@@ -11,6 +11,7 @@
   "RequiresNetworkMagic": "RequiresNoMagic",
   "ShelleyGenesisFile": "mainnet-shelley-genesis.json",
   "ShelleyGenesisHash": "1a3be38bcbb7911969283716ad7aa550250226b76a61fc51cc9a9a35d9276d81",
+  "ShelleyHardForkNotBeforeEpoch": 207,
```
mrBliss, line 14: Edsko confirmed to me that we should actually use 208 here, not 207.

erikd: Updated.

cardano-node/src/Cardano/Node/Configuration/POM.hs
```diff
@@ -222,11 +222,11 @@ instance FromJSON PartialNodeConfiguration where
              }
 
       parseHardForkProtocol v = do
-        npcTestShelleyHardForkAtEpoch   <- v .:? "TestShelleyHardForkAtEpoch"
+        npcShelleyHardForkAtEpoch   <- v .:? "ShelleyHardForkAtEpoch"
         npcTestShelleyHardForkAtVersion <- v .:? "TestShelleyHardForkAtVersion"
```
mrBliss, line 226: Hmm, I'm not sure about the `Test-` prefix thing. `ShelleyHardForkAtEpoch` should *never* be used for mainnet, only for testing/testnets, because for mainnet we want to trigger the hard fork *at a version*. Following that reasoning, `TestShelleyHardForkAtVersion` should become `ShelleyHardForkAtVersion`, but that *is* used for mainnet.

erikd: Ok, I will revert that part.

dcoutts (approved):

erikd:
bors merge

PR #1907, opened 2020-10-01
Update index state

newhoggy:


mrBliss:
The title is a bit misleading, this doesn't just update the index state (only change + some potential fallout), but it makes the code compatible with GHC 8.10. Maybe that should be done in a separate PR.

Moreover, @dcoutts do you still want to hold off on updating the index state until we have a 1.21 release?

mrBliss:
@newhoggy FYI, I'll soon need an update of the index state. Concretely, the next update of the dependency of `ouroboros-network`, *after* the one in https://github.com/input-output-hk/cardano-node/pull/1928, requires it (to bring in the new `nothunks` package). Maybe you can spin off a separate PR that just updates the index state? I can do that as well.

Also, the other thing this PR does is mostly the removal of redundant imports. As such changes are backwards-compatible with GHC 8.6.5, we don't have to wait for the switch to GHC 8.10 (.2 or .3) to merge them. If you rebase and update the PR, I'd be happy to review/approve it. By the time we switch to GHC 8.10, we might require some more of these changes, but that's not a problem.

mrBliss:

PR #1912, opened 2020-10-01
Fix severity level for TraceLabelPeer.

karknu:
Before this all TraceBlockFetch messages was at Debug level rather than
the expected Info level.


karknu:
Fixes #1584.

karknu:
Dupe of #1910.

PR #1916, opened 2020-10-01
Fix chairman test output

newhoggy:
This fixes the test output in CI.

For some reason `nix` makes the source code of tests unavailable if they are in a library, which then causes the test output to be unable to find them and print useless `forallN` lines instead.

The fix is to move all the test code from the library into the test component (which is done by use of a `common` stanza)

See [before](https://hydra.iohk.io/build/4286853/nixlog/12) and [after](https://hydra.iohk.io/build/4286428/nixlog/36) for comparison.

### DIFF
--- a/cardano-node-chairman/cardano-node-chairman.cabal
+++ b/cardano-node-chairman/cardano-node-chairman.cabal
@@ -10,13 +10,38 @@ license-files:         LICENSE
                        NOTICE
 build-type:            Simple
 
-library
+common common-modules
   hs-source-dirs:       src
-  build-depends:        base >= 4.12 && < 5
+  other-modules:        Test.Process
+                        Test.Base
+                        Testnet.ByronShelley
+                        Testnet.Conf
+                        Testnet.Shelley
+
+executable cardano-node-chairman
+  hs-source-dirs:       app
+  main-is:              cardano-node-chairman.hs
+  other-modules:        Cardano.Chairman
+                        Cardano.Chairman.Options
+                        Paths_cardano_node_chairman
+  default-language:     Haskell2010
+  ghc-options:          -threaded
+                        -Wall
+                        -rtsopts
+                        "-with-rtsopts=-T"
+                        -Wno-unticked-promoted-constructors
+  build-depends:        base >=4.12 && <5
                       , aeson
                       , async
+                      , bytestring
                       , call-stack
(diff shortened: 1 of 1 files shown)


rvl:
Which files do the tests need? In the Haskell.nix build, the library, test-suite component, and the execution of tests are all separate derivations. By default, the tests are executed in an empty directory.

There is a module option to change that which you can add to `nix/haskell.nix`:
```
packages.cardano-node-chairman.components.tests.chairman-tests.keepSource = true;
```

newhoggy:
The `test` component needs access to the source from the `library` component in the same file.

newhoggy:
> Which files do the tests need? In the Haskell.nix build, the library, test-suite component, and the execution of tests are all separate derivations. By default, the tests are executed in an empty directory.
> 
> There is a module option to change that which you can add to `nix/haskell.nix`:
> 
> ```
> packages.cardano-node-chairman.components.tests.chairman-tests.keepSource = true;
> ```

I'm trying out your suggestion here: https://github.com/input-output-hk/cardano-node/pull/1921

newhoggy:
Unfortunately the suggested nix fix didn't work.  See https://github.com/input-output-hk/cardano-node/pull/1921 and the result here: https://hydra.iohk.io/build/4290502/nixlog/46

erikd (approved):

newhoggy:
bors merge

cardano-node-chairman/cardano-node-chairman.cabal
```diff
@@ -10,22 +10,54 @@ license-files:         LICENSE
                        NOTICE
 build-type:            Simple
 
-library
```
newhoggy, line 13: Previously the source files were under here, but hedgehog can't find them, so I converted this into a `common` module.

cardano-node-chairman/cardano-node-chairman.cabal
```diff
@@ -34,72 +66,27 @@ library
...
 
   if os(windows)
-     build-depends:    Win32
+     build-depends:     Win32
   else
-     build-depends:    unix
+     build-depends:     unix
 
 test-suite chairman-tests
+  import:               common-modules
```
newhoggy, line 82: This means the source files are included with the test module via this import.

PR #1920, opened 2020-10-02
Catch em all

newhoggy:


cardano-node-chairman/test/Spec/Network.hs
```diff
@@ -56,3 +64,9 @@ hprop_isPortOpen_True = H.propertyOnce . H.workspace "temp/network" $ \_ -> do
             case socketResult of
               Right socket -> return (socket, n)
               Left (_ :: IOException) -> openOnePortFrom ns
+
+hprop_moo :: Property
+hprop_moo = H.propertyOnce . H.workspace "temp/network" $ \_ -> do
+  e <- H.catchAssertion H.failure return
+  H.noteShow_ e
+  H.throwAssertion e
```
newhoggy, line 72: Example of catching the error then showing it, then re-throwing it.

PR #1921, opened 2020-10-02
Fix chairman test output via nix

newhoggy:
We should be able to see the results here when it completes: https://github.com/input-output-hk/cardano-node/pull/1922


newhoggy:
Unfortunately it didn't work: https://hydra.iohk.io/build/4290502/nixlog/46

newhoggy:
bors merge

PR #1927, opened 2020-10-05
Remove default socket path in node configuration

Jimbo4350:


jbgi (approved):
tested sucessully on shelley_qa

dcoutts (approved):
LGTM

And I agree that this should not have a default.

PR #1929, opened 2020-10-05
Update tx-metadata.md

olgahryniuk:
Adding an extended overview of tx metadata, its structure, and some characteristics. Worked on this with Jordan Millar.

### DIFF
--- a/doc/reference/tx-metadata.md
+++ b/doc/reference/tx-metadata.md
@@ -1,6 +1,6 @@
 # Transaction Metadata
 
-Transaction metadata (tx metadata) contains details about a specific transaction, including sender and receiver IDs, transaction comments, and tags. Adding metadata to transactions  provides transaction information by adding arbitrarily structured data onto the chain and is a useful feature in Cardano Shelley. Tx metadata is stored on-chain and is carried along with each transaction. A factor in its design is that the on-chain metadata is not stored in the ledger state and does not influence transaction validation, thereby not compromising ledger performance.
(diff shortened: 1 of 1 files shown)


doc/reference/tx-metadata.md
```diff
@@ -1,16 +1,75 @@
...
-into the internal format. We provide two different mappings between tx
-metadata and JSON, useful for different purposes.
-In both schemes, the top level JSON is an object indexed by integers which
-are mapped to JSON values.
+For example, tx metadata can be used to:
+
++ certify ownership and exchange of assets specifying asset owners in different time periods, transfer details, or asset value at the time of transfer
++ certify documents and signatures, by using a public hash that proves the document’s existence 
+
+Transactions can contain metadata whose hash is part of the body of the transaction. This hash is read by all transaction signatures. The metadata value is kept outside the transaction to allow checking its integrity and authentication.
```
Jimbo4350, line 12: Instead of:
> This hash is read by all transaction signatures. The metadata value is kept outside the transaction to allow checking its integrity and authentication.

I would say something like:
> Because the metadata hash is in the tx body, this allows for integrity checking and authentication of the metadata

doc/reference/tx-metadata.md
```diff
@@ -1,16 +1,75 @@
...
+                       	metadata.
+  --metadata-json-file FILE
+                       	Filepath of the metadata file, in JSON format.
+  --metadata-cbor-file FILE
+                       	Filepath of the metadata, in raw CBOR format.
+```
+
+## Metadata schemas - mappings and formats
+
+On-chain metadata that is carried along with transactions is encoded according to CBOR. To create a transaction, you can add the metadata with pre-encoded CBOR. Alternatively, you can add metadata in JSON format, which will be converted into the internal format. In this section, we provide examples of the two different mappings: tx metadata and JSON, which are useful for different purposes. In both schemas, the top-level JSON is an object indexed by integers, which are mapped to JSON values.
```
Jimbo4350, line 72: Instead of 
> In this section, we provide examples of the two different mappings: tx metadata and JSON

I would say:
> In this section, we provide examples of the two different mappings _between_ tx metadata and JSON

PR #1932, opened 2020-10-05
Refine Cardano.Api.{Byron,Shelley}

Jimbo4350:


cardano-api/src/Cardano/Api/Shelley.hs
```diff
@@ -1,95 +1,159 @@
...
+    Witness (ShelleyBootstrapWitness, ShelleyKeyWitness, ShelleyScriptWitness),
+    ShelleyWitnessSigningKey (WitnessPaymentKey,
+                              WitnessPaymentExtendedKey,
+                              WitnessStakeKey,
+                              WitnessStakeExtendedKey,
+                              WitnessStakePoolKey,
+                              WitnessGenesisKey,
+                              WitnessGenesisExtendedKey,
+                              WitnessGenesisDelegateKey,
+                              WitnessGenesisDelegateExtendedKey),
```
intricate, line 44: Overall, I don't think we should be aligning our code in this way and I'll just give a couple of reasons:

1. If the name `ShelleyWitnessSigningKey` changes, then all of these lines need to be realigned (adds `diff` noise and is error prone in the sense that some may not notice that they need to re-align if they're doing a project-wide search-and-replace, for example)

2. I'd say that, generally, we should avoid variable-length indentation. I think the `cardano-wallet` wiki makes a good point on this: https://github.com/input-output-hk/cardano-wallet/wiki/Coding-Standards#avoid-variable-length-indentation. Although, I do acknowledge that we already have a bunch of variable-length indentation in the codebase.

3. Also, generally, I think we should stick with the "comma-leading" style:
```
    ShelleyWitnessSigningKey 
      ( WitnessPaymentKey
      , WitnessPaymentExtendedKey
      , WitnessStakeKey
      ...
      ),
```

intricate: Also, side note, in the case above, aren't those just normal constructor exports? So couldn't you just export `ShelleyWitnessSigningKey (..)`? Or are you doing it this way for added readability?

If so, the trouble with that comes when a new constructor is added and one may forget to export it.

erikd: As a further points to the ones Luke has already made, align should aim to be invariant with other changes.

Eg:
```
import A      (A (..), toA)
import B      (B (..), toB)
```
will require re-indenting these import lists as soon as we import say `Cardano.DbSync.Era.Shelley.Insert`.

This means the following is better
```
import A (A (..), toA)
import B (B (..), toB)
```
because the alignment rule would not require any changes to the existing imports if `Cardano.DbSync.Era.Shelley.Insert` were added.

This rule can be boiled down to "use indent rules that as maximally resistant to requiring semantically null changes".

I also prefer not to needlessly consume vertical code space (from working too long on laptops no doubt) so I would probably format that as:
```
    ShelleyWitnessSigningKey (WitnessPaymentKey, WitnessPaymentExtendedKey, WitnessStakeKey,
        WitnessStakeExtendedKey, WitnessStakePoolKey, WitnessGenesisKey, WitnessGenesisExtendedKey,
        WitnessGenesisDelegateKey, WitnessGenesisDelegateExtendedKey),
```

This reminds me of another formatting rule I like is that indentation should suggest nesting. Once thing I see all the time in out code base which I dislike is:
```
wibble x = case x of
   A -> ....
   B -> ....
```
which afaiac is far less clear than:
```
wibble x =
  case x of
     A -> ....
     B -> ....
 ```
where the `A` and `B` are much more obviously heirarchically subserviant to the `case`.

Jimbo4350: > 

Also, side note, in the case above, aren't those just normal constructor exports? So couldn't you just export ShelleyWitnessSigningKey (..)? Or are you doing it this way for added readability?

If so, the trouble with that comes when a new constructor is added and one may forget to export it.

I went for an explicit approach because it makes it clearer which era the constructors belong to. If we do decide to export more upstream library modules through `cardano-api` this would be preferable as you could instantly figure out where constructors/types are coming, saving us the usual wild goose chase. I think this is the better trade off.

Jimbo4350: >Also, side note, in the case above, aren't those just normal constructor exports? So couldn't you just export ShelleyWitnessSigningKey (..)? Or are you doing it this way for added readability?

>If so, the trouble with that comes when a new constructor is added and one may forget to expo

 When I was under the impression that we were also exporting upstream library modules via `cardano-api`, I explicitly exported constructors to avoid the usual wild goose chase when trying to figure out what came from which module. I still think its of value at the potential expense of forgetting to export a constructor.

dcoutts (approved):
LGTM!

PR #1928, opened 2020-10-05
Refactor node addresses and updated ouroboros-network dependency

coot:
Updated version of `ouroboros-network`.  This PR also includes a refactoring of address types that the node is using to a more consistent form.

PR #1935, opened 2020-10-06
Update dependency on ouroboros-network

mrBliss:
Notable changes:
* https://github.com/input-output-hk/ouroboros-network/pull/2640
* https://github.com/input-output-hk/ouroboros-network/pull/2649

### DIFF
--- a/cardano-node/src/Cardano/Node/Protocol/Byron.hs
+++ b/cardano-node/src/Cardano/Node/Protocol/Byron.hs
@@ -108,7 +108,7 @@ mkConsensusProtocolByron NodeByronProtocolConfiguration {
                                 npcByronSupportedProtocolVersionAlt)
         (Update.SoftwareVersion npcByronApplicationName
                                 npcByronApplicationVersion)
-        optionalLeaderCredentials
+        (maybeToList optionalLeaderCredentials)
 
 
 readGenesis :: GenesisFile
@@ -177,9 +177,14 @@ readLeaderCredentials genesisConfig
                          . hoistEither
                          $ canonicalDecodePretty delegCertFileBytes
 
+          -- The label is used to distinguish multiple sets of credentials
+          -- during tracing. We use the public signing key as the label. TODO
+          -- @deepfire works for you? Or should we hash it first?
+         let label = Text.pack $ show $ Byron.Crypto.toVerification signingKey
+
          bimapExceptT CredentialsError Just
            . hoistEither
-           $ mkByronLeaderCredentials genesisConfig signingKey delegCert
+           $ mkByronLeaderCredentials genesisConfig signingKey delegCert label
 
   where
     deserialiseSigningKey :: LB.ByteString
--- a/cardano-node/src/Cardano/Node/Protocol/Shelley.hs
+++ b/cardano-node/src/Cardano/Node/Protocol/Shelley.hs
@@ -33,6 +33,7 @@ import           Control.Monad.Trans.Except.Extra (firstExceptT, handleIOExceptT
                      newExceptT)
 
 import qualified Cardano.Crypto.Hash.Class as Crypto
+import           Cardano.Crypto.KES (deriveVerKeyKES)
 
 import qualified Ouroboros.Consensus.Cardano as Consensus
 import           Ouroboros.Consensus.Cardano.ShelleyHFC
@@ -110,7 +111,7 @@ mkConsensusProtocolShelley NodeShelleyProtocolConfiguration {
         (ProtVer npcShelleySupportedProtocolVersionMajor
                  npcShelleySupportedProtocolVersionMinor)
         (MaxMajorProtVer npcShelleyMaxSupportedProtocolVersion)
-        optionalLeaderCredentials
+        (maybeToList optionalLeaderCredentials)
 
 genesisHashToPraosNonce :: GenesisHash -> Nonce
 genesisHashToPraosNonce (GenesisHash h) = Nonce (Crypto.castHash h)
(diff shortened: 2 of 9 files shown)


mrBliss:
@deepfire I suppose you have made these changes already somewhere? Can you take a look at the TODOs that mention "deepfire"?

coot (approved):
LGTM, but there are TODO's left for @deepfire.

mrBliss:
> LGTM, but there are TODO's left for @deepfire.

Indeed, see my comment above.

@coot I thought the goal was not to merge it in master ASAP, but to make it possible for you to bring in the network-related changes that come after the consensus changes that I have just propagated. I thought you could use this PR as the starting point, and have you -- if you want --  push your changes to this branch. Or did I misunderstand your goal?

coot:
Either way will work for me.

cardano-node/src/Cardano/Tracing/Metrics.hs
```diff
@@ -79,29 +78,12 @@ instance HasKESMetricsData ByronBlock where
...
-      -- Multiple eras can have 'KESMetricsData', e.g., Shelley and Shelley +
-      -- native assets. We don't know which era we're in, so we have to make a
-      -- choice: we choose the last era's 'KESMetricsData'.
-      --
-      -- TODO this might need revision when we have updated everything for the
-      -- first hard fork after Shelley, but before it actually happened. At
-      -- that moment, we'll trace the KESMetricsData of the upcoming era,
-      -- instead of the KESMetricsData of the current Shelley era.
-      combineKESMetrics :: [KESMetricsData] -> KESMetricsData
-      combineKESMetrics = lastDef NoKESMetricsData
```
mrBliss, line 107: This PR solves the problem described in the TODO. For each Shelley-based era, the KES metrics will be traced independently. @deepfire this means that when a node is started with credentials for both Shelley and Allegra, that the two KES metrics will be traced per slot. Maybe they should be groupedd per label or something?

cardano-node/src/Cardano/Node/Protocol/Byron.hs
```diff
@@ -177,9 +177,14 @@ readLeaderCredentials genesisConfig
                          . hoistEither
                          $ canonicalDecodePretty delegCertFileBytes
 
+          -- The label is used to distinguish multiple sets of credentials
+          -- during tracing. We use the public signing key as the label. TODO
+          -- @deepfire works for you? Or should we hash it first?
+         let label = Text.pack $ show $ Byron.Crypto.toVerification signingKey
```
deepfire, line 183: @mrBliss, this looks fine, as we don't have a basis for preference yet.

cardano-node/src/Cardano/Node/Protocol/Shelley.hs
```diff
@@ -163,14 +164,20 @@ readLeaderCredentials (Just ProtocolFilepaths {
     KesSigningKey kesKey <-
       firstExceptT FileError . newExceptT $ readFileTextEnvelope (AsSigningKey AsKesKey) kesFile
 
+    -- The label is used to distinguish multiple sets of credentials during
+    -- tracing. We use the public KES key as the label. TODO @deepfire works for
+    -- you? Or should we hash it first?
+    let label = T.pack (show (deriveVerKeyKES kesKey))
```
deepfire, line 170: @mrBliss, this looks fine, as we don't have a basis for preference yet.

cardano-node/src/Cardano/Tracing/Metrics.hs
deepfire: @mrBliss, we don't particularly care for KES metrics in dense pools, I'd say -- so the less resources we spend here the better.

mrBliss: Unlike the other two comments, this question is more about having two eras instead of having multiple stake pools.

> this means that when a node is started with credentials for both Shelley and Allegra, that the two KES metrics will be traced per slot.

Around the transition point from Shelley to Allegra, all block producing nodes should have credentials for both eras. I imagine the operators will want to see the metrics of both KES keys independently.

deepfire: Indeed, misunderstood you initially.

>  I imagine the operators will want to see the metrics of both KES keys independently.

Absolutely, yes.

mrBliss: Right. As mentioned on Slack, this doesn't have to been done right away, it should just be done before the transition to the Allegra era (and any testnets that exercise such a transition). You can create an issue for it to not forget about it.

cardano-node/src/Cardano/Node/Protocol/Byron.hs
mrBliss: I'm curious what this looks like when traced, I'm afraid that the type doesn't have a user-friendly `Show`..

cardano-node/src/Cardano/Node/Protocol/Shelley.hs
mrBliss: Same as for Byron, I'm curious what this looks like when traced.

coot:
Closing as it is included in #1928

mrBliss:
> Closing as it is included in #1928

Hmm, https://github.com/input-output-hk/cardano-node/pull/1937 had this PR as target. @deepfire For me it's fine if you push your commits from #1937 to the new PR.

cardano-node/src/Cardano/Tracing/Metrics.hs
deepfire: @mrBliss, so in https://github.com/input-output-hk/cardano-node/tree/cad-1920-bulk-creds I've properly wired the credentials label into logs, so that label should now be displayed in structured logs.  I still need to test it though.

@mrBliss, now that I re-evaluate your question, it seems that tracking KES metrics separately per-epoch is fine -- they'll be labelled appropriately.

We'll need to accomodate it in RTView, though.

PR #1945, opened 2020-10-08
Update dependencies

newhoggy:


cardano-node/src/Cardano/Tracing/OrphanInstances/Shelley.hs
```diff
@@ -317,9 +317,6 @@ instance Era era => ToObject (UtxowPredicateFailure era) where
              , "txBodyMetaDataHash" .= txBodyMetaDataHash
              , "fullMetaDataHash" .= fullMetaDataHash
              ]
-  toObject _verb InvalidMetaData =
-    mkObject [ "kind" .= String "InvalidMetaData"
-             ]
```
intricate, line 322: Why are we removing this case? It still appears to be a constructor of the `UtxowPredicateFailure era` type.

newhoggy: Not sure, the constructor is gone altogether.

mrBliss:
Please see https://github.com/input-output-hk/cardano-node/pull/1928

mrBliss:

PR #1951, opened 2020-10-08
Independent means of checking if port is open on MacOS

newhoggy:


cardano-node-chairman/src/Testnet/ByronShelley.hs
```diff
@@ -644,6 +644,9 @@ testnet H.Conf {..} = do
     H.onFailure . H.noteM_ $ H.readFile nodeStdoutFile
     H.onFailure . H.noteM_ $ H.readFile nodeStderrFile
 
+    when (OS.os == "darwin") $ do
+      H.onFailure . H.noteIO_ $ IO.readProcess "lsof" ["-iTCP:" <> portString, "-sTCP:LISTEN", "-n", "-P"] ""
```
erikd, line 648: Ewww, that is a bit nasty, but ..... ok.

erikd (approved):

newhoggy:
bors merge

PR #1955, opened 2020-10-09
Use fail instead of panic in some JSON parsers throughout cardano-node

intricate:


intricate:
Subsumed by https://github.com/input-output-hk/cardano-node/pull/2088

PR #1967, opened 2020-10-12
Disable edge tests

newhoggy:


coot:
Is required before merging: https://github.com/input-output-hk/cardano-node/pull/1967

dcoutts (approved):

newhoggy:
Merged https://github.com/input-output-hk/cardano-node/pull/1970 instead.  This is because disabling the node test leaves a gap in the testing.  It is better to revert to using the chairman cluster test instead of just disabling the edge test to ensure there is no gap.

coot:
bors merge

PR #1970, opened 2020-10-12
Switch back to chairman cluster test

newhoggy:
The edge tests were failing in CI on every build for me, so I would like to revert to the chairman cluster test which works most of the time until we figure out why the edge tests have such a high failure rate and have a fix for it.


erikd (approved):
I hardly even know what I am looking at but :man_shrugging:

newhoggy:
bors merge

PR #1981, opened 2020-10-14
General chairman test code quality

newhoggy:
* Some of the chairman tests are older code that don't use the new system allocated random ports allocation.  Updating all tests to do this.
* Running `lsof` for MacOS and Linux
* Use a randomly generated test magic.  This is to protect against possible misconfiguration that might allow separate testnets to merge.
* Disable port test.  We will be relying on the chairman test process to verify everything is working.
* Catch `IOException` in sprocket test so as to not fail test prematurely.


erikd (approved):

newhoggy:
bors merge

PR #1990, opened 2020-10-19
Switch to GHC 8.10.2

mrBliss:
This is done in a number of steps, see the individual commit messages for more details.


newhoggy (approved):

cardano-node/cardano-node.cabal
```diff
@@ -133,6 +133,12 @@ library
                        -Wpartial-fields
                        -Wcompat
 
+  -- Pattern match checking has improved since 8.6.5. GHC now considers more
+  -- pattern matches as redundant than before, so disable this warning for older
+  -- versions of GHC that still consider them incomplete.
+  if impl(ghc < 8.10)
+    ghc-options:       -Wno-incomplete-patterns
+
```
mrBliss, line 141: This doesn't work the way it should because for some reasons the warning flags are duplicated/overridden in `cabal.project`.

https://github.com/input-output-hk/cardano-node/blob/8d110471462f296f66bc91735426047a4cb1909a/cabal.project#L11-L24

I guess the advantage of putting them there is that it's easy to change all of them at once. However, I'm not sure whether that's the right thing to do: now the flags are in two places (`cabal.project` and the separate `.cabal` files). People wanting to change the flags in the `.`cabal` files (like me) will be surprised when their flags have no effect (as they're overridden by the flags in the `cabal.project` file).

I propose the following:
* We make sure all the warnings listed in `cabal.project` are enabled for each package component in this project.
* In `cabal.project`, we trim `ghc-options` to `-Werror`. All the warnings are already in the `.cabal` files, but `-Werror` is a switch that you don't want to have in each `.cabal` file.

I'll add a commit with this change. We can of course discuss this.

mrBliss: Oh, I just discovered that the `.cabal` files include more warnings than the `cabal.project`, e.g., `-Wcompat`. So this duplication was already out of sync/date.

mrBliss: See https://github.com/input-output-hk/cardano-node/pull/1990/commits/49699c5422e970de776e462f33915318046a3f67

nix/haskell.nix
```diff
@@ -53,7 +53,8 @@ let
           "hpc"
           "mtl" "parsec" "text" "transformers"
           "xhtml"
-          # "stm" "terminfo"
+          # "stm"
+          "terminfo"
```
mrBliss, line 57: I now think this only has effect on Windows (see the `isWindows` on line 34), although my understanding of Nix is limited.

mrBliss: Alternatively, we just wait for #1977 to be merged, as it will remove the dependency on `vty` (which relies on `terminfo`) :grin:.

deepfire: @mrBliss, your change should be sufficient -- `isWindows` only applies to the first, small attrset:
https://github.com/input-output-hk/cardano-node/blob/70d6ab9ceae2fe9929b640ec6b066543d4584e6a/nix/haskell.nix#L34-L38

mrBliss:
Now that #1977 has been bors'ed, I have rebased this branch on `bors/staging`. I believe those commits will disappear from this PR when they are part of master.

I also added 232744f and 523da4b2 to clean up after #1977.

Since we no longer depend on `vty` and thus `terminfo` after #1977, I have dropped the `terminfo` workarounds for Nix since they didn't seem to fix the problem.

mrBliss:
bors merge

mrBliss:
Let's give it another go.

mrBliss:
bors merge

intricate (approved):

mrBliss:
Hmm, I'm still seeing some failing jobs: https://hydra.iohk.io/eval/970884.

For example https://hydra.iohk.io/build/4513297/nixlog/1:
```
unpacking sources
unpacking source archive /nix/store/96229sf9xxwixrarg8nz32r2z2w1lv7k-generic-data-0.9.1.0.tar.gz
source root is generic-data-0.9.1.0
setting SOURCE_DATE_EPOCH to timestamp 1000000000 of file generic-data-0.9.1.0/test/unit.hs
patching sources
building
Compiling package Setup.hs
[1 of 1] Compiling Main             ( Setup.hs, Setup.o )

Setup.hs:1:1: error:
    Could not find module `Distribution.Extra.Doctest'
    Perhaps you meant Distribution.Simple.Doctest (from Cabal-3.2.0.0)
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
  |
1 | import Distribution.Extra.Doctest (defaultMainWithDoctests)
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

builder for '/nix/store/3x12bvxzjln714qp4wwf94w79p6jy6i2-generic-data-0.9.1.0-setup.drv' failed with exit code 1
```
I thought that f51256d9725b05d81b63eedad62f5ace4a4ae1ab would fix that, just like https://github.com/input-output-hk/ouroboros-network/commit/0d3c71e9532ed1d08367cbc1fc44d3ea29fb084e.

@hamishmack Thoughts?

mrBliss:
I have pushed cd420cb612e361aa86f5ff4116ec77b214973896 to see whether that fixes the current Nix issues.

(I'm a Haskell developer, not a Nix developer, so I'm just trying things without really understanding them. It's getting quite frustrating to see all GitHub Actions CI jobs green, yet some Nix issue is still wasting my time :slightly_frowning_face:)

mrBliss:
Uggh: https://hydra.iohk.io/jobset/Cardano/cardano-node-pr-1990#tabs-errors
```
error: unable to fork: Cannot allocate memory
```
:roll_eyes:

mrBliss:
bors merge

mrBliss:
bors merge

mark-stopka:
FYI: @mimi1vx, we should no longer encounter any GHC 8.10.2 problems in our OBS project...

PR #1998, opened 2020-10-21
Publish chairman test artifacts

newhoggy:
This uploads test configuration and logs as artefacts so they are easy to get when things go wrong.

image


erikd (approved):

newhoggy:
bors merge

PR #2003, opened 2020-10-21
Switch to ghc-8.10

newhoggy:


erikd (approved):

erikd:
Might be work cleaning up that commit history a bit. Lot of stuff there that we will not care about in a couple of months.

mrBliss:
Done in #1990

PR #2008, opened 2020-10-22
Fix nix tests on wine and musl64

hamishmack:


mrBliss (approved):
Cheers!

hamishmack:
These changed were included in 6eb72b3e

PR #2013, opened 2020-10-23
Extended gitattributes file

coot:
We need to protect the genesis files when one clones the repository on
Windows.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,3 +5,18 @@
 nix/.stack.nix/*.nix linguist-generated=true
 .stack-to-nix.cache linguist-generated=true
 nix/.stack-pkgs.nix linguist-generated=true
+
+# genesis files
+configuration/cardano/mainnet-byron-genesis.json -text
+configuration/cardano/mainnet-shelley-genesis.json -text
+configuration/cardano/shelley_qa-byron-genesis.json -text
+configuration/cardano/shelley_qa-shelley-genesis.json -text
+configuration/defaults/byron-mainnet/genesis.json -text
+configuration/defaults/byron-staging/genesis.json -text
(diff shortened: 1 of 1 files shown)


newhoggy:
I'm assuming this ensures `git` tracks these files as `text`.

newhoggy (approved):

coot:
I checked that it works when cloning a fresh repository, but it seems it does not work when switching branches which is a bit disappointing.

.gitattributes
```diff
@@ -5,3 +5,18 @@
...
+configuration/cardano/shelley_qa-shelley-genesis.json -text
+configuration/defaults/byron-mainnet/genesis.json -text
+configuration/defaults/byron-staging/genesis.json -text
+configuration/defaults/byron-testnet/genesis.json -text
+configuration/defaults/mainnet-silent/genesis.json -text
+configuration/defaults/mainnet-via-fetcher/genesis.json -text
+configuration/mainnet-ci/genesis.json -text
+configuration/mainnet-ci/shelley-staging-genesis.json -text
+configuration/mainnet-ci/shelley-staging-short-genesis.json -text
+configuration/mainnet-ci/testnet-genesis.json -text
```
mrBliss, line 22: Would `configuration/**.json` or `**.json` work?

coot: That would be future proof, and I think harmless for other `json` files.

coot: Can one use `**` in `gitattribute` files, and does it work on `Windows`?

mrBliss: Reading through https://git-scm.com/docs/gitattributes, I think `*.json` should work.

coot: thanks, I'll double check if that works.

PR #2030, opened 2020-10-28
Allow for longer start up time in chairman tests.

newhoggy:
Very occasionally 'until genesis start time at' doesn't get logged, which could be because we haven't allowed enough time for the node to reach the point where it would log

This will hopefully address this failure: https://github.com/input-output-hk/cardano-node/pull/2028/checks?check_run_id=1319609263


erikd (approved):

intricate (approved):

newhoggy:
bors merge

PR #2032, opened 2020-10-29
doco (retire.md) - Extracting current epoch is outdated for network type

rdlrt:
For combinator networks, the slotsPerEpoch is different for Byron and Shelley eras. As such the method to simply divide current unSlot by slotsPerEpoch would only be valid for TPraos (or Byron) network. Suggest a change to query EKG instead

### DIFF
--- a/doc/stake-pool-operations/retire_stakepool.md
+++ b/doc/stake-pool-operations/retire_stakepool.md
@@ -7,7 +7,7 @@ To retire a pool we need to:
 
 The deregistration certificate contains the _epoch_ in which we want to retire the pool. This epoch must be _after_ the current epoch and _not later than_ `eMax` epochs in the future, where `eMax` is a protocol parameter.
 
-So we first need to figure out the current epoch. For this, the easiest way to query epoch would be by querying using EKG port (part of node config). To query the current epoch , you can use the command (where jq is a command line tool for parsing JSON, part of EPEL package on linux):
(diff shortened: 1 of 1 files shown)


doc/stake-pool-operations/retire_stakepool.md
```diff
@@ -7,22 +7,10 @@ To retire a pool we need to:
...
-    export CARDANO_NODE_SOCKET_PATH=relay-db/node-socket
-    cardano-cli shelley query tip --mainnet
-
-    > Tip (SlotNo {unSlotNo = 856232}) ...
-
-This gives us
-
-    expr 856232 / 21600
-    > 39
+    curl -s -H 'Accept: application/json' http://127.0.0.1:12788/ | jq -r .cardano.node.ChainDB.metrics.epoch.int.val
```
erikd, line 12: The only thing missing here is that `hasEKG` needs to be enable in the config file and that the port number needs to match.

Add a note about `hasEKG` and this is good to go.

erikd (changes requested):

doc/stake-pool-operations/retire_stakepool.md
rdlrt: Apologies, totally missed notification about this comment. Added now

Jimbo4350 (approved):

Jimbo4350:
bors merge

Jimbo4350:
bors merge

erikd (approved):

erikd:
bors merge

newhoggy:
bors merge

newhoggy:
`master` already includes this text:

> You can also get the current epoch if you are monitoring a node with [EKG](../logging-monitoring/ekg.md):
> 
> ```
> curl -s -m 3 -H 'Accept: application/json' http://127.0.0.1:12788/ | jq '.cardano.node.metrics.epoch.int.val'
> > 51
> ```
> 
> So we are currently in epoch 51.

So some variant of these changes has already been merged.

PR #2033, opened 2020-10-29
Delete old chairman Haskell code

newhoggy:
The new chairman Haskell code lives under `cardano-node-chairman` now.

This can be deleted because the old chairman tests that ran in Hydra are now disabled.  Deleting this code means we won't be able to turn on the old chairman tests again.


erikd (approved):

jbgi (approved):

newhoggy:
bors merge

PR #2047, opened 2020-11-02
CAD-2167: TraceForwarder with a callback.

denisshevchenko:
The node has important information about itself:

1. Protocol
2. Version
3. Commit
4. Node start time
5. Slot length
6. Epoch length
7. KES period length
8. System start time

All this information should be traced to RTView (by `TraceForwarderBK` plugin) automatically, once after each connection with TraceAcceptorBK is established. Now `TraceForwarderBK` plugin provides a callback function for it.

_IMPORANT: Now it doesn't compile, I need help with Byron start time._

### DIFF
--- a/cardano-node/src/Cardano/Node/Configuration/Logging.hs
+++ b/cardano-node/src/Cardano/Node/Configuration/Logging.hs
@@ -1,6 +1,6 @@
 {-# LANGUAGE CPP #-}
-{-# LANGUAGE GADTs #-}
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE GADTs #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE Rank2Types #-}
 {-# LANGUAGE ScopedTypeVariables #-}
@@ -62,9 +62,9 @@ import           Cardano.Slotting.Slot (EpochSize (..))
 import           Ouroboros.Consensus.Block (BlockProtocol)
 import qualified Ouroboros.Consensus.BlockchainTime.WallClock.Types as WCT
 import           Ouroboros.Consensus.Byron.Ledger.Conversions
+import qualified Ouroboros.Consensus.Cardano as Consensus
 import           Ouroboros.Consensus.Cardano.Block
 import           Ouroboros.Consensus.Cardano.CanHardFork
-import qualified Ouroboros.Consensus.Cardano as Consensus
 import qualified Ouroboros.Consensus.Config as Consensus
 import           Ouroboros.Consensus.Config.SupportsNode (ConfigSupportsNode (..))
 import           Ouroboros.Consensus.HardFork.Combinator.Degenerate
--- a/cardano-node/src/Cardano/Node/Configuration/Logging.hs
+++ b/cardano-node/src/Cardano/Node/Configuration/Logging.hs
@@ -1,4 +1,6 @@
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE Rank2Types #-}
 {-# LANGUAGE ScopedTypeVariables #-}
@@ -6,6 +8,7 @@
 module Cardano.Node.Configuration.Logging
   ( LoggingLayer (..)
   , createLoggingLayer
+  , nodeBasicInfo
   , shutdownLoggingLayer
   -- re-exports
   , Trace
@@ -22,6 +25,10 @@ import           Cardano.Prelude hiding (trace)
 import qualified Control.Concurrent.Async as Async
 import           Control.Exception.Safe (MonadCatch)
 import           Control.Monad.Trans.Except.Extra (catchIOExceptT)
+import           Data.List (nub)
+import           Data.Text (pack)
+import           Data.Time.Clock (UTCTime, getCurrentTime)
+import           Data.Version (showVersion)
 
 import           Cardano.BM.Backend.Aggregation (plugin)
 import           Cardano.BM.Backend.EKGView (plugin)
@@ -35,7 +42,7 @@ import qualified Cardano.BM.Configuration.Model as Config
 import           Cardano.BM.Counters (readCounters)
 import           Cardano.BM.Data.Backend (Backend, BackendKind)
 import           Cardano.BM.Data.Counter
-import           Cardano.BM.Data.LogItem (LOContent (..), LOMeta (..), LoggerName,
+import           Cardano.BM.Data.LogItem (LOContent (..), LOMeta (..), LogObject (..), LoggerName,
                      PrivacyAnnotation (..), mkLOMeta)
 import           Cardano.BM.Data.Observable
 import           Cardano.BM.Data.Severity (Severity (..))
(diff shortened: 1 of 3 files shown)


cardano-node/src/Cardano/Node/Configuration/Logging.hs
```diff
@@ -211,3 +232,56 @@ createLoggingLayer ver nodeConfig' = do
...
+              -> Consensus.TopLevelConfig (CardanoBlock c)
+              -> UTCTime
+              -> IO [LogObject Text]
+nodeBasicInfo nc cfg nodeStartTime' = do
+  let CardanoLedgerConfig cfgByron cfgShelley cfgAllegra cfgMary = Consensus.configLedger cfg
+      protocolDependentItems =
+        case ncProtocol nc of
+          ByronProtocol ->
+             let genesis = byronLedgerConfig cfgByron
+             in [ ("systemStartTime",  show (getSystemStart $ Consensus.configBlock cfg))
```
denisshevchenko, line 249: It doesn't compile. @mrBliss could you help me please?

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -177,19 +175,6 @@ setupTrace loggingLayer = do
...
-  -> Word64
-  -> IO ()
-traceNodeUpTime tr nodeLaunchTime = do
-  now <- getMonotonicTimeNSec
-  let upTimeInNs = now - nodeLaunchTime
-  meta <- mkLOMeta Notice Public
-  traceNamedObject tr (meta, LogValue "upTime" (Nanoseconds upTimeInNs))
-  threadDelay 1000000
-  traceNodeUpTime tr nodeLaunchTime
-
```
mrBliss, line 192: Why is this removed? I suppose because you're now tracing `nodeStartTime`? Maybe make this unrelated change in a separate commit explaining why instead of sneaking it in here.

cardano-node/src/Cardano/Node/Configuration/Logging.hs
mrBliss: > It doesn't compile.
What's the error?

cardano-node/src/Cardano/Node/Configuration/Logging.hs
```diff
@@ -211,3 +232,56 @@ createLoggingLayer ver nodeConfig' = do
...
+            in [ ("systemStartTime",          show (SL.sgSystemStart genesisShelley))
+               , ("slotLengthShelley",        show (mkSlotLength $ SL.sgSlotLength genesisShelley))
+               , ("slotLengthAllegra",        show (mkSlotLength $ SL.sgSlotLength genesisAllegra))
+               , ("slotLengthMary",           show (mkSlotLength $ SL.sgSlotLength genesisMary))
+               , ("epochLengthShelley",       show (SL.sgEpochLength genesisShelley))
+               , ("epochLengthAllegra",       show (SL.sgEpochLength genesisAllegra))
+               , ("epochLengthMary",          show (SL.sgEpochLength genesisMary))
+               , ("slotsPerKESPeriodShelley", show (SL.sgSlotsPerKESPeriod genesisShelley))
+               , ("slotsPerKESPeriodAllegra", show (SL.sgSlotsPerKESPeriod genesisAllegra))
+               , ("slotsPerKESPeriodMary",    show (SL.sgSlotsPerKESPeriod genesisMary))
```
mrBliss, line 273: There's quite a bit of duplication here. Repeating all three for all the upcoming eras is not so nice. I would write a function that takes a `ShelleyGenesis era` and a `String` or `Text` ("Shelley", "Allegra", or "Mary").

cardano-node/src/Cardano/Node/Configuration/Logging.hs
```diff
@@ -211,3 +232,56 @@ createLoggingLayer ver nodeConfig' = do
...
+                , ("epochLengthByron", show (fromByronEpochSlots $ Gen.configEpochSlots genesis))
+                ]
+          ShelleyProtocol ->
+            let genesis = shelleyLedgerGenesis $ shelleyLedgerConfig cfgShelley
+            in [ ("systemStartTime",          show (SL.sgSystemStart genesis))
+               , ("slotLengthShelley",        show (mkSlotLength $ SL.sgSlotLength genesis))
+               , ("epochLengthShelley",       show (SL.sgEpochLength genesis))
+               , ("slotsPerKESPeriodShelley", show (SL.sgSlotsPerKESPeriod genesis))
+               ]
+          CardanoProtocol ->
```
mrBliss, line 260: What about Byron?

cardano-node/src/Cardano/Node/Configuration/Logging.hs
```diff
@@ -211,3 +232,56 @@ createLoggingLayer ver nodeConfig' = do
...
+            let genesisShelley = shelleyLedgerGenesis $ shelleyLedgerConfig cfgShelley
+                genesisAllegra = shelleyLedgerGenesis $ shelleyLedgerConfig cfgAllegra
+                genesisMary    = shelleyLedgerGenesis $ shelleyLedgerConfig cfgMary
+            in [ ("systemStartTime",          show (SL.sgSystemStart genesisShelley))
```
mrBliss, line 264: You are using `show` here, what does the output look like? Best to make sure it doesn't look like `SlotLength { getSlotLength :: ... }` or something, because that makes parsing brittler.

denisshevchenko:
@mrBliss The error is this one:

```
src/Cardano/Node/Configuration/Logging.hs:249:46: error:
    • Couldn't match type ‘Cardano.Ledger.Crypto.HASH c’
                     with ‘Cardano.Crypto.Hash.Blake2b.Blake2b_256’
        arising from a use of ‘getSystemStart’
```

mrBliss:
> IMPORANT: Now it doesn't compile, I need help with Byron start time.
Best to make this a draft. That was an option when opening the PR, but you can still do it now, there's a link below the list of reviewers.

mrBliss:
> @mrBliss The error is this one:
> 
> ```
> src/Cardano/Node/Configuration/Logging.hs:249:46: error:
>     • Couldn't match type ‘Cardano.Ledger.Crypto.HASH c’
>                      with ‘Cardano.Crypto.Hash.Blake2b.Blake2b_256’
>         arising from a use of ‘getSystemStart’
> ```

Try adding the `CardanoHardForkConstraints c` constraint. You can import it from `Ouroboros.Consensus.Cardano.CanHardFork`.

cardano-node/src/Cardano/Node/Run.hs
denisshevchenko: Fixed. It will be removed in a separate PR.

cardano-node/src/Cardano/Node/Configuration/Logging.hs
denisshevchenko: @mrBliss Now I see another error:

```
      Expected type: Consensus.TopLevelConfig (Consensus.CardanoBlock c)
        Actual type: Consensus.TopLevelConfig blk
```

denisshevchenko:
@mrBliss It is ready for review (no more compile errors anymore).

deepfire:
@denisshevchenko, can we also re-use `nodeBasicInfo` during startup, instead of triggering individual traces separately?

deepfire (changes requested):
@denisshevchenko, can we also re-use nodeBasicInfo during startup, instead of triggering individual traces (for commit/release/version and other similar startup-only parameters traces) separately?

deepfire (commented):
@denisshevchenko, can we also re-use `nodeBasicInfo` during startup, instead of triggering individual traces (for commit/release/version and other similar startup-only parameters traces) separately?

deepfire (approved):
LGTM now, thanks Denis!

cardano-node/src/Cardano/Node/Configuration/Logging.hs
```diff
@@ -211,3 +240,51 @@ createLoggingLayer ver nodeConfig' = do
...
+       , ("slotLengthByron",  show (WCT.getSlotLength . fromByronSlotLength $ genesisSlotLength genesis))
+       , ("epochLengthByron", show (unEpochSize . fromByronEpochSlots $ Gen.configEpochSlots genesis))
+       ]
+  getGenesisValues era config =
+    let genesis = shelleyLedgerGenesis $ shelleyLedgerConfig config
+    in [ ("systemStartTime",          show (SL.sgSystemStart genesis))
+       , ("slotLength" <> era,        show (WCT.getSlotLength . WCT.mkSlotLength $ SL.sgSlotLength genesis))
+       , ("epochLength" <> era,       show (unEpochSize . SL.sgEpochLength $ genesis))
+       , ("slotsPerKESPeriod" <> era, show (SL.sgSlotsPerKESPeriod genesis))
+       ]
```
intricate, line 290: Is it appropriate to just use `show` on these values? What does the output look like?

I'd much prefer `render*` functions that ensure these values are appropriately formatted and rendered to `Text`.

intricate (approved):
If the text representation of the values in `nodeBasicInfo` looks okay, then we can merge this.

cardano-node/src/Cardano/Node/Configuration/Logging.hs
denisshevchenko: > If the text representation of the values in nodeBasicInfo looks okay, then we can merge this.

@intricate Yes, the representation of these values is okay: it's the simplest possible. :-) For example, `slotLength` will be rendered not as:

```
SlotLength { getSlotLength = 1 }
```

but as:

```
1
```

PR #2052, opened 2020-11-03
CAD-1920 staked genesis:  bulk pools support

deepfire:
1. Generation of staked geneses -- geneses including pools that are active immediately since blockchain start.
1. Extended with generation of bulk pool credentials files.

### DIFF
--- a/cardano-api/src/Cardano/Api/Typed.hs
+++ b/cardano-api/src/Cardano/Api/Typed.hs
@@ -426,6 +426,8 @@ module Cardano.Api.Typed (
     toByronProtocolMagicId,
     toByronRequiresNetworkMagic,
     toShelleyNetwork,
+    toShelleyPoolParams,
+    toShelleyStakeAddr,
     toNetworkMagic,
 
     Shelley.Addr(..),
--- a/cardano-cli/cardano-cli.cabal
+++ b/cardano-cli/cardano-cli.cabal
@@ -68,6 +68,7 @@ library
                      , aeson-pretty
                      , attoparsec
                      , bech32 >= 1.1.0
+                     , binary
                      , bytestring
                      , base16-bytestring
                      , base58-bytestring
@@ -87,6 +88,7 @@ library
                      , contra-tracer
                      , cborg >= 0.2.4 && < 0.3
                      , containers
+                     , cryptonite
                      , directory
                      , filepath
                      , formatting
@@ -103,10 +105,12 @@ library
                      , ouroboros-consensus-cardano
                      , ouroboros-consensus-shelley
                      , ouroboros-network
+                     , primitive
                      , process
                      , scientific
                      , shelley-spec-ledger
                      , small-steps
+                     , split
                      , text
                      , time
                      , transformers
--- a/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
+++ b/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
@@ -782,6 +782,11 @@ pGenesisCmd =
              Opt.progDesc ("Create a Shelley genesis file from a genesis "
                         ++ "template and genesis/delegation/spending keys."))
 
+      , Opt.command "create-staked"
+          (Opt.info pGenesisCreateStaked $
+             Opt.progDesc ("Create a staked Shelley genesis file from a genesis "
+                        ++ "template and genesis/delegation/spending keys."))
+
       , Opt.command "hash"
           (Opt.info pGenesisHash $
              Opt.progDesc "Compute the hash of a genesis file")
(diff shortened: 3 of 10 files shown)


cardano-cli/cardano-cli.cabal
```diff
@@ -103,9 +103,12 @@ library
                      , ouroboros-consensus-shelley
                      , ouroboros-network
                      , process
+                     , QuickCheck
                      , scientific
                      , shelley-spec-ledger
+                     , shelley-spec-ledger-test
```
mrBliss, line 109: We don't want to depend on testing-related libraries in library code.

deepfire: These dependencies are only due for the `Arbitrary` instance for `Shelley.PaymentCredential StandardShelley`.
..which is basically a `data` wrapper around a `Hash`.

So I perhaps should look harder.

deepfire: To clarify -- in this case I don't care about the credential this `Hash` is supposed to represent -- I only need a hash, to fill in this field, to make a completely fake entry.

mrBliss: An example from consensus you can base yourself on:
```haskell
import           Cardano.Crypto.Hash (HashAlgorithm)
import qualified Cardano.Crypto.Hash as Hash

import           Cardano.Ledger.Crypto (ADDRHASH, Crypto)

import qualified Shelley.Spec.Ledger.API as SL
import qualified Shelley.Spec.Ledger.Keys as SL (hashWithSerialiser)

mkDummyHash :: forall h a. HashAlgorithm h => Proxy h -> Int -> Hash.Hash h a
mkDummyHash _ = coerce . SL.hashWithSerialiser @h toCBOR

mkKeyHash :: forall c discriminator. Crypto c => Int -> SL.KeyHash discriminator c
mkKeyHash = SL.KeyHash . mkDummyHash (Proxy @(ADDRHASH c))
```
(The list of imports is probably incomplete)

deepfire: Thank you a lot, Thomas!

deepfire: Fixed, thank you, Thomas!

cardano-api/src/Cardano/Api/Typed.hs
```diff
@@ -326,6 +326,7 @@ module Cardano.Api.Typed (
     toByronTxOut,
     toShelleyNetwork,
     toShelleyPoolParams,
+    toShelleyStakeAddr,
```
erikd, line 329: Commit message could be improved. I have a good idea what all the words in the commit message mean, but I am still unsure what the commit actually does.

erikd (dismissed):

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
```diff
@@ -339,8 +672,8 @@ getCurrentTimePlus30 =
     plus30sec = addUTCTime (30 :: NominalDiffTime)
 
 
-readShelleyGenesis
-  :: FilePath
+readShelleyGenesis ::
+     FilePath
```
Jimbo4350, line 676: Why the move?

deepfire: This way you can grep for the definition -- by simply adding the `::`.

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -863,18 +884,73 @@ pGenesisCmd =
...
+          )
+
+    pInitialSupplyDelegated :: Parser Lovelace
+    pInitialSupplyDelegated =
+      fmap (Lovelace . fromMaybe 0) $ Opt.optional $
+        Opt.option Opt.auto
+          (  Opt.long "supply-delegated"
+          <> Opt.metavar "LOVELACE"
+          <> Opt.help "The initial coin supply in Lovelace which will be evenly distributed across initial, delegating stake holders."
+          <> Opt.value (fromIntegral (0 :: Integer))
```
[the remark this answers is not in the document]
deepfire: Right, thank you @Jimbo4350!

cardano-api/src/Cardano/Api/Typed.hs
deepfire: Done.

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
Jimbo4350: Please revert this (and other instances in this PR where you have done this). All other functions in this repo do not follow this style.

Jimbo4350: Will green light once this is done.

deepfire: @Jimbo4350, done.

Jimbo4350 (approved):

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -863,18 +884,73 @@ pGenesisCmd =
...
+          )
+
+    pInitialSupplyDelegated :: Parser Lovelace
+    pInitialSupplyDelegated =
+      fmap (Lovelace . fromMaybe 0) $ Opt.optional $
+        Opt.option Opt.auto
+          (  Opt.long "supply-delegated"
+          <> Opt.metavar "LOVELACE"
+          <> Opt.help "The initial coin supply in Lovelace which will be evenly distributed across initial, delegating stake holders."
+          <> Opt.value 0
```
Jimbo4350, line 935: @deepfire Same thing here re: `Lovelace . fromMaybe 0` and `Opt.value 0`

PR #2054, opened 2020-11-03
Add multi-asset tx out value type

intricate:


cardano-api/src/Cardano/Api/Typed.hs
```diff
@@ -946,13 +985,130 @@ newtype Lovelace = Lovelace Integer
...
+  -> TxOutValue era
+pattern SimpleTxOutValue l <-
+  ( (\txOut ->
+      case txOut of
+        ByronTxOutValue l -> Just (Lovelace $ Byron.lovelaceToInteger l)
+        ShelleyTxOutValue (Shelley.Coin l) -> Just (Lovelace l)
+        AllegraTxOutValue (Shelley.Coin l) -> Just (Lovelace l)
+        _ -> Nothing
+    ) -> Just l
+  )
```
intricate, line 1069: Unidirectional because constructing a simple lovelace-only tx out value can fail in the Byron case.

intricate:
Subsumed by https://github.com/input-output-hk/cardano-node/pull/2083.

PR #2057, opened 2020-11-03
new CLI command to query the shelley protocol state

JaredCorduan:
I am adding a new CLI shelley query command to make use of the mini-protocol DebugChainDepState recently added to the ouroboros-network repository (see https://github.com/input-output-hk/ouroboros-network/pull/2694).

This command returns the operational certificate counters and the four nonces of interest to the protocol (epoch, last applied block, evolving, and candidate).

I'm open to different names if there is something better than "protocol state".


intricate (approved):
LGTM :100:

PR #2066, opened 2020-11-06
Use generated ports in Byron testnet

newhoggy:
The Byron testnet was previously using static port numbers.  Make it use generated port numbers to bring it in line with other testnets, which will help avoid port collision.

### DIFF
--- a/cardano-node-chairman/src/Testnet/Byron.hs
+++ b/cardano-node-chairman/src/Testnet/Byron.hs
@@ -17,7 +17,6 @@ import           Data.Maybe
 import           Data.Ord
 import           Data.Semigroup
 import           Data.String
-import           Data.Time.Clock
 import           GHC.Num
 import           Hedgehog.Extras.Stock.Aeson (rewriteObject)
 import           Hedgehog.Extras.Stock.IO.Network.Sprocket (Sprocket (..))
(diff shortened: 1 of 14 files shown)


cardano-node-chairman/src/Testnet/Byron.hs
```diff
@@ -112,7 +120,26 @@ testnet H.Conf {..} = do
     H.createDirectoryIfMissing dbDir
     H.createDirectoryIfMissing $ tempBaseAbsPath </> "" <> socketDir
 
-    H.copyFile (baseConfig </> "topology-node-" <> si <> ".json") (tempAbsPath </> "topology-node-" <> si <> ".json")
+    let otherPorts = let (as, bs) = L.splitAt i allPorts in as <> L.drop 1 bs
+
+    H.lbsWriteFile (tempAbsPath </> "topology-node-" <> si <> ".json") $ J.encode $
+      J.object
```
erikd, line 126: Probably should pull that out as a separate top level or `where` clause value rather than defining it inline.

erikd (dismissed):
See comment.

cardano-node-chairman/src/Testnet/Byron.hs
newhoggy: Move to top-level `dropNth` function.

newhoggy:
This was merged as part of https://github.com/input-output-hk/cardano-node/pull/2048

PR #2068, opened 2020-11-06
CAD-1877 node:  multi-pool support in non-Cardano mode

deepfire:
Accept bulk credential files.


cardano-node/src/Cardano/Node/Parsers.hs
```diff
@@ -43,6 +43,7 @@ nodeRunParser = do
   shelleyKESFile  <- optional parseKesKeyFilePath
   shelleyVRFFile  <- optional parseVrfKeyFilePath
   shelleyCertFile <- optional parseOperationalCertFilePath
+  shelleyBulkCredsFile <- optional parseBulkCredsFilePath
```
erikd, line 46: We need a better commit message on this. I am **TOTALLY** uninterested in the files that conflicted in some merge.

The commit message **should** contain information about what this commit actually does and referring to a Jira ticket that people outside IOHK cannot view.

erikd (changes requested):

cardano-node/src/Cardano/Node/Protocol/Shelley.hs
```diff
@@ -174,14 +231,15 @@ readLeaderCredentials (Just ProtocolFilepaths {
...
-
--- But not OK to supply some of the files without the others.
-readLeaderCredentials (Just ProtocolFilepaths {shelleyCertFile = Nothing}) =
-    throwError OCertNotSpecified
-readLeaderCredentials (Just ProtocolFilepaths {shelleyVRFFile = Nothing}) =
-    throwError VRFKeyNotSpecified
-readLeaderCredentials (Just ProtocolFilepaths {shelleyKESFile = Nothing}) =
-    throwError KESKeyNotSpecified
+  where
+    parseEnvelope ::
```
Jimbo4350, line 235: You shouldn't need `parseEnvelope` as we have `readFileTextEnvelope` in `cardano-api` already. You can lift that into `ExceptT`.

Jimbo4350 (changes requested):

cardano-node/src/Cardano/Node/Protocol/Shelley.hs
deepfire: there isn't quite a match there:

1. The code depends on `readEnvelope` not decoding the envelope which `readFileTextEnvelope` does:
```
   readEnvelope :: FilePath -> ExceptT ShelleyProtocolInstantiationError IO
                                       (TextEnvelope, FilePath)
   readEnvelope fp = do
     content <- handleIOExceptT (CredentialsReadError fp) $
                  BS.readFile fp
     firstExceptT (EnvelopeParseError fp) $ hoistEither $
       (, fp) <$> Aeson.eitherDecodeStrict' content
```
2. `readFileTextEnvelope` does `runExceptT`, pushes the computation outside of the `ExceptT` monad, which means it'll have to be rewrapped back into `ExceptT` by the code.. which is suboptimal.

erikd: @deepfire Please provide a reasonable alternative and let @Jimbo4350 chose between them.

erikd (changes requested):

deepfire:
Addressing this suggestion:

> You shouldn't need `readEnvelope` and `parseEnvelope` as we have `readFileTextEnvelope` in `cardano-api` already. You can lift that into ExceptT.

> @deepfire Please provide a reasonable alternative and let @Jimbo4350 chose between them.

> I would also prefer sticking with `readFileTextEnvelope` for reading the credential files provided via the CLI.

`readFileTextEnvelope` reads the file and expects it to conform to a particular format.

Have you considered what bulk credentials files are?  They are a list of content triplets of what the usual credentials files are.

That means that `readFileTextEnvelope` cannot be used -- at all.

So there is no reasonable alternative.

cc @intricate @Jimbo4350 @erikd

cardano-node/src/Cardano/Node/Protocol/Shelley.hs
```diff
@@ -137,34 +137,91 @@ readGenesis (GenesisFile file) mbExpectedGenesisHash = do
...
+     content <- handleIOExceptT (CredentialsReadError fp) $
+                  BS.readFile fp
+     envelopes <- firstExceptT (EnvelopeParseError fp) $ hoistEither $
+                    Aeson.eitherDecodeStrict' content
+     pure $ uncurry mkCredentials <$> zip [0..] envelopes
+    where
+      mkCredentials :: Int -> (TextEnvelope, TextEnvelope, TextEnvelope)
+                    -> ShelleyCredentials
+      mkCredentials ix (teCert, teVrf, teKes) =
+       let loc = fp <> ":" <> show ix
```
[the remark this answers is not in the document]
deepfire: @intricate, replaced `:` with a dot.

cardano-node/src/Cardano/Node/Protocol/Shelley.hs
deepfire: `readFileTextEnvelope` reads the file and expects it to conform to a particular format.

Bulk credentials files a list of content triplets corresponding to what the usual credentials files contain, individually.

That means that `readFileTextEnvelope` cannot be used -- at all.

So there is no reasonable alternative.

cc @intricate @Jimbo4350 @erikd

deepfire:
We've agreed with @intricate that in the light of upcoming API changes it would make sense to separate processing of bulk and singleton credentials.

And so that change was made.

intricate (approved):
LGTM.

erikd (approved):

deepfire:
bors r+

To clarify -- the change that was decided on with @intricate was different from what was initially proposed by other reviewers.

PR #2038, opened 2020-10-30
Start using the real Allegra and Mary eras

mrBliss:
Up until now, the Allegra and Mary eras part of `CardanoBlock` were simply
aliases of the Shelley era.

We now switch them over to the real Allegra and Mary eras. This means that some
things will be different for the eras, e.g., `TxBody`, `Script`, `Value`,
`UtxoPredicateFailure`, etc.

PR #2073, opened 2020-11-09
Remove the HardForkNotBeforeEpoch parameter

jutaro:
This propagates https://github.com/input-output-hk/ouroboros-network/pull/2736

The parameter was introduced because we expected it would improve the
performance by not having to recalculate some things until the epoch of the
transition has been reached. However, disabling it had no noticeable impact in
recent microbenchmarks.

This field is also needed for each new era, adding to the boilerplate. Moreover,
setting it incorrectly can cause *real* problems that are not always easy to
debug.

So remove it altogether.

### DIFF
--- a/cardano-node/src/Cardano/Node/Types.hs
+++ b/cardano-node/src/Cardano/Node/Types.hs
@@ -310,18 +310,13 @@ data NodeByronProtocolConfiguration =
 data NodeHardForkProtocolConfiguration =
      NodeHardForkProtocolConfiguration {
 
-       -- | If we have knowledge about when the Shelley hard fork is then we
-       -- have an opportunity to optimise the bulk sync slightly.
-       --
-       npcShelleyHardForkNotBeforeEpoch :: Maybe EpochNo
-
        -- | For testing purposes we support specifying that the hard fork
        -- happens at an exact epoch number (ie the first epoch of the new era).
        --
        -- Obviously if this is used, all the nodes in the test cluster must be
        -- configured the same, or they will disagree.
        --
-     , npcTestShelleyHardForkAtEpoch :: Maybe EpochNo
+       npcTestShelleyHardForkAtEpoch :: Maybe EpochNo
 
        -- | For testing purposes we support specifying that the hard fork
        -- happens at a given major protocol version. For example this can be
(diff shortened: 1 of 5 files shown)


configuration/cardano/mainnet-config.json
```diff
@@ -11,7 +11,6 @@
   "RequiresNetworkMagic": "RequiresNoMagic",
   "ShelleyGenesisFile": "mainnet-shelley-genesis.json",
   "ShelleyGenesisHash": "1a3be38bcbb7911969283716ad7aa550250226b76a61fc51cc9a9a35d9276d81",
-  "ShelleyHardForkNotBeforeEpoch": 208,
```
mrBliss, line 14: :+1:

cardano-node/src/Cardano/Node/Types.hs
```diff
@@ -310,18 +310,13 @@ data NodeByronProtocolConfiguration =
...
-       --
-       npcShelleyHardForkNotBeforeEpoch :: Maybe EpochNo
-
        -- | For testing purposes we support specifying that the hard fork
        -- happens at an exact epoch number (ie the first epoch of the new era).
        --
        -- Obviously if this is used, all the nodes in the test cluster must be
        -- configured the same, or they will disagree.
        --
-     , npcTestShelleyHardForkAtEpoch :: Maybe EpochNo
```
mrBliss, line 324: Two more spaces

mrBliss (commented):
The build is failing on CI

intricate:
Related to https://github.com/input-output-hk/ouroboros-network/pull/2736.

mrBliss:
Status: the corresponding [change in `ouroboros-network`](https://github.com/input-output-hk/ouroboros-network/pull/2736) was merged after https://github.com/input-output-hk/ouroboros-network/pull/2679, which means that the latter change first has to be propagated. This is being done in https://github.com/input-output-hk/cardano-node/pull/2038. When that PR has been merged, we can update this one.

erikd:
#2038 has been merged. Can someone please move this through? I need this updated to make progress on `db-sync`.

mrBliss:
> #2038 has been merged. Can someone please move this through? I need this updated to make progress on `db-sync`.

We'll work on this today. However, I don't think this PR is blocking you, as this propagates a change in `ouroboros-network` that happened *after* the one propagated in #2038. So you can just use the commit of `ouroboros-network` that the current master refers to.

mrBliss (approved):

intricate (approved):

PR #2080, opened 2020-11-12
Adapt to the reorganised Node.run arguments in consensus

mrBliss:
In https://github.com/input-output-hk/ouroboros-network/pull/2720 we reorganised
the arguments to `Node.run` so that the consensus ThreadNet tests can use the
same `Node.run` function as the real node while using a custom network layer.

In this commit we propagate that change.

For people that want to override some arguments that now seem to have
disappeared: they have been moved to the `LowLevelRunNodeArgs` record in
consensus. See the `stdLowLevelRunNodeArgsIO` and `Node.runWith` functions for
constructing and running with custom `LowLevelRunNodeArgs`.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -159,8 +159,8 @@ source-repository-package
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/ouroboros-network
-  tag: da2130c6b19d38cb9ec519ed4fb2d644b35e55cb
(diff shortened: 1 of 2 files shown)


mrBliss:
Draft until #2720 has been merged.

intricate (approved):
LGTM.

newhoggy (approved):

PR #2086, opened 2020-11-13
Slightly more lenient chairman blocks assertion.

newhoggy:
Reducing from `55` to `53` blocks for Byron testnet.

This is to avoid this failure in CI: https://github.com/input-output-hk/cardano-node/runs/1393451411?check_suite_focus=true


erikd (approved):

newhoggy:
bors merge

PR #2101, opened 2020-11-18
Improved tracing option parsing

karknu:
TraceSelection contains 30+ bools for enabling/disabling different
tracers. This turns a bug in the order of those fields in
traceConfigParser from a run time error into a compile time error.
Based on an idea from @coot .

### DIFF
--- a/cardano-node/src/Cardano/Tracing/Config.hs
+++ b/cardano-node/src/Cardano/Tracing/Config.hs
@@ -1,15 +1,22 @@
+{-# LANGUAGE DataKinds         #-}
+{-# LANGUAGE KindSignatures      #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications    #-}
+
 
 module Cardano.Tracing.Config
   ( TraceOptions (..)
   , TraceSelection (..)
   , traceConfigParser
+  , OnOff (..)
   ) where
 
 import           Cardano.Prelude
 
 import           Data.Aeson
 import           Data.Aeson.Types (Parser)
+import           Data.Text (pack)
 
 import           Cardano.BM.Tracing (TracingVerbosity (..))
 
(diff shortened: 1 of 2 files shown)


cardano-node/src/Cardano/Tracing/Config.hs
```diff
@@ -21,78 +28,185 @@ data TraceOptions
   | TracingOn TraceSelection
   deriving (Eq, Show)
 
+type TraceAcceptPolicy = ("TraceAcceptPolicy" :: Symbol)
```
coot, line 31: We might not need brackets here.

A vim macro to get rid of it :D
```
qqdi(va("-pq
```

karknu: We don't need them for 8.10.2. I just added them so that it compiles with 8.6.3.

coot: Hmm, I didn't noticed that `GHC` parsers changed.

cardano-node/src/Cardano/Tracing/Config.hs
```diff
@@ -21,78 +28,185 @@ data TraceOptions
...
+type TraceMempool = ("TraceMempool" :: Symbol)
+type TraceMux = ("TraceMux" :: Symbol)
+type TraceTxInbound = ("TraceTxInbound" :: Symbol)
+type TraceTxOutbound = ("TraceTxOutbound" :: Symbol)
+type TraceTxSubmissionProtocol = ("TraceTxSubmissionProtocol" :: Symbol)
+
+newtype OnOff (name :: Symbol) = OnOff { isOn :: Bool } deriving (Eq, Show)
+
+class TracerName tm where
+    getName :: tm -> Text
```
coot, line 66: @CodiePP you might now better if having such a class is useful for other context or not, if just for the `OnOff` type I would prefer just a function.

coot (approved):
LGTM

cardano-node/src/Cardano/Tracing/Config.hs
```diff
@@ -21,78 +28,182 @@ data TraceOptions
...
+type TraceTxOutbound = ("TraceTxOutbound" :: Symbol)
+type TraceTxSubmissionProtocol = ("TraceTxSubmissionProtocol" :: Symbol)
+
+newtype OnOff (name :: Symbol) = OnOff { isOn :: Bool } deriving (Eq, Show)
+
+instance FromJSON (OnOff a) where
+    parseJSON (Data.Aeson.Bool b)= return $ OnOff b
+    parseJSON _ = mzero
+
+getName :: forall name. KnownSymbol name => OnOff name -> Text
```
Jimbo4350, line 69: :100:

Jimbo4350 (approved):
:100:

intricate (approved):

CodiePP (approved):
I like it! :meetoooo:

PR #2114, opened 2020-11-23
Update dependency on ouroboros-network

mrBliss:
Notable changes:

* Share a single set of credentials for all Shelley-based eras:
  https://github.com/input-output-hk/ouroboros-network/pull/2753
* Add a pure Mary mode
  https://github.com/input-output-hk/ouroboros-network/pull/2754

PR #2116, opened 2020-11-24
Use common shelley credentials

newhoggy:
See https://github.com/input-output-hk/ouroboros-network/commit/2194d81d745c75dda57063ba89e56e1ec6b37753


mrBliss:
Let's use #2114 instead of duplicating it.

PR #2117, opened 2020-11-24
Adapt to single forging thread for all eras

newhoggy:


mrBliss:
(Yet another PR?) Let's use #2114 instead of duplicating it.

PR #2122, opened 2020-11-25
Flag to launch non-Byron-only testnet with custom activeSlotCoeff

newhoggy:


erikd (approved):
Code is fine, but the message in the PR:
```
$ for x in {1..1}; do (export KEEP_WORKSPACE=1; cd carda .......
```
is probably not something we want in git history forever.

newhoggy:
Change the text to use a slightly better command.  Note this just in the PR description and not in the git history.

erikd:
Yes, but if you use github to merge it, it will be in the git history.

newhoggy:
That's awkward.  Moved it into a comment:

The following shows that the genesis files get updated to have the provided `activeSlotCoeff` value:

```
$ (cd cardano-node-chairman; export KEEP_WORKSPACE=1; cabal run -- cardano-testnet shelley --active-slots-coeff 0.25)
$ for x in $(find /private/var/folders/zh/ln41q4zs52x2fd61rxccmq640000gn/T/chairman/test-2dbb537814cf6ce0/ -name 'g*.json'); do echo "== $x =="; cat $x | jq . | grep eff; done
== /private/var/folders/zh/ln41q4zs52x2fd61rxccmq640000gn/T/chairman/test-2dbb537814cf6ce0//genesis.json ==
  "activeSlotsCoeff": 0.25,
== /private/var/folders/zh/ln41q4zs52x2fd61rxccmq640000gn/T/chairman/test-2dbb537814cf6ce0//genesis.spec.json ==
  "activeSlotsCoeff": 0.25,
```

newhoggy:
bors merge

PR #2135, opened 2020-11-27
Better thread handling when starting main node action

coot:
`handleSimpleNode` is a non-terminating function: it can only error.  We
need to use `finally` clause to ensure that an action runs when it
errors.  It is also more pragmatic to use `withAsync` to run peer list
logging action, this way it will be cancelled once `handleSimjpleNode`
errors.  Since we don't care if that action fails we do not await on its
termination.

### DIFF
--- a/cardano-node/src/Cardano/Node/Run.hs
+++ b/cardano-node/src/Cardano/Node/Run.hs
@@ -127,11 +127,14 @@ runNode cmdPc = do
 
     tracers <- mkTracers (ncTraceConfig nc) trace nodeKernelData
 
-    peersThread <- Async.async $ handlePeersListSimple trace nodeKernelData
-    handleSimpleNode p trace tracers nc (setNodeKernel nodeKernelData)
-    Async.uninterruptibleCancel peersThread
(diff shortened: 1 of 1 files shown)


dcoutts (approved):
LGTM

coot:
bors merge

deepfire:
I don't think that `handleSimpleNode` is necessarily a non-terminating function:

- `handleSimpleNode`
- `withShutdownHandling (Node.run)`
- `withShutdownHandler`
- which is then either:
  - `Async.race_` `Node.run` 
  - `Node.run`

So when exit is requestible, it should be terminating.

deepfire:
Basically, if `withShutdownHandling` still works as intended (which it should), the terminability of `handleSimpleNode` should depend on configuration -- whether or not one of the two ways (intra/extra-process) of termination signalling has been made active.

coot:
The point is that `Node.run` can only error (it never terminates), if that happens clean shutdown of logging would not be executed.

deepfire:
That is a larger, and yes -- more important point, indeed!

PR #2142, opened 2020-11-30
Upgrade to cabal 3.4

newhoggy:


newhoggy:
Looks like it doesn't work yet for Windows.

PR #2164, opened 2020-12-03
CAD-2359 cli:  construct a full cert for annotation in 'byron delegation check-delegation'

deepfire:


cardano-cli/src/Cardano/CLI/Byron/Delegation.hs
```diff
@@ -103,10 +103,22 @@ checkDlgCert cert magic issuerVK' delegateVK' =
...
+      , Dlg.signature  = Dlg.signature cert
+      , Dlg.annotation = ()
+      }
+
+    cert'' :: Dlg.ACertificate ByteString
+    cert'' =
+      cert'
+      { Dlg.aEpoch = Annotated epoch (serialize' epoch)
+      , Dlg.annotation = serialize' $ cert'
+      }
```
mrBliss, line 121: I would have gone for something like (didn't type check it):
```haskell
cert' =
  let unannotated = cert { Dlg.aEpoch = Annotated epoch (serialize' epoch) }
  in unannotated { Dlg.annotation = serialize' unannotated }
```
(Maybe it needs a `unannotated :: Dlg.Certificate` signature?)

mrBliss (approved):
Thanks!

PR #2178, opened 2020-12-05
Network tracers

coot:
This PR improves network tracers; originally done in `coot/p2p-integration` branch,
but this part is indepentend of p2p changes in `ouroboros-network`.  This PR consists of two patches:

- Improved mini-protocol tracers
  Which includes agency together with the mini-protocol message tag.

- networking trace using show instances rather than json values

### DIFF
--- a/cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
+++ b/cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
@@ -257,7 +257,7 @@ instance HasSeverityAnnotation (Identity (SubscriptionTrace LocalAddress)) where
 instance Transformable Text IO (Identity (SubscriptionTrace LocalAddress)) where
   trTransformer = trStructuredText
 instance HasTextFormatter (Identity (SubscriptionTrace LocalAddress)) where
-  formatText _ = pack . show . toList
+  formatText a _ = pack (show a)
 
 
 instance ToObject (Identity (SubscriptionTrace LocalAddress)) where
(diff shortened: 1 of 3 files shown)


cardano-api/src/Cardano/Api/Typed.hs
```diff
@@ -6,6 +6,7 @@
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE QuantifiedConstraints #-}
```
intricate, line 9: Is this necessary?

coot: It might be not, since there's `ShowQuery` wrapper.

coot: Fixed.

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
```diff
@@ -84,7 +86,13 @@ instance HasSeverityAnnotation NtN.AcceptConnectionsPolicyTrace where
...
 instance HasPrivacyAnnotation (TraceFetchClientState header)
 instance HasSeverityAnnotation (TraceFetchClientState header) where
-  getSeverityAnnotation _ = Info
+  getSeverityAnnotation BlockFetch.AddedFetchRequest {} = Info
+  getSeverityAnnotation BlockFetch.AcknowledgedFetchRequest {} = Info
+  getSeverityAnnotation BlockFetch.StartedFetchBatch {} = Info
+  getSeverityAnnotation BlockFetch.CompletedBlockFetch {} = Info
+  getSeverityAnnotation BlockFetch.CompletedFetchBatch {} = Info
+  getSeverityAnnotation BlockFetch.RejectedFetchBatch {} = Info
+  getSeverityAnnotation BlockFetch.ClientTerminating {} = Notice
```
deepfire, line 96: @coot, do you think it can be:
```suggestion
  getSeverityAnnotation BlockFetch.ClientTerminating {} = Notice
  getSeverityAnnotation _ = Info
```
?

deepfire (approved):
LGTM with one small nit.

Thank you, @coot!

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
coot: We will have something like 20 hot peers, so these messages will rather be sparse. I thought that initially its better to start with `Notice` and we can move to `Info` later. Are you  ok with it @deepfire ?

coot:
bors merge

newhoggy (approved):

coot:
bors merge

dcoutts (commented):
LGTM!

PR #2213, opened 2020-12-14
Continue on error for chairman test artifact upload

newhoggy:


newhoggy:
This is to avoid the following being marked as an error:

image

The upload is for diagnosing chairman test failures and is "best effort" anyway, and most of the time we don't care if logs and config are uploaded successfully.

dcoutts (approved):

newhoggy:
bors merge

PR #2214, opened 2020-12-14
Select optimal cabal version

newhoggy:
This reduces the build time for Linux in Github Actions from 28 minutes to 16 minutes.

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -27,7 +27,7 @@ jobs:
         echo "CABAL_BUILDDIR=$CABAL_BUILDDIR" >> $GITHUB_ENV
 
     - name: Set cache version
-      run: echo "CACHE_VERSION=9w76Z3Q" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/workflows/haskell.yml
```diff
@@ -61,11 +61,18 @@ jobs:
           flags: -external-libsodium-vrf
         EOF
 
+    - name: Select optimal cabal version
+      run: |
+        case "$OS" in
+          Windows_NT)   echo "CABAL_VERSION=3.2.0.0"      >> $GITHUB_ENV;;
+          *)            echo "CABAL_VERSION=3.4.0.0-rc4"  >> $GITHUB_ENV;;
```
erikd, line 68: `>>` (append) instead of `>`?

newhoggy: Yep.  As per: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions

PR #2160, opened 2020-12-03
MetaData -> Metadata

rvl:
Makes the spelling consistent. See also: input-output-hk/cardano-ledger-specs#2036 and input-output-hk/ouroboros-network#2786.

```
git ls-files -z | xargs -0 sed -i -e 's/MetaData/Metadata/g' -e 's/MetaDatum/Metadatum/g'
```

PR #2216, opened 2020-12-14
Update dependencies

mrBliss:
* `MetaData` -> `Metadata`

* rename auxiliary data
  https://github.com/input-output-hk/cardano-ledger-specs/pull/2052
  + Rename the type family `Core.Metadata` to `Core.AuxiliaryData`.
  + Rename the Allegra & Mary `Metadata` to `AuxiliaryData`.
  + Rename ValidityInterval accessors:
    `validFrom` -> `invalidBefore`
    `validTo` -> `invalidHereafter`

* Parameterise era-independent types over the crypto instead of era
  https://github.com/input-output-hk/cardano-ledger-specs/pull/2038

  A bunch of types were parameterised over the era, while they were really
  era-independent. It is confusing to parameterise something over the era while
  it really is era-independent. Moreover, it requires much more needless
  conversions/coercions.

  (Nearly?) all of them are now parameterised by the crypto instead of the era.
  The Cardano API has been updated to reflect this. In most places, the
  `ledgerera` parameter has been replaced by `StandardCrypto`.

### DIFF
--- a/cardano-api/src/Cardano/Api/Address.hs
+++ b/cardano-api/src/Cardano/Api/Address.hs
@@ -1,8 +1,9 @@
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE PatternSynonyms #-}
-{-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 
 -- | Cardano addresses: payment and stake addresses.
@@ -64,7 +65,6 @@ module Cardano.Api.Address (
 
 import           Prelude
 
-import           Data.Coerce
 import           Data.Text (Text)
 import qualified Data.Text.Encoding as Text
 import qualified Data.ByteString.Base58 as Base58
@@ -73,8 +73,6 @@ import           Control.Applicative
 
 import qualified Cardano.Chain.Common as Byron
 
-import qualified Cardano.Ledger.Era as Ledger
-import           Ouroboros.Consensus.Shelley.Eras (StandardShelley)
 import           Ouroboros.Consensus.Shelley.Protocol.Crypto (StandardCrypto)
 
 import qualified Shelley.Spec.Ledger.Address as Shelley
@@ -165,8 +163,8 @@ data Address addrtype where
      --
      ShelleyAddress
        :: Shelley.Network
-       -> Shelley.PaymentCredential StandardShelley
-       -> Shelley.StakeReference    StandardShelley
+       -> Shelley.PaymentCredential StandardCrypto
+       -> Shelley.StakeReference    StandardCrypto
        -> Address ShelleyAddr
        -- Note that the two ledger credential types here are parametrised by
        -- the era, but in fact this is a phantom type parameter and they are
@@ -197,7 +195,7 @@ instance SerialiseAsRawBytes (Address ByronAddr) where
       $ addr
 
     deserialiseFromRawBytes (AsAddress AsByronAddr) bs =
-        case Shelley.deserialiseAddr bs :: Maybe (Shelley.Addr StandardShelley) of
+        case Shelley.deserialiseAddr bs :: Maybe (Shelley.Addr StandardCrypto) of
           Nothing             -> Nothing
           Just Shelley.Addr{} -> Nothing
           Just (Shelley.AddrBootstrap (Shelley.BootstrapAddress addr)) ->
@@ -381,14 +379,14 @@ anyAddressInShelleyBasedEra (AddressByron   addr) = byronAddressInEra addr
 anyAddressInShelleyBasedEra (AddressShelley addr) = shelleyAddressInEra addr
 
 
-anyAddressInEra :: IsCardanoEra era
+anyAddressInEra :: forall era. IsCardanoEra era
                 => AddressAny
                 -> Maybe (AddressInEra era)
 anyAddressInEra (AddressByron addr) =
     Just (AddressInEra ByronAddressInAnyEra addr)
 
 anyAddressInEra (AddressShelley addr) =
-    case cardanoEraStyle cardanoEra of
+    case cardanoEraStyle (cardanoEra @era) of
       LegacyByronEra      -> Nothing
       ShelleyBasedEra era -> Just (AddressInEra (ShelleyAddressInEra era) addr)
 
@@ -420,7 +418,7 @@ data StakeAddress where
 
      StakeAddress
        :: Shelley.Network
-       -> Shelley.StakeCredential StandardShelley
+       -> Shelley.StakeCredential StandardCrypto
        -> StakeAddress
   deriving (Eq, Ord, Show)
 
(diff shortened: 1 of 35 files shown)


mrBliss:
Typically, I would squash the two commits into one, as all changes are needed to get to a buildable state (otherwise git bisect is broken). But to make the review easier, I cherry-picked #2160 and made the remaining changes in another commit.

So if we want a buildable history, we should squash before merging.

cardano-api/src/Cardano/Api/Address.hs
```diff
@@ -381,14 +379,14 @@ anyAddressInShelleyBasedEra (AddressByron   addr) = byronAddressInEra addr
...
-anyAddressInEra :: IsCardanoEra era
+anyAddressInEra :: forall era. IsCardanoEra era
                 => AddressAny
                 -> Maybe (AddressInEra era)
 anyAddressInEra (AddressByron addr) =
     Just (AddressInEra ByronAddressInAnyEra addr)
 
 anyAddressInEra (AddressShelley addr) =
-    case cardanoEraStyle cardanoEra of
+    case cardanoEraStyle (cardanoEra @era) of
```
mrBliss, line 389: I needed this for some reason, it would be nice if we could avoid it.

cardano-api/src/Cardano/Api/Address.hs
```diff
@@ -532,58 +525,37 @@ toShelleyStakeReference  NoStakeAddress =
...
   StakeAddressByValue (fromShelleyStakeCredential stakecred)
 fromShelleyStakeReference (Shelley.StakeRefPtr ptr) =
   StakeAddressByPointer ptr
 fromShelleyStakeReference Shelley.StakeRefNull =
   NoStakeAddress
-
--- The era parameter in these types is a phantom type so it is safe to cast.
--- We choose to cast because we need to use an era-independent address
--- representation, but have to produce an era-dependent format used by the
--- Shelley ledger lib.
```
mrBliss, line 578: Fixed :slightly_smiling_face:

cardano-api/src/Cardano/Api/Eras.hs
```diff
@@ -239,7 +241,8 @@ deriving instance Show (ShelleyBasedEra era)
 -- of Shelley-based eras, but also non-uniform by making case distinctions on
 -- the 'ShelleyBasedEra' constructors.
 --
-class IsCardanoEra era => IsShelleyBasedEra era where
+class (IsCardanoEra era, Ledger.Crypto (ShelleyLedgerEra era) ~ StandardCrypto)
```
mrBliss, line 244: I needed this `Ledger.Crypto (ShelleyLedgerEra era) ~ StandardCrypto` constraint in `Cardano.CLI.Shelley.Run.Query`.

cardano-api/src/Cardano/Api/Tx.hs
```diff
@@ -513,7 +515,7 @@ makeSignedTransaction witnesses (ShelleyTxBody era txbody txmetadata) =
           txbody'
           (Shelley.WitnessSet
             (Set.fromList [ w | ShelleyKeyWitness _ w <- witnesses ])
-            (Map.fromList [ (Shelley.hashScript sw, sw)
+            (Map.fromList [ (Shelley.hashScript @ledgerera sw, sw)
```
mrBliss, line 518: `Shelley.hashScript` is ambiguous. I'm not a big fan of ambiguous types, but the ledger uses them.

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -719,12 +717,12 @@ data TxBody era where
        :: ShelleyBasedEra era
        -> Ledger.TxBody (ShelleyLedgerEra era)
 
-          -- The 'Ledger.Metadata' is really the /auxiliary/ data, it consists
-          -- of one or several things, depending on era:
+          -- The 'Ledger.AuxiliaryData' consists of one or several things,
+          -- depending on era:
```
mrBliss, line 721: :relieved:

cardano-cli/src/Cardano/CLI/Shelley/Orphans.hs
```diff
@@ -104,82 +106,64 @@ deriving newtype instance ToJSON BlockNo
...
-deriving instance ToJSON (Ledger.PState StandardShelley)
-deriving instance ToJSON (Ledger.StakeReference StandardShelley)
 deriving instance ToJSON (Ledger.UTxOState StandardShelley)
 
-deriving instance ToJSON (Ledger.DPState StandardAllegra)
-deriving instance ToJSON (Ledger.DState StandardAllegra)
-deriving instance ToJSON (Ledger.InstantaneousRewards StandardAllegra)
-deriving instance ToJSON (Ledger.SnapShot StandardAllegra)
-deriving instance ToJSON (Ledger.SnapShots StandardAllegra)
-deriving instance ToJSON (Ledger.NonMyopic StandardAllegra)
```
mrBliss, line 160: A bunch of these are now parameterised by the crypto, so only a single instance is needed instead of one for each era.

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -967,10 +965,11 @@ makeShelleyTransactionBody era@ShelleyBasedEraShelley
           (case txUpdateProposal of
              TxUpdateProposalNone -> SNothing
              TxUpdateProposal _ p -> SJust (toShelleyUpdate p))
-          (maybeToStrictMaybe (hashAuxiliaryData <$> txAuxData)))
+          (maybeToStrictMaybe
+            (Ledger.hashAuxiliaryData @StandardShelley <$> txAuxData)))
```
mrBliss, line 969: Same comment about ambiguous types here.

cardano-cli/src/Cardano/CLI/Byron/UpdateProposal.hs
```diff
@@ -141,10 +141,11 @@ createUpdateProposal
...
     signProposal (toByronProtocolMagicId nw) proposalBody noPassSigningKey
   where
-    proposalBody = ProposalBody pVer protocolParamsUpdate sVer metaData
+    proposalBody = ProposalBody pVer protocolParamsUpdate sVer updateMetadata
+
+    updateMetadata :: M.Map SystemTag InstallerHash
+    updateMetadata = M.singleton sysTag inshash
 
-    metaData :: M.Map SystemTag InstallerHash
-    metaData = M.singleton sysTag inshash
```
mrBliss, line 147: Rename `metaData` to `metadata` gives a name shadowing warning, so I picked something else (`metaData` would be wrong, IMO).

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -279,6 +279,7 @@ handleSimpleNode p trace nodeTracers nc onKernel = do
     , dtHandshakeTracer = handshakeTracer nodeTracers'
     , dtHandshakeLocalTracer = localHandshakeTracer nodeTracers'
     , dtDiffusionInitializationTracer = diffusionInitializationTracer nodeTracers'
+    , dtLedgerPeersTracer = nullTracer -- TODO network team
```
mrBliss, line 282: @karknu @coot: If you'd like a tracer, feel free to push a commit adding one.

cardano-node/src/Cardano/Tracing/Kernel.hs
```diff
@@ -52,7 +54,8 @@ mapNodeKernelDataIO f (NodeKernelData ref) =
   readIORef ref >>= traverse f
 
 nkQueryLedger ::
-     (ExtLedgerState blk -> a)
+     IsLedger (LedgerState blk)
```
mrBliss, line 57: This is because of https://github.com/input-output-hk/ouroboros-network/pull/2795

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
```diff
@@ -287,6 +287,7 @@ instance HasSeverityAnnotation (WithMuxBearer peer MuxTrace) where
     MuxTraceStartEagerly _ _ -> Debug
     MuxTraceStartOnDemand _ _ -> Debug
     MuxTraceStartedOnDemand _ _ -> Debug
+    MuxTraceTerminating {} -> Debug
```
mrBliss, line 290: @karknu @coot Is this the right verbosity?

cardano-node/src/Cardano/Tracing/OrphanInstances/Shelley.hs
```diff
@@ -420,8 +420,8 @@ instance ToJSONKey MA.AssetName where
 instance ToJSON MA.ValidityInterval where
   toJSON vi =
     Aeson.object $
-        [ "validFrom" .= x | x <- mbfield (MA.validFrom vi) ]
-     ++ [ "validTo"   .= x | x <- mbfield (MA.validTo   vi) ]
+        [ "invalidBefore"    .= x | x <- mbfield (MA.invalidBefore    vi) ]
+     ++ [ "invalidHereafter" .= x | x <- mbfield (MA.invalidHereafter vi) ]
```
mrBliss, line 424: I'm changing these fields, is that breaking something?

deepfire: Not for the benchmarking team.

cardano-cli/src/Cardano/CLI/Byron/UpdateProposal.hs
deepfire: `metadatum`?

cardano-node/src/Cardano/Node/Run.hs
karknu: `nullTracer` is fine for now. The code that use ledger peers aren't in master.

mrBliss:
I pushed another update, which brings in https://github.com/input-output-hk/ouroboros-network/pull/2811.

That commit can be squashed before merging.

newhoggy (approved):

mrBliss:
Can I squash and merge this? I'm afraid it will get stale and that someone else will try to duplicate this work.

erikd:
Please do!

mrBliss:
> Please do!

Ok, will do :slightly_smiling_face: 

I have rebased, but not squashed. I don't know what the preference is for this repo: buildable commits or logical commits that can be reviewed in isolation.

I went for the latter in this case, because of the big "MetaData -> Metadata" renaming.

mrBliss:
bors merge

erikd:
For a dependency update like this, I would have just squashed it into a single commit, but whatever.

PR #2219, opened 2020-12-15
Remove byron delegation cli commands

Jimbo4350:
We delegate solely from Byron genesis on mainnet so there is no need for Byron era delegation commands in the cli.


dcoutts (approved):
Yay!

PR #2239, opened 2020-12-23
Fix outdated calc process in doc for KES period

jun0tpyrc:


Jimbo4350:
Thanks, this has already been fixed in `master`

PR #2240, opened 2020-12-23
Update shelley tx submission with new cardano-api IPC

Jimbo4350:


Jimbo4350:
This already made its way into `master`

PR #2242, opened 2020-12-23
CAD-2440 supervisord:  consolidate non-service parametrisation in profiles

deepfire:
This:
- adds a notion of "profile" to the `supervisord` clusters, and uses it to consolidate their parametrisation
- splits out genesis generation and re-parametrises it in terms of profiles


jonringer (approved):
the nix LGTM

not sure how to test :(

PR #2253, opened 2021-01-05
Update fromShelleyAddr to also require a ShelleyBasedEra parameter

Jimbo4350:


newhoggy (approved):

dcoutts (approved):
LGTM!

I agree it makes sense for this to be an explicit param, rather than an implicit class-based param.

PR #2275, opened 2021-01-12
Integrate new api queries and tx submission into cardano-cli

Jimbo4350:
The following queries from the new Api are integrated into `cardano-cli`:
- Protocol parameters query 
- Stake distribution query
- Stake address info query

### DIFF
--- a/cardano-api/cardano-api.cabal
+++ b/cardano-api/cardano-api.cabal
@@ -192,6 +192,7 @@ test-suite cardano-api-test
                         Test.Cardano.Api.Typed.CBOR
                         Test.Cardano.Api.Typed.Envelope
                         Test.Cardano.Api.Typed.Gen
+                        Test.Cardano.Api.Typed.JSON
                         Test.Cardano.Api.Typed.MultiSig.Allegra
                         Test.Cardano.Api.Typed.MultiSig.Mary
                         Test.Cardano.Api.Typed.MultiSig.Shelley
--- a/cardano-api/cardano-api.cabal
+++ b/cardano-api/cardano-api.cabal
@@ -27,6 +27,7 @@ library
                         Cardano.Api.LocalChainSync
                         Cardano.Api.Modes
                         Cardano.Api.Protocol
+                        Cardano.Api.ProtocolParameters
                         Cardano.Api.Protocol.Byron
                         Cardano.Api.Protocol.Cardano
                         Cardano.Api.Protocol.Shelley
@@ -60,7 +61,7 @@ library
                         Cardano.Api.KeysPraos
                         Cardano.Api.NetworkId
                         Cardano.Api.OperationalCertificate
-                        Cardano.Api.ProtocolParameters
+-- TODO: move here      Cardano.Api.ProtocolParameters
                         Cardano.Api.Query
                         Cardano.Api.Script
                         Cardano.Api.SerialiseBech32
--- a/cardano-api/cardano-api.cabal
+++ b/cardano-api/cardano-api.cabal
@@ -25,12 +25,14 @@ library
                         Cardano.Api.Crypto.Ed25519Bip32
                         Cardano.Api.Eras
                         Cardano.Api.LocalChainSync
+                        Cardano.Api.Modes
                         Cardano.Api.Protocol
                         Cardano.Api.Protocol.Byron
                         Cardano.Api.Protocol.Cardano
                         Cardano.Api.Protocol.Shelley
                         Cardano.Api.Protocol.Types
                         Cardano.Api.Shelley.Genesis
+                        Cardano.Api.TxInMode
                         Cardano.Api.TxSubmit
                         Cardano.Api.Typed
 
@@ -56,7 +58,6 @@ library
                         Cardano.Api.KeysByron
                         Cardano.Api.KeysShelley
                         Cardano.Api.KeysPraos
-                        Cardano.Api.Modes
                         Cardano.Api.NetworkId
                         Cardano.Api.OperationalCertificate
                         Cardano.Api.ProtocolParameters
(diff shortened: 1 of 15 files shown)


cardano-api/src/Cardano/Api/Query.hs
```diff
@@ -88,6 +89,14 @@ data QueryInMode mode result where
 
 deriving instance Show (QueryInMode mode result)
 
+createQueryInMode :: EraInMode era mode
+                  -> QueryInShelleyBasedEra era result
+                  -> QueryInMode mode (Either EraMismatch result)
+createQueryInMode eraInMode' query =
+  case cardanoEraStyle $ eraInModeToEra eraInMode' of
+    LegacyByronEra -> error $ "Expecting shelley based era when trying to create query. \
```
Jimbo4350, line 97: @dcoutts it would be nice if I could figure out how to tell the compiler the `LegacyByronEra` case isn't possible. I tried `case eraInMode of {}` but that didn't work :thinking:

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -122,30 +122,23 @@ runQueryProtocolParameters
...
-                              ByronProtocolNotSupportedError)
-
+runQueryProtocolParameters anyEra@(AnyCardanoEra era) anyCmodeParams@(AnyConsensusModeParams cModeParams)
+                           network mOutFile = do
+  SocketPath sockPath <- firstExceptT ShelleyQueryCmdEnvVarSocketErr
+                           readEnvSocketPath
+  eraInMode <- hoistMaybe (ShelleyQueryCmdEraConsensusModeMismatch anyEra anyCmodeParams) $ toEraInMode cModeParams era
+
+  let localNodeConnInfo = NewIPC.LocalNodeConnectInfo cModeParams network sockPath
+      qInMode = NewIPC.createQueryInMode eraInMode NewIPC.QueryProtocolParameters
```
dcoutts, line 132: I think we can avoid using `createQueryInMode` here, which has the problem that it doesn't know the era is a Shelley era.

We can use `cardanoEraStyle era` right here, and fail in the Byron case, and otherwise we get a `ShelleyBasedEra` directly, and we can use `QueryInEra eraInMode (QueryInShelleyBasedEra era ...)` directly.

dcoutts (changes requested):
I think this is looking very good.

A number of relatively minor comments below.

cardano-api/cardano-api.cabal
```diff
@@ -24,12 +24,14 @@ library
                         -- modules above
                         Cardano.Api.Crypto.Ed25519Bip32
                         Cardano.Api.LocalChainSync
+                        Cardano.Api.Modes
```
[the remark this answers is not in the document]
Jimbo4350: We can but I think it would be more straightforward to expose thing appropriately after we remove all the old IPC stuff.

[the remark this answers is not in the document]
dcoutts: Ok. Sure, that's reasonable.

cardano-api/src/Cardano/Api/Modes.hs
```diff
@@ -171,6 +193,14 @@ data ConsensusModeParams mode where
        :: Byron.EpochSlots
        -> ConsensusModeParams CardanoMode
 
+deriving instance Show (ConsensusModeParams mode)
+
+instance TestEquality ConsensusModeParams where
+  testEquality (ByronModeParams _) (ByronModeParams _) = Just Refl
+  testEquality ShelleyModeParams ShelleyModeParams = Just Refl
+  testEquality (CardanoModeParams _) (CardanoModeParams _) = Just Refl
+  testEquality _ _ = Nothing
```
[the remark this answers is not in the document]
Jimbo4350: So we need an `Eq` instance (and therefore a `TestEquality` instance) for `ConsensusModeParam`s because we use it in the cli (the command types need an `Eq` instance) to make a `LocalNodeConnectInfo mode` for the queries.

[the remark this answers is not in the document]
Jimbo4350: Success :smile:

cardano-api/src/Cardano/Api/Query.hs
```diff
@@ -128,8 +128,9 @@ data QueryInShelleyBasedEra era result where
 
      QueryStakeAddresses
        :: Set StakeCredential
-       -> QueryInShelleyBasedEra era (Map StakeCredential Lovelace,
-                                      Map StakeCredential PoolId)
+       -> NetworkId
```
dcoutts, line 131: Why did we need to add the network id? We don't seem to be using it below.

cardano-api/src/Cardano/Api/IPC.hs
```diff
@@ -43,6 +43,13 @@ module Cardano.Api.IPC (
     QueryInEra(..),
     QueryInShelleyBasedEra(..),
     queryNodeLocalState,
+
+    -- *** Tip query
+    getLocalChainTip,
+
+    -- *** Helpers
+    ConsensusMode(..),
+    consensusModeOnly,
```
dcoutts, line 52: In the end we should export these only via Cardano.Api.Mode, and that'll get picked up and re-exported by Cardano.Api. This is ok for now (though we can add a TODO to remind us if you like).

dcoutts (approved):
LGTM!

cardano-api/src/Cardano/Api/Query.hs
Jimbo4350: We use it here: https://github.com/input-output-hk/cardano-node/pull/2275/files#diff-12e79fbc433fabcfcd46baf0c85e3986c1e0a2460d5a28aac0a5bab5ca3df5ddR446

This allows us to create the `StakeAddress` so we can serialise it more conveniently (vs dealing with a `StakeCredential`)

PR #2276, opened 2021-01-13
Fix haddock publishing

newhoggy:
See https://input-output-hk.github.io/cardano-node/


dcoutts (approved):
Can we add a link in the README?

newhoggy:
bors merge

PR #2277, opened 2021-01-13
Add ProtocolState and LedgerState queries to api

Jimbo4350:


cardano-api/src/Cardano/Api/Query.hs
```diff
@@ -23,35 +23,40 @@ module Cardano.Api.Query (
     -- * Internal conversion functions
     toConsensusQuery,
     fromConsensusQueryResult,
+
+    -- TODO: Move me
+    LedgerState(..),
+    ProtocolState(..),
```
DavidEichmann, line 29: Not sure where to move these. Perhaps under `Cardano.Api.Ledger`? Simply because the underlying types are:

* `Shelley.Spec.Ledger.LedgerState` `NewEpochState`
* `Shelley.Spec.Ledger.API.Protocol` `ChainDepState`

DavidEichmann (changes requested):
Just need to decide where to put Ledger/ProtocolState. This is otherwise pretty straightforward.

cardano-api/src/Cardano/Api/Query.hs
```diff
@@ -23,35 +23,40 @@ module Cardano.Api.Query (
...
+import qualified Ouroboros.Consensus.Shelley.Ledger as Consensus
+import           Ouroboros.Network.Block (Serialised)
 
 import qualified Cardano.Chain.Update.Validation.Interface as Byron.Update
 
 import qualified Cardano.Ledger.Era as Ledger
 import qualified Cardano.Ledger.Shelley.Constraints as Ledger
 
-import qualified Shelley.Spec.Ledger.API         as Shelley
+import qualified Shelley.Spec.Ledger.API as Shelley
```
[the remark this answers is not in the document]
Jimbo4350: Yep we use `stylish-haskell`

cardano-api/src/Cardano/Api/Query.hs
Jimbo4350: I think we should leave them here for now under the `Wrapper types used in queries` heading. I don't want to create a new module for 2 `newtypes`.  We can break up the module and group them appropriately once we are a little further along.

DavidEichmann (approved):
Looking good to me

dcoutts (approved):
LGTM!

PR #2281, opened 2021-01-14
CAD-2477:  clean up metric name space

deepfire:
1. Consolidate all metrics under `cardano.node.metrics`.
2. Update configs.
3. Eliminate mentions of stale namespace references.

Now:
Screenshot from 2021-01-15 19-27-04

Before:
Screenshot from 2021-01-15 01-25-50

### DIFF
--- a/cardano-node/src/Cardano/Tracing/Tracers.hs
+++ b/cardano-node/src/Cardano/Tracing/Tracers.hs
@@ -400,7 +400,7 @@ traceChainMetrics tr = Tracer $ \ev ->
    doTrace :: ChainInformation -> IO ()
    doTrace ChainInformation { slots, blocks, density, epoch, slotInEpoch } = do
      -- TODO this is executed each time the chain changes. How cheap is it?
-     meta <- mkLOMeta Critical Confidential
+     meta <- mkLOMeta Critical Public
      let traceD :: Text -> Double -> IO ()
          traceD msg d = traceNamedObject tr (meta, LogValue msg (PureD d))
          traceI :: Integral a => Text -> a -> IO ()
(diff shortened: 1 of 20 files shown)


configuration/defaults/byron-staging/configuration.yaml
```diff
@@ -224,8 +224,6 @@ options:
       - EKGViewBK  
     # Uncomment it to send 'cardano.node.metrics' to 'TraceForwarderBK' as well.
     # - TraceForwarderBK
-    cardano.node-metrics:
-      - EKGViewBK
```
erikd, line 228: Is there some mechanism to make sure these get fixed in the config tempate devops uses?

deepfire: I've made @jbgi aware of this, yes.

erikd: So this is currently a manual process? Can it not be automated in some way to make sure it always gets done?

deepfire: This template is in a different repository -- `iohk-nix`, and it contains defaults for multiple services.

We've actually been discussing this topic a while ago, but it wasn't feasible then.

cc @disassembler

denisshevchenko (approved):
LGTM

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -353,12 +354,13 @@ teeTraceChainTip
   => TraceOptions
   -> MVar (Maybe (WithSeverity (ChainDB.TraceEvent blk)), Integer)
   -> Trace IO Text
+  -> Trace IO Text
   -> Tracer IO (WithSeverity (ChainDB.TraceEvent blk))
-teeTraceChainTip TracingOff _ _ = nullTracer
-teeTraceChainTip (TracingOn trSel) elided tr =
+teeTraceChainTip TracingOff _ _ _ = nullTracer
+teeTraceChainTip (TracingOn trSel) elided trTrc trMet =
```
Jimbo4350, line 360: Why are we adding an additional parameter to `teeTraceChainTip` when both parameters accept the same tracer?  Can't we move `appendName` into `teeTraceChainTip`?

deepfire: @Jimbo4350, it's to make it a bit more apparent at the top level of the call chain that we're posting metrics to two different namespaces.

Jimbo4350 (approved):
LGTM with the caveat that devops should look at this PR wrt changes made to metrics

jbgi (approved):

PR #2274, opened 2021-01-12
Update dependencies

mrBliss:
Note that this does not incorporate the latest changes in cardano-ledger-specs.

Notable changes:
* https://github.com/input-output-hk/ouroboros-network/pull/2807
* https://github.com/input-output-hk/ouroboros-network/pull/2811
* https://github.com/input-output-hk/ouroboros-network/pull/2832

PR #2283, opened 2021-01-18
niv update haskell.nix

hamishmack:


erikd:
This was pulled into #2274 .

PR #2285, opened 2021-01-18
niv update haskell.nix to fix isProject

hamishmack:


erikd:
Pulled the Niv/Nix fix into #2274, so closing this.

PR #2287, opened 2021-01-19
Use traceChainSyncBlockServer instead of traceChainSyncHeaderServer

newhoggy:


newhoggy:
bors merge

Jimbo4350 (approved):

karknu (approved):
LGTM

nfrisby:
Who owns the chairman tests? Is that failure an obvious CI hiccup, or is it possible that the fix you're making here is breaking something downstream that was perhaps-unintentionally a workaround for this typo?

newhoggy:
> Who owns the chairman tests? Is that failure an obvious CI hiccup, or is it possible that the fix you're making here is breaking something downstream that was perhaps-unintentionally a workaround for this typo?

That would be me.  I think it's likely just a flaky CI issue.  As for why it's flaky, it's been that way for many months now and the fix will need collaboration with consensus/network devs.

newhoggy:
bors merge

PR #2292, opened 2021-01-20
New served.header.count metric

newhoggy:
This implements "Counter, monotonically increasing, number of block headers by chain sync served" from https://github.com/input-output-hk/cardano-node/issues/2261

Requires https://github.com/input-output-hk/ouroboros-network/pull/2886 upstream.

image

### DIFF
--- a/cardano-node/src/Cardano/Tracing/Tracers.hs
+++ b/cardano-node/src/Cardano/Tracing/Tracers.hs
@@ -93,8 +93,10 @@ import           Cardano.Node.Configuration.Logging
 -- For tracing instances
 import           Cardano.Node.Protocol.Byron ()
 import           Cardano.Node.Protocol.Shelley ()
+import           Ouroboros.Consensus.MiniProtocol.ChainSync.Server
 
 import qualified Ouroboros.Network.Diffusion as ND
+import qualified Control.Concurrent.STM as STM
 
 {- HLINT ignore "Redundant bracket" -}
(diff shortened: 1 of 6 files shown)


cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -441,13 +449,23 @@ mkConsensusTracers
...
   -> IO (Consensus.Tracers' peer localPeer blk (Tracer IO))
 mkConsensusTracers trSel verb tr nodeKern fStats = do
+  let trmet = appendName "metrics" tr
+
   blockForgeOutcomeExtractor <- mkOutcomeExtractor
   elidedFetchDecision <- newstate  -- for eliding messages in FetchDecision tr
   forgeTracers <- mkForgeTracers
+  meta <- mkLOMeta Critical Public
+
+  headerMessageCount <- STM.newTVarIO 0
```
newhoggy, line 459: @deepfire I'm looking for a way to avoid having to create a TVar here and let the monitoring framework handle the counting.

deepfire: @newhoggy, question is -- we have multiple peers we're syncing from, so a thread-local counter wouldn't make a lot of sense, which means you'd need shared state anyway.

Or do I misunderstand your intent here?

newhoggy: That makes sense.  I'm happy to keep the shared state.

PR #2294, opened 2021-01-20
Optimize Hydra Eval

hamishmack:
Three changes that the hydra eval faster.  In order of importance:

### 1) Split `native` jobs into `linux` and `darwin`

Reduces the number of times hydra restarts the eval process because
memory is running low.  When both linux and darwin versions are
in the same attribute the evaluation runs out of memory more often
as it alternates between evaluating linux and darwin jobs.

Grouping all the linux and darwin jobs together reduces the number
of times the memory limit is reached.  This is makes the
"no change" eval more than 4x faster (270s vs 1300s).

### 2) Avoid duplicate `cabal configure`

To get a list of package names in the project we call
`cabalProject`, before using those names to provide the
overrides for another call to `cabalProject`.
`cabalProjectLocal` is not the same for the two calls and
as a result `cabal configure` is run twice (and it is very
slow for this project.  Passing the same `cabalProjectLocal`
value to both calls should improve evaluation greatly
when changes are made to the `cabal.project` or `.cabal`
files.

### 3) Include jobs to pin plan-nix

Pinning the `plan-nix` output means it is copied to the
hydra cache.  This may not improve eval on the hydra
machine itself, it will make testing evaluation on
other machines faster (since they will have access to the
cached plans).

### Jobset link

https://hydra.iohk.io/jobset/Cardano/cardano-node-pr-2294#tabs-jobs

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -40,7 +40,8 @@
 
 with (import pkgs.iohkNix.release-lib) {
   inherit pkgs;
-  inherit supportedSystems supportedCrossSystems scrubJobs projectArgs;
+  supportedSystems = [ "x86_64-linux" ];
+  inherit supportedCrossSystems scrubJobs projectArgs;
   packageSet = import cardano-node;
   gitrev = cardano-node.rev;
 };
@@ -48,10 +49,18 @@ with (import pkgs.iohkNix.release-lib) {
 with pkgs.lib;
 
 let
+  darwinRelease = (import pkgs.iohkNix.release-lib) {
+    inherit pkgs;
+    supportedSystems = [ "x86_64-darwin" ];
+    supportedCrossSystems = [];
+    inherit scrubJobs projectArgs;
+    packageSet = import cardano-node;
+    gitrev = cardano-node.rev;
+  };
   makeScripts = cluster: let
     getScript = name: {
       x86_64-linux = (pkgsFor "x86_64-linux").scripts.${cluster}.${name};
-      x86_64-darwin = (pkgsFor "x86_64-darwin").scripts.${cluster}.${name};
+      x86_64-darwin = (darwinRelease.pkgsFor "x86_64-darwin").scripts.${cluster}.${name};
     };
   in {
     node = getScript "node";
@@ -125,11 +134,16 @@ let
 
   jobs = {
     inherit dockerImageArtifact;
-    native =
+    linux =
       let filteredBuilds = mapAttrsRecursiveCond (a: !(isList a)) (path: value:
         if (any (p: take (length p) path == p) onlyBuildOnDefaultSystem) then filter (s: !(elem s nonDefaultBuildSystems)) value else value)
         (packagePlatforms project);
       in (mapTestOn (__trace (__toJSON filteredBuilds) filteredBuilds));
+    darwin =
+      let filteredBuilds = mapAttrsRecursiveCond (a: !(isList a)) (path: value:
+        if (any (p: take (length p) path == p) onlyBuildOnDefaultSystem) then filter (s: !(elem s nonDefaultBuildSystems)) value else value)
+        (darwinRelease.packagePlatforms darwinRelease.project);
+      in (darwinRelease.mapTestOn (__trace (__toJSON filteredBuilds) filteredBuilds));
     musl64 = mapTestOnCross musl64 (packagePlatformsCross (filterProject noMusl64Build));
     ifd-pins = mkPins {
       inherit (sources) iohk-nix "haskell.nix";
@@ -137,9 +151,9 @@ let
       inherit (pkgs.haskell-nix) hackageSrc stackageSrc;
     };
     cardano-node-macos = import ./nix/binary-release.nix {
-      inherit pkgs project;
+      inherit (darwinRelease) pkgs project;
       platform = "macos";
-      exes = filter (p: p.system == "x86_64-darwin") (collectJobs jobs.native.exes);
+      exes = filter (p: p.system == "x86_64-darwin") (collectJobs jobs.darwin.exes);
     };
     cardano-node-linux = import ./nix/binary-release.nix {
       inherit pkgs project;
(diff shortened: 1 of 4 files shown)


hamishmack:
I'm confused by why hydra eval is still taking 1300+ seconds.  Locally the linux parts take less than 180 seconds.  I'm going to push some commits to the branch to try to work out what is going on.

hamishmack:
I am pretty happy with this now.  The missing ingredient was separating the native `darwin` and `linux` jobs (rather than grouping them all under `native`).   The eval for this branch is now close to 5x as fast when no changes are made and I suspect it will be at least 15 minutes faster in when there are changes.

jbgi:
@hamishmack does adding caching for `plan-nix` makes https://github.com/input-output-hk/cardano-node/pull/2302/files obsolete or not?

hamishmack:
The plan only depends on the dummy version of ghc (we capture enough of the GHC output to run `cabal configure` and make a stand in for the compiler that allows the plan to be built even without GHC).  The `roots` include the full GHC.

release.nix
```diff
@@ -40,18 +40,27 @@
 
 with (import pkgs.iohkNix.release-lib) {
   inherit pkgs;
-  inherit supportedSystems supportedCrossSystems scrubJobs projectArgs;
+  supportedSystems = [ "x86_64-linux" ];
```
rvl, line 43: For clarity, could this release-lib call be put alongside the macos one?
So have two variables: `linuxRelease` and `macosRelease`.

release.nix
```diff
@@ -125,21 +134,26 @@ let
 
   jobs = {
     inherit dockerImageArtifact;
-    native =
+    linux =
```
rvl, line 137: `musl64` is also linux. Could it be like this?
```suggestion
    linux.native = ...;
    linux.musl64 = ...;
```

release.nix
```diff
@@ -125,21 +134,26 @@ let
...
       let filteredBuilds = mapAttrsRecursiveCond (a: !(isList a)) (path: value:
         if (any (p: take (length p) path == p) onlyBuildOnDefaultSystem) then filter (s: !(elem s nonDefaultBuildSystems)) value else value)
         (packagePlatforms project);
       in (mapTestOn (__trace (__toJSON filteredBuilds) filteredBuilds));
+    darwin =
```
rvl, line 142: I would prefer to call it `macos` instead of `darwin` where possible - less jargon.

release.nix
```diff
@@ -125,21 +134,26 @@ let
...
+      let filteredBuilds = mapAttrsRecursiveCond (a: !(isList a)) (path: value:
+        if (any (p: take (length p) path == p) onlyBuildOnDefaultSystem) then filter (s: !(elem s nonDefaultBuildSystems)) value else value)
+        (darwinRelease.packagePlatforms darwinRelease.project);
+      in (darwinRelease.mapTestOn (__trace (__toJSON filteredBuilds) filteredBuilds));
```
rvl, line 146: Now that there are top-level attrsets per platform, the `.x86_64-darwin` suffix is redundant. How difficult would it be to remove that?

rvl (commented):
Looks like a huge improvement. How did you find out about that eval restarting due to large attrsets is the cause of slowness?

jbgi:
@hamishmack I actually started yesterday to work on most of the comments @rvl did, so unless you also worked on it already, I can push my changes tomorrow morning.

hamishmack:
> Looks like a huge improvement. How did you find out about that eval restarting due to large attrsets is the cause of slowness?

I was frustrated with the eval time when fixing the tests the haskell.nix had broken.  While doing that I noticed that running `hydra-eval-jobs` locally was much faster.  I did a kind of binary search on differences between the local eval I was doing and the one on hydra.

I suspected it was related to macOS at first because I had switched off macOS builds on the linux machine I was using to run the tests (it did not have access to a macOS build agent).  So I disabled the parts of the nix code that required macOS build agent during eval (git rev and the derivation that strips debug code both use the `haskellBuildUtils` stuff from `iohk-nix` and it uses `cabal2Nix` from `haskellPackages` at eval time).  That way I could enable macOS for and run `hydra-eval-jobs` on my test machine.  It was fast (slower than linux only, but not as slow as on hydra)!

That left me confused.  I was wondering if we needed to switch those `iohk-nix` to use `haskell.nix` for `haskellBuildUtils` instead of nixpkgs `haskellPackages` to avoid some kind of the eval overhead.

Luckily before I went too far down that path I pushed the version without the `haskellBuildUtils` to hydra and found that it was still really slow!

I wondered about the GC restart issue because I noticed it was doing that quite a lot and it seemed to be a factor in the haskell.nix eval time issues we have had.  I reorganised the order and found it was a little faster locally, but a lot faster on hydra.

release.nix
hamishmack: Funnily enough I had both `linuxRelease` and `darwinRelease` at first to make sure I found all the places things needed to be changed to `darwinRelease`.  Then I stripped out `linuxRelase` to make the PR smaller.

release.nix
hamishmack: This combined with the `linuxRelease` suggestion, makes me wonder if we should bite the bullet and remove use of `release-lib`.  I think it mostly just adds the suffix?

hamishmack:
> @hamishmack I actually started yesterday to work on most of the comments @rvl did, so unless you also worked on it already, I can push my changes tomorrow morning.

I was off yesterday so I have not done anything on it yet. I'll spend some time today trying to understand `release-lib` better and seeing if we can remove it easily.

release.nix
rvl: Yes, I was thinking the same. The other thing that release-lib adds is difficulty.
The package platforms filtering should be factored out into a mini-release-lib however. We don't want to have the same jobset flitering code sprayed everywhere.

jbgi:
@hamishmack I implemented suggestions (and also rebased) in #2315. Maybe work on a `release-lib`-free PR can be done separately? In any case, feel free to cherry-pick my commit 812ceaacf6db7a9de7c5431bc2ea0322769610c1 (need to rebase first) in this PR.

hamishmack:
> Maybe work on a release-lib-free PR can be done separately?

I think that would be best.  I'll make a new PR that uses haskell.nix/ci-lib.nix stuff @michaelpj used to refactor the haskell.nix release.nix.

hamishmack:
This was merged in #2315.

PR #2300, opened 2021-01-22
Update cli multi-asset parser to parse ASCII asset names

### LINKED ISSUE #2297: [FR] - Extend CLI syntax for multi-asset names
When minting a new token with cardano-cli there are restrictions on the characters that can be used for the asset name. It is not possible to use any symbols or spaces even though these are allowed by the ledger. There are examples in this native token document that are not currently possible to create, e.g. “Voting shares” and “Non-voting shares”

Example of issue:
```
$ cardano-cli transaction build-raw --tx-in 0f53b8a419b1505389700b66b5194e0ceefe6aecec91b7d41743764d96374f74#1 --tx-out="$(cat policy.addr)+999999800000+100 $(cat policyId).Non-voting shares" --mint="100 $(cat policyId).Non-voting shares" --fee 200000 --out-file txbody --mary-era
option --tx-out: 
unexpected '-'
expecting letter or digit, white space, "+" or end of input
```
Desired result:
Cardano-cli allows any character to be used for asset name. It will match ledger's validation rules.

Jimbo4350:
Resolves #2297

### DIFF
--- a/cardano-api/test/Test/Cardano/Api/Typed/Gen.hs
+++ b/cardano-api/test/Test/Cardano/Api/Typed/Gen.hs
@@ -162,9 +162,13 @@ genAssetName =
   Gen.frequency
     -- mostly from a small number of choices, so we get plenty of repetition
     [ (9, Gen.element ["", "a", "b", "c"])
-    , (1, AssetName <$> Gen.utf8 (Range.singleton  32) Gen.alphaNum)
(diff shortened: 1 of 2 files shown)


Jimbo4350:
Subsumed by: https://github.com/input-output-hk/cardano-node/pull/3211

PR #2303, opened 2021-01-25
CAD-2196: simplify node configuration for trace forwarder

denisshevchenko:
Now, to configure the node for working with RTView, the user should make only one single change in the node's configuration file - add `traceForwardTo` section, for example:

```
  "traceForwardTo": {
    "tag": "RemoteSocket",
    "contents": [
      "0.0.0.0",
      "3000"
    ]
  }
```

In this case:

1. `TraceForwarder` plugin will be activated,
2. all metrics/peers/errors will be forwarded by adding to `mapBackends` section.

Please note that `TurnOnLogMetrics` flag should be enabled anyway (for tracing resources metrics), but in the default configuration this flag is already `true`.

### DIFF
--- a/cardano-node/src/Cardano/Node/Configuration/Logging.hs
+++ b/cardano-node/src/Cardano/Node/Configuration/Logging.hs
@@ -42,7 +42,7 @@ import           Cardano.BM.Configuration (Configuration)
 import qualified Cardano.BM.Configuration as Config
 import qualified Cardano.BM.Configuration.Model as Config
 import           Cardano.BM.Data.Aggregated (Measurable (..))
-import           Cardano.BM.Data.Backend (Backend, BackendKind)
+import           Cardano.BM.Data.Backend (Backend, BackendKind (..))
 import           Cardano.BM.Data.LogItem (LOContent (..), LOMeta (..), LoggerName)
 import qualified Cardano.BM.Observer.Monadic as Monadic
 import qualified Cardano.BM.Observer.STM as Stm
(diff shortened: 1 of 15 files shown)


cardano-node/src/Cardano/Node/Configuration/Logging.hs
```diff
@@ -187,6 +193,26 @@ createLoggingLayer ver nodeConfig' p = do
                                                   (nodeBasicInfo nodeConfig p nodeStartTime)
            >>= loadPlugin switchBoard
 
+         -- Forward all the metrics/peers/errors to 'TraceForwarderBK' using 'mapBackends'.
+         let metricsLogger = "cardano.node.metrics"
+             peersLogger   = "cardano.node.metrics.peersFromNodeKernel"
```
deepfire, line 198: @denisshevchenko, `cardano.node.metrics.peersFromNodeKernel` should be implied by `cardano.node.metrics` -- is it really the case that you still need the former?

cardano-node/src/Cardano/Node/Configuration/Logging.hs
```diff
@@ -187,6 +193,26 @@ createLoggingLayer ver nodeConfig' p = do
...
+             errorsLoggers = [ "cardano.node.AcceptPolicy"
+                             , "cardano.node.ChainDB"
+                             , "cardano.node.DnsResolver"
+                             , "cardano.node.DnsSubscription"
+                             , "cardano.node.ErrorPolicy"
+                             , "cardano.node.Handshake"
+                             , "cardano.node.IpSubscription"
+                             , "cardano.node.LocalErrorPolicy"
+                             , "cardano.node.LocalHandshake"
+                             , "cardano.node.Mux"
```
deepfire, line 208: @denisshevchenko, are we not interested in all errors?If so -- then getting all traces for cardano.node with `Error`+ severity could be indeed the way to go?

I'm a bit uneasy about this magic declaration of a list by fiat..

denisshevchenko: Yes, we're interested in all errors. And I took this list from here: https://github.com/input-output-hk/cardano-node/blob/master/cardano-node/src/Cardano/Tracing/Tracers.hs#L278

cardano-node/src/Cardano/Node/Configuration/Logging.hs
denisshevchenko: Thanks, fixed.

cardano-node/src/Cardano/Node/Configuration/Logging.hs
denisshevchenko: Done.

deepfire (approved):
LGTM

Jimbo4350 (approved):

PR #2313, opened 2021-01-28
Install Haskell tools before checkout

newhoggy:
Just today, Github Actions builds starting being flaky.  For example:

image

image

This PR re-organises the Github Actions CI so that Haskell setup doesn't fail with the above error, which could potentially cause the error to appear in `cabal configure` instead, so a `retry` is added there.

I'm reasonably certain the flakiness is within Github Actions infrastructure.


Jimbo4350 (commented):
Approving with the caveat that I am not familiar with GitHub Actions

PR #2319, opened 2021-01-28
Remove old IPC api from cardano-api

Jimbo4350:
- Removes old IPC
- Adjusts modules exposed by `cardano-api`


DavidEichmann (approved):
Looking Good :-)

dcoutts (commented):
LGTM!

PR #2321, opened 2021-01-29
CAD-2371 Metrics for blocks lost due to switching

newhoggy:


cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -388,31 +389,44 @@ ignoringSeverity tr = Tracer $ \(WithSeverity _ ev) -> traceWith tr ev
...
+  tSwitchingBlocksLost <- STM.newTVarIO 0
+  return . Tracer $ \ev ->
+    fromMaybe (pure ()) $
+      doTrace tSwitchingBlocksGained tSwitchingBlocksLost <$> chainTipInformation ev
+  where
+    chainTipInformation :: ChainDB.TraceEvent blk -> Maybe ChainInformation
+    chainTipInformation = \case
+      ChainDB.TraceAddBlockEvent ev -> case ev of
+        ChainDB.SwitchedToAFork _warnings newTipInfo _ chain ->
+          Just $ chainInformation newTipInfo chain 0 0
```
newhoggy, line 405: I need to insert some logic here to count number of blocks lost and gained due to switching.  Also need a way to tell which blocks are mine.

nfrisby:
```
-- | The new block fits onto some fork and we have switched to that fork
-- (second fragment), as it is preferable to our (previous) current chain
-- (first fragment).
SwitchedToAFork
  [LedgerEvent blk]
  (NewTipInfo blk)
  (AnchoredFragment (Header blk))
  (AnchoredFragment (Header blk))
  ```
  
  So you want this pattern:
  
  ```
  ChainDB.SwitchedToAFork _warnings newTipInfo headersLost headersGained
  ```

Comment (nfrisby) [reply cardano-node/src/Cardano/Tracing/Tracers.hs:405]: Off the top of my head, I don't think the `HasHeader (Header blk)` constraint here is enough to determine directly if the header was one you forged or not. Maybe indirectly, eg if that list of points you forged is already being maintained nearby.

Comment (newhoggy) [reply cardano-node/src/Cardano/Tracing/Tracers.hs:405]: @nc6 informs me that I can get it from here:

https://github.com/input-output-hk/cardano-ledger-specs/blob/master/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/BlockChain.hs#L412

I suppose I need extra constraints to access these?

Comment (newhoggy) [reply cardano-node/src/Cardano/Tracing/Tracers.hs:405]: I feel like I need something similar to this type family:

https://github.com/input-output-hk/ouroboros-network/blob/b3730282ea065d0f902cac1dcb3237e01fecd1cf/ouroboros-consensus/src/Ouroboros/Consensus/Protocol/Signed.hs#L12

It's not clear to me there is enough to do this without changes to `ouroboros-network`.

Comment (newhoggy) [reply cardano-node/src/Cardano/Tracing/Tracers.hs:405]: Also, where can I get my own keys that I can compare to the ones in the block?

File: cardano-node/src/Cardano/Tracing/Tracers.hs [hunk 2]
```diff
@@ -386,33 +388,54 @@ ignoringSeverity :: Tracer IO a -> Tracer IO (WithSeverity a)
 ignoringSeverity tr = Tracer $ \(WithSeverity _ ev) -> traceWith tr ev
 {-# INLINE ignoringSeverity #-}
 
+countMyBlocks
+  :: forall blk. HasHeader (Header blk)
+  => AF.AnchoredFragment (Header blk)
+  -> Word64
+countMyBlocks = fromIntegral . length . AS.filter isMine
+  where isMine :: Header blk -> Bool
+        isMine _ = True
```
newhoggy, line 397: Added a new `countMyBlocks` function.  All that's left is to implement `isMine`.

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -969,19 +996,23 @@ data ChainInformation = ChainInformation
...
+  -> Word64
   -> ChainInformation
-chainInformation newTipInfo frag = ChainInformation
-    { slots       = unSlotNo $ fromWithOrigin 0 (AF.headSlot frag)
-    , blocks      = unBlockNo $ fromWithOrigin (BlockNo 1) (AF.headBlockNo frag)
-    , density     = fragmentChainDensity frag
-    , epoch       = ChainDB.newTipEpoch newTipInfo
-    , slotInEpoch = ChainDB.newTipSlotInEpoch newTipInfo
+chainInformation newTipInfo frag blocksLost = ChainInformation
+    { slots         = unSlotNo $ fromWithOrigin 0 (AF.headSlot frag)
```
erikd, line 1010: Probably better to remove all the alignment between the identified and the `=`.

Its suboptimal to have to change the indention everything one of those identifiers changes or one is added/removed.

erikd (approved):

cardano-node/src/Cardano/Tracing/Tracers.hs
newhoggy: Removed!

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -386,33 +392,55 @@ ignoringSeverity :: Tracer IO a -> Tracer IO (WithSeverity a)
...
+      doTrace tSwitchingBlocksLost <$> chainTipInformation ev
+  where
+    chainTipInformation :: ChainDB.TraceEvent blk -> Maybe ChainInformation
+    chainTipInformation = \case
+      ChainDB.TraceAddBlockEvent ev -> case ev of
+        ChainDB.SwitchedToAFork _warnings newTipInfo oldChain newChain ->
+          let blocksLost = (countMyBlocks blockConfig oldChain) - (countMyBlocks blockConfig newChain) in
+          Just $ chainInformation newTipInfo newChain blocksLost
+        ChainDB.AddedToCurrentChain _warnings newTipInfo _ chain ->
+          Just $ chainInformation newTipInfo chain 0
```
nfrisby, line 426: I think you want to do the same thing here: count it as "-1 blocks lost" if the new block is one we forged.

More specifically: I'm not certain about this next claim, but I'd be surprised if the new chain had more than one additional block. So you should be able to look at just the latest header in `chain`.

Alternatively, you could so the same counting within each chain and then the subtraction as in the `SwitchedToAFork` case.

https://github.com/input-output-hk/ouroboros-network/blob/1efb25d657ca24da00fdcefa17fb6c1eb6909b49/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs#L592-L599

nfrisby: If you don't want to consider extensions, then I think you're doing what Duncan had proposed. But he had also proposed bounding the delta at 0, so it only ever increases, IIRC. (Maybe you're chooisng some middle ground that I don't recall from the Slack thread?)

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -969,19 +997,23 @@ data ChainInformation = ChainInformation
   , slotInEpoch :: Word64
     -- ^ Relative slot number of the tip of the current chain within the
     -- epoch.
+  , blocksLost :: Int64
+    -- ^ Blocks lost due to switch at fork
```
nfrisby, line 1001: If you do end up following my overall proposal, then this comment is wrong and also the name is misleading.

I'd call it `myBlocksSelected`, an the comment would say something along the lines of: it's the difference between the numbers of my blocks on the latest selection minus that same number as of the last restart. Thus comparing `myBlocksSelected` to the number of blocks I've forged since the last restart shows how many blocks I've forged for naught.

nfrisby: Looks like that ^^^ plan would require an additional metric that simply just increments with each `TraceForgedBlock` event.

cardano-node/src/Cardano/Tracing/Tracers.hs
newhoggy: Thanks for the suggestion.  I make the code the same as the `SwitchedToAFork` case.  I think that should work.

cardano-node/src/Cardano/Tracing/Tracers.hs
newhoggy: I've updated the code to what I think you mean.

nfrisby: OK, after a little more thinking, I actually think one metric will suffice, and it can be "blocks forged since last restart but not currently selected".

`TraceForgedBlock` should increment the metric. `SwitchedToAFork` and `AddedToCurrentChain` should both increment it for each block of ours on the old chain and also decrement it for each block of ours on the new chain.

If we restart and then immediately switch to a fork with more of ours blocks, then this metric might dip negative. But over time, we should generally see this metric gradually increase.

Does that make sense? I'm headed to bed now -- talk to you tomorrow.

newhoggy: I feel like `TraceForgedBlock` shouldn't increment the metric because when we forge a block it is appended to the current chain and is therefore not one of the unselected ones?

nfrisby: The `TraceForgedBlock` just means we created the block and are about to add it to the ChainDB. Once it's added, then the ChainDB might select it, which would incur a separate `AddedToCurrentChain` event. So decrement for the forge, then increment for the subsequent selection, thus giving us the desired delta of 0.

newhoggy: Done.  I've used the term "uncoupled" to mean blocks forged not on the current chain, which is the thing being counted.

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -705,11 +738,15 @@ notifyBlockForging fStats tr = Tracer $ \case
...
                       ftsNodeIsLeaderNum fts + 1))
-  Consensus.TraceForgedBlock {} ->
+  Consensus.TraceForgedBlock {} -> do
     traceCounter "blocksForgedNum" tr
       =<< mapForgingCurrentThreadStats fStats
             (\fts -> (fts { ftsBlocksForgedNum = ftsBlocksForgedNum fts + 1 },
                        ftsBlocksForgedNum fts + 1))
+    meta <- mkLOMeta Critical Public
+    let tBlocksUncoupled = fsBlocksUncoupled fStats
+    traceI tr meta "mySelectedBlocks" =<< STM.modifyReadTVarIO tBlocksUncoupled (+ 1)
```
nfrisby, line 748: `mySelectedBlocks` sounds like the number of blocks I have forged that _are_ selected. But the metric is the number of such blocks that _are not_ selected. Just sharing my reaction, FWIW.

nfrisby, line 748: Yep, it's a `+1` because `blocksUncoupledDelta` counts blocks not on the current chain. And even though we just forged it, it's not on the chain until we get a corresponding `AddedToCurrentChain, which is handled higher up in the diff. 👍

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -386,33 +393,59 @@ ignoringSeverity :: Tracer IO a -> Tracer IO (WithSeverity a)
...
+countMyBlocks
+  :: forall blk. ()
+  => HasHeader (Header blk)
+  => BlockSupportsMetrics blk
+  => BlockConfig blk
+  -> AF.AnchoredFragment (Header blk)
+  -> Int64
+countMyBlocks blockConfig = fromIntegral . length . AS.filter isMine
+  where isMine :: Header blk -> Bool
+        isMine header = isSelfIssued blockConfig header == IsSelfIssued
```
nfrisby, line 405: One last caveat: @disassembler on Slack asked which identity the `isSelfIssued` is using to determine "self". It's the cold key. I'm not sure if that's what he expected/wanted/needed; we haven't continued the discussion.

nfrisby (commented):
The logic looks correct to me 👍. I'm only not Approving because input-output-hk/ouroboros-network#2930 is not yet merged --- I was waiting for this PR to validate my upstream diff (the one that adds `isSelfIssued`).

There's one last caveat about what `isSelfIssued` should use to identify "self", see my comment on this PR that mentions Samuel.

cardano-node/src/Cardano/Tracing/Tracers.hs
newhoggy: Oops.  The string was not updated.  Will update it to `blocksUncoupled` - the number of such blocks that are not selected.

newhoggy: Fixed!

nfrisby:
@newhoggy Have you discussed this concern with @disassembler ? I haven't.

https://github.com/input-output-hk/cardano-node/pull/2321#discussion_r577201839

newhoggy:
Reproducing this comment from https://github.com/input-output-hk/ouroboros-network/pull/2930#issuecomment-785435436

> Thanks for the update. You mentioned we were using the cold keys, but @disassembler says we should be using the vrf keys.
>
> This makes sense because other nodes use the vrf keys to verify that blocks a legit and it isn't cycled as I understand.

nfrisby:
After a group convo on Slack, I merged https://github.com/input-output-hk/ouroboros-network/pull/2930 as-is. So I think this PR will also be ready once it bumps the `ouroboros-network` dep to something including the current tip of `master` there:5https://github.com/input-output-hk/ouroboros-network/commit/7cb1380ce08721ef9957ad2db019274fdbc52968

newhoggy:
bors merge

cardano-node/src/Cardano/Tracing/Tracers.hs
nfrisby: (we resolved this: _the public half of_ cold key we're using is fine)

nfrisby:
I re-reviewed the new metric itself: still LGTM. I did not review the other eg testing changes.

PR #2324, opened 2021-01-31
Update building-the-node-using-nix.md

modir:
The install script needs to be executable.

### DIFF
--- a/doc/getting-started/building-the-node-using-nix.md
+++ b/doc/getting-started/building-the-node-using-nix.md
@@ -4,6 +4,7 @@ The [Nix Package Manager][nix] can be installed on most Linux distributions by d
 running the installation script:
 ```
 curl -L https://nixos.org/nix/install > install-nix.sh
+chmod +x install-nix.sh
 ./install-nix.sh
 ```
 and following the directions.


doc/getting-started/building-the-node-using-nix.md
```diff
@@ -4,6 +4,7 @@ The [Nix Package Manager][nix] can be installed on most Linux distributions by d
 running the installation script:
 ```
 curl -L https://nixos.org/nix/install > install-nix.sh
+chmod +x install-nix.sh
```
yihuang, line 7: Why not one-liner: `curl -L https://nixos.org/nix/install | sh`

modir: This would work as well. But I am not here to propose the best solution. I am only want to show that the current documentation has a bug. You can not execute the shell script if it doesn't have the executable bit set. If you want you can propose a merge request with your solution and then I take mine back.

deepfire (approved):
LGTM, thank you @modir!

Jimbo4350:
Thanks, subsumed by: https://github.com/input-output-hk/cardano-node/pull/2442

PR #2327, opened 2021-02-01
CAD-2568 supervisord:  extend with pool-enabled Cardano-mode, with configuration profiles

deepfire:
1. Port an extended variant of the configuration profile machinery from `cardano-ops` benchmarking
2. Switch to `Cardano` protocol support with pools -- and drop the `-hfc` variant.

This is a step on a road to allow sharing cluster configuration language/functionality across the board:

- in `cardano-node` CI
- in `cardano-benchmarking`, both as local clusters and in CI
- in `cardano-ops` benchmarking setup, as used in AWS benchmarks

For description of profiles, please see https://github.com/input-output-hk/cardano-node/blob/cad-2568-supervisord/nix/supervisord-cluster/profiles.jq#L1-L40

This affects (but not really, since no parameters are changed) the `start-cluster` functionality.

Testing:

```
nix-shell -A devops --run "start-cluster; supervisorctl tail -f bft1; stop-cluster" | grep TraceStartLeadershipCheck
```

..and after a minute or so, you'll see blocks being made:

```
[andromed:cardano.node.LeadershipCheck:Info:44] [2021-01-13 22:49:09.00 UTC] {"credentials":"Shelley_Shelley","kind":"TraceStartLeadershipCheck","delegMapSize":3,"slot":49,"chainDensity":0,"utxoSize":9}
[andromed:cardano.node.LeadershipCheck:Info:44] [2021-01-13 22:49:10.00 UTC] {"credentials":"Shelley_Shelley","kind":"TraceStartLeadershipCheck","delegMapSize":3,"slot":50,"chainDensity":0,"utxoSize":9}
[andromed:cardano.node.LeadershipCheck:Info:44] [2021-01-13 22:49:11.00 UTC] {"credentials":"Shelley_Shelley","kind":"TraceStartLeadershipCheck","delegMapSize":3,"slot":51,"chainDensity":2.0e-2,"utxoSize":9}
[andromed:cardano.node.LeadershipCheck:Info:44] [2021-01-13 22:49:12.00 UTC] {"credentials":"Shelley_Shelley","kind":"TraceStartLeadershipCheck","delegMapSize":3,"slot":52,"chainDensity":2.0e-2,"utxoSize":9}
```

# TL;DR

Imagine being able to use the exact same configuration between a local cluster, CI and in AWS benchmarking environment.

Imagine also having a multitude of such configurations, referred to by descriptive names.

NOTE: this is insufficient to get there, just yet -- but it's an important step.


newhoggy:
"varianto" in the PR description is probably a typo.

newhoggy:
https://github.com/input-output-hk/cardano-node/blob/cad-2426-jq-profiles/nix/supervisord-cluster/profiles.jq#L1-L40 doesn't exist.

nix/supervisord-cluster/default.nix
```diff
@@ -61,7 +70,7 @@ let
         addr = "127.0.0.1";
         port = p;
         valency = 1;
-      }) (lib.filter (p: p != selfPort) (lib.genList (i: basePort + i + 1) (composition.numBft + composition.numPools)));
+      }) (lib.filter (p: p != selfPort) (lib.genList (i: basePort + i + 1) (composition.n_bft_hosts + composition.n_pools)));
```
newhoggy, line 73: Are we adopting new coding styles for variable names in nix?

deepfire:
@newhoggy, thanks, fixed on both counts!

nix/supervisord-cluster/default.nix
deepfire: @newhoggy, this is mirroring the profile structure.

nix/supervisord-cluster/profiles.jq
```diff
@@ -0,0 +1,88 @@
...
+
+         ## Compute the derived params.
+         | add_derived_params
+
+         ## Finally, assembly into a dictionary..
+         | { "\(.name)":
+               ## ..and cleanup:
+               . | delpaths ([["generator", "epochs"]])}
+        )
+  | add;
```
newhoggy, line 88: I've never seen `jq` files before.  This is cool.  😁

newhoggy (approved):

nix/supervisord-cluster/profiles.jq
deepfire: As soon as this stops evolving too fast, we'll probably migrate them to Haskell : -)

cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
```diff
@@ -190,7 +190,7 @@ data TransactionCmd
   | TxCalculateMinFee
       TxBodyFile
       (Maybe NetworkId)
-      ProtocolParamsFile
```
Jimbo4350, line 193: Please change this to a sum type.

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -1197,13 +1201,13 @@ pEpochNoUpdateProp =
       <> Opt.help "The epoch number in which the update proposal is valid."
       )
 
-pGenesisFile :: Parser GenesisFile
```
Jimbo4350, line 1200: This change seems unnecessary.

Jimbo4350 (changes requested):

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
deepfire: @Jimbo4350, why do you think so?

This changes allows `pGenesisFile` to be reused in two slightly different contexts.

cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
deepfire: Can I ask for a justification for this suggestion?

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
Jimbo4350: Ah right, didn't catch the other usage.

cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
Jimbo4350: `Either` implies that `Left` is an error and in this case `Left` isn't an error. A sum type is more appropriate here and makes it more clear what's happening.

deepfire: > `Either` implies that `Left` is an error

I don't see that in the documentation of Either, to be frank.

I've seen more than one introduction into algebraic types that explains them as reducible to `(,)` and `Either`.

So, I'm inclned to see `Either` possessing the same generality as `Maybe`.

Now, on the clarity grounds I could agree actually. That would allow per-sum-element documentation of the intent.

deepfire: However, when I agree that clarity can be improved, take a look at the rest of `TransactionCmd`:

```
data TransactionCmd
  = TxBuildRaw
      AnyCardanoEra
      [TxIn]
      [TxOutAnyEra]
      (Maybe Value)
      -- ^ Multi-Asset value
      (Maybe SlotNo)
      -- ^ Transaction lower bound
      (Maybe SlotNo)
      -- ^ Transaction upper bound
      (Maybe Lovelace)
      -- ^ Tx fee
      [CertificateFile]
      [(StakeAddress, Lovelace)]
      TxMetadataJsonSchema
      [ScriptFile]
      -- ^ Auxillary scripts
      [MetadataFile]
      (Maybe UpdateProposalFile)
      TxBodyFile
  | TxSign TxBodyFile [WitnessSigningData] (Maybe NetworkId) TxFile
  | TxCreateWitness TxBodyFile WitnessSigningData (Maybe NetworkId) OutputFile
  | TxAssembleTxBodyWitness TxBodyFile [WitnessFile] OutputFile
  | TxSubmit AnyConsensusModeParams NetworkId FilePath
  | TxMintedPolicyId ScriptFile
  | TxCalculateMinFee
      TxBodyFile
      (Maybe NetworkId)
      (Either GenesisFile ProtocolParamsFile)
      TxInCount
      TxOutCount
      TxShelleyWitnessCount
      TxByronWitnessCount
  | TxGetTxId (Either TxBodyFile TxFile)
```

You already can see `TxGetTxId` use `Either`, in the exact same fashion.

deepfire: @Jimbo4350, for documentation purposes, I could duplicate the explanation for the semantics I provided in https://github.com/input-output-hk/cardano-node/pull/2327/files#diff-cc9b567f3acd813c81e105c074ebd783719c74364c88a78c3c22638322354f6dR542-R545

Would you prefer that?

deepfire: Actually, did that, as that's a good idea.

Jimbo4350: In the `Data.Either` docs:
```
The Either type is sometimes used to represent a value which is either correct or an error; by convention, the Left constructor is used to hold an error value and the Right constructor is used to hold a correct value (mnemonic: "right" also means "correct").
```
Please make the change and when you do I will approve this PR.

deepfire: "Sometimes", I think means something that's different from "always", right?

deepfire: @Jimbo4350, the change you asked for has been made.

It would really help, in the future, to distinguish between _optional improvements_ and _hard requirements_.

*Especially* in the context, where the directly surrounding code does not follow the requested structure.

Jimbo4350 (approved):
LGTM!

PR #2343, opened 2021-02-09
Fix example file generation script to adapt for macOS

cblp:
Must work on Linux, too, due to BSD compatibility

### DIFF
--- a/scripts/byron-to-alonzo/mkfiles.sh
+++ b/scripts/byron-to-alonzo/mkfiles.sh
@@ -67,6 +67,12 @@ case $UNAME in
                 DATE="date";;
 esac
 
+UNAME=$(uname -s) SED=
+case $UNAME in
+  Darwin )      SED="gsed";;
+  Linux )       SED="sed";;
+esac
+
 sprocket() {
   if [ "$UNAME" == "Windows_NT" ]; then
     # Named pipes names on Windows must have the structure: "\\.\pipe\PipeName"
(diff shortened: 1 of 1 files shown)


Jimbo4350 (changes requested):
If you want to make the `mkfiles.sh` script macOS compatible then you need to utilize `uname` to do the correct thing given a particular os.

cblp:
> If you want to make the `mkfiles.sh` script macOS compatible then you need to utilize `uname` to do the correct thing given a particular os.

@Jimbo4350 I think these changes are compatible with GNU/Linux, too, since GNU utilities are compatible with the BSD interface. I'll double-check.

cblp:
It works correctly on both linux and macOS

scripts/byron-to-alonzo/mkfiles.sh
```diff
@@ -312,7 +313,7 @@ cp ../configuration/cardano/shelley_qa-alonzo-genesis.json shelley/genesis.alonz
 # We're going to use really quick epochs (300 seconds), by using short slots 0.2s
 # and K=10, but we'll keep long KES periods so we don't have to bother
 # cycling KES keys
-sed -i shelley/genesis.spec.json \
+sed -i~ \
```
Jimbo4350, line 316: This now creates two `genesis.spec.json` files which makes things ambiguous

Jimbo4350 (changes requested):

scripts/byron-to-alonzo/mkfiles.sh
cblp: done

Jimbo4350 (approved):
Nice :+1:

cblp:
bors merge

PR #2355, opened 2021-02-11
CAD-2572 Direct submission of the ChainSync headers served metric to EKG

jutaro:
1. Bump `iohk-monitoring-framework` to the latest `master`, that includes EKG server extraction.
2. Define `EKGDirect` -- a direct handle onto the EKG server/store used by the EKG backend of the logging framework.
3. Stop submitting `cardano.node.metrics.served.header.counter.int` to the switchboard.
4. Submit it via `EKGDirect`, instead.

### DIFF
--- a/cardano-node/src/Cardano/Tracing/Tracers.hs
+++ b/cardano-node/src/Cardano/Tracing/Tracers.hs
@@ -279,7 +279,7 @@ mkTracers
   => TraceOptions
   -> Trace IO Text
   -> NodeKernelData blk
-  -> EKGDirect
+  -> Maybe EKGDirect
   -> IO (Tracers peer localPeer blk)
 mkTracers tOpts@(TracingOn trSel) tr nodeKern ekgDirect = do
   fStats <- mkForgingStats
@@ -368,7 +368,7 @@ teeTraceChainTip
      )
   => TraceOptions
   -> MVar (Maybe (WithSeverity (ChainDB.TraceEvent blk)), Integer)
-  -> EKGDirect
+  -> Maybe EKGDirect
   -> Trace IO Text
   -> Trace IO Text
   -> Tracer IO (WithSeverity (ChainDB.TraceEvent blk))
@@ -398,10 +398,10 @@ ignoringSeverity tr = Tracer $ \(WithSeverity _ ev) -> traceWith tr ev
 
 traceChainMetrics
   :: forall blk. HasHeader (Header blk)
-  => EKGDirect -> Trace IO Text -> Tracer IO (ChainDB.TraceEvent blk)
-traceChainMetrics ekgDirect _tr = Tracer $ \ev ->
-  fromMaybe (pure ()) $
-    doTrace <$> chainTipInformation ev
+  => Maybe EKGDirect -> Trace IO Text -> Tracer IO (ChainDB.TraceEvent blk)
+traceChainMetrics Nothing _ = nullTracer
+traceChainMetrics (Just ekgDirect) _tr = Tracer $ \ev ->
+  fromMaybe (pure ()) $ doTrace <$> chainTipInformation ev
  where
    chainTipInformation :: ChainDB.TraceEvent blk -> Maybe ChainInformation
    chainTipInformation = \case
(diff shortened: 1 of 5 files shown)


deepfire (approved):
LGTM, thank you Jurgen!

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -420,6 +430,28 @@ traceD tr meta msg d = traceNamedObject tr (meta, LogValue msg (PureD d))
...
+ case SMap.lookup name registeredMap of
+   Just gauge -> Gauge.set gauge (fromIntegral val)
+   Nothing -> do
+     gauge <- EKG.getGauge name (ekgServer ekgDirect)
+     let registeredMap' = SMap.insert name gauge registeredMap
+     writeIORef (ekgGauges ekgDirect) registeredMap'
+     Gauge.set gauge (fromIntegral val)
+
+_sendEKGDirectDouble :: EKGDirect -> Text -> Double -> IO ()
+_sendEKGDirectDouble ekgDirect name val = do
```
newhoggy, line 445: Why does this function start with an underscore?

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -420,6 +430,28 @@ traceD tr meta msg d = traceNamedObject tr (meta, LogValue msg (PureD d))
...
+sendEKGDirectInt :: Integral a => EKGDirect -> Text -> a -> IO ()
+sendEKGDirectInt ekgDirect name val = do
+ registeredMap <- readIORef (ekgGauges ekgDirect)
+ case SMap.lookup name registeredMap of
+   Just gauge -> Gauge.set gauge (fromIntegral val)
+   Nothing -> do
+     gauge <- EKG.getGauge name (ekgServer ekgDirect)
+     let registeredMap' = SMap.insert name gauge registeredMap
+     writeIORef (ekgGauges ekgDirect) registeredMap'
+     Gauge.set gauge (fromIntegral val)
```
newhoggy, line 442: What happens if this is run from multiple threads and `getGauge` gets called multiple times for the same name?

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -544,6 +572,13 @@ mkConsensusTracers trSel verb tr nodeKern fStats = do
        <*> counting (liftCounting staticMeta name "slot-is-immutable" tr)
        <*> counting (liftCounting staticMeta name "node-is-leader" tr)
 
+   traceServedCount :: Maybe EKGDirect -> STM.TVar Int -> TraceChainSyncServerEvent blk -> IO ()
+   traceServedCount Nothing _ _ = pure ()
+   traceServedCount (Just ekgDirect) tHeadersServed ev =
+     when (isRollForward ev) $ do
+       count <- STM.modifyReadTVarIO tHeadersServed (+1)
```
newhoggy, line 579: Are we still using TVars?  I had the impression that we were going to replace the use of TVars.

cardano-node/src/Cardano/Tracing/Tracers.hs
deepfire: @newhoggy, this is currently unused, but will be immediately used by anyone who needs to post a `Double` metric.

We don't want then to be running around for recipes.. : -)

cardano-node/src/Cardano/Tracing/Tracers.hs
newhoggy: Mostly due to this comment https://input-output-rnd.slack.com/archives/GR599HMFX/p1613382544290700

deepfire: @newhoggy, can you tell a bit more -- what should we be using instead?  I'm afraid I'm not up-to-date..

newhoggy: Found the comment https://github.com/input-output-hk/cardano-node/pull/2366#discussion_r576062283

newhoggy: I added @coot to seek clarification on this question.

deepfire: Good point, we should be using https://hackage.haskell.org/package/ekg-core-0.1.1.7/docs/System-Metrics-Counter.html

newhoggy:
💯

newhoggy (approved):

PR #2358, opened 2021-02-13
*Update simple_transaction.md with era flag

tigrpoolcom:
Added --mary-era to it


Jimbo4350:
Thanks, this has already been fixed in `master`

PR #2359, opened 2021-02-13
*Replace type in register_stakepool.md and extend section with url sh…

tigrpoolcom:
…ortener

The typo lead to multiple entries in the cardano forum, to avoid that i fixed the error.
I also extended the part with the git and git.io to make it easier for people who are not familiar with gist.


Jimbo4350:
Thanks, this PR is subsumed by: https://github.com/input-output-hk/cardano-node/pull/2442

PR #2366, opened 2021-02-15

newhoggy:


erikd (approved):

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -502,6 +502,10 @@ mkConsensusTracers trSel verb tr nodeKern fStats = do
               traceI trmet meta "submissions.rejected.count" =<<
                 STM.modifyReadTVarIO tSubmissionsRejected (+ ptxcRejected processed)
 
+            TraceLabelPeer _ TraceTxInboundTerminated -> return ()
+            TraceLabelPeer _ (TraceTxInboundCanRequestMoreTxs _) -> return ()
+            TraceLabelPeer _ (TraceTxInboundCannotRequestMoreTxs _) -> return ()
```
newhoggy, line 507: @coot Mainly I wanted to know if something should be done here.

newhoggy:
bors merge

cardano-node/src/Cardano/Tracing/OrphanInstances/Consensus.hs
```diff
@@ -156,6 +156,7 @@ instance HasSeverityAnnotation (TraceChainSyncClientEvent blk) where
   getSeverityAnnotation (TraceFoundIntersection _ _ _) = Info
   getSeverityAnnotation (TraceRolledBack _) = Notice
   getSeverityAnnotation (TraceException _) = Warning
+  getSeverityAnnotation (TraceTermination _) = Warning
```
coot, line 159: I don't think other mini-protocols use `Warning` for tracing termination.  Terminating min-protocols will be a normal mode of operation, so `Warning` seems to high severity for me.  Could you check what other mini-protocols are using?

cardano-node/src/Cardano/Tracing/Tracers.hs
coot: No, I don't think we need to do anything here.

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -502,6 +502,10 @@ mkConsensusTracers trSel verb tr nodeKern fStats = do
               traceI trmet meta "submissions.rejected.count" =<<
                 STM.modifyReadTVarIO tSubmissionsRejected (+ ptxcRejected processed)
```
coot, line 503: I know it's unrelated to the patch here, but using `STM` for counters will have performance penalty. Using something like `ekg`'s `Gauge` for that should be better.

PR #2373, opened 2021-02-16
CAD-2564: trace KES metrics even if there is error

denisshevchenko:
Previously, KES-related information wasn't traced if the node has expired KES-key. Now KES-related information will be traced anyway.

Now, even if KES-key is poisoned, KES-metrics is here:

```
$ curl 0.0.0.0:12798/metrics | grep KES
cardano_node_metrics_operationalCertificateExpiryKESPeriod_int 72
cardano_node_metrics_currentKESPeriod_int 0
cardano_node_metrics_remainingKESPeriods_int 0
cardano_node_metrics_operationalCertificateStartKESPeriod_int 10
```


deepfire (approved):
LGTM!

Jimbo4350 (approved):
LGTM

PR #2374, opened 2021-02-16
nixos: allow multiple instances to bind to same port

jbgi:


nix/nixos/cardano-node-service.nix
```diff
@@ -90,9 +106,16 @@ in {
...
+        type = types.int;
+        default = 1;
+        description = ''
+          Number of instance of the service to run.
+        '';
+      };
       script = mkOption {
         type = types.str;
-        default = mkScript cfg;
+        default = mkScript cfg 0;
```
disassembler, line 118: shouldn't this 0 be cfg.instances?

nix/nixos/cardano-node-service.nix
```diff
@@ -39,37 +49,43 @@ let
...
         echo "..or, once again, in a single line:"
         echo "${toString cmd}"
         ${lib.optionalString (cfg.environment == "selfnode") ''
-          echo "Wiping all data in ${cfg.stateDir}"
-          rm -rf ${cfg.stateDir}/*
+          echo "Wiping all data in ${instanceDbPath}"
+          rm -rf ${instanceDbPath}/*
           GENESIS_FILE=$(${pkgs.jq}/bin/jq -r .GenesisFile < ${cfg.nodeConfigFile})
           START_TIME=$(date +%s --date="30 seconds")
           ${pkgs.jq}/bin/jq -r --arg startTime "''${START_TIME}" '. + {startTime: $startTime|tonumber}' < $GENESIS_FILE > ${cfg.stateDir}/genesis.json
```
johnalotoski, line 75: ${cfg.stateDir} should also be ${instanceDbPath} in this line?

nix/nixos/cardano-node-service.nix
```diff
@@ -39,37 +49,43 @@ let
...
           ${pkgs.jq}/bin/jq -r --arg GenesisFile genesis.json '. + {GenesisFile: $GenesisFile}' < ${cfg.nodeConfigFile} > ${realNodeConfigFile}
         ''}
         ${lib.optionalString (cfg.environment == "shelley_selfnode") ''
-          echo "Wiping all data in ${cfg.stateDir}"
-          rm -rf ${cfg.stateDir}/*
+          echo "Wiping all data in ${instanceDbPath}"
+          rm -rf ${instanceDbPath}/*
           GENESIS_FILE=$(${pkgs.jq}/bin/jq -r .GenesisFile < ${cfg.nodeConfigFile})
           START_TIME=$(date --utc +"%Y-%m-%dT%H:%M:%SZ" --date="30 seconds")
           ${pkgs.jq}/bin/jq -r --arg startTime "''${START_TIME}" '. + {startTime: $startTime}' < $GENESIS_FILE > ${cfg.stateDir}/genesis.json
```
johnalotoski, line 83: ${cfg.stateDir} should also be ${instanceDbPath} in this line?

nix/nixos/cardano-node-service.nix
jbgi: no, `0` is for the first instance, which will be the instance config used by standalone scripts.

nix/nixos/cardano-node-service.nix
jbgi: actually `selfnode` env is not supported with multi-instance for now so I'm going to bother doing adaptation for it (reverting above change).

johnalotoski (approved):

Issue #2360, opened 2021-02-14
[FR] - Provide high quality multiarch docker images


tdiesler:
**Internal/External**
*External* 

**Area**
*Docker/Kubernetes*   

**Describe the feature you'd like**

This is an umbrella issue that combines various feature requests and bug fixes in order to provide a high quality Cardano Docker image. In the not too distant future it may become a reality that Cardano nodes need to get integrated in corporate  IT infrastructure. They may become destinations or sources for various corporate data streams and as such, it'd be good to have an offering that allows these organizations to integrate Cardano with already existing cloud infrastructure. 

With Kubernetes as the de-facto standard for container deployment, orchestration, monitoring, scaling , etc, it should be as easy as this to integrate a Cardano node ...

```
kubectl apply -f https://raw.githubusercontent.com/input-output-hk/cardano-node/master/cardano-node.yaml
```

This issue suggests the following improvements ...

* [#2361] Added support for arm64 and perhaps for other target platforms to follow
* [#2353] Support for graceful shutdown in container environments
* [#2362] Self contained topology updates (until there is native p2p support provided by the node)
* [#2364] Comprehensive documentation on how to run Cardano in managed container environments

**Describe alternatives you've considered**

An alternative would be to have advanced image functionality be provided by 3rd party and let this repository continue to focus on bare metal installations and more or less leave the existing (x86_64) docker image as is.

IMHO, it'd however be much better to have this tech be provided upstream.

**Additional context / screenshots**

As far as I can tell, much of the above is low hanging fruit and if it's ok with you, I'd volunteer to do this over the course of the following weeks/months. The multiarch bit may initially require a workaround until we have the regular nix build that works on arm64. The image cannot rely on an external process to do topology updates - this has to be done within. Initially we could perhaps use guild-operators for that (to be discussed). Kubernetes support and getting started docs should not be a big deal.

Preskton:
Hi, @tdiesler - I'd also offer up https://github.com/input-output-hk/cardano-node/issues/2423 as an optimization after we work through some of your above list.

tdiesler:
Status ...

PR [#2456] is the fivth and last installment of [#2360] Provide high quality multiarch docker image and k8s support

Now we have ...

* Preserved functionality via -e NETWORK
* Node configuration through env vars, which is later needed for k8s integration
* Deployment scripts for Kubernetes and Docker Compose
* Integrated configurable topology update
* Graceful shutdown on docker stop
* Improved access to the CLI
[...]

PR #2375, opened 2021-02-16
Docker container does not shutdown cleanly on SIGTERM

tdiesler:


tdiesler:
This is the second installment of [#2360] Provide high quality multiarch docker image and k8s support

* It preserves existing functionality via -e NETWORK
* Allows the node to get configured through env vars, which is later needed for k8s integration
* Improves access to the CLI

There is a README about this [here](https://github.com/tdiesler/cardano-node/tree/ghi2353/nix/docker).

tdiesler:
This is now part of #2393

PR #2382, opened 2021-02-18
Fix inline expression in docs

thbenschop:
Inline expression had incorrect result and mis-matched the description text.


Jimbo4350:
This has been taken care of in: https://github.com/input-output-hk/cardano-node/pull/2435/files#diff-35e5e2569b35abc715f089d7ce1b8096241de8eacc53a5aaf3ad570a32855eaeR79

PR #2383, opened 2021-02-18
Remove IPv6 options and make the host IP address omniversioned

cblp:


karknu (changes requested):
This removes the support for nodes running both IPv4 and IPv6 at the same time with manually configured addresses.

cblp:
@karknu is it a documented feature to run the node on multiple IP interfaces?

JamesRobertKelley:
> @karknu is it a documented feature to run the node on multiple IP interfaces?

I was wondering if it would be possible to bind the cardano-node to multiple IPs because I'm getting a Starlink dish soon and was going to multihome the server so that it would communicate on both the Starlink and local cable Internet (because Starlink is in beta and may drop connections).

coot:
^ @karknu is it a documented feature to run the node on multiple IP interfaces?

This feature is independent whether one is using multiple interfaces or not.

^ I was wondering if it would be possible to bind the cardano-node to multiple IPs

We are not supporting this at the moment, beside allowing to configure different `ipv6` and `ipv4` addresses / ports which may or may not be served by a different interface.  We might support multiple inbound addresses in the future.

cblp:
wrong direction

PR #2414, opened 2021-02-25
Update Stake Pool Operations Doc

0percentpool:
I've updated the docs for `v1.25.1`. I've also included a commit in preparation for Mary :)

With regards to calculating the current epoch, I've included an EKG query as `slotNo / epochLength` does not produce the correct epoch.

If you need any updates, or removal of the additional commits, please let me know.


Jimbo4350:
Thanks, subsumed by: https://github.com/input-output-hk/cardano-node/pull/2435

PR #2424, opened 2021-03-01
nixos service: add ipv6 support. Better control of port sharing.

jbgi:
use-case: ipv6 addr can be used in multi-instances mode so that instances
 on same machine can be included each other topology's, while still
 sharing ipv4 port.

### DIFF
--- a/nix/nixos/cardano-node-service.nix
+++ b/nix/nixos/cardano-node-service.nix
@@ -56,12 +56,13 @@ let
           "--config ${realNodeConfigFile}"
           "--database-path ${instanceDbPath}"
           "--topology ${topology}"
-        ] ++ (lib.optionals (!cfg.systemdSocketActivation) [
+        ] ++ (lib.optionals (!cfg.systemdSocketActivation) ([
           "--host-addr ${cfg.hostAddr}"
-          "--port ${toString cfg.port}"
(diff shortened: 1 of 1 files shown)


nix/nixos/cardano-node-service.nix
```diff
@@ -56,12 +56,13 @@ let
           "--config ${realNodeConfigFile}"
           "--database-path ${instanceDbPath}"
           "--topology ${topology}"
-        ] ++ (lib.optionals (!cfg.systemdSocketActivation) [
+        ] ++ (lib.optionals (!cfg.systemdSocketActivation) ([
           "--host-addr ${cfg.hostAddr}"
-          "--port ${toString cfg.port}"
+          "--port ${toString (cfg.port + i)}"
```
manveru, line 61: Shouldn't this also check `shareIpv(4|6)port`?

manveru (approved):
LGTM

Jimbo4350 (approved):

nix/nixos/cardano-node-service.nix
jbgi: no because port sharing us only supported via systemd sockets. On the command line, ports must be different owtherwise it will fail to bind.

PR #2430, opened 2021-03-02
added indication that latest GHC version may fail

dizgotti:
when trying to build with GHC 8.10.4, the following error was produced when running "cabal build all":
cardano cabal: Failed to build ListLike-4.7.2

this was remedied by using GHC 8.10.2 instead


Jimbo4350:
Subsumed by: https://github.com/input-output-hk/cardano-node/pull/2432

PR #2446, opened 2021-03-04
Avoid reorder goals on platforms that don't need it 

newhoggy:
The build optimisation that saves up to 10 minutes per configuration where it is disabled (ie. 4x per job).

This is useful not only to reduce feedback time but also because builds are a bit clogged today.


newhoggy:
Before:
Screen Shot 2021-03-04 at 1 52 57 pm

After:
Screen Shot 2021-03-04 at 1 52 37 pm

erikd (approved):

newhoggy:
bors merge

PR #2448, opened 2021-03-04
Add technical writers to CODEOWNERS

Jimbo4350:


erikd (approved):
Probably worth removing the actual user names from the commit title.

Its un-necessary detail.

dcoutts (approved):

PR #2449, opened 2021-03-04
Make release manager a codeowner for cabal.project

nc6:
We do not envision that the release manager will actually review the PR
here, but adding as a codeowner will ensure he gets tagged on any
dependency bumps.


Jimbo4350 (approved):

nc6:
bors merge

PR #2456, opened 2021-03-07
[Part 5] - Provide high quality multiarch docker image and k8s support

tdiesler:
This is the fivth and last installment of [#2360] Provide high quality multiarch docker image and k8s support

Now we have ...

* Preserved functionality via -e NETWORK
* Node configuration through env vars, which is later needed for k8s integration
* Deployment scripts for Kubernetes and Docker Compose
* Integrated configurable topology update
* Graceful shutdown on docker stop
* Improved access to the CLI
* Comprehensive documentation for Kubernetes & Compose

There is a README about this [here](https://github.com/tdiesler/cardano-node/tree/docker/nix/docker).

PS: Once this is merged we'd need to cleanup a few links and perhaps find a better home for the docs.

### DIFF
--- a/nix/docker/context/bin/topologyUpdate
+++ b/nix/docker/context/bin/topologyUpdate
@@ -1,0 +1,174 @@
+#!/bin/bash
+# shellcheck disable=SC2086,SC2034
+# shellcheck source=/dev/null
+
+# This is borrowed from https://github.com/cardano-community/guild-operators
+# and mainly just removes the auto-update of the script itself
+#
+# i.e. we want a stable/known version of this script
+#
+# Also, this is hopefully only a temporary thing until
+# the node can take care about it's p2p2 updates itself
+
+PARENT="$(dirname $0)"
+
+######################################
+# User Variables - Change as desired #
+######################################
+
+#CNODE_HOSTNAME="CHANGE ME"  # (Optional) Must resolve to the IP you are requesting from
+#CUSTOM_PEERS="None"         # Additional custom peers to (IP:port[:valency]) to add to your target topology.json
+                             # eg: "10.0.0.1:3001|10.0.0.2:3002|relays.mydomain.com:3003:3"
+
+CNODE_VALENCY=1             # (Optional) for multi-IP hostnames
+MAX_PEERS=15                # Maximum number of peers to return on successful fetch
+
+if [[ ! -f ${PARENT}/env ]]; then
+  echo "[Error] Generated env file missing: ${PARENT}/env"
+  echo "This is a mandatory prerequisite\n"
+  exit 1
+fi
+
+# source generated env variables
+if ! source ${PARENT}/env offline; then
+  echo "[Error] Cannot source: ${PARENT}/env"
+  exit 1;
+fi
+
+# NetworkMagic extracted from the genesis file
+NWMAGIC=$(jq -r .networkMagic < ${GENESIS_JSON})
+
+# Installed by pkgs.cacert
+# https://curl.se/docs/sslcerts.html
+export CURL_CA_BUNDLE="/etc/ssl/certs/ca-bundle.crt"
+
+######################################
+# Do NOT modify code below           #
+######################################
+
+usage() {
+  cat <<EOF
+Usage: $(basename "$0") [-f] [-p] [-l]
+Topology Updater - Build topology with community pools
+
+-f    Disable fetch of a fresh topology file
+-p    Disable node alive push to Topology Updater API
+-l    Enable an endless topology update loop (one call per hour)
+
+EOF
+  exit 1
+}
+
+TU_FETCH='Y'
+TU_PUSH='Y'
+LOOP='N'
+
+while getopts :fpl opt; do
+  case ${opt} in
+    f ) TU_FETCH='N' ;;
+    p ) TU_PUSH='N' ;;
+    l ) LOOP='Y' ;;
+    \? ) usage ;;
+  esac
+done
+shift $((OPTIND -1))
+
+if [[ ${CNODE_HOSTNAME} != "" ]]; then
+  T_HOSTNAME="&hostname=${CNODE_HOSTNAME}"
+else
+  T_HOSTNAME=''
+fi
+
+if [[ ${CUSTOM_PEERS} != "" ]]; then
+  CUSTOM_PEERS_PARAM="&customPeers=${CUSTOM_PEERS}"
+fi
+
+#####################################################################
+#
+# Fetch the current block number from the EKG endpoint
+#
+fetchBlockNumber () {
+
+  fail_cnt=0
+  while ! blockNo=$(curl -s -m ${EKG_TIMEOUT} -H 'Accept: application/json' "http://${EKG_HOST}:${EKG_PORT}/" 2>/dev/null | jq -er '.cardano.node.metrics.blockNum.int.val //0' ); do
+    ((fail_cnt++))
+    if [[ ${fail_cnt} -eq 5 ]]; then echo "5 consecutive EKG queries failed, aborting!"; exit 1; fi
+    echo "(${fail_cnt}/5) Failed to grab blockNum from node EKG metrics, sleeping for 30s before retrying..."
+    sleep 30
+  done
+}
+
+#####################################################################
+#
+# Push the node config info to the topology endpoint
+#
+# If you run your node in IPv4/IPv6 dual stack network configuration and want announced the
+# IPv4 address only please add the -4 parameter to the curl command below  (curl -4 -s ...)
+pushNodeInfo () {
+
+  if [[ ${TU_PUSH} = "Y" ]]; then
+
+      echo "[`date +'%Y-%m-%d %H:%M:%S %Z'`] Updating topology ..."
+
+      fetchBlockNumber
+
+      echo "curl -s -f https://api.clio.one/htopology/v1/?port=${CNODE_PORT}&blockNo=${blockNo}&valency=${CNODE_VALENCY}&magic=${NWMAGIC}${T_HOSTNAME}"
+      curl -s -f "https://api.clio.one/htopology/v1/?port=${CNODE_PORT}&blockNo=${blockNo}&valency=${CNODE_VALENCY}&magic=${NWMAGIC}${T_HOSTNAME}" | \
+        tee -a ${LOG_DIR}/topologyUpdateResult
+  else
+
+    echo "[Warning] Topology update push disabled"
+  fi
+}
+
+#####################################################################
+#
+# Fetch the next topology update
+#
+fetchTopology () {
+
+  if [[ ${TU_FETCH} = "Y" ]]; then
+
+    echo "curl -s -f https://api.clio.one/htopology/v1/fetch/?max=${MAX_PEERS}&magic=${NWMAGIC}${CUSTOM_PEERS_PARAM}"
+    curl -s -f -o ${TOPOLOGY}.tmp "https://api.clio.one/htopology/v1/fetch/?max=${MAX_PEERS}&magic=${NWMAGIC}${CUSTOM_PEERS_PARAM}" \
+      && mv ${TOPOLOGY}.tmp ${TOPOLOGY} && cat ${TOPOLOGY}
+  else
+
+    echo "[Warning] Topology update fetch disabled"
+  fi
+}
+
+##########################################################################################################################################
+#
+# Run a single topology update cycle
+#
+if [[ ${LOOP} = "N" ]]; then
+
+    pushNodeInfo
+    fetchTopology
+    exit 0
+fi
+
+##########################################################################################################################################
+#
+# Run an endless topology update loop
+#
+
+MIN=`date +"%-M"`
+INITIAL_WAIT=10
+MIN=$(($MIN + $INITIAL_WAIT))
+if (( $MIN > 59 )); then MIN=$(($MIN - 60)); fi
+
+CRONJOB="$MIN * * * * root topologyUpdate"
+echo "Topology update: $CRONJOB"
+
+echo "Initially waiting for ${INITIAL_WAIT} minutes ..."
+sleep $(($INITIAL_WAIT * 60))
+
+while true; do
+  pushNodeInfo
+  fetchTopology
+  sleep 3600
+done
+
+echo "[Warning] Exiting topology update loop"
--- a/nix/docker/k8s/cardano-nodes.yaml
+++ b/nix/docker/k8s/cardano-nodes.yaml
@@ -64,7 +64,7 @@ spec:
           name: relaycfg
           key:  customPeers
     - name: CARDANO_UPDATE_TOPOLOGY
-      value: "true"
+      value: "false"
     args: [ "run" ]
     ports:
       - containerPort: 3001
--- a/nix/docker/k8s/cardano-nodes.yaml
+++ b/nix/docker/k8s/cardano-nodes.yaml
@@ -1,0 +1,190 @@
+#
+# NAMESPACE=cardano
+# kubectl create namespace $NAMESPACE
+#
+
+# GCE Storage Class
+---
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: cardano-standard-rwo
+provisioner: kubernetes.io/gce-pd
+parameters:
+  type: pd-standard
+  fstype: ext4
+  replication-type: none
+
+# Relay ###############################################################################################################
+
+# Relay Volume Claim
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: relay-data
+  labels:
+    app: cardano
+    pvc: relay-data
+spec:
+  storageClassName: cardano-standard-rwo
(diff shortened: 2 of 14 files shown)


nix/docker/k8s/cardano-nodes.yaml
```diff
@@ -0,0 +1,190 @@
...
+  resources:
+    requests:
+      storage: 10G
+
+# Block Producer Pod
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: bprod
```
Preskton, line 139: For the names of various objects, in most cases, Kubernetes allows for longer names. You may want to consider fully spelling this out as `block-producer` for clarity and readability for people who aren't as familiar with Cardano's architecture.

nix/docker/k8s/cardano-nodes.yaml
```diff
@@ -0,0 +1,190 @@
...
+  - name: tcp
+    port: 3001
+    nodePort: 30010
+
+# Relay TCP Service
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: relay-clip
```
Preskton, line 104: What's the significance/meaning of `-clip`? I'm new to Cardano, so I'm not sure if this is a piece of the architecture I missed, or something I need to read up on more...?

nix/docker/k8s/cardano-nodes.yaml
```diff
@@ -0,0 +1,190 @@
...
+  labels:
+    app: cardano
+    pod: relay
+spec:
+  containers:
+  - name: relay
+    image: inputoutput/cardano-node:dev
+    imagePullPolicy: Always
+    env:
+    - name: CARDANO_PUBLIC_IP
```
Preskton, line 51: You may consider using `envFrom` (https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#configure-all-key-value-pairs-in-a-configmap-as-container-environment-variables) to take all of the keys from a given config map and configure them as env vars to reduce repetition/improve readability.

nix/docker/k8s/cardano-nodes.yaml
```diff
@@ -0,0 +1,190 @@
...
+  name: relay-data
+  labels:
+    app: cardano
+    pvc: relay-data
+spec:
+  storageClassName: cardano-standard-rwo
+  accessModes: [ ReadWriteOnce ]
+  resources:
+    requests:
+      storage: 10G
```
Preskton, line 34: Curious how `10G` was chosen - should this align to the "Stake Pool minimum system requirements" of `30G` -- https://iohk.zendesk.com/hc/en-us/articles/900001208966-Stake-Pool-Minimum-System-Requirements?

nix/docker/k8s/cardano-nodes.yaml
```diff
@@ -0,0 +1,190 @@
...
+    app: cardano
+spec:
+  type: NodePort
+  selector:
+    app: cardano
+    pod: relay
+  ports:
+  - name: tcp
+    port: 3001
+    nodePort: 30010
```
Preskton, line 97: I'm having trouble finding documentation on the ports and what needs to be exposed and how stateful the connections need to be. Do you have a suggestion/chunk of code I could look at? Trying to determine if we absolutely need NodePort which limits scaling w/i a node, or if there's a way we could expose these via a load balancer + ingress.

Preskton (commented):
Not a maintainer, so not explicitly approving - but overall this looks like a nice start.

A couple of questions sprinkled throughout, mostly because I don't really understand the Cardano architecture/statefulness model yet.

Overview of comments:

- Prefer using the "full name" for things for readability, discoverability, and approachability.
- Consider not specifying specific `clusterIP` for portability across clusters that might have different CIDR ranges internally.
- Few other minor questions.

nix/docker/k8s/cardano-nodes.yaml
```diff
@@ -0,0 +1,190 @@
...
+# kubectl create namespace $NAMESPACE
+#
+
+# GCE Storage Class
+---
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: cardano-standard-rwo
+provisioner: kubernetes.io/gce-pd
```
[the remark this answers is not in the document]
tdiesler: Yes, this is the first cut designed for entry as easy as possible. The full docs are [here](https://tdiesler.gitbook.io/cardano/v/iohk/k8s/cardano-k8s). For now, users can spin up a cluster like this ...

```
$ gcloud container clusters create $CLUSTER_NAME \
  --disk-size=32GB \
  --machine-type=e2-medium \
  --node-locations=$CLUSTER_ZONE \
  --zone=$CLUSTER_ZONE \
  --num-nodes=2
```

To enable portability, which I agree we should in the future, we'd have to provide a ton of docs that address how to setup k8s for various providers and their specifics around storage, firewall, etc.

For now, I'd suggest we stick with GKE until this has matured enough, so that we feel comfortable with adding an additional layer of complexity (i.e. script portability)

nix/docker/k8s/cardano-nodes.yaml
```diff
@@ -0,0 +1,190 @@
...
+# Relay TCP Service
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: relay-clip
+  labels:
+    app: cardano
+spec:
+  clusterIP: 10.3.240.100
```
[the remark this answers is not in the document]
tdiesler: Again, portability to other clusters is considered out of scope in the context of this initial PR. 

For the networking perspective, we want the Relay to be reachable for external ingress (i.e. via NodePort 30010). In the topology for the relay, we point to the internal network IP of the block producer. In the topology for the block producer, we point to the internal network IP of the relay (i.e. what you see here)

nix/docker/k8s/cardano-nodes.yaml
tdiesler: Yes, we can do that. How about a PR in the incubator space [over here](https://github.com/tdiesler/nessus-cardano/blob/master/nix/docker/k8s/cardano-nodes.yaml#L139).

This is where I try things out before I dare to create a PR with this project.

nix/docker/k8s/cardano-nodes.yaml
tdiesler: The ClusterIP for the relay

nix/docker/k8s/cardano-nodes.yaml
tdiesler: This also needs to work for [plain Docker and Docker Compose](https://tdiesler.gitbook.io/cardano/v/iohk/)

nix/docker/k8s/cardano-nodes.yaml
tdiesler: The Cardano block store currently needs ...

```
 $ du -h /mnt/disks/data00/
392M    /mnt/disks/data00/ledger
6.2G    /mnt/disks/data00/immutable
6.6G    /mnt/disks/data00/
```

It is set to 10G to safe cost for these initial test clusters. It will probably be enough for quite a bit longer

nix/docker/k8s/cardano-nodes.yaml
tdiesler: Currently, there is a (central) topology updating service, which every node has to call exactly once per hour. A node is accepted after three successful calls with the same IP. Therefore, this IP has to be stable or be targeted by some stable DNS name.

This will change in the near future with Alonzo when Cardano nodes will be able to find their peers on their own (i.e. without the help of an external topology update service).

tdiesler: Much of what you see here can probably be improved quite considerably by using a StatefulSet instead of individual Pods. This is what I'm currently looking at over at [nessus-cardano](https://github.com/tdiesler/nessus-cardano).

As for now, I'm not sure whether IOHK appreciates this work let alone plans to merge it upstream eventually. Your comments are the first piece of technical feedback that I got - and thanks a million for that.

Yes, please come over to  nessus-cardano to provide more feedback and help with improved versions of this. That project is not meant to be a fork of the iohk image, but rather an incubating space for future PRs like this.

The nessusio/cardano-node images are [here](https://hub.docker.com/r/nessusio/cardano-node). They have seen quite a bit of adoption already and hopefully can see a lot less when this tech becomes available here.

nix/docker/context/bin/topologyUpdate
```diff
@@ -0,0 +1,174 @@
...
+# IPv4 address only please add the -4 parameter to the curl command below  (curl -4 -s ...)
+pushNodeInfo () {
+
+  if [[ ${TU_PUSH} = "Y" ]]; then
+
+      echo "[`date +'%Y-%m-%d %H:%M:%S %Z'`] Updating topology ..."
+
+      fetchBlockNumber
+
+      echo "curl -s -f https://api.clio.one/htopology/v1/?port=${CNODE_PORT}&blockNo=${blockNo}&valency=${CNODE_VALENCY}&magic=${NWMAGIC}${T_HOSTNAME}"
```
disassembler, line 115: Can you make this an env variable please? We can't have the default docker image connect to a 3rd party service that the CF or IOHK doesn't control.

nix/docker/k8s/cardano-nodes.yaml
```diff
@@ -0,0 +1,190 @@
...
+      valueFrom:
+        configMapKeyRef:
+          name: relaycfg
+          key:  topology
+    - name: CARDANO_CUSTOM_PEERS
+      valueFrom:
+        configMapKeyRef:
+          name: relaycfg
+          key:  customPeers
+    - name: CARDANO_UPDATE_TOPOLOGY
```
disassembler, line 66: please default this to false throughout the examples

disassembler (changes requested):
I'll have @jbgi look at this in further detail. But primarily we can't link to a service IOHK or the CF doesn't control in the docker image (topology updater) and all examples should default to it being disabled. Aside from that, this improvement looks great.

nix/docker/context/bin/topologyUpdate
tdiesler: Thanks, for looking at this @disassembler. The topology update is disabled by default and can be [enabled like this](https://tdiesler.gitbook.io/cardano/v/iohk/plain-docker/running-the-nodes#topology-updater)

Do you mean to have `api.clio.one/htopology/v1` configurable as separate env var?

tdiesler:
@Preskton I sketched out some K8S options in [this document](https://docs.google.com/document/d/1JF6nAWk3D0zNWu8npzPIZ_3u2v18v1AiC3qWvFF_-g4/edit?usp=sharing). I'll ask a few other folks who are interested in this this topic to comment as well. Perhaps, we can come to a common understand on how to take this further (and even learn from each other). Cheers.

nix/docker/k8s/cardano-nodes.yaml
tdiesler: ok

Preskton:
@tdiesler I took a look at the `StatefulSet` version earlier in the week, and it looked like a great start. Started making some comments in the gdocs. Again, great start - looking forward to learning more about the architecture so I can better contribute! Thanks for kicking this off.

tdiesler:
@disassembler good morning. Is there anything else you'd like me to do with this?

nix/docker/k8s/cardano-nodes.yaml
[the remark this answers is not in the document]
Asafffff: Hey, I'm new to this but wanted to help. Sorry if I'm not understanding some broader scope, but, if it is all internal communication, why using an IP instead of a service name? This is also the case for the `docker-compose` file. Docker / K8S can translate the service name to the matching IP address, without having to manage it explicitly.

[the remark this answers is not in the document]
tdiesler: It's communication across nodes on a private network. With StatefulSets we have the hostname. Indeed, the clusterIP thing shown here is out of date.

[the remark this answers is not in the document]
Asafffff: No problem, just wanted to confirm that it should also be fixed in the docker-compose file. GJ :)

PyurOfHeart:
Hi, thank you for your contribution. Having a docker image and ready-to-deploy kubernetes manifests is a great leap forward for Cardano -- it will allow people to spin up their own cardano stake pool with much less effort and expertise than they need today.

We run our pool on Kubernetes, and there were a few problems we had to solve that overlap with what's happening in this PR. One particular issue that we had to solve -- and you may care about -- is running rootless containers. What I mean by that is ensuring that every single container running in your kubernetes cluster does so without root privileges.

Containers are a great way to isolate workloads from each other, but a lot of those benefits are put at risk when they run as root, because being root inside the container allows one to become root on the underlying hosts. This opens up the cluster itself to vulnerabilities via malware originating from the workloads running within them.

The best-practice way to solve this is to run containers as non-root users, and to set up pod security policies in kubernetes that prevent privileged containers from running and specifying run-as users for each pod. This way, if you happen to have a single compromised container, the blast radius is most likely limited to that container. This can have additional benefits, such as making all binaries read/execute-only for the container user so that malicious software on the host can not alter the binaries.

I do realize this is an initial PR, but I wanted to call this out here because some of the patterns used here simply do not work in a non-root container. The major one is the use of cron as a daemon within the container: Daemons have to be started by the root user, and if the container runs as an unprivileged user it can not start cron. This interferes with the ability to do automated topology updates and other scheduled tasks.

In our containers we overcame this problem by using supervisord. Supervisor is PID 1 in our containers and launches cardano-node and other processes as its children. We use the pattern described [here](https://advancedweb.hu/supervisor-with-docker-lessons-learned/#simple-scheduling) to schedule things like topology updates.

I do see that you're tracking non-root images as an enhancement in your playground project (https://github.com/tdiesler/nessus-cardano/issues/33) but this is such an impactful change that if it's not there from the beginning, it will be very challenging to migrate to it later.

I'm of no authority here so I just wanted to chime in and share what I learned from my experience managing a containerized pool running on Kubernetes. I think supporting rootless images from the get-go will be a major step in protecting the stake pools that will utilize this infrastructure.

Preskton:
@PyurOfHeart would you be up for submitting a PR as well or pointing us to an example of how y'all have laid things out? I've been familiarizing myself with the node architecture so that I can be of help.

One thing I'm trying to understand better/wrap my head around is why introduce `supervisord` rather than dedicate a container within the pod for topology updates. Or if the output from the update can be shared - could this instead get written to a common PV/configmap and then mounted accordingly?

tdiesler:
Great feedback thanks.

> such an impactful change ...

I'll see if I can get it into the 1.26.0 release in time

PyurOfHeart:
> One thing I'm trying to understand better/wrap my head around is why introduce `supervisord` rather than dedicate a container within the pod for topology updates. Or if the output from the update can be shared - could this instead get written to a common PV/configmap and then mounted accordingly?

Happy to explain!

The topology update process that is used in this PR is based on a service some folks are hosting online. The service facilitates discovery between various relays in cardano — the nodes that the P2P network is composed of. This service has two functions: a push and a pull. 

The pull process is very simple: You hit the API asking for peers, it gives you a JSON you can stick into your topology file. This could very well be done in a separate container  (eg a CronJob) in kubernetes if it weren’t for the API’s vetting mechanism. (More on that below)

The push process is a bit more complicated: You send a request to the API telling it where the tip of the blockchain is, so that it knows the request originates from a healthy, well-synced relay. It wants you to send this information to it every hour before it will allow you to use the pull API, and before it will start advertising you as a peer to others. Because getting the tip requires access to a local relay, this whole thing has to happen inside the relay container. 

The pull API will not let you retrieve peers unless you’re a client that is also doing the “push” operation and advertising yourself as a healthy relay. It somewhat assumes you’re using a standard non-k8s deployment, in that when you do the “push” operation it looks at the source IP of the request to determine who you are. When you send a “pull” request, it won’t give you peers unless you’re calling from that same IP. 

You can see where this gets tricky with kubernetes: If you run in a private network with a NAT gateway for egress traffic, there’s no way to use this mechanism to update your topology. If you run in a public network where each node of your cluster has a public IP, then you need to open up a NodePort Service in your cluster so that requests can be received by your relays at the IP address the API sees them at. 

This whole requirement around the IPs for each API having to match means that you really want to make sure that you’re making the “push” request from the same IP address (host) as the one you’re pulling from, and the fact that “push” requires access to a live relay every hour means that you pretty much need to do this inside the same container as your relays. There are a few ways to cheat the push requirement (publish the tip of the blockchain somewhere and then retrieve it from a separate pod, or use one of the website keeping track of the tip, etc) but we chose not to invest in those workaround in our pool.

nix/docker/k8s/cardano-nodes.yaml
PyurOfHeart: In the cardano-node release notes 20G is listed as the “required” free disk space for a pool. Those requirements are usually over-zealous, but disk is cheap and having some headroom would be preferable to running out of disk space soon 🙂

tdiesler:
I don't think we need to worry too much about topology updater, this is now being phased out and replaced by the P2P Manager in Alonzo.

tdiesler:
Superseded by #2608

PR #2459, opened 2021-03-09
WIP Fork at support for more eras

newhoggy:


cardano-node-chairman/src/Testnet/ByronShelley.hs
```diff
@@ -91,7 +93,9 @@ defaultTestnetOptions = TestnetOptions
   , numPoolNodes = 1
   , activeSlotsCoeff = 0.1
   , epochLength = 1500
-  , forkPoint = AtVersion 1
+  , forkPointShelley = AtVersion 1
+  , forkPointAllegra = AtVersion 1
+  , forkPointMary = AtVersion 1
```
yihuang, line 98: It won't transition from byron to mary at the same time if we set to the same version or epoch right?
To setup a mary testnet as fast as possible, the best I can do is(https://github.com/yihuang/cardano-devnet/blob/master/mkfiles.sh#L89):
```
TestShelleyHardForkAtEpoch: 1
TestAllegraHardForkAtEpoch: 2
TestMaryHardForkAtEpoch: 3
```

newhoggy: Thanks for the hint.

I'll give that a try.

yihuang:
```
TestShelleyHardForkAtEpoch: 0
TestAllegraHardForkAtEpoch: 0
TestMaryHardForkAtEpoch: 0
```
This config seems to work.

PR #2464, opened 2021-03-09
Re-export chain sync client definitions

DavidEichmann:
This allows me to remove dependencies on ouroboros-network and protocols in the cardano-client-demo package.

### DIFF
--- a/cardano-api/cardano-api.cabal
+++ b/cardano-api/cardano-api.cabal
@@ -18,6 +18,8 @@ library
   exposed-modules:      Cardano.Api
                         Cardano.Api.Byron
                         Cardano.Api.Shelley
+                        Cardano.Api.ChainSync.Client
+                        Cardano.Api.ChainSync.ClientPipelined
 
                         Cardano.Api.Crypto.Ed25519Bip32
                         -- TODO: Eliminate in the future when
@@ -114,6 +116,7 @@ library
                       , time
                       , transformers
                       , transformers-except
+                      , typed-protocols
                       , unordered-containers >= 0.2.11
                       , vector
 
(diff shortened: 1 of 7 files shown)


cardano-api/src/Cardano/Api/Client.hs
```diff
@@ -0,0 +1,17 @@
...
+    , ClientStIntersect(..)
+
+      -- * Null chain sync client
+    , chainSyncClientNull
+
+      -- * Utilities
+    , mapChainSyncClient
+    ) where
+
+import Ouroboros.Network.Protocol.ChainSync.Client
```
DavidEichmann, line 17: Add newline

DavidEichmann:
I'll try to summarize chats with @dcoutts and @Jimbo4350:

*  [x] We do want to export enough from `cardano-api` such that we need only depend on `cardano-api` e.g. in the demos. So the new exports are desired, but the question is from which module(s) should they be exported.
* [ ] There are name clashes between `Ouroboros.Network.Protocol.ChainSync.Client` and `Ouroboros.Network.Protocol.ChainSync.ClientPipelined` e.g. the constructor name `SendMsgRequestNext` exists for `ClientStIdle` and for `ClientPipelinedStIdle`. This means we must separate the exports into different modules
* [ ] As there are 2 other client-node protocols, the module name `Cardano.Api.Client` is not really specific enough. I should change this to `Cardano.Api.ChainSync.Client` and likewise for pipelined. 
* [ ] I'll look into exporting the other 2 client-node protocols too. If easy enough, I'll add it to this PR.

DavidEichmann:
I think I'll leave the `TxSubmission` and `StateQuery` stuff for followup PRs. No need to cram it all into this one. I'm waiting for comment now.

dcoutts (approved):
LGTM!

I notice in the demo we're also having to
```
import Cardano.Slotting.Slot
import Cardano.Chain.Slotting
```
which suggests targets for API warts to excise in future.

DavidEichmann:
bors merge

PR #2465, opened 2021-03-09
More friendly transaction printing

cblp:


cardano-cli/src/Cardano/CLI/Shelley/Run/Pretty.hs
```diff
@@ -1,107 +1,162 @@
...
 import qualified Shelley.Spec.Ledger.API as Shelley
 
-prettyTx :: TxBody era -> LByteString
-prettyTx body0 =
-  encodePretty' defConfig{confCompare = compare} $
-  object $
-  case body0 of
-    ByronTxBody tx -> ["era" .= ("Byron" :: Text), "tx" .= tx]
+-- | User-friendly text representation
+class FriendlyJson a where
```
Jimbo4350, line 32: What is the justification for creating the `FriendlyJson` typeclass? I think it would make more sense to extend `prettyTx` to suit our needs before creating another typeclass.

cblp: I had a fight with the compiler on type families and GADT matching, but after you pointed, I figured a way to make it without an extra class.

But there are many types in consideration here, and there will be more soon. Typeclasses are a safer way to dispatch on a type than a set of functions.

Jimbo4350: > But there are many types in consideration here, and there will be more soon
When there are more types to consider we can revisit implementing a class. I think for now its OK to rely on existing `ToJSON` instances and `encodePretty` https://hackage.haskell.org/package/aeson-pretty-0.8.8/docs/Data-Aeson-Encode-Pretty.html

cblp: Yes, OK.

cardano-cli/src/Cardano/CLI/Shelley/Run/Pretty.hs
```diff
@@ -1,107 +1,180 @@
...
-          mint =
-            body
+friendlyTxBodyBS :: Api.TxBody era -> ByteString
+friendlyTxBodyBS =
+  encodePretty (setConfCompare compare defConfig) . friendlyTxBody
+
+friendlyTxBody :: Api.TxBody era -> Value
+friendlyTxBody = \case
+  ByronTxBody tx ->
+    _Object (HashMap.insert "era" "Byron") $ friendlyTxBodyByron tx
```
Jimbo4350, line 42: I think a more concise was of doing this would be:
```
mergeEra :: CardanoEra era -> JSON.Value -> JSON.Value
mergeEra era (JSON.Object obj) =
  case toJSON era of 
    JSON.Object e -> JSON.Object $ e <> obj
    _ -> JSON.Object obj
mergeEra _   _  = JSON.Null
```
And defining a suitable `ToJSON` instance for `CardanoEra era` i.e
```
toJSON ByronEra = object [ "era" .= Aeson.String "ByronEra" ]
```
etc

Have a look at https://github.com/input-output-hk/cardano-node/pull/2465/files#diff-cc9b567f3acd813c81e105c074ebd783719c74364c88a78c3c22638322354f6dR789
You will notice that we have access to `CardanoEra era` (`_era`). We can use this to simplify the pattern matching.

cardano-cli/src/Cardano/CLI/Shelley/Run/Pretty.hs
```diff
@@ -1,107 +1,180 @@
...
+import           Cardano.Binary (Annotated)
 import           Cardano.CLI.Helpers (textShow)
+import qualified Cardano.Chain.UTxO as Byron
+import           Cardano.Ledger.Shelley as Ledger (ShelleyEra)
+import           Cardano.Ledger.ShelleyMA (MaryOrAllegra (Allegra, Mary), ShelleyMAEra)
 import qualified Cardano.Ledger.ShelleyMA.TxBody as ShelleyMA
-import           Cardano.Prelude
-import           Data.Aeson as JSON (Value, object, (.=))
-import           Data.Aeson.Encode.Pretty (Config (confCompare), defConfig, encodePretty')
+import           Data.Aeson as JSON (Object, Value (..), object, toJSON, (.=))
```
Jimbo4350, line 28: When qualifying and import, be sure to search the repo to see how its usually qualified. In this instance we usually qualify `Data.Aeson` as `Aeson`

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -290,17 +291,17 @@ instance IsCardanoEra era => ToJSON (TxOut era) where
       ShelleyAddressInEra sbe ->
         case sbe of
           ShelleyBasedEraShelley ->
-            let hexAddr = serialiseToRawBytesHexText addr
+            let hexAddr = serialiseToBech32 addr
```
Jimbo4350, line 294: :+1:

erikd: Since is now Bech32` we might want to rename the variable from `hexAddr`

cardano-cli/src/Cardano/CLI/Shelley/Run/Pretty.hs
cblp: The user-friendly name for the Byron era is "Byron", not "ByronEra".

I'm not sure this is concise, but I rewrote it a bit differently, please see.

cardano-api/src/Cardano/Api/TxBody.hs
cblp: done

Jimbo4350:
@cblp Have a look at this commit: https://github.com/input-output-hk/cardano-node/commit/998421ce55df3901aaf60e6a2fc810cdd2753638

cardano-api/src/Cardano/Api/Eras.hs
```diff
@@ -137,6 +138,12 @@ deriving instance Eq   (CardanoEra era)
 deriving instance Ord  (CardanoEra era)
 deriving instance Show (CardanoEra era)
 
+instance ToJSON (CardanoEra era) where
+   toJSON ByronEra   = "Byron"
```
Jimbo4350, line 142: :+1:

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -292,20 +293,20 @@ instance IsCardanoEra era => ToJSON (TxOut era) where
       ShelleyAddressInEra sbe ->
         case sbe of
           ShelleyBasedEraShelley ->
-            let hexAddr = serialiseToRawBytesHexText addr
-            in object [ "address" .= hexAddr
-                      , "value" .= toJSON val
-                      ]
+            object
+              [ "address" .= serialiseToBech32 addr
```
Jimbo4350, line 297: :+1:

.github/workflows/github-page.yml
```diff
@@ -70,6 +65,11 @@ jobs:
     - name: Cabal Configure
       run: cabal configure --builddir="$CABAL_BUILDDIR" --write-ghc-environment-files=always
 
+    - name: Use cabal.project.local.github-pages
```
Jimbo4350, line 68: Why this move?

cardano-cli/src/Cardano/CLI/Shelley/Run/Pretty.hs
```diff
@@ -1,107 +1,166 @@
...
+              , "credential" .= cred
+              , "stake reference" .= textShow ref
+              , "Bech32" .= addressBech32
+              ]
+        ]
+      AddrBootstrap _ ->
+        ["bootstrap address" .= object ["Bech32" .= String addressBech32]]
+  where
+    addressBech32 =
+      case fromShelleyAddr @Api.ShelleyEra addr of
```
Jimbo4350, line 159: We generally try not to use `TypeApplication`s and prefer to pass type level information via values. This is ok for now but in future please avoid `TypeApplication`s in the cli.

cardano-cli/src/Cardano/CLI/Shelley/Run/Pretty.hs
```diff
@@ -1,107 +1,166 @@
...
+        AddressInEra (ShelleyAddressInEra _) a -> serialiseToBech32 a
+        AddressInEra ByronAddressInAnyEra _ -> panic "expected Shelley address"
+
+assertObject :: Value -> Object
+assertObject = \case
+  Object obj -> obj
+  _ -> panic "expected JSON Object"
```
Jimbo4350, line 166: It would be more useful to show the incorrect `Value`
`panic $ "Expected JSON Object but got: " <> show incorrect`

cardano-cli/test/Test/Golden/TxView.hs
```diff
@@ -0,0 +1,179 @@
...
+        , "--mary-era"
+        , "--tx-in"
+        ,   "fe5dd07fb576bff960d6e066eade5b26cdb5afebe29f76ea58d0a098bce5d891\
+            \#135"
+        , "--tx-out"
+        ,   "addr_test1qrefnr4k09pvge6dq83v6s67ruter8sftmky8qrmkqqsxy7q5psgn8tgqmupq4r79jmxlyk4eqt6z6hj5g8jd8393msqaw47f4\
+            \+138"
+        , "--fee", "139"
+        , "--invalid-before", "140"
+        -- , "--mint", "141" -- TODO(cblp, 2021-03-19) Transaction cannot mint ada, only non-ada assets
```
Jimbo4350, line 142: Transactions in the Mary era can mint multi-assets. Is the plan to pretty print multi assets in the future? Or did you intend to implement this in the current PR?

cardano-cli/src/Cardano/CLI/Shelley/Run/Pretty.hs
```diff
@@ -1,107 +1,166 @@
...
+              , "Bech32" .= addressBech32
+              ]
+        ]
+      AddrBootstrap _ ->
+        ["bootstrap address" .= object ["Bech32" .= String addressBech32]]
+  where
+    addressBech32 =
+      case fromShelleyAddr @Api.ShelleyEra addr of
+        AddressInEra (ShelleyAddressInEra _) a -> serialiseToBech32 a
+        AddressInEra ByronAddressInAnyEra _ -> panic "expected Shelley address"
```
Jimbo4350, line 161: You should use `serialiseAddress` in both cases (see `SerialiseAddress`) . Byron addresses still exist in the Shelley era. We still need to be able to render them.

Jimbo4350 (commented):
Much better, a few minor fixes and then I will approve.

.github/workflows/github-page.yml
newhoggy: Ah, yes.  This is correct.  Unfortunately `cabal configure` clobbers `cabal.project.local` so this has to go after `cabal.project.local`.

cblp: Yes. One can see the effect in https://github.com/input-output-hk/cardano-node/runs/2173237313

cardano-cli/src/Cardano/CLI/Shelley/Run/Pretty.hs
cblp: But why? TypeApplications make code both simple and easy.

cardano-cli/src/Cardano/CLI/Shelley/Run/Pretty.hs
cblp: The value may be big. Is it enough to print JSON-type only? (I mean Object, or Array, or Number...)

Jimbo4350: Yep that would make sense :+1:

cardano-api/src/Cardano/Api/Eras.hs
gitmachtl: This would be the perfect output also for the query tip or query protocol-parameters about the current era the node is in right now.

cardano-cli/src/Cardano/CLI/Shelley/Run/Pretty.hs
cblp: It must be done separately. I've created #2533.

cardano-cli/test/Test/Golden/TxView.hs
cblp: It must be done separately. I've created #2533.

cardano-api/src/Cardano/Api/Eras.hs
cblp: @gitmachtl Could you please open an issue for that?

gitmachtl: > @gitmachtl Could you please open an issue for that?

https://github.com/input-output-hk/cardano-node/issues/2509

already there as a FR

Jimbo4350:
@cblp After you squash the commits I will give this one final look over.

cblp:
@Jimbo4350 squashed

cardano-cli/src/Cardano/CLI/Shelley/Run/Pretty.hs
```diff
@@ -1,107 +1,175 @@
...
+import           Cardano.Binary (Annotated)
 import           Cardano.CLI.Helpers (textShow)
+import qualified Cardano.Chain.UTxO as Byron
+import           Cardano.Ledger.Shelley as Ledger (ShelleyEra)
+import           Cardano.Ledger.ShelleyMA (MaryOrAllegra (Allegra, Mary), ShelleyMAEra)
 import qualified Cardano.Ledger.ShelleyMA.TxBody as ShelleyMA
-import           Cardano.Prelude
-import           Data.Aeson as JSON (Value, object, (.=))
-import           Data.Aeson.Encode.Pretty (Config (confCompare), defConfig, encodePretty')
+import           Data.Aeson (Object, Value (..), object, toJSON, (.=))
```
Jimbo4350, line 27: Have a look at: https://github.com/input-output-hk/cardano-node/blob/master/cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs#L20

We tend to separate hackage packages from local packages.

Jimbo4350 (approved):
LGTM! Please adjust the imports  (see comments) and then you are free to bors :smile:

cardano-cli/src/Cardano/CLI/Shelley/Run/Pretty.hs
cblp: Oops, my bad. Sure.

Jimbo4350 (approved):
LGTM!

cblp:
bors merge

PR #2466, opened 2021-03-09
CAD-2693 supervisord:  port topology generation from cardano-ops

deepfire:
This implements dynamic topology generation for `supervisord` local cluster profiles.

### DIFF
--- a/nix/supervisord-cluster/supervisor.nix
+++ b/nix/supervisord-cluster/supervisor.nix
@@ -23,11 +23,15 @@ let
       extra = {
         services.cardano-node = {
           enable = true;
-          inherit (envConfig) operationalCertificate kesKey topology vrfKey nodeConfig nodeConfigFile port dbPrefix socketPath;
+          inherit (envConfig) topology nodeConfig nodeConfigFile port dbPrefix socketPath;
           inherit stateDir;
-        } // lib.optionalAttrs useCabalRun
-          { executable = "cabal run exe:cardano-node --";
-          };
+        }
+        // lib.optionalAttrs (__hasAttr "vrfKey" envConfig)
+        { inherit (envConfig) operationalCertificate kesKey vrfKey;
+        }
+        // lib.optionalAttrs useCabalRun
+        { executable = "cabal run exe:cardano-node --";
+        };
       };
     in lib.evalModules {
       prefix = [];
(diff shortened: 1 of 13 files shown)


nix/supervisord-cluster/supervisor.nix
```diff
@@ -23,12 +23,15 @@ let
...
           inherit stateDir;
-        } // lib.optionalAttrs useCabalRun
-          { executable = "cabal run exe:cardano-node --";
-          };
+        }
+        // lib.optionalAttrs (__hasAttr "vrfKey" envConfig)
+        { inherit (envConfig) operationalCertificate kesKey vrfKey;
+        }
+        // lib.optionalAttrs useCabalRun
+        { executable = "cabal run exe:cardano-node --";
```
MarcFontaine, line 33: Is it not the case that cabal and nix are somehow orthogonal ?
Why is the executable called with 'cabal run exe:cardano-node' ?
I would have expected that the executable is defined in a nix-derivation ?

nix/supervisord-cluster/topology.nix
```diff
@@ -0,0 +1,106 @@
...
+            ( . *
+              { pools:
+                (if .pools == null then 0 else
+                 if .pools == 1    then 1 else
+                    ([$density, 1] | max) end end)
+              }
+            )
+          )
+        };
+
```
MarcFontaine, line 55: The is some Haskell code to generate the topology.
Why is the output of the Haskell tool patched with jq ?
Is it not better to do everything in Haskell ?

nix/supervisord-cluster/topology/cardano-topology.hs
```diff
@@ -0,0 +1,247 @@
+#! /usr/bin/env nix-shell
+#! nix-shell -p "haskellPackages.ghcWithPackages (pkgs: with pkgs; [aeson graphviz optparse-applicative split])" -i runhaskell
+
```
MarcFontaine, line 3: Why do we have this code as she-bang Haskell script ?
Why not just a regular executable ?

MarcFontaine (commented):
To me it looks like everything is mixed up:
Haskell, nix, shell-scripts and jq are all used in a non trivial way.
It looks like all four languages are used to mangle the final topology file.
Is it not possible to have some kind of separation of concerns ? 
This looks like a nightmare to maintain

nix/supervisord-cluster/supervisor.nix
deepfire: First, that's not the new part here -- it's just shuffling code around.  This code was introduced in the previous PR.

Second, the idea is that we're reusing the `cardano-node` service definition here -- precisely to be able to use the Nix infrastructure to give use all the dependencies, but _also_ use `cabal build/run` to get the "last mile" -- the node itself.

nix/supervisord-cluster/topology.nix
deepfire: The short answer is that this specific format does not exist in isolation.

The topology as generated by the haskell tool can be consumed by the AWS deployment infra, without any changes -- and that's what is valuable here.

nix/supervisord-cluster/topology/cardano-topology.hs
deepfire: Good point -- that's no longer necessary.  It was used like this in the `cardano-ops` deployment, but now that I had to introduce the cabal thing, that's obsolete.

Good catch!

deepfire:
@MarcFontaine, you're looking at one small piece -- the important thing is that it's a step towards unified cluster configuration.

Once we have that major step done, we can go after making things pretty.

Think of this as a way to get rid of the `shelley3pools` nightmare -- but also getting all the benefits of:

- same benchmark profile specification that we enjoy on AWS (actually even improving on that)
- reusing the same Nix `cardano-node.service` Nix configuration
- reusing the same profile-driven genesis generation
- setting the stage to using the same log analysis pipeline
- etc.

Yes, it's not pretty just yet.

Baby steps!

MarcFontaine (approved):

LaurenceIO (approved):

PR #2468, opened 2021-03-10
Test support for running CLI commands with environment variables

newhoggy:
This is necessary for running cardano-cli commands that need a socket file.


erikd (approved):

newhoggy:
bors merge

PR #2481, opened 2021-03-11
CAD-2746 tracers:  add the local mux tracer

deepfire:
Local socket mux used to be `nullTracer`.

This adds a more useful one, that is also off by default.


karknu (approved):
LGTM

Jimbo4350 (approved):

PR #2486, opened 2021-03-15
Merge script address build command into address build command

newhoggy:
Addresses this ticket: https://github.com/input-output-hk/cardano-node/issues/2333


newhoggy:
Help text for the `address build` command:

```
cabal exec -- cardano-cli address build
Usage: cardano-cli address build (--payment-verification-key STRING |
                                   --payment-verification-key-file FILE |
                                   --script-file FILE)
                                 [--stake-verification-key STRING |
                                   --stake-verification-key-file FILE]
                                 [--out-file FILE]
  Build a Shelley payment address, with optional delegation to a stake address.

Available options:
  --payment-verification-key STRING
                           Payment verification key (Bech32-encoded)
  --payment-verification-key-file FILE
                           Filepath of the payment verification key.
  --script-file FILE       Filepath of the script.
  --stake-verification-key STRING
                           Stake verification key (Bech32 or hex-encoded).
  --stake-verification-key-file FILE
                           Filepath of the staking verification key.
```

cardano-cli/src/Cardano/CLI/Shelley/Key.hs
```diff
@@ -381,6 +383,11 @@ readVerificationKeyOrTextEnvFile asType verKeyOrFile =
     VerificationKeyFilePath (VerificationKeyFile fp) ->
       readKeyFileTextEnvelope (AsVerificationKey asType) fp
 
+data PaymentSource
```
Jimbo4350, line 386: Let's move this to `cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs`

cardano-cli/src/Cardano/CLI/Shelley/Run/Address.hs
```diff
@@ -213,25 +224,4 @@ runAddressBuildScript
...
-                   --TODO: add support for referring to stake addresses
-
-      scriptAddrText :: Text
-      scriptAddrText = serialiseAddress scriptAddr
-
-  liftIO $ case mOutFp of
-    Just (OutputFile oFp) -> Text.writeFile oFp scriptAddrText
-    Nothing               -> Text.putStr        scriptAddrText
-
+runAddressBuildScript scriptFile = runAddressBuild (SourcePaymentScript scriptFile) Nothing
```
Jimbo4350, line 227: I'm wondering if we also print a deprecation warning here as well.

Jimbo4350 (approved):

cardano-cli/src/Cardano/CLI/Shelley/Run/Address.hs
newhoggy: Do we have an example of a deprecation warning?

cardano-cli/src/Cardano/CLI/Shelley/Key.hs
newhoggy: Unfortunately, I can't move it because `VerificationKeyTextOrFile` is the type of the constructor argument which is defined in this file.

newhoggy: Unfortunately I can't move it to `Parsers` because `AddressCmd` definition in `Commands` needs it.  Moving the type into `Parsers` would create a cyclic import dependency between `Parsers` and `Commands`.

cardano-cli/src/Cardano/CLI/Shelley/Run/Address.hs
Jimbo4350: https://github.com/input-output-hk/cardano-node/blob/0dac653ca71133abc7a5cc01d31d72aabba3d9d9/cardano-cli/src/Cardano/CLI/Parsers.hs#L75

Jimbo4350: Or https://github.com/input-output-hk/cardano-node/blob/94f79cc9d637af569836cd30828681875c2f986a/cardano-api/src/Cardano/Api/Eras.hs#L95

newhoggy:
bors merge

PR #2492, opened 2021-03-16
fix IPv4/v6 use message

gufmar:
minor message fix, to prevent mistakes when users consider the "failed to parse IPvX" console error message

### DIFF
--- a/cardano-node/src/Cardano/Node/Parsers.hs
+++ b/cardano-node/src/Cardano/Node/Parsers.hs
@@ -116,7 +116,7 @@ parseNodeHostIPv4Address str =
   maybe
     (Left $
       "Failed to parse: " ++ str ++
-      ". Use --host-ipv6-addr to specify IPv6 adddress")
+      ". Use --host-addr to specify IPv4 adddress")
     (Right . NodeHostIPv4Address)
     (readMaybe str)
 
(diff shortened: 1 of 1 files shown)


karknu (commented):
LGTM.

nc6:
I think this goes against the point of the message, which is to refer the user to the IPv6 syntax if they accidentally use the IPv4 one and vice verse

gufmar:
I see. from this point of view it makes sense. 
If one has a typo/mistake in the address, without using the wrong ip-type the message is a bit misleading. 
May say: "_If you want to use an IPv4 address, please use ..._" ?

Jimbo4350:
@gufmar 
>I think this goes against the point of the message, which is to refer the user to the IPv6 syntax if they accidentally use the IPv4 one and vice verse
I believe this was the original intention @cblp

cblp:
Yes, the current message suggests _another_ option. Sorry, I'm not a native English speaker, if somebody knows how to express this intention better please write it.

cblp (changes requested):
No need to double help. It's needed to guide the user.

gufmar:
sorry wrong click (re-opened)
I would suggest expressing if the value for the used option is wrong, that this value is not valid for this option:
_This is not a valid IPv4 address_
Then if one by mistake has used an IPv6 address he should understand that this option is not for this address type and he must look for the right option.

cblp:
Good idea! Please write the code.

newhoggy:
Related PR https://github.com/input-output-hk/cardano-node/pull/2691

cblp:
@gufmar please evaluate the new wording

gufmar:
Sorry, I missed this (in-between so many cardano-node notifications)
looks good to me

PR #2497, opened 2021-03-16
fix inline code in README

uwe:
Github seems to be confused by the ```:code:``` inline style.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -146,8 +146,8 @@ The general synopsis is as follows:
 
    Usage: cardano-cli (Era based commands | Byron specific commands | Miscellaneous commands)
 
-> NOTE: the exact invocation command depends on the environment.  If you have only built ``cardano-cli``, without installing it, then you have to prepend :code:`cabal run -- `
-before :code:`cardano-cli`.  We henceforth assume that the necessary environment-specific adjustment has been made, so we only mention ``cardano-cli``.
+> NOTE: the exact invocation command depends on the environment.  If you have only built ``cardano-cli``, without installing it, then you have to prepend ``cabal run -- ``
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -146,8 +146,8 @@ The general synopsis is as follows:
 
    Usage: cardano-cli (Era based commands | Byron specific commands | Miscellaneous commands)
 
-> NOTE: the exact invocation command depends on the environment.  If you have only built ``cardano-cli``, without installing it, then you have to prepend :code:`cabal run -- `
-before :code:`cardano-cli`.  We henceforth assume that the necessary environment-specific adjustment has been made, so we only mention ``cardano-cli``.
+> NOTE: the exact invocation command depends on the environment.  If you have only built ``cardano-cli``, without installing it, then you have to prepend ``cabal run -- ``
```
Preskton, line 149: Hi, @uwe, it seems like even after this change, the markdown parser is still somewhat confused.

I think the output you want is: If you have only built `cardano-cli` without installing it, then you have to prepend `cabal run -- ` before `cardano-cli`.

Which you can achieve with this:

```markdown
If you have only built `cardano-cli` without installing it, then you have to prepend `cabal run --` before `cardano-cli`.
```

For clarity and readability, though, you might consider simplifying the sentence by saying:

```markdown
If you've built `cardano-cli` without installing it, you should instead use `cabal run -- cardano-cli`.
```

newhoggy (approved):

newhoggy:
bors merge

README.rst
uwe: @newhoggy Preskton is right, this is still broken. Shall I make a PR with his proposal (last sentence)? Or do you want to fix this directly on master?

Preskton: I'm happy to raise a PR. 👍🏼

Preskton: Raised #2522.

PR #2498, opened 2021-03-16
Adapt packages to use Protocol type class

jhartzell42:
This PR depends on https://github.com/input-output-hk/ouroboros-network/pull/2978

In that PR, the `Protocol` type that the cardano-node packages use is removed. This is so that ouroboros-network does not need to enumerate all the consensus modes it implements. The majority of changes in this PR are in reaction to that, creating a type class to replace it.

We found that it would take a significant increase in code complexity to remove the explicit enumeration of supported consensus modes in two places: 
- cardano-node/src/Cardano/Node/Configuration/Logging.hs:nodeBasicInfo
- cardano-node/src/Cardano/Node/Query.hs

For that purpose, we have written the GADT `BlockType` which provides the little bit of type discovery these places need. 

In addition, we have removed the `Enum` instance for `AnyCardanoEra`. The motivation is that we anticipate the inclusion of experimental eras and consensus modes in the code base which cannot reasonably be said to be enumerable among the mainnet eras. This instance does not appear to be used anywhere. If it needs to be preserved then we will need an answer on how to handle eras we don't expect to be part of mainnet (at least when they are first implemented).

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -174,8 +174,8 @@ source-repository-package
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/ouroboros-network
-  tag: 7f90c8c59ffc7d61a4e161e886d8962a9c26787a
-  --sha256: 0hnw6hvbyny3wniaqw8d37l4ysgp8xrq5d84fapxfm525a4hfs0x
+  tag: 66dd5eb0f4bf20272e30a8c5c0a0c9d3992de039
+  --sha256: 1z23l4l81afdzmsh0bgs4ivg0nv58l4mp7g25qa4nfsczl6j07cf
   subdir:
     io-sim
     io-sim-classes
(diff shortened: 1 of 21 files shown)


jhartzell42:
Per @nc6 , we don't have to support GHC 8.6, and those are the only tests that are failing, so we can go ahead with this.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -5,6 +5,7 @@
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies #-}
```
Jimbo4350, line 8: Is this pragma necessary?

Jimbo4350 (commented):
I like the changes but I would get a sign off from @dcoutts

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
intricate: Doesn't seem like it.

jhartzell42: Fixed

dcoutts (approved):
LGTM!

Just in case there are conflicts, this should go in after PR #2547. There probably won't be, but just to be safe.

jhartzell42:
bors merge

jhartzell42:
Sorry I didn't see the comment at first. Thanks!

PR #2505, opened 2021-03-18
tracers:  drop `ForgeTime` and outcome-extractor-based microbenchmarking

deepfire:
On our way to replace the tracing infrastructure, we should also take the opportunity to deprecate the mechanisms that could be moved outside the node.


Jimbo4350:
@deepfire Nice :+1: Just squash the commits and I will approve

deepfire:
@Jimbo4350, done.

Jimbo4350 (approved):

PR #2522, opened 2021-03-21
Readme: clarify prepend of `cabal run` if you didn't install

Preskton:
Updated language in the `README` around when to use `cardano-cli` vs `cabal run`.  Update to #2497.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -146,8 +146,7 @@ The general synopsis is as follows:
 
    Usage: cardano-cli (Era based commands | Byron specific commands | Miscellaneous commands)
 
-> NOTE: the exact invocation command depends on the environment.  If you have only built ``cardano-cli``, without installing it, then you have to prepend ``cabal run -- ``
-before ``cardano-cli``.  We henceforth assume that the necessary environment-specific adjustment has been made, so we only mention ``cardano-cli``.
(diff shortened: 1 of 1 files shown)


uwe:
Maybe it makes sense to change all double backticks to single backticks. This would avoid further changes down the line. Maybe Github confuses things when a single line has two inline sections. For example ```cardano-cli``` is still in double backticks (and many more).

Preskton:
Thanks for the suggestion, @uwe. I'll fix this paragraph and then submit an issue for a larger, generalized cleanup of the `README` as a whole.

Preskton:
Ok, this should be good to go. Didn't realize that RST behaves differently than markdown, so I changed these to all be double backticks, and it appears correctly in the GH markdown viewer.

newhoggy:
I see on `master` the following:

```
NOTE: the exact invocation command depends on the environment.  If you have only built ``cardano-cli``, without installing it, then you have to prepend ``cabal run -- ``
-before ``cardano-cli``.  We henceforth assume that the necessary environment-specific adjustment has been made, so we only mention ``cardano-cli``.
```

so it looks like some variant of this change has already been merged.

Thanks for the contribution!

PR #2536, opened 2021-03-24
stake-snapshot and pool-params query commands

kevinhammond:
This PR provides two new query commands that should be useful to SPOs who are using the CNCLI set of tools, for example (as well as for internal IOG purposes):

`cardano-cli query stake-snapshot --stake-pool-id <poolid>`

`cardano-cli query pool-params --stake-pool-id <poolid>`

These commands allow users to access two pieces of information from the ledger state

1. stake snapshots
2. pool parameters, including retirements

The stake snapshot returns information about mark, set, go snapshots for a pool, plus the current total stake that can be used in a 'sigma' calculation:

```
cardano-cli query stake-snapshot --stake-pool-id a3e71398e80739522e907c35b8f652064e5268e62e9b78a9603f6b98 --mainnet
{
    "activeStakeGo": 200921353802160,
    "activeStakeMark": 209901353802160,
    "activeStakeTotal": 32334956959700987,
    "activeStakeSet": 201571353802160
}
```
The pool parameters return three pieces of information: current parameters, future parameters and retiring information.  They may be `null` if eg the parameters are not changing.

```
cardano-cli query pool-params --stake-pool-id a3e71398e80739522e907c35b8f652064e5268e62e9b78a9603f6b98 --mainnet
[
    {
        "publicKey": "a3e71398e80739522e907c35b8f652064e5268e62e9b78a9603f6b98",
        "cost": 340000000,
        "metadata": {
            "hash": "196eb1b2937af2c88647204920d19985501c99da3fc69e421a942b9b8f817e94",
            "url": "https://git.io/JTI4o"
        },
        "owners": [
            "6877dcb866858b2a4941351c98cd67202262bfd82a016cbbaa073927"
        ],
        "vrf": "11ad7c225b48721dbd9642de1ab05640fe53ab29af2df513f236cc7ac4617da6",
        "pledge": 0,
        "margin": 1,
        "rewardAccount": {
            "network": "Mainnet",
            "credential": {
                "key hash": "6877dcb866858b2a4941351c98cd67202262bfd82a016cbbaa073927"
            }
        },
        "relays": [
            {
                "single host address": {
                    "IPv6": null,
                    "port": 3001,
                    "IPv4": "34.91.217.120"
                }
            },
            {
                "single host address": {
                    "IPv6": null,
                    "port": 3001,
                    "IPv4": "34.91.176.78"
                }
            },
            {
                "single host address": {
                    "IPv6": null,
                    "port": 3001,
                    "IPv4": "34.91.150.76"
                }
            }
        ]
    },
    null,
    null
]

```

The main advantage over using `query ledger-state` is that these commands avoid the need to dump the full ledger state (time consuming and memory intensive - meaning they reduce the total system demands for SPOs), and will make it easier to support CNCLI and other tools.  They also use existing internal operations (such as the ledger pool stake calculation), meaning that the information is guaranteed to be identical to that which the ledger is using (and without having to write scripts to extract/correlate the information).

I have decided to put these under 'query' rather than 'pool' since they are extracting info from the ledger state.

### DIFF
--- a/cardano-api/src/Cardano/Api/TxBody.hs
+++ b/cardano-api/src/Cardano/Api/TxBody.hs
@@ -159,6 +159,7 @@ import           Cardano.Api.KeysShelley
 import           Cardano.Api.NetworkId
 import           Cardano.Api.ProtocolParameters
 import           Cardano.Api.Script
+import           Cardano.Api.SerialiseBech32
 import           Cardano.Api.SerialiseCBOR
 import           Cardano.Api.SerialiseRaw
 import           Cardano.Api.SerialiseTextEnvelope
@@ -292,20 +293,20 @@ instance IsCardanoEra era => ToJSON (TxOut era) where
       ShelleyAddressInEra sbe ->
         case sbe of
           ShelleyBasedEraShelley ->
-            let hexAddr = serialiseToRawBytesHexText addr
-            in object [ "address" .= hexAddr
-                      , "value" .= toJSON val
-                      ]
+            object
+              [ "address" .= serialiseToBech32 addr
+              , "value" .= toJSON val
+              ]
           ShelleyBasedEraAllegra ->
-            let hexAddr = serialiseToRawBytesHexText addr
-            in object [ "address" .= hexAddr
-                      , "value" .= toJSON val
-                      ]
+            object
+              [ "address" .= serialiseToBech32 addr
+              , "value" .= toJSON val
+              ]
           ShelleyBasedEraMary ->
-            let hexAddr = serialiseToRawBytesHexText addr
-            in object [ "address" .= hexAddr
-                      , "value" .= toJSON val
-                      ]
+            object
+              [ "address" .= serialiseToBech32 addr
+              , "value" .= toJSON val
+              ]
 
 
 
--- a/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
+++ b/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
@@ -673,23 +673,20 @@ pQueryCmd =
     pQueryProtocolParameters :: Parser QueryCmd
     pQueryProtocolParameters =
       QueryProtocolParameters'
-        <$> pCardanoEra
-        <*> pConsensusModeParams
+        <$> pConsensusModeParams
         <*> pNetworkId
         <*> pMaybeOutputFile
 
     pQueryTip :: Parser QueryCmd
     pQueryTip = QueryTip
-                  <$> pCardanoEra
-                  <*> pConsensusModeParams
+                  <$> pConsensusModeParams
                   <*> pNetworkId
                   <*> pMaybeOutputFile
 
     pQueryUTxO :: Parser QueryCmd
     pQueryUTxO =
       QueryUTxO'
-        <$> pCardanoEra
-        <*> pConsensusModeParams
+        <$> pConsensusModeParams
         <*> pQueryFilter
         <*> pNetworkId
         <*> pMaybeOutputFile
@@ -697,44 +694,38 @@ pQueryCmd =
     pQueryStakeDistribution :: Parser QueryCmd
     pQueryStakeDistribution =
       QueryStakeDistribution'
-        <$> pCardanoEra
-        <*> pConsensusModeParams
+        <$> pConsensusModeParams
         <*> pNetworkId
         <*> pMaybeOutputFile
 
     pQueryStakeAddressInfo :: Parser QueryCmd
     pQueryStakeAddressInfo =
       QueryStakeAddressInfo
-        <$> pCardanoEra
-        <*> pConsensusModeParams
+        <$> pConsensusModeParams
         <*> pFilterByStakeAddress
         <*> pNetworkId
         <*> pMaybeOutputFile
 
     pQueryLedgerState :: Parser QueryCmd
     pQueryLedgerState = QueryLedgerState'
-                          <$> pCardanoEra
-                          <*> pConsensusModeParams
+                          <$> pConsensusModeParams
                           <*> pNetworkId
                           <*> pMaybeOutputFile
 
     pQueryProtocolState :: Parser QueryCmd
     pQueryProtocolState = QueryProtocolState'
-                            <$> pCardanoEra
-                            <*> pConsensusModeParams
+                            <$> pConsensusModeParams
                             <*> pNetworkId
                             <*> pMaybeOutputFile
 
     pQueryStakeSnapshot :: Parser QueryCmd
     pQueryStakeSnapshot = QueryStakeSnapshot'
-                          <$> pCardanoEra
                           <*> pConsensusModeParams
                           <*> pNetworkId
                           <*> pStakePoolVerificationKeyHash
 
     pQueryPoolParams :: Parser QueryCmd
     pQueryPoolParams = QueryPoolParams'
-                          <$> pCardanoEra
                           <*> pConsensusModeParams
                           <*> pNetworkId
                           <*> pStakePoolVerificationKeyHash
@@ -1514,19 +1505,6 @@ pNetworkId =
       (  Opt.long "mainnet"
       <> Opt.help "Use the mainnet magic id."
       )
-
-{-
-pPoolId :: Parser PoolId
-pPoolId =
-  pMainnet <|> fmap Testnet pTestnetMagic
- where
-   pMainnet :: Parser NetworkId
-   pMainnet =
-    Opt.flag' Mainnet
-      (  Opt.long "mainnet"
-      <> Opt.help "Use the mainnet magic id."
-      )
--}
 
 pTestnetMagic :: Parser NetworkMagic
 pTestnetMagic =
--- a/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
+++ b/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
@@ -720,13 +720,13 @@ pQueryCmd =
 
     pQueryStakeSnapshot :: Parser QueryCmd
     pQueryStakeSnapshot = QueryStakeSnapshot'
-                          <*> pConsensusModeParams
+                          <$> pConsensusModeParams
                           <*> pNetworkId
                           <*> pStakePoolVerificationKeyHash
 
     pQueryPoolParams :: Parser QueryCmd
     pQueryPoolParams = QueryPoolParams'
-                          <*> pConsensusModeParams
+                          <$> pConsensusModeParams
                           <*> pNetworkId
                           <*> pStakePoolVerificationKeyHash
 
--- a/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
+++ b/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
@@ -664,25 +664,32 @@ pQueryCmd =
         (Opt.info pQueryLedgerState $ Opt.progDesc "Dump the current ledger state of the node (Ledger.NewEpochState -- advanced command)")
     , subParser "protocol-state"
         (Opt.info pQueryProtocolState $ Opt.progDesc "Dump the current protocol state of the node (Ledger.ChainDepState -- advanced command)")
+    , subParser "stake-snapshot"
+        (Opt.info pQueryStakeSnapshot $ Opt.progDesc "Obtain the stake snapshot for a pool (mark,set,go,total -- advanced command)")
+    , subParser "pool-params"
+        (Opt.info pQueryPoolParams $ Opt.progDesc "Dump the pool parameters (Ledger.NewEpochState.esLState._delegationState._pState._pParams -- advanced command)")
     ]
   where
     pQueryProtocolParameters :: Parser QueryCmd
     pQueryProtocolParameters =
       QueryProtocolParameters'
-        <$> pConsensusModeParams
+        <$> pCardanoEra
+        <*> pConsensusModeParams
         <*> pNetworkId
         <*> pMaybeOutputFile
 
     pQueryTip :: Parser QueryCmd
     pQueryTip = QueryTip
-                  <$> pConsensusModeParams
+                  <$> pCardanoEra
+                  <*> pConsensusModeParams
                   <*> pNetworkId
                   <*> pMaybeOutputFile
 
     pQueryUTxO :: Parser QueryCmd
     pQueryUTxO =
       QueryUTxO'
-        <$> pConsensusModeParams
+        <$> pCardanoEra
+        <*> pConsensusModeParams
         <*> pQueryFilter
         <*> pNetworkId
         <*> pMaybeOutputFile
@@ -690,29 +697,48 @@ pQueryCmd =
     pQueryStakeDistribution :: Parser QueryCmd
     pQueryStakeDistribution =
       QueryStakeDistribution'
-        <$> pConsensusModeParams
+        <$> pCardanoEra
+        <*> pConsensusModeParams
         <*> pNetworkId
         <*> pMaybeOutputFile
 
     pQueryStakeAddressInfo :: Parser QueryCmd
     pQueryStakeAddressInfo =
       QueryStakeAddressInfo
-        <$> pConsensusModeParams
+        <$> pCardanoEra
+        <*> pConsensusModeParams
         <*> pFilterByStakeAddress
         <*> pNetworkId
         <*> pMaybeOutputFile
 
     pQueryLedgerState :: Parser QueryCmd
     pQueryLedgerState = QueryLedgerState'
-                          <$> pConsensusModeParams
+                          <$> pCardanoEra
+                          <*> pConsensusModeParams
                           <*> pNetworkId
                           <*> pMaybeOutputFile
 
     pQueryProtocolState :: Parser QueryCmd
     pQueryProtocolState = QueryProtocolState'
-                            <$> pConsensusModeParams
+                            <$> pCardanoEra
+                            <*> pConsensusModeParams
                             <*> pNetworkId
                             <*> pMaybeOutputFile
+
+    pQueryStakeSnapshot :: Parser QueryCmd
+    pQueryStakeSnapshot = QueryStakeSnapshot'
+                          <$> pCardanoEra
+                          <*> pConsensusModeParams
+                          <*> pNetworkId
+                          <*> pStakePoolVerificationKeyHash
+
+    pQueryPoolParams :: Parser QueryCmd
+    pQueryPoolParams = QueryPoolParams'
+                          <$> pCardanoEra
+                          <*> pConsensusModeParams
+                          <*> pNetworkId
+                          <*> pStakePoolVerificationKeyHash
+
 
 pGovernanceCmd :: Parser GovernanceCmd
 pGovernanceCmd =
@@ -1488,6 +1514,19 @@ pNetworkId =
       (  Opt.long "mainnet"
       <> Opt.help "Use the mainnet magic id."
       )
+
+{-
+pPoolId :: Parser PoolId
+pPoolId =
+  pMainnet <|> fmap Testnet pTestnetMagic
+ where
+   pMainnet :: Parser NetworkId
+   pMainnet =
+    Opt.flag' Mainnet
+      (  Opt.long "mainnet"
+      <> Opt.help "Use the mainnet magic id."
+      )
+-}
 
 pTestnetMagic :: Parser NetworkMagic
 pTestnetMagic =
--- a/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
+++ b/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
@@ -665,7 +665,7 @@ pQueryCmd =
     , subParser "protocol-state"
         (Opt.info pQueryProtocolState $ Opt.progDesc "Dump the current protocol state of the node (Ledger.ChainDepState -- advanced command)")
     , subParser "stake-snapshot"
-        (Opt.info pQueryStakeSnapshot $ Opt.progDesc "Obtain the stake snapshot for a pool (mark,set,go,total -- advanced command)")
+        (Opt.info pQueryStakeSnapshot $ Opt.progDesc "Obtain the three stake snapshots for a pool, plus the total active stake (advanced command)")
     , subParser "pool-params"
         (Opt.info pQueryPoolParams $ Opt.progDesc "Dump the pool parameters (Ledger.NewEpochState.esLState._delegationState._pState._pParams -- advanced command)")
     ]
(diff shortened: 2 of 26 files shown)


cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -1489,6 +1515,19 @@ pNetworkId =
...
+  pMainnet <|> fmap Testnet pTestnetMagic
+ where
+   pMainnet :: Parser NetworkId
+   pMainnet =
+    Opt.flag' Mainnet
+      (  Opt.long "mainnet"
+      <> Opt.help "Use the mainnet magic id."
+      )
+-}
+
```
erikd, line 1530: If we do not need this it should be deleted instead of just commented out.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -57,6 +57,17 @@ import           Ouroboros.Network.Protocol.LocalStateQuery.Type as LocalStateQu
...
 import           Shelley.Spec.Ledger.Scripts ()
 
+-- These imports are used specifically for stake snapshots
+import           Shelley.Spec.Ledger.LedgerState hiding (LedgerState,_delegations)
+import           Shelley.Spec.Ledger.EpochBoundary
+import           Shelley.Spec.Ledger.Coin
+import           Shelley.Spec.Ledger.Keys (KeyHash(..))
+import           Data.Maybe
+import           Cardano.Ledger.Era
+
```
erikd, line 67: The convention for this repo is that all imports from outside this repo *MUST* be explcit or qualified.

These imports should be distributed amongst the above imports.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -90,90 +103,102 @@ renderShelleyQueryCmdError err =
...
+  -> AnyConsensusModeParams
   -> NetworkId
   -> Maybe OutputFile
   -> ExceptT ShelleyQueryCmdError IO ()
-runQueryTip (AnyConsensusModeParams cModeParams) network mOutFile = do
-  SocketPath sockPath <- firstExceptT ShelleyQueryCmdEnvVarSocketErr readEnvSocketPath
-  let localNodeConnInfo = LocalNodeConnectInfo cModeParams network sockPath
-      consensusMode = consensusModeOnly cModeParams
+runQueryTip (AnyCardanoEra era) (AnyConsensusModeParams cModeParams) network mOutFile = do
+    SocketPath sockPath <- firstExceptT ShelleyQueryCmdEnvVarSocketErr readEnvSocketPath
```
erikd, line 173: We should avoid semantically null changes, like indentation changes.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -188,35 +213,37 @@ runQueryTip (AnyConsensusModeParams cModeParams) network mOutFile = do
    toObject _ _ = Aeson.Null
 
 
+
+
```
erikd, line 217: Should not be part of this PR.

erikd (changes requested):

AndrewWestberg:
I believe we need `activeStakeMarkTotal`, `activeStakeSetTotal`, and `activeStakeGoTotal`. The total active stake can change between each of the snapshots.

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -664,56 +664,82 @@ pQueryCmd =
...
     pQueryProtocolState = QueryProtocolState'
-                            <$> pConsensusModeParams
+                            <$> pCardanoEra
+                            <*> pConsensusModeParams
                             <*> pNetworkId
                             <*> pMaybeOutputFile
 
+    pQueryStakeSnapshot :: Parser QueryCmd
+    pQueryStakeSnapshot = QueryStakeSnapshot'
+                          <$> pCardanoEra
```
Jimbo4350, line 730: should be 2 spaces for indentation

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -664,56 +664,82 @@ pQueryCmd =
...
+                          <*> pConsensusModeParams
+                          <*> pNetworkId
+                          <*> pStakePoolVerificationKeyHash
+
+    pQueryPoolParams :: Parser QueryCmd
+    pQueryPoolParams = QueryPoolParams'
+                          <$> pCardanoEra
```
Jimbo4350, line 737: should be 2 spaces for indentation

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -90,90 +103,102 @@ renderShelleyQueryCmdError err =
...
   -> NetworkId
   -> Maybe OutputFile
   -> ExceptT ShelleyQueryCmdError IO ()
-runQueryProtocolParameters (AnyConsensusModeParams cModeParams) network mOutFile = do
+runQueryProtocolParameters anyEra@(AnyCardanoEra era) (AnyConsensusModeParams cModeParams)
+                           network mOutFile = do
   SocketPath sockPath <- firstExceptT ShelleyQueryCmdEnvVarSocketErr
                            readEnvSocketPath
+
+  let consensusMode = consensusModeOnly cModeParams
```
Jimbo4350, line 136: I'm not sure what's happened here. If you look at the diff we have clearly undone some refactoring.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
Jimbo4350: Same as above. If you look at the diff we have clearly undone some refactoring.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -188,35 +213,37 @@ runQueryTip (AnyConsensusModeParams cModeParams) network mOutFile = do
...
   -> QueryFilter
   -> NetworkId
   -> Maybe OutputFile
   -> ExceptT ShelleyQueryCmdError IO ()
-runQueryUTxO (AnyConsensusModeParams cModeParams)
+runQueryUTxO anyEra@(AnyCardanoEra era) (AnyConsensusModeParams cModeParams)
              qfilter network mOutFile = do
   SocketPath sockPath <- firstExceptT ShelleyQueryCmdEnvVarSocketErr readEnvSocketPath
+  let consensusMode = consensusModeOnly cModeParams
+  eraInMode <- hoistMaybe (ShelleyQueryCmdEraConsensusModeMismatch anyEra (AnyConsensusMode consensusMode))
```
Jimbo4350, line 233: Same as above. If you look at the diff we have clearly undone some refactoring.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -232,93 +259,176 @@ runQueryUTxO (AnyConsensusModeParams cModeParams)
   maybeFiltered NoFilter = Nothing
 
 
+runQueryPoolParams
+  :: AnyCardanoEra
+  -> AnyConsensusModeParams
+  -> NetworkId
+  -> Hash StakePoolKey
+  -> ExceptT ShelleyQueryCmdError IO ()
+
```
Jimbo4350, line 268: Unneccesary newline here

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -232,93 +259,176 @@ runQueryUTxO (AnyConsensusModeParams cModeParams)
...
+runQueryPoolParams anyEra@(AnyCardanoEra era) (AnyConsensusModeParams cModeParams)
+                    network poolid = do
+    SocketPath sockPath <- firstExceptT ShelleyQueryCmdEnvVarSocketErr readEnvSocketPath
+
+    let consensusMode = consensusModeOnly cModeParams
```
Jimbo4350, line 273: So this will work, but this can be refactored. If you look at my earlier comments that point out where refactoring was undone, it will be clearer as to how this can be refactored.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -232,93 +259,176 @@ runQueryUTxO (AnyConsensusModeParams cModeParams)
...
+  -> NetworkId
+  -> Hash StakePoolKey
+  -> ExceptT ShelleyQueryCmdError IO ()
+runQueryStakeSnapshot anyEra@(AnyCardanoEra era) (AnyConsensusModeParams cModeParams)
+                    network poolid = do
+    SocketPath sockPath <- firstExceptT ShelleyQueryCmdEnvVarSocketErr readEnvSocketPath
+
+    let consensusMode = consensusModeOnly cModeParams
+
+    eraInMode <- hoistMaybe (ShelleyQueryCmdEraConsensusModeMismatch anyEra (AnyConsensusMode consensusMode))
```
Jimbo4350, line 308: Same here, this can be refactored. See above.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -232,93 +259,176 @@ runQueryUTxO (AnyConsensusModeParams cModeParams)
...
-                  cModeParams
-                  localNodeConnInfo
-                  qInMode
-      obtainLedgerEraClassConstraints sbe (writeLedgerState mOutFile) result
-    Nothing -> left . ShelleyQueryCmdEraConsensusModeMismatch anyE $ AnyConsensusMode cMode
+    eraInMode <- hoistMaybe (ShelleyQueryCmdEraConsensusModeMismatch anyEra (AnyConsensusMode consensusMode))
+                  $ toEraInMode era consensusMode
+
+    let localNodeConnInfo = LocalNodeConnectInfo cModeParams network sockPath
+    sbe <- getSbe $ cardanoEraStyle era
```
Jimbo4350, line 344: More undone refactoring.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -232,93 +259,176 @@ runQueryUTxO (AnyConsensusModeParams cModeParams)
...
-  sbe <- getSbe $ cardanoEraStyle era
+    SocketPath sockPath <- firstExceptT ShelleyQueryCmdEnvVarSocketErr readEnvSocketPath
+    let consensusMode = consensusModeOnly cModeParams
+
+    eraInMode <- hoistMaybe (ShelleyQueryCmdEraConsensusModeMismatch anyEra (AnyConsensusMode consensusMode))
+                  $ toEraInMode era consensusMode
+
+    let localNodeConnInfo = LocalNodeConnectInfo cModeParams network sockPath
+
+    sbe <- getSbe $ cardanoEraStyle era
```
Jimbo4350, line 378: Undone refactoring

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -232,93 +259,176 @@ runQueryUTxO (AnyConsensusModeParams cModeParams)
...
+
+  let consensusMode = consensusModeOnly cModeParams
+
+  eraInMode <- hoistMaybe (ShelleyQueryCmdEraConsensusModeMismatch anyEra (AnyConsensusMode consensusMode))
+                 $ toEraInMode era consensusMode
+
   let localNodeConnInfo = LocalNodeConnectInfo cModeParams network sockPath
+  qInMode <- case cardanoEraStyle era of
+               LegacyByronEra -> left ShelleyQueryCmdByronEra
+               ShelleyBasedEra sbe ->
```
Jimbo4350, line 417: And here

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -380,6 +490,132 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
      first (const ls) (decodeFull ls)
 
 
+-- Data Structure to hold stake information so that it can be printed nicely
+data Stakes =  Stakes {
```
Jimbo4350, line 494: Let's add some docs as to that `mark, set, go, total` mean. Both here and in the cli docs.
`-- Data Structure to hold stake information so that it can be printed nicely` is unnecessary and should be deleted.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -380,6 +490,132 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
...
+      mark, set, go, total :: Integer
+    } deriving Show
+
+instance ToJSON Stakes where
+    -- this generates a Value
```
Jimbo4350, line 499: `-- this generates a Value` is unnecessary

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -380,6 +490,132 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
...
+    -- this encodes directly to a bytestring Builder
+    toEncoding (Stakes m s g t) =
+        pairs ("mark" .= m <> "set" .= s  <> "go" .= g  <> "total" .= t)
+
+
+writeStakeSnapshot :: forall era ledgerera.
+                    ShelleyLedgerEra era ~ ledgerera
+                 => Era ledgerera
+                 => FromCBOR (LedgerState era)
+                 =>
```
Jimbo4350, line 512: unnecessary newline

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -380,6 +490,132 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
...
+    toJSON (Stakes m s g t) =
+        object ["activeStakeMark" .= m, "activeStakeSet" .= s, "activeStakeGo" .= g, "activeStakeTotal" .= t]
+
+    -- this encodes directly to a bytestring Builder
+    toEncoding (Stakes m s g t) =
+        pairs ("mark" .= m <> "set" .= s  <> "go" .= g  <> "total" .= t)
+
+
+writeStakeSnapshot :: forall era ledgerera.
+                    ShelleyLedgerEra era ~ ledgerera
```
Jimbo4350, line 509: Indentation should be 2 spaces

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -380,6 +490,132 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
...
+                 => Era ledgerera
+                 => FromCBOR (LedgerState era)
+                 =>
+                    SerialisedLedgerState era
+                 -> Hash StakePoolKey
+                 -> ExceptT ShelleyQueryCmdError IO ()
+
+writeStakeSnapshot qState poolId =
+       case decodeLedgerState qState of
```
Jimbo4350, line 518: indentation should be 2 spaces

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -380,6 +490,132 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
...
+           Left bs ->
```
Jimbo4350, line 519: Why the newline here?

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -380,6 +490,132 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
...
+                       hk = KeyHash $ fromJust $ maybehk
+
+ where
+   decodeLedgerState
+     :: SerialisedLedgerState era
+     -> Either LBS.ByteString (LedgerState era)
+   decodeLedgerState (SerialisedLedgerState (Serialised ls)) =
+     first (const ls) (decodeFull ls)
+
+--   getPoolStake
```
Jimbo4350, line 553: Why is this type signature commented out?

Jimbo4350 (changes requested):
Right overall direction but some things need to be addressed.

Scitz0:
1. I would like to request a change to the pool-params output to make it more clear what the three different sets returned mean. See suggestion below.
2. In addition to this, I would like for the future set an additional field called `activeInEpoch` containing what epoch the future pool update becomes active.

**Current output:**
An array containing three unnamed sets for current, future, and retired.
```
[
  {
    "publicKey": "0b08b536ba3afe21fbd826fb92bee897a3d1e575a7849b2e6c4659ab",
    "...": "rest of the fields"
  },
  {
    "publicKey": "0b08b536ba3afe21fbd826fb92bee897a3d1e575a7849b2e6c4659ab",
    "...": "rest of the fields"
  },
  null
]
```

**Requested output:**
My idea would be to have a schema like this instead, replacing the JSON array containing three unnamed objects with a key/value schema where `active` and `future` contain a JSON object and `retired` the epoch number or null.
Makes it clean and easy to understand what is what.
```
{
  "active": {
    "publicKey": "0b08b536ba3afe21fbd826fb92bee897a3d1e575a7849b2e6c4659ab",
    "...": "rest of the fields"
  },
  "future": {
    "publicKey": "0b08b536ba3afe21fbd826fb92bee897a3d1e575a7849b2e6c4659ab",
    "activeInEpoch": 125,
    "...": "rest of the fields"
  },
  "retired": null
}
```

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
kevinhammond: OK.  There's another Cardano.Ledger.Era import in Genesis.hs (not touched by me).  I'll fix that one as well

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
kevinhammond: > 1. I would like to request a change to the pool-params output to make it more clear what the three different sets returned mean. See suggestion below.
> 2. In addition to this, I would like for the future set an additional field called `activeInEpoch` containing what epoch the future pool update becomes active.
> 
> **Current output:**
> An array containing three unnamed sets for current, future, and retired.
> 
> ```
> [
>   {
>     "publicKey": "0b08b536ba3afe21fbd826fb92bee897a3d1e575a7849b2e6c4659ab",
>     "...": "rest of the fields"
>   },
>   {
>     "publicKey": "0b08b536ba3afe21fbd826fb92bee897a3d1e575a7849b2e6c4659ab",
>     "...": "rest of the fields"
>   },
>   null
> ]
> ```
> 
> **Requested output:**
> My idea would be to have a schema like this instead, replacing the JSON array containing three unnamed objects with a key/value schema where `active` and `future` contains a JSON object(null for future if there are none) and `retired` the epoch number or null.
> Makes it clean and easy to understand what is what.
> 
> ```
> {
>   "active": {
>     "publicKey": "0b08b536ba3afe21fbd826fb92bee897a3d1e575a7849b2e6c4659ab",
>     "...": "rest of the fields"
>   },
>   "future": {
>     "publicKey": "0b08b536ba3afe21fbd826fb92bee897a3d1e575a7849b2e6c4659ab",
>     "activeInEpoch": 125,
>     "...": "rest of the fields"
>   },
>   "retired": null
> }
> ```

Added

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
kevinhammond: > 1. I would like to request a change to the pool-params output to make it more clear what the three different sets returned mean. See suggestion below.
> 2. In addition to this, I would like for the future set an additional field called `activeInEpoch` containing what epoch the future pool update becomes active.
> 
> **Current output:**
> An array containing three unnamed sets for current, future, and retired.
> 
> ```
> [
>   {
>     "publicKey": "0b08b536ba3afe21fbd826fb92bee897a3d1e575a7849b2e6c4659ab",
>     "...": "rest of the fields"
>   },
>   {
>     "publicKey": "0b08b536ba3afe21fbd826fb92bee897a3d1e575a7849b2e6c4659ab",
>     "...": "rest of the fields"
>   },
>   null
> ]
> ```
> 
> **Requested output:**
> My idea would be to have a schema like this instead, replacing the JSON array containing three unnamed objects with a key/value schema where `active` and `future` contains a JSON object(null for future if there are none) and `retired` the epoch number or null.
> Makes it clean and easy to understand what is what.
> 
> ```
> {
>   "active": {
>     "publicKey": "0b08b536ba3afe21fbd826fb92bee897a3d1e575a7849b2e6c4659ab",
>     "...": "rest of the fields"
>   },
>   "future": {
>     "publicKey": "0b08b536ba3afe21fbd826fb92bee897a3d1e575a7849b2e6c4659ab",
>     "activeInEpoch": 125,
>     "...": "rest of the fields"
>   },
>   "retired": null
> }
> ```

Done

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
kevinhammond: I think this was done by the git merge.  I've remerged on the latest sources

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
kevinhammond: I think it's important to explain why this data structure is introduced here.  It really has no purpose other than displaying the stake information.  Generally, I find it's good practice to comment data structures.  I have added some comments about mark, set, go.  They're a bit esoteric

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
kevinhammond: OK.  I copied the comment from another instance

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -292,20 +293,20 @@ instance IsCardanoEra era => ToJSON (TxOut era) where
       ShelleyAddressInEra sbe ->
         case sbe of
           ShelleyBasedEraShelley ->
-            let hexAddr = serialiseToRawBytesHexText addr
-            in object [ "address" .= hexAddr
-                      , "value" .= toJSON val
-                      ]
+            object
+              [ "address" .= serialiseToBech32 addr
```
Jimbo4350, line 297: Unnecessary diff here and in other places.

Jimbo4350 (changes requested):
There is still some unnecessary diff in this PR that needs to be sorted out before I review the rest.

erikd:
Closing this in favor of #2560 because no one else has.

PR #2539, opened 2021-03-25
Nixos service fixes

jbgi:


nix/nixos/cardano-node-service.nix
```diff
@@ -102,14 +100,6 @@ in {
...
       };
-      instanced = mkOption {
-        type = types.bool;
-        default = false;
-        description = ''
-          Whether to enable systemd service instancing.
-          For details see https://fedoramagazine.org/systemd-template-unit-files/
-        '';
-      };
       instances = mkOption {
```
deepfire, line 103: Isn't this also coming along with `instanced`, or is this the other instancing implementation?

deepfire (approved):
LGTM, thank you @jbgi!

nix/nixos/cardano-node-service.nix
jbgi: no, `instances`` is not implemented through systemd template, just regular nix-generated services.

PR #2549, opened 2021-03-29
Switch to use separate hedgehog-extras repository

newhoggy:
This would allow this project to cache better in `cardano-node`, and also allow for this `hedgehog-extras` to be used elsewhere, for example `cardano-entropy`.


erikd (approved):

newhoggy:
bors merge

PR #2560, opened 2021-03-30
Updated stake-snapshot and pool-params CLI commands

kevinhammond:
This PR provides two new query commands that should be useful to SPOs who are using the CNCLI set of tools, for example (as well as for internal IOG purposes):

`cardano-cli query stake-snapshot --stake-pool-id <poolid>`

`cardano-cli query pool-params --stake-pool-id <poolid>`

These commands allow users to access two pieces of information from the ledger state

1. stake snapshots
2. pool parameters, including retirements

The stake snapshot returns information about the mark, set, go ledger snapshots for a pool, plus the total active stake for each snapshot that can be used in a 'sigma' calculation:

```
cardano-cli query stake-snapshot --stake-pool-id 00beef0a9be2f6d897ed24a613cf547bb20cd282a04edfc53d477114 --mainnet
{
    "poolStakeGo": 40278547538358,
    "activeStakeGo": 22753958467474959,
    "poolStakeMark": 40424218559492,
    "activeStakeMark": 22670949084364797,
    "poolStakeSet": 39898761956772,
    "activeStakeSet": 22488877070796904
}

```
Each snapshot is taken at the end of a different era.  The `go` snapshot is the current one and was taken two epochs earlier, `set` was taken one epoch ago, and `mark` was taken immediately before the start of the current epoch.

The pool parameters command returns three pieces of information: current parameters, future parameters and retiring information.  They may be `null` if eg the parameters are not changing.

```
cardano-cli query pool-params --stake-pool-id d785ff6a030ae9d521770c00f264a2aa423e928c85fc620b13d46eda --mainnet
{
    "poolParams": {
        "publicKey": "d785ff6a030ae9d521770c00f264a2aa423e928c85fc620b13d46eda",
        "cost": 340000000,
        "metadata": {
            "hash": "b150b12a1301c4b1510ac8b9f53f7571cabb43455f6fd244cd8fd97504b1c869",
            "url": "https://adalite.io/ADLT4-metadata.json"
        },
        "owners": [
            "463a9695c9222183ee6e1523478722bebcb332fa3769f1d8ef40c7d0",
            "5049c1dac0e597ee902f27a74a167cf135ae7c1717b0d3a417cd6c67"
        ],
        "vrf": "0a21e37b1917ce37a897eb2a8dc6715973a18d0586f7ab4962e3975561151348",
        "pledge": 30000000000,
        "margin": 3.0e-2,
        "rewardAccount": {
            "network": "Mainnet",
            "credential": {
                "key hash": "b1bc146a5fb0683c4e3836712d115b98619048bc307cc059b6adc76e"
            }
        },
        "relays": [
            {
                "single host address": {
                    "IPv6": null,
                    "port": 3003,
                    "IPv4": "54.228.75.154"
                }
            },
            {
                "single host address": {
                    "IPv6": null,
                    "port": 3001,
                    "IPv4": "54.228.75.154"
                }
            },
            {
                "single host address": {
                    "IPv6": null,
                    "port": 3003,
                    "IPv4": "34.249.11.89"
                }
            },
            {
                "single host address": {
                    "IPv6": null,
                    "port": 3001,
                    "IPv4": "34.249.11.89"
                }
            }
        ]
    },
    "futurePoolParams": null,
    "retiring": null
}

```

The main advantage of these commands over using `query ledger-state` is that they avoid the need to dump the full ledger state (which is both time consuming and memory intensive - meaning they reduce the total system demands for SPOs), and will make it easier to support CNCLI and other tools.  They also use existing internal operations (such as the ledger pool stake and active stake calculations), meaning that the information is guaranteed to be identical to that which the ledger is using (and without having to write scripts to extract/correlate the information).

I have decided to put these under 'query' rather than 'pool' since they are extracting info from the ledger state.

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
+++ b/cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
@@ -48,6 +48,7 @@ import           Cardano.Binary (decodeFull)
 import           Cardano.Crypto.Hash (hashToBytesAsHex)
 
 import qualified Cardano.Ledger.Crypto as Crypto
+import qualified Cardano.Ledger.Era as Era
 import qualified Cardano.Ledger.Shelley.Constraints as Ledger
 import           Ouroboros.Consensus.Cardano.Block as Consensus (EraMismatch (..))
 import           Ouroboros.Consensus.Shelley.Protocol (StandardCrypto)
@@ -55,6 +56,10 @@ import           Ouroboros.Network.Block (Serialised (..))
 import           Ouroboros.Network.Protocol.LocalStateQuery.Type as LocalStateQuery
                    (AcquireFailure (..))
 import qualified Shelley.Spec.Ledger.API.Protocol as Ledger
+import           Shelley.Spec.Ledger.Coin
+import           Shelley.Spec.Ledger.EpochBoundary
+import           Shelley.Spec.Ledger.Keys (KeyHash (..), KeyRole (..))
+import           Shelley.Spec.Ledger.LedgerState hiding (LedgerState, _delegations)
 import           Shelley.Spec.Ledger.Scripts ()
 
 {- HLINT ignore "Reduce duplication" -}
@@ -68,6 +73,7 @@ data ShelleyQueryCmdError
   | ShelleyQueryCmdAcquireFailure !AcquireFailure
   | ShelleyQueryCmdEraConsensusModeMismatch !AnyConsensusMode !AnyCardanoEra
   | ShelleyQueryCmdByronEra
+  | ShelleyQueryCmdPoolIdError (Hash StakePoolKey)
   | ShelleyQueryCmdEraMismatch !EraMismatch
   deriving Show
 
@@ -80,6 +86,7 @@ renderShelleyQueryCmdError err =
     ShelleyQueryCmdHelpersError helpersErr -> renderHelpersError helpersErr
     ShelleyQueryCmdAcquireFailure aqFail -> Text.pack $ show aqFail
     ShelleyQueryCmdByronEra -> "This query cannot be used for the Byron era"
+    ShelleyQueryCmdPoolIdError poolId -> "The pool id does not exist: " <> show poolId
     ShelleyQueryCmdEraConsensusModeMismatch (AnyConsensusMode cMode) (AnyCardanoEra era) ->
       "Consensus mode and era mismatch. Consensus mode: " <> show cMode <>
       " Era: " <> show era
@@ -100,6 +107,10 @@ runQueryCmd cmd =
       runQueryStakeAddressInfo consensusModeParams addr network mOutFile
     QueryLedgerState' consensusModeParams network mOutFile ->
       runQueryLedgerState consensusModeParams network mOutFile
+    QueryStakeSnapshot' consensusModeParams network poolid ->
+      runQueryStakeSnapshot consensusModeParams network poolid
+    QueryPoolParams' consensusModeParams network poolid ->
+      runQueryPoolParams consensusModeParams network poolid
     QueryProtocolState' consensusModeParams network mOutFile ->
       runQueryProtocolState consensusModeParams network mOutFile
     QueryUTxO' consensusModeParams qFilter networkId mOutFile ->
@@ -233,6 +244,55 @@ runQueryUTxO (AnyConsensusModeParams cModeParams)
   maybeFiltered :: QueryFilter -> Maybe (Set AddressAny)
   maybeFiltered (FilterByAddress as) = Just as
   maybeFiltered NoFilter = Nothing
+
+
+-- | Query the current and future parameters for a stake pool, including the retirement date.
+-- Any of these may be empty (in which case a null will be displayed).
+--
+
+runQueryPoolParams
+  :: AnyConsensusModeParams
+  -> NetworkId
+  -> Hash StakePoolKey
+  -> ExceptT ShelleyQueryCmdError IO ()
+runQueryPoolParams (AnyConsensusModeParams cModeParams) network poolid = do
+  SocketPath sockPath <- firstExceptT ShelleyQueryCmdEnvVarSocketErr readEnvSocketPath
+  let localNodeConnInfo = LocalNodeConnectInfo cModeParams network sockPath
+
+  anyE@(AnyCardanoEra era) <- determineEra cModeParams localNodeConnInfo
+  let cMode = consensusModeOnly cModeParams
+  sbe <- getSbe $ cardanoEraStyle era
+
+  eInMode <- toEraInMode era cMode
+    & hoistMaybe (ShelleyQueryCmdEraConsensusModeMismatch (AnyConsensusMode cMode) anyE)
+
+  let qInMode = QueryInEra eInMode . QueryInShelleyBasedEra sbe $ QueryLedgerState
+  result <- executeQuery era cModeParams localNodeConnInfo qInMode
+  obtainLedgerEraClassConstraints sbe (writePoolParams poolid) result
+
+
+-- | Obtain stake snapshot information for a pool, plus information about the total active stake.
+-- This information can be used for leader slot calculation, for example, and has been requested by SPOs.
+-- Obtaining the information directly is significantly more time and memory efficient than using a full ledger state dump.
+runQueryStakeSnapshot
+  :: AnyConsensusModeParams
+  -> NetworkId
+  -> Hash StakePoolKey
+  -> ExceptT ShelleyQueryCmdError IO ()
+runQueryStakeSnapshot (AnyConsensusModeParams cModeParams) network poolid = do
+  SocketPath sockPath <- firstExceptT ShelleyQueryCmdEnvVarSocketErr readEnvSocketPath
+  let localNodeConnInfo = LocalNodeConnectInfo cModeParams network sockPath
+
+  anyE@(AnyCardanoEra era) <- determineEra cModeParams localNodeConnInfo
+  let cMode = consensusModeOnly cModeParams
+  sbe <- getSbe $ cardanoEraStyle era
+
+  eInMode <- toEraInMode era cMode
+    & hoistMaybe (ShelleyQueryCmdEraConsensusModeMismatch (AnyConsensusMode cMode) anyE)
+
+  let qInMode = QueryInEra eInMode . QueryInShelleyBasedEra sbe $ QueryLedgerState
+  result <- executeQuery era cModeParams localNodeConnInfo qInMode
+  obtainLedgerEraClassConstraints sbe (writeStakeSnapshot poolid) result
 
 
 runQueryLedgerState
@@ -375,13 +435,80 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
     Just (OutputFile fpath) ->
       handleIOExceptT (ShelleyQueryCmdWriteFileError . FileIOError fpath)
         $ LBS.writeFile fpath $ unSerialised serLedgerState
- where
-   decodeLedgerState
-     :: SerialisedLedgerState era
-     -> Either LBS.ByteString (LedgerState era)
-   decodeLedgerState (SerialisedLedgerState (Serialised ls)) =
-     first (const ls) (decodeFull ls)
 
+writeStakeSnapshot :: forall era ledgerera. ()
+  => ShelleyLedgerEra era ~ ledgerera
+  => Era.Crypto ledgerera ~ StandardCrypto
+  => FromCBOR (LedgerState era)
+  => PoolId
+  -> SerialisedLedgerState era
+  -> ExceptT ShelleyQueryCmdError IO ()
+writeStakeSnapshot (StakePoolKeyHash hk) qState =
+  case decodeLedgerState qState of
+    -- In the event of decode failure print the CBOR instead
+    Left bs -> firstExceptT ShelleyQueryCmdHelpersError $ pPrintCBOR bs
+
+    Right ledgerState -> do
+      -- Ledger State
+      let (LedgerState snapshot) = ledgerState
+
+      -- The three stake snapshots, obtained from the ledger state
+      let (SnapShots markS setS goS _) = esSnapshots $ nesEs snapshot
+
+      -- Calculate the three pool and active stake values for the given pool
+      liftIO . LBS.putStrLn $ encodePretty $ Stakes
+        { markPool = getPoolStake hk markS
+        , setPool = getPoolStake hk setS
+        , goPool = getPoolStake hk goS
+        , markTotal = getAllStake markS
+        , setTotal = getAllStake setS
+        , goTotal = getAllStake goS
+        }
+
+-- | Sum all the stake that is held by the pool
+getPoolStake :: KeyHash Shelley.Spec.Ledger.Keys.StakePool crypto -> SnapShot crypto -> Integer
+getPoolStake hash ss = pStake
+  where
+    Coin pStake = fold s
+    (Stake s) = poolStake hash (_delegations ss) (_stake ss)
+
+-- | Sum the active stake from a snapshot
+getAllStake :: SnapShot crypto -> Integer
+getAllStake (SnapShot stake _ _) = activeStake
+  where
+    Coin activeStake = fold . unStake $ stake
+
+-- | This function obtains the pool parameters, equivalent to the following jq query on the output of query ledger-state
+--   .nesEs.esLState._delegationState._pstate._pParams.<pool_id>
+writePoolParams :: forall era ledgerera. ()
+  => ShelleyLedgerEra era ~ ledgerera
+  => FromCBOR (LedgerState era)
+  => Crypto.Crypto (Era.Crypto ledgerera)
+  => Era.Crypto ledgerera ~ StandardCrypto
+  => PoolId
+  -> SerialisedLedgerState era
+  -> ExceptT ShelleyQueryCmdError IO ()
+writePoolParams (StakePoolKeyHash hk) qState =
+  case decodeLedgerState qState of
+    -- In the event of decode failure print the CBOR instead
+    Left bs -> firstExceptT ShelleyQueryCmdHelpersError $ pPrintCBOR bs
+
+    Right ledgerState -> do
+      let LedgerState snapshot = ledgerState
+      let poolState = _pstate $ _delegationState $ esLState $ nesEs snapshot
+
+      -- Pool parameters
+      let poolParams = Map.lookup hk $ _pParams poolState
+      let fPoolParams = Map.lookup hk $ _fPParams poolState
+      let retiring = Map.lookup hk $ _retiring poolState
+
+      liftIO . LBS.putStrLn $ encodePretty $ Params poolParams fPoolParams retiring
+
+decodeLedgerState :: forall era. ()
+  => FromCBOR (LedgerState era)
+  => SerialisedLedgerState era
+  -> Either LBS.ByteString (LedgerState era)
+decodeLedgerState (SerialisedLedgerState (Serialised ls)) = first (const ls) (decodeFull ls)
 
 writeProtocolState :: Crypto.Crypto StandardCrypto
                    => Maybe OutputFile
@@ -629,6 +756,7 @@ obtainLedgerEraClassConstraints
   -> ((Ledger.ShelleyBased ledgerera
       , ToJSON (LedgerState era)
       , FromCBOR (LedgerState era)
+      , Era.Crypto ledgerera ~ StandardCrypto
       ) => a) -> a
 obtainLedgerEraClassConstraints ShelleyBasedEraShelley f = f
 obtainLedgerEraClassConstraints ShelleyBasedEraAllegra f = f
--- a/cardano-cli/src/Cardano/CLI/Types.hs
+++ b/cardano-cli/src/Cardano/CLI/Types.hs
@@ -15,16 +15,23 @@ module Cardano.CLI.Types
   , TxOutAnyEra (..)
   , UpdateProposalFile (..)
   , VerificationKeyFile (..)
+  , Stakes (..)
+  , Params (..)
   ) where
 
 import           Cardano.Prelude
 
+import           Data.Aeson (ToJSON (..), object, pairs, (.=))
 import qualified Data.Aeson as Aeson
 import qualified Data.Text as Text
 
 import qualified Cardano.Chain.Slotting as Byron
 
 import           Cardano.Api
+
+import qualified Cardano.Ledger.Crypto as Crypto
+
+import           Shelley.Spec.Ledger.TxBody (PoolParams (..))
 
 -- | Specify what the CBOR file is
 -- i.e a block, a tx, etc
@@ -61,6 +68,65 @@ data QueryFilter
   = FilterByAddress !(Set AddressAny)
   | NoFilter
   deriving (Eq, Show)
+
+-- | This data structure is used to allow nicely formatted output within the query stake-snapshot command.
+--
+-- "markPool", "setPool", "goPool" are the three ledger state stake snapshots (from most recent to least recent)
+-- go is the snapshot that is used for the current epoch, set will be used in the next epoch,
+-- mark for the epoch after that.  "markTotal", "setTotal", "goTotal" record the total active stake for each snapshot.
+--
+-- This information can be used by community tools to calculate upcoming leader schedules.
+data Stakes =  Stakes
+  { markPool :: Integer
+  , setPool :: Integer
+  , goPool :: Integer
+  , markTotal :: Integer
+  , setTotal :: Integer
+  , goTotal :: Integer
+  } deriving Show
+
+-- | Pretty printing for stake information
+instance ToJSON Stakes where
+  toJSON (Stakes m s g mt st gt) = object
+    [ "poolStakeMark" .= m
+    , "poolStakeSet" .= s
+    , "poolStakeGo" .= g
+    , "activeStakeMark" .= mt
+    , "activeStakeSet" .= st
+    , "activeStakeGo" .= gt
+    ]
+
+  toEncoding  (Stakes m s g mt st gt) = pairs $ mconcat
+    [ "poolStakeMark" .= m
+    , "poolStakeSet" .= s
+    , "poolStakeGo" .= g
+    , "activeStakeMark" .= mt
+    , "activeStakeSet" .= st
+    , "activeStakeGo" .= gt
+    ]
+
+-- | This data structure is used to allow nicely formatted output in the query pool-params command.
+-- params are the current pool parameter settings, futureparams are new parameters, retiringEpoch is the
+-- epoch that has been set for pool retirement.  Any of these may be Nothing.
+data Params crypto = Params
+  { poolParameters :: Maybe (PoolParams crypto)
+  , futurePoolParameters :: Maybe (PoolParams crypto)
+  , retiringEpoch :: Maybe EpochNo
+  } deriving Show
+
+-- | Pretty printing for pool parameters
+instance Crypto.Crypto crypto =>  ToJSON (Params crypto) where
+  toJSON (Params p fp r) = object
+    [ "poolParams" .= p
+    , "futurePoolParams" .= fp
+    , "retiring" .= r
+    ]
+
+  toEncoding (Params p fp r) = pairs $ mconcat
+    [ "poolParams" .= p
+    , "futurePoolParams" .= fp
+    , "retiring" .= r
+    ]
 
 newtype SigningKeyFile = SigningKeyFile
   { unSigningKeyFile :: FilePath }
--- a/doc/stake-pool-operations/query_stakepool.md
+++ b/doc/stake-pool-operations/query_stakepool.md
@@ -1,0 +1,112 @@
+# Querying a Stake Pool
+
+Two queries are available for querying your stakepool:
+
+* `stake-snapshot` (advanced): Get the stake snapshot information for a stake pool
+* `pool-params` (advanced): Get the current and future parameters for a stake pool,
+  including retirement
+
+## Querying for stake snapshot
+
+The stake snapshot returns information about the mark, set, go ledger snapshots for a pool, plus
+the total active stake for each snapshot that can be used in a 'sigma' calculation:
+
+```bash
+$ cardano-cli query stake-snapshot \
+    --stake-pool-id 00beef0a9be2f6d897ed24a613cf547bb20cd282a04edfc53d477114 \
+    --mainnet
+{
+    "poolStakeGo": 40278547538358,
+    "activeStakeGo": 22753958467474959,
+    "poolStakeMark": 40424218559492,
+    "activeStakeMark": 22670949084364797,
+    "poolStakeSet": 39898761956772,
+    "activeStakeSet": 22488877070796904
+}
+```
+
+Each snapshot is taken at the end of a different era.  The `go` snapshot is the current one and
+was taken two epochs earlier, `set` was taken one epoch ago, and `mark` was taken immediately
+before the start of the current epoch.
+
+This command if for debugging purposes only and may fail when used in a memory constrained
+environment due to the size of the ledger state.
+
+# Querying for pool pparameters
+
+The pool parameters command returns three pieces of information: current parameters, future
+parameters and retiring information.
+
+They may be `null` if eg the parameters are not changing.
+
+```bash
+$ cardano-cli query pool-params \
+    --stake-pool-id d785ff6a030ae9d521770c00f264a2aa423e928c85fc620b13d46eda \
+    --mainnet
+{
+    "poolParams": {
+        "publicKey": "d785ff6a030ae9d521770c00f264a2aa423e928c85fc620b13d46eda",
+        "cost": 340000000,
+        "metadata": {
+            "hash": "b150b12a1301c4b1510ac8b9f53f7571cabb43455f6fd244cd8fd97504b1c869",
+            "url": "https://adalite.io/ADLT4-metadata.json"
+        },
+        "owners": [
+            "463a9695c9222183ee6e1523478722bebcb332fa3769f1d8ef40c7d0",
+            "5049c1dac0e597ee902f27a74a167cf135ae7c1717b0d3a417cd6c67"
+        ],
+        "vrf": "0a21e37b1917ce37a897eb2a8dc6715973a18d0586f7ab4962e3975561151348",
+        "pledge": 30000000000,
+        "margin": 3.0e-2,
+        "rewardAccount": {
+            "network": "Mainnet",
+            "credential": {
+                "key hash": "b1bc146a5fb0683c4e3836712d115b98619048bc307cc059b6adc76e"
+            }
+        },
+        "relays": [
+            {
+                "single host address": {
+                    "IPv6": null,
+                    "port": 3003,
+                    "IPv4": "54.228.75.154"
+                }
+            },
+            {
+                "single host address": {
+                    "IPv6": null,
+                    "port": 3001,
+                    "IPv4": "54.228.75.154"
+                }
+            },
+            {
+                "single host address": {
+                    "IPv6": null,
+                    "port": 3003,
+                    "IPv4": "34.249.11.89"
+                }
+            },
+            {
+                "single host address": {
+                    "IPv6": null,
+                    "port": 3001,
+                    "IPv4": "34.249.11.89"
+                }
+            }
+        ]
+    },
+    "futurePoolParams": null,
+    "retiring": null
+}
+```
+
+The main advantage of these commands over using `query ledger-state` is that they avoid the need
+to dump the full ledger state (which is both time consuming and memory intensive - meaning they
+reduce the total system demands for SPOs), and will make it easier to support CNCLI and other
+tools.
+
+They also use existing internal operations (such as the ledger pool stake and active stake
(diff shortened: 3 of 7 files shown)


cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -80,6 +89,7 @@ renderShelleyQueryCmdError err =
     ShelleyQueryCmdHelpersError helpersErr -> renderHelpersError helpersErr
     ShelleyQueryCmdAcquireFailure aqFail -> Text.pack $ show aqFail
     ShelleyQueryCmdByronEra -> "This query cannot be used for the Byron era"
+    ShelleyQueryCmdPoolIdError -> "The pool id does not exist"
```
Jimbo4350, line 92: What would be very useful to the user would be to include the `PoolId` and render it here.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -232,6 +246,62 @@ runQueryUTxO (AnyConsensusModeParams cModeParams)
...
+  SocketPath sockPath <- firstExceptT ShelleyQueryCmdEnvVarSocketErr readEnvSocketPath
+  let localNodeConnInfo = LocalNodeConnectInfo cModeParams network sockPath
+
+  anyE@(AnyCardanoEra era) <- determineEra cModeParams localNodeConnInfo
+  let cMode = consensusModeOnly cModeParams
+  sbe <- getSbe $ cardanoEraStyle era
+
+  case toEraInMode era cMode of
+    Just eInMode -> do
+      let qInMode = QueryInEra eInMode
```
Jimbo4350, line 265: :+1: Nice

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -380,6 +450,150 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
      first (const ls) (decodeFull ls)
 
 
+-- | This local data structure is introduced purely to allow nicely formatted output and has no other purpose.
```
Jimbo4350, line 453: Great :+1:

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -380,6 +450,150 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
...
+-- mark, set, go are the three ledger state stake snapshots (most recent to least recent)
+-- go is the snapshot that is used for the current epoch, set will be used in the next epoch,
+-- mark for the epoch after that.  marktot etc record the total active stake for each snapshot
```
Jimbo4350, line 456: I think spelling error here: "marktot"

erikd:
How does this differ from #2536 ? Is that PR not addressing the same issue as this one?

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -380,6 +450,150 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
...
+-- This information can be used by community tools to calculate upcoming leader schedules
+data Stakes =  Stakes {
```
Jimbo4350, line 458: Let's move this type (along with its JSON instance) where it belongs, in `Cardano.CLI.Types`

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -380,6 +450,150 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
...
+                   -> ExceptT ShelleyQueryCmdError IO ()
+
+writeStakeSnapshot poolId qState =
+  case decodeLedgerState qState of
+    Left bs -> firstExceptT ShelleyQueryCmdHelpersError $ pPrintCBOR bs
+    Right ledgerState ->
+      if maybehk == Nothing then
+        left ShelleyQueryCmdPoolIdError
+      else
+        liftIO . LBS.putStrLn $  encodePretty $ Stakes {mark = markStake,set = setStake,go = goStake,marktot = markTotal,settot = setTotal,gotot = goTotal}
```
Jimbo4350, line 489: This line is pretty long, let's indent it so its less than 100 chars long.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -380,6 +450,150 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
...
+      where
+        -- Ledger State
+        (LedgerState snapshot)  = ledgerState
+
+        -- The three stake snapshots, obtained from the ledger state
+        (SnapShots markS setS goS _) = esSnapshots $ nesEs snapshot
+
+        -- Calculate the three pool and active stake values for the given pool
+        markStake =   getPoolStake hk markS
```
Jimbo4350, line 498: There is code alignment and some irregular spacing in this where clause. We don't align code in the cli (changes whilst keeping the alignment results in semantically null changes). Let's remove the alignment and fix the spacing before and after some of the `=`s in this clause.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -380,6 +450,150 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
...
+  case decodeLedgerState qState of
+    Left bs -> firstExceptT ShelleyQueryCmdHelpersError $ pPrintCBOR bs
+    Right ledgerState ->
+      if maybehk == Nothing then
+        left ShelleyQueryCmdPoolIdError
+      else
+        liftIO . LBS.putStrLn $  encodePretty $ Stakes {mark = markStake,set = setStake,go = goStake,marktot = markTotal,settot = setTotal,gotot = goTotal}
+      where
+        -- Ledger State
+        (LedgerState snapshot)  = ledgerState
```
Jimbo4350, line 492: irregulat spacing before equal sign

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -380,6 +450,150 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
...
+        markTotal =   getAllStake markS
+        setTotal =    getAllStake setS
+        goTotal =     getAllStake goS
+
+        -- Convert the hash string into a KeyHash for use by the ledger
+        maybehk = hashFromStringAsHex $ filter (/= '"') $ show poolId
+        hk = KeyHash $ fromJust $ maybehk
+
+ where
+   decodeLedgerState
```
Jimbo4350, line 511: Hm...this command may hang because of the size of the ledger state. Context: https://input-output-rnd.slack.com/archives/CC0CLNEH1/p1617065989042000

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -380,6 +450,150 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
...
+   getAllStake :: SnapShot crypto
+               -> Integer
+   getAllStake (SnapShot stake _ _) = activeStake
+      where
+        Coin activeStake = fold . unStake $ stake
+
+-- | This local data structure is introduced purely to allow nicely formatted output and has no other purpose.
+-- params are the current pool parameter settings, futureparams are new parameters, retiringEpoch is the
+-- epoch that has been set for pool retirement.  Any of these may be Nothing
+data Params crypto =  Params {
```
Jimbo4350, line 537: This also belongs in `Cardano.CLI.Types`

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -380,6 +450,150 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
...
+
+writePoolParams :: forall era ledgerera.
+                   ShelleyLedgerEra era ~ ledgerera
+                => FromCBOR (LedgerState era)
+                => Crypto.Crypto (Era.Crypto ledgerera)
+                => Hash StakePoolKey
+                -> SerialisedLedgerState era
+                -> ExceptT ShelleyQueryCmdError IO ()
+
+-- | This function obtains the pool parameters, equivalent to the following jq query on the output of query ledger-state
```
Jimbo4350, line 559: Please move comment to the top of the type signature.

Jimbo4350 (changes requested):
Moving in the right direction :+1:

Jimbo4350:
@kevinhammond We have a problem with these commands. They both depend on the ledger state query which currently hangs when used on mainnet. We don't have immediate plans to fix this now as the focus is eUTxO things. Context: https://input-output-rnd.slack.com/archives/CC0CLNEH1/p1617065989042000

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
kevinhammond: Done

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
kevinhammond: No it doesn't :).  A major point of the command is to produce a specialised and more efficient version of the ledger-state query that allows SPOs to reduce their memory requirements.   It uses a fraction of the memory.  Example call on Mainnet.

```
time getsnapshot 00beef0a9be2f6d897ed24a613cf547bb20cd282a04edfc53d477114
{
    "poolStakeGo": 40278547538358,
    "activeStakeGo": 22753958467474959,
    "poolStakeMark": 40424218559492,
    "activeStakeMark": 22670949084364797,
    "poolStakeSet": 39898761956772,
    "activeStakeSet": 22488877070796904
}

real	0m20.085s
user	0m8.874s
sys	0m2.585s

```

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
kevinhammond: > @kevinhammond We have a problem with these commands. They both depend on the ledger state query which currently hangs when used on mainnet. We don't have immediate plans to fix this now as the focus is eUTxO things. Context: https://input-output-rnd.slack.com/archives/CC0CLNEH1/p1617065989042000
> 
> > Please avoid in future any production use of the debug state dump features.It's there for debug and hacking / prototyping purposes. But it's not a stable format. It's too big. It's not designed for production use.
> 
> The way forward would likely be bugging the consensus folk to expose a suitable query

Here is the evidence that they do not (run live on mainnet).  The commands deliberately select a subset of the ledger state information, so reducing time and memory requirements for SPOs (who are currently using the ledger-state dump for production reasons).  The commands can, of course, be changed if the ledger state format changes or the consensus team exposes the required state.  The SPOs are very keen to have these available - they will help reduce memory requirements when provisioning a stake pool.

````
time cardano-cli query stake-snapshot --stake-pool-id 00beef0a9be2f6d897ed24a613cf547bb20cd282a04edfc53d477114 --mainnet
{
    "poolStakeGo": 40278547538358,
    "activeStakeGo": 22753958467474959,
    "poolStakeMark": 40424218559492,
    "activeStakeMark": 22670949084364797,
    "poolStakeSet": 39898761956772,
    "activeStakeSet": 22488877070796904
}

real	0m20.085s
user	0m8.874s
sys	0m2.585s
````

````
time cardano-cli query pool-params --stake-pool-id  00beef0a9be2f6d897ed24a613cf547bb20cd282a04edfc53d477114 --mainnet
{
    "poolStakeGo": 40278547538358,
    "activeStakeGo": 22753958467474959,
    "poolStakeMark": 40424218559492,
    "activeStakeMark": 22670949084364797,
    "poolStakeSet": 39898761956772,
    "activeStakeSet": 22488877070796904
}

real	0m20.085s
user	0m8.874s
sys	0m2.585s
````

**Comment by kevinhammond** [comment]
> How does this differ from #2536 ? Is that PR not addressing the same issue as this one?
> 
> Just creating a new PR creates more work for the people who have to review it.

The other PR was created as a draft (I had in fact only intended Jordan to look at it).  I did try updating it, but the code has moved on, and I was seeing many irrelevant and confusing changes to configuration files etc  I'll delete #2536

**Comment by Jimbo4350** [comment]
> @kevinhammond We have a problem with these commands. They both depend on the ledger state query which currently hangs when used on mainnet. We don't have immediate plans to fix this now as the focus is eUTxO things. Context: https://input-output-rnd.slack.com/archives/CC0CLNEH1/p1617065989042000
> 
> > Please avoid in future any production use of the debug state dump features.It's there for debug and hacking / prototyping purposes. But it's not a stable format. It's too big. It's not designed for production use.
> 
> The way forward would likely be bugging the consensus folk to expose a suitable query

After speaking with Kevin, his queries don't decode the entire ledger state and therefore have a smaller memory footprint.

**Comment by erikd** [comment]
> I did try updating it, but the code has moved on, and I was seeing many irrelevant and confusing changes to configuration files 

Git provides all the functionality needed to over come issues like this.

**Comment by kevinhammond** [comment]
> > I did try updating it, but the code has moved on, and I was seeing many irrelevant and confusing changes to configuration files
> 
> Git provides all the functionality needed to over come issues like this.

Yes, but it's notorious for having corners that are difficult/impossible to move out of.  In this case, I followed the recommended approach and hit one of those corners.  This avoided the risk of messing up the repo (the draft PR can be deleted now)

**Comment by Scitz0** [comment]
As a community tool builder, I would just like to add that this PR is very welcome and will help make our tools better and the end-user experience much improved. So anything that can be done to help push this over the finish line and get it included in an upcoming release is appreciated.

**Comment by AndrewWestberg** [comment]
Let's please get this through. We're currently taking 15 minutes in some cases to dump the entire ledger state to json (2gb) and takes even longer to decode it and sum up the active stake and total active stake for our pools.  This has become so painful that we have an external API built so not everybody has to go through this decoding process.  This PR, while not perfect, gets us what we need in around 20 seconds. It's head and shoulders over what we do today and we won't have to rely on an external API to get the info needed for cncli leaderlogs.

**Comment by kevinhammond** [comment]
> > @kevinhammond We have a problem with these commands. They both depend on the ledger state query which currently hangs when used on mainnet. We don't have immediate plans to fix this now as the focus is eUTxO things. Context: https://input-output-rnd.slack.com/archives/CC0CLNEH1/p1617065989042000
> > > Please avoid in future any production use of the debug state dump features.It's there for debug and hacking / prototyping purposes. But it's not a stable format. It's too big. It's not designed for production use.
> > 
> > 
> > The way forward would likely be bugging the consensus folk to expose a suitable query
> 
> After speaking with Kevin, his queries don't decode the entire ledger state and therefore have a smaller memory footprint.

Yes, seems to be 2.5GB for these commands versus 12GB for `query ledger-state` on my machine. 

I think the memory cost is essentially all "sunk cost" (for the ledger state), since it's only necessary to extract maps etc whose results are then folded over to calculate the sums etc.  In contrast, the full ledger state dump needs to traverse the whole structure and convert everything into JSON before it is output (it looks like it might even be producing the output strictly?).

In the long term, we can look at exposing more things through the consensus interface, but this should relieve the short term pressure to do that (and help SPOs reduce their memory requirements and reliance on external code)

**Comment by erikd** [comment]
@kevinhammond This PR is nowhere near ready to be merged yet, There are numerous "fixup" commits that from what I can see, should all be squashed down to a single commit. If you don't know how to do that, please ask one of the other devs.

File: cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs [hunk 12]
```diff
@@ -379,6 +457,109 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
...
+                   -> ExceptT ShelleyQueryCmdError IO ()
+
+writeStakeSnapshot poolId qState =
+  case decodeLedgerState qState of
+    Left bs -> firstExceptT ShelleyQueryCmdHelpersError $ pPrintCBOR bs
+    Right ledgerState ->
+      if isNothing maybehk then
+        left $ ShelleyQueryCmdPoolIdError poolId
+      else
+        liftIO . LBS.putStrLn $  encodePretty $ Stakes {markpool = markStake, setpool = setStake, gopool = goStake, marktot = markTotal, settot = setTotal, gotot = goTotal}
```
Jimbo4350, line 475: This line is well over our 100 char limit, lets indent it. Something like below would be more reasonable:
```
liftIO . LBS.putStrLn $ encodePretty 
  $ Stakes { markpool = markStake
           , setpool = setStake
           , gopool = goStake
           , marktot = markTotal
           , settot = setTotal
           , gotot = goTotal
           }
```

cardano-cli/src/Cardano/CLI/Types.hs
```diff
@@ -62,6 +69,42 @@ data QueryFilter
...
+-- mark for the epoch after that.  "marktot", "setttot", "gotot" record the total active stake for each snapshot
+-- This information can be used by community tools to calculate upcoming leader schedules
+data Stakes =  Stakes {
+      markpool, setpool, gopool :: Integer,
+      marktot, settot, gotot :: Integer
+    } deriving Show
+
+-- | Pretty printing for stake information
+instance ToJSON Stakes where
+    toJSON (Stakes m s g mt st gt) =
```
Jimbo4350, line 84: Should be 2 spaces indentation for the `ToJSON` instance

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -379,6 +457,109 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
...
+     -> Either LBS.ByteString (LedgerState era)
+   decodeLedgerState (SerialisedLedgerState (Serialised ls)) =
+     first (const ls) (decodeFull ls)
+
+   -- Sum all the stake that is held by the pool
+   getPoolStake :: KeyHash Shelley.Spec.Ledger.Keys.StakePool crypto
+                -> SnapShot crypto
+                -> Integer
+   getPoolStake hk ss = pStake
+      where
```
Jimbo4350, line 507: We don't need to nest these `where` clauses. We can write it as one `where` clause:
```
      where
        Coin pStake = fold s
        (Stake s) = poolStake hk (_delegations ss) (_stake ss)
  ```

File: cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs [hunk 15]
```diff
@@ -45,18 +46,24 @@ import           Cardano.CLI.Shelley.Parsers (OutputFile (..), QueryCmd (..))
...
 import           Ouroboros.Network.Protocol.LocalStateQuery.Type as LocalStateQuery
                    (AcquireFailure (..))
 import qualified Shelley.Spec.Ledger.API.Protocol as Ledger
+import           Shelley.Spec.Ledger.Coin
+import           Shelley.Spec.Ledger.EpochBoundary
+import           Shelley.Spec.Ledger.Keys (KeyHash (..), KeyRole (..))
+import           Shelley.Spec.Ledger.LedgerState hiding (LedgerState, _delegations)
 import           Shelley.Spec.Ledger.Scripts ()
 
+
```
Jimbo4350, line 66: unnecessary line addition here

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -379,6 +457,109 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
...
+                -> ExceptT ShelleyQueryCmdError IO ()
+writePoolParams poolId qState =
+  case decodeLedgerState qState of
+    Left bs -> firstExceptT ShelleyQueryCmdHelpersError $ pPrintCBOR bs
+    Right ledgerState ->
+      if isNothing maybehk then
+        left $ ShelleyQueryCmdPoolIdError poolId
+      else
+        liftIO . LBS.putStrLn $  encodePretty $ Params poolparams fpoolparams retiring
+      where
```
Jimbo4350, line 536: Let's move the functions/values from this where clause to the outer where clause. No need to have two.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -379,6 +457,109 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
...
+
+writeStakeSnapshot poolId qState =
+  case decodeLedgerState qState of
+    Left bs -> firstExceptT ShelleyQueryCmdHelpersError $ pPrintCBOR bs
+    Right ledgerState ->
+      if isNothing maybehk then
+        left $ ShelleyQueryCmdPoolIdError poolId
+      else
+        liftIO . LBS.putStrLn $  encodePretty $ Stakes {markpool = markStake, setpool = setStake, gopool = goStake, marktot = markTotal, settot = setTotal, gotot = goTotal}
+      where
```
Jimbo4350, line 476: Let's move the functions/values from this where clause to the outer where clause. No need to have two.

Jimbo4350 (changes requested):
Looking a lot better, final changes requested and then I'll be happy to approce :+1:

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
kevinhammond: Apart from `hk`, the definitions depend on the match to  `Right ledgerstate`, so cannot be lifted out.  I have pushed the function definitions into the inner scope since they are not currently used in the Left branch

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
kevinhammond: Again, this is not possible, so I have pushed the function definitions into the inner scope

kevinhammond: Scrap that.  decodeLedgerState is used to generate the Right so can't be pushed inwards.  I've lifted it to the module level instead

newhoggy:
After doing some testing, it looks like only `stake-snapshot` has an issue.  Node usage grows to `5GB`.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -375,13 +441,89 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
...
+  => Hash StakePoolKey
+  -> SerialisedLedgerState era
+  -> ExceptT ShelleyQueryCmdError IO ()
+writeStakeSnapshot poolId qState =
+  case decodeLedgerState qState of
+    Left bs -> {- firstExceptT ShelleyQueryCmdHelpersError $ pPrintCBOR bs -} do
+      result <- lift $ runExceptT (pPrintCBOR bs)
+      case result of
+        Left e -> left (ShelleyQueryCmdHelpersError e)
+        Right _ -> return ()
```
newhoggy, line 458: As far as I can tell, the commented code `firstExceptT ShelleyQueryCmdHelpersError $ pPrintCBOR bs` is equivalent to the expansion below it.

This seems wrong.  We tried to decode the ledger state and failed, and yet there still seems to be a way to return a success.

Jimbo4350: We should have something like:
```
        liftIO $ putTextLn "Version mismatch between node and consensus, so dumping this as generic CBOR."
        firstExceptT ShelleyQueryCmdHelpersError $ pPrintCBOR lbs
        ```

Comment (newhoggy) [reply cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs:458]: Right, but shouldn't it still fail?

Comment (Jimbo4350) [reply cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs:458]: I don't think it should fail as it's a debugging command and there was never a guarantee that the versions would match. Printing why you're getting a CBOR dump instead of JSON is good enough IMO.

Comment (newhoggy) [reply cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs]: Ok cool.  I've restored the original and added an explanatory comment.

**Comment by newhoggy** [comment]
Summary of changes:

* Minimise differences with code on `master`.
* Tidy up coding style for change and new code
* Use hoistMaybe to flatten code to minimise `case` matching make the happy path more obvious.
* Reduce use of where clauses, especially nested where clauses.
* Add documentation.

File: cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs [hunk 19]
```diff
@@ -375,13 +435,87 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
...
+  -> SerialisedLedgerState era
+  -> ExceptT ShelleyQueryCmdError IO ()
+writeStakeSnapshot poolId qState =
+  case decodeLedgerState qState of
+    -- In the event of decode failure print the CBOR instead
+    Left bs -> firstExceptT ShelleyQueryCmdHelpersError $ pPrintCBOR bs
+
+    Right ledgerState -> do
+      -- Convert the hash string into a KeyHash for use by the ledger
+      hk <- KeyHash <$> hashFromStringAsHex (filter (/= '"') (show poolId))
```
Jimbo4350, line 453: We _should_ be able to use `serialiseToRawBytesHexText` here instead of `hashFromStringAsHex (filter (/= '"') (show poolId)`

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -375,13 +435,87 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
...
-     :: SerialisedLedgerState era
-     -> Either LBS.ByteString (LedgerState era)
-   decodeLedgerState (SerialisedLedgerState (Serialised ls)) =
-     first (const ls) (decodeFull ls)
 
+writeStakeSnapshot :: forall era ledgerera. ()
+  => ShelleyLedgerEra era ~ ledgerera
+  => Era.Era ledgerera
+  => FromCBOR (LedgerState era)
+  => Hash StakePoolKey
```
Jimbo4350, line 443: It may be clearer to use the type synonym `PoolId` here

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -375,13 +435,87 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
...
+  case decodeLedgerState qState of
+    -- In the event of decode failure print the CBOR instead
+    Left bs -> firstExceptT ShelleyQueryCmdHelpersError $ pPrintCBOR bs
+
+    Right ledgerState -> do
+      let LedgerState snapshot = ledgerState
+      let poolState = _pstate $ _delegationState $ esLState $ nesEs snapshot
+
+      -- Convert the hash string into a KeyHash for use by the ledger
+      hk <- KeyHash <$> hashFromStringAsHex (filter (/= '"') (show poolId))
```
Jimbo4350, line 504: Same here

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -375,13 +435,87 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
...
+  where
+    Coin activeStake = fold . unStake $ stake
+
+-- | This function obtains the pool parameters, equivalent to the following jq query on the output of query ledger-state
+--   .nesEs.esLState._delegationState._pstate._pParams.<pool_id>
+writePoolParams :: forall era ledgerera. ()
+  => ShelleyLedgerEra era ~ ledgerera
+  => FromCBOR (LedgerState era)
+  => Crypto.Crypto (Era.Crypto ledgerera)
+  => Hash StakePoolKey
```
Jimbo4350, line 491: Same here

doc/stake-pool-operations/query_stakepool.md
```diff
@@ -0,0 +1,107 @@
+# Querying a Stake Pool
+
```
Jimbo4350, line 2: We should add something to the effect of "this command if for debugging purposes only and is not guaranteed to succeed due to the size of the ledger state"

Jimbo4350 (commented):
Looking good, a few minor comments

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: This seems to not work as `serialiseToRawBytesHexText` returns `Text` but I need a `Maybe (Hash (ADDRHASH (Crypto ledgerera)) (VerKeyDSIGN (DSIGN (Crypto ledgerera))))`

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: Fixed!

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: Fixed!

doc/stake-pool-operations/query_stakepool.md
newhoggy: Done!

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
Jimbo4350: [1mdiff --git a/cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs b/cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs[m
[1mindex 10c695228..a61afcf84 100644[m
[1m--- a/cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs[m
[1m+++ b/cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs[m
[36m@@ -438,22 +438,17 @@[m [mwriteLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =[m
 [m
 writeStakeSnapshot :: forall era ledgerera. ()[m
   => ShelleyLedgerEra era ~ ledgerera[m
[31m-  => Era.Era ledgerera[m
[32m+[m[32m  => Era.Crypto ledgerera ~ StandardCrypto[m
   => FromCBOR (LedgerState era)[m
   => PoolId[m
   -> SerialisedLedgerState era[m
   -> ExceptT ShelleyQueryCmdError IO ()[m
[31m-writeStakeSnapshot poolId qState =[m
[32m+[m[32mwriteStakeSnapshot (StakePoolKeyHash hk) qState =[m
   case decodeLedgerState qState of[m
     -- In the event of decode failure print the CBOR instead[m
     Left bs -> firstExceptT ShelleyQueryCmdHelpersError $ pPrintCBOR bs[m
 [m
     Right ledgerState -> do[m
[31m-      let x = hashFromStringAsHex (filter (/= '"') (show poolId))[m
[31m-      -- Convert the hash string into a KeyHash for use by the ledger[m
[31m-      hk <- KeyHash <$> x[m
[31m-        & hoistMaybe (ShelleyQueryCmdPoolIdError poolId)[m
[31m-[m
       -- Ledger State[m
       let (LedgerState snapshot) = ledgerState[m
 [m
[36m@@ -764,6 +759,7 @@[m [mobtainLedgerEraClassConstraints[m
   -> ((Ledger.ShelleyBased ledgerera[m
       , ToJSON (LedgerState era)[m
       , FromCBOR (LedgerState era)[m
[32m+[m[32m      , Era.Crypto ledgerera ~ StandardCrypto[m
       ) => a) -> a[m
 obtainLedgerEraClassConstraints ShelleyBasedEraShelley f = f[m
 obtainLedgerEraClassConstraints ShelleyBasedEraAllegra f = f[m

Jimbo4350:
```diff
[1mdiff --git a/cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs b/cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs[m
[1mindex 10c695228..a61afcf84 100644[m
[1m--- a/cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs[m
[1m+++ b/cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs[m
[36m@@ -438,22 +438,17 @@[m [mwriteLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =[m
 [m
 writeStakeSnapshot :: forall era ledgerera. ()[m
   => ShelleyLedgerEra era ~ ledgerera[m
[31m-  => Era.Era ledgerera[m
[32m+[m[32m  => Era.Crypto ledgerera ~ StandardCrypto[m
   => FromCBOR (LedgerState era)[m
   => PoolId[m
   -> SerialisedLedgerState era[m
   -> ExceptT ShelleyQueryCmdError IO ()[m
[31m-writeStakeSnapshot poolId qState =[m
[32m+[m[32mwriteStakeSnapshot (StakePoolKeyHash hk) qState =[m
   case decodeLedgerState qState of[m
     -- In the event of decode failure print the CBOR instead[m
     Left bs -> firstExceptT ShelleyQueryCmdHelpersError $ pPrintCBOR bs[m
 [m
     Right ledgerState -> do[m
[31m-      let x = hashFromStringAsHex (filter (/= '"') (show poolId))[m
[31m-      -- Convert the hash string into a KeyHash for use by the ledger[m
[31m-      hk <- KeyHash <$> x[m
[31m-        & hoistMaybe (ShelleyQueryCmdPoolIdError poolId)[m
[31m-[m
       -- Ledger State[m
       let (LedgerState snapshot) = ledgerState[m
 [m
[36m@@ -764,6 +759,7 @@[m [mobtainLedgerEraClassConstraints[m
   -> ((Ledger.ShelleyBased ledgerera[m
       , ToJSON (LedgerState era)[m
       , FromCBOR (LedgerState era)[m
[32m+[m[32m      , Era.Crypto ledgerera ~ StandardCrypto[m
       ) => a) -> a[m
 obtainLedgerEraClassConstraints ShelleyBasedEraShelley f = f[m
 obtainLedgerEraClassConstraints ShelleyBasedEraAllegra f = f[m

```

Jimbo4350:
```diff
diff --git a/cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs b/cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
index 10c695228..a61afcf84 100644
--- a/cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
+++ b/cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
@@ -438,22 +438,17 @@ writeLedgerState mOutFile qState@(SerialisedLedgerState serLedgerState) =
 
 writeStakeSnapshot :: forall era ledgerera. ()
   => ShelleyLedgerEra era ~ ledgerera
-  => Era.Era ledgerera
+  => Era.Crypto ledgerera ~ StandardCrypto
   => FromCBOR (LedgerState era)
   => PoolId
   -> SerialisedLedgerState era
   -> ExceptT ShelleyQueryCmdError IO ()
-writeStakeSnapshot poolId qState =
+writeStakeSnapshot (StakePoolKeyHash hk) qState =
   case decodeLedgerState qState of
     -- In the event of decode failure print the CBOR instead
     Left bs -> firstExceptT ShelleyQueryCmdHelpersError $ pPrintCBOR bs
 
     Right ledgerState -> do
-      let x = hashFromStringAsHex (filter (/= '"') (show poolId))
-      -- Convert the hash string into a KeyHash for use by the ledger
-      hk <- KeyHash <$> x
-        & hoistMaybe (ShelleyQueryCmdPoolIdError poolId)
-
       -- Ledger State
       let (LedgerState snapshot) = ledgerState
 
@@ -764,6 +759,7 @@ obtainLedgerEraClassConstraints
   -> ((Ledger.ShelleyBased ledgerera
       , ToJSON (LedgerState era)
       , FromCBOR (LedgerState era)
+      , Era.Crypto ledgerera ~ StandardCrypto
       ) => a) -> a
 obtainLedgerEraClassConstraints ShelleyBasedEraShelley f = f
 obtainLedgerEraClassConstraints ShelleyBasedEraAllegra f = f
```

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: Fixed!

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: Yep.  Worked.  Thanks!

newhoggy:
I've updated and retested.  Looks good.

Jimbo4350 (approved):
Awesome! Thanks @kevinhammond & @newhoggy. Just squash the commits before merging & notify QA

newhoggy:
bors merge

newhoggy (approved):

newhoggy:
bors merge

AndrewWestberg:
Whoo Hoo!

Crypto2099:
This is a huge and very welcome step. Thank you very much @kevinhammond for working to get this included natively in the node!

PR #2562, opened 2021-03-31
Use ToJSON instance for getting era in query tip command

newhoggy:


newhoggy:
Using `ToJSON` instance instead.  This results in the era name being capitalised:

```
CARDANO_NODE_SOCKET_PATH='/Users/jky/Library/Application Support/Daedalus Mainnet/cardano-node.socket' ./dist-newstyle/build/x86_64-osx/ghc-8.10.4/cardano-cli-1.26.1/x/cardano-cli/build/cardano-cli/cardano-cli query tip --mainnet
{
    "epoch": 256,
    "hash": "c7b692a73d16a8b3bc518c6a87fd9050376f3d0a389212bae90f34af08cd297f",
    "slot": 25608927,
    "block": 5530601,
    "era": "Mary"
}
```

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -19,6 +19,7 @@ import           Prelude (String)
 
 import           Data.Aeson (ToJSON (..), (.=))
 import qualified Data.Aeson as Aeson
+import qualified Data.Aeson as J
```
Jimbo4350, line 22: Is the qualified import necessary? We usually qualify as `Aeson`. I think we're ok to just use `toJSON` here.

newhoggy: Oops.  I missed that.  Fixed!

Jimbo4350 (approved):
LGTM!

PR #2581, opened 2021-04-07
Sync progress

newhoggy:


cardano-cli/src/Cardano/CLI/Slot.hs
```diff
@@ -0,0 +1,329 @@
...
+import qualified Ouroboros.Consensus.HardFork.History.Qry as HF
+
+data TimeInterpreterLog = MsgInterpreterPastHorizon
+  (Maybe String) -- ^ Reason for why the failure should be impossible
+  StartTime
+  PastHorizonException
+  deriving (Show)
+
+-- | Number of slots in a single epoch
+newtype EpochLength = EpochLength { unEpochLength :: Word32 }
```
nfrisby, line 62: I think you can use `EpochSize` from https://github.com/input-output-hk/cardano-base/blob/master/slotting/src/Cardano/Slotting/Slot.hs

cardano-cli/src/Cardano/CLI/Slot.hs
```diff
@@ -0,0 +1,329 @@
...
+    deriving (Show, Eq, Generic)
+
+-- | Duration of a single slot.
+newtype SlotLength = SlotLength { unSlotLength :: NominalDiffTime }
```
nfrisby, line 66: I'm in the midst of opening a `cardano-base` PR that will provide this definition in a new `Cardano.Slotting.Time` module.

It will also include `SystemStart`, which is the same as `StartTime`.

At the moment, both of these are exported by https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/src/Ouroboros/Consensus/BlockchainTime/WallClock/Types.hs but my imminent PR will move almost all of that module upstream to `cardano-base`.

cardano-cli/src/Cardano/CLI/Slot.hs
```diff
@@ -0,0 +1,329 @@
...
+  deriving stock (Generic, Eq, Ord)
+  deriving (Read, Show) via (Quiet (Hash tag))
+  deriving anyclass (NFData, Hashable)
+
+newtype SyncTolerance = SyncTolerance NominalDiffTime
+  deriving stock (Generic, Eq, Show)
+
+data SyncProgress
+  = Ready
+  | Syncing !(Quantity "percent" Percentage)
```
nfrisby, line 95: I don't think the `Quantity "percent"` layer adds any additional clarity to the `Percentage` type. What do you think?

cardano-cli/src/Cardano/CLI/Slot.hs
```diff
@@ -0,0 +1,329 @@
...
+--
+-- @HF.interpretQuery@ could interpret both individually, but
+--
+-- @
+--    q3 = q1 >> q2
+-- @
+--
+-- would fail.
+--
+-- This wrapper exists to fix this.
```
nfrisby, line 125: Do you still need this extra layer of `Qry`? In particular, I only see `interpretQuery ti $ slotToRelTime tipSlotNo` used here, and that should only require a single era. (I do see the PR is marked as Draft, so maybe there's more coming.)

cardano-cli/src/Cardano/CLI/Slot.hs
```diff
@@ -0,0 +1,329 @@
...
+-- | Set up a 'TimeInterpreter' for a given start time, and an 'Interpreter'
+-- queried from the ledger layer.
+mkTimeInterpreter
+    :: Monad m
+    => StartTime
+    -> EpochSize
+    -> Cardano.SlotLength
+    -> TimeInterpreter (ExceptT PastHorizonException m)
+mkTimeInterpreter start sz len = TimeInterpreter
+    { interpreter = pure (mkInterpreter (neverForksSummary sz len))
```
nfrisby, line 327: If I understand the context correctly, then the use of `neverForksSummary` is not the right way to acquire an `Interpreter`. IE We're concerned about syncing the Cardano chain, and that should does fork, so the "never forks" constructor for `Summary` doesn't fit this use-case.

I'll write my in-a-vacuum expectations in another comment.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -155,13 +158,34 @@ runQueryTip (AnyConsensusModeParams cModeParams) network mOutFile = do
   anyEra <- determineEra cModeParams localNodeConnInfo
   mEpoch <- mEpochQuery anyEra consensusMode  localNodeConnInfo
   tip <- liftIO $ getLocalChainTip localNodeConnInfo
+
+  let timeInterpreter = mkTimeInterpreter @IO
```
nfrisby, line 162: I think the anticipated path is to do something similar to what `mEpochQuery` already does, but submitting the `GetInterpreter` query instead of `QueryEpoch`.

https://github.com/input-output-hk/ouroboros-network/blob/dea09273c87578a2047445e4434ba30df6570ac2/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger/Query.hs#L312

But now that I'm looking more closely, I see there's a whole other query data type in the `cardano-api` that I wasn't expecting. So I'm less sure about what path to actually suggest. I'll ping Duncan on our Slack thread.

Summary: you almost certainly want to get the `Interpreter` by querying the local node, and the Consensus-defined `GetInterpreter` is ultimately the only query I know for doing that.

cardano-cli/src/Cardano/CLI/Slot.hs
newhoggy: There's not more coming.  This code was adapted from what I found in `cardano-wallet`, so I preserved the structure and am looking to untangle and simplify it.  If I can get away with single era for this query that would be great!

cardano-cli/src/Cardano/CLI/Slot.hs
newhoggy: I have removed `Quantity` as recommended.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: In attempting to add `GetInterpreter` here: https://github.com/input-output-hk/cardano-node/blob/3c556c5b09811c159dc7a92021f4d1c6aac3a104/cardano-api/src/Cardano/Api/Query.hs#L117

I would need to update `toConsensusQueryShelleyBased` https://github.com/input-output-hk/cardano-node/blob/3c556c5b09811c159dc7a92021f4d1c6aac3a104/cardano-api/src/Cardano/Api/Query.hs#L289

Does that mean I need extra constructors here?

https://github.com/input-output-hk/ouroboros-network/blob/ec086166c5492b1a407d8ea20b4e8c18589046cf/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Query.hs#L82

newhoggy: I can see [this definition](https://github.com/input-output-hk/ouroboros-network/blob/dea09273c87578a2047445e4434ba30df6570ac2/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger/Query.hs#L312):

```
data QueryHardFork xs result where
  GetInterpreter :: QueryHardFork xs (History.Interpreter xs)
```



data QueryHardFork xs result where
312 | GetInterpreter :: QueryHardFork xs (History.Interpreter xs)
…




data QueryHardFork xs result where
312 | GetInterpreter :: QueryHardFork xs (History.Interpreter xs)
…




data QueryHardFork xs result where
312 | GetInterpreter :: QueryHardFork xs (History.Interpreter xs)
…




data QueryHardFork xs result where
312 | GetInterpreter :: QueryHardFork xs (History.Interpreter xs)
…




data QueryHardFork xs result where
312 | GetInterpreter :: QueryHardFork xs (History.Interpreter xs)
…




data QueryHardFork xs result where
312 | GetInterpreter :: QueryHardFork xs (History.Interpreter xs)
…

newhoggy: [here](https://github.com/input-output-hk/cardano-node/blob/3b5cc66ef7607fc4cd85d3276eca7b47b913f7a1/cardano-api/src/Cardano/Api/Query.hs#L274)

```
toConsensusQuery (QueryCurrentEra CardanoModeIsMultiEra) =
    Some (Consensus.QueryHardFork Consensus.GetCurrentEra)
```

[here](https://github.com/input-output-hk/cardano-node/blob/3b5cc66ef7607fc4cd85d3276eca7b47b913f7a1/cardano-api/src/Cardano/Api/IPC.hs#L448)

```
convLocalStateQueryClient mode =
    Net.Query.mapLocalStateQueryClient
      (toConsensusPointInMode mode)
      toConsensusQuery
      fromConsensusQueryResult
```

[here](https://github.com/input-output-hk/cardano-node/blob/3b5cc66ef7607fc4cd85d3276eca7b47b913f7a1/cardano-api/src/Cardano/Api/IPC.hs#L394)

```
convLocalNodeClientProtocols
    mode
    LocalNodeClientProtocols {
      localChainSyncClient,
      localTxSubmissionClient,
      localStateQueryClient
    } =
    LocalNodeClientProtocolsForBlock {
      localChainSyncClientForBlock    = case localChainSyncClient of
        NoLocalChainSyncClient -> NoLocalChainSyncClient
        LocalChainSyncClientPipelined clientPipelined -> LocalChainSyncClientPipelined $ convLocalChainSyncClientPipelined mode clientPipelined
        LocalChainSyncClient client -> LocalChainSyncClient $ convLocalChainSyncClient mode client,

      localTxSubmissionClientForBlock = convLocalTxSubmissionClient mode <$>
                                          localTxSubmissionClient,

      localStateQueryClientForBlock   = convLocalStateQueryClient mode <$>
                                          localStateQueryClient
    }
```

[here](https://github.com/input-output-hk/cardano-node/blob/3b5cc66ef7607fc4cd85d3276eca7b47b913f7a1/cardano-api/src/Cardano/Api/IPC.hs#L344)

```
mkLocalNodeClientParams :: forall mode block.
                           ConsensusBlockForMode mode ~ block
                        => ConsensusModeParams mode
                        -> LocalNodeClientProtocolsInMode mode
                        -> LocalNodeClientParams
mkLocalNodeClientParams modeparams clients =
    -- For each of the possible consensus modes we pick the concrete block type
    -- (by picking the appropriate 'Consensus.ProtocolClient' value).
    --
    -- Though it is not immediately visible, this point where we use
    -- 'LocalNodeClientParams' is also where we pick up the necessary class
    -- instances. This works because in each case we have a monomorphic block
    -- type and the instances are all in scope. This is why the use of
    -- LocalNodeClientParams is repeated within each branch of the case:
    -- because it is only within each branch that the GADT match makes the
    -- block type monomorphic.
    --
    case modeparams of
      ByronModeParams epochSlots ->
        LocalNodeClientParams
          (Consensus.ProtocolClientByron epochSlots)
          (convLocalNodeClientProtocols ByronMode clients)

      ShelleyModeParams ->
        LocalNodeClientParams
          Consensus.ProtocolClientShelley
          (convLocalNodeClientProtocols ShelleyMode clients)

      CardanoModeParams epochSlots ->
        LocalNodeClientParams
          (Consensus.ProtocolClientCardano epochSlots)
          (convLocalNodeClientProtocols CardanoMode clients)
```

[here](https://github.com/input-output-hk/cardano-node/blob/3b5cc66ef7607fc4cd85d3276eca7b47b913f7a1/cardano-api/src/Cardano/Api/IPC.hs#L179)

```
connectToLocalNode LocalNodeConnectInfo {
                     localNodeSocketPath,
                     localNodeNetworkId,
                     localConsensusModeParams
                   } clients =
    Net.withIOManager $ \iomgr ->
      Net.connectTo
        (Net.localSnocket iomgr localNodeSocketPath)
        Net.NetworkConnectTracers {
          Net.nctMuxTracer       = nullTracer,
          Net.nctHandshakeTracer = nullTracer
        }
        versionedProtocls
        localNodeSocketPath
  where
    versionedProtocls =
      -- First convert from the mode-parametrised view of things to the
      -- block-parametrised view and then do the final setup for the versioned
      -- bundles of mini-protocols.
      case mkLocalNodeClientParams localConsensusModeParams clients of
        LocalNodeClientParams ptcl clients' ->
          mkVersionedProtocols localNodeNetworkId ptcl clients'
```

[here](https://github.com/input-output-hk/cardano-node/blob/3b5cc66ef7607fc4cd85d3276eca7b47b913f7a1/cardano-api/src/Cardano/Api/IPC.hs#L463)

```
queryNodeLocalState :: forall mode result.
                       LocalNodeConnectInfo mode
                    -> Maybe ChainPoint
                    -> QueryInMode mode result
                    -> IO (Either Net.Query.AcquireFailure result)
queryNodeLocalState connctInfo mpoint query = do
    resultVar <- newEmptyTMVarIO
    connectToLocalNode
      connctInfo
      LocalNodeClientProtocols {
        localChainSyncClient    = NoLocalChainSyncClient,
        localStateQueryClient   = Just (singleQuery mpoint resultVar),
        localTxSubmissionClient = Nothing
      }
    atomically (takeTMVar resultVar)
```

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -183,13 +203,56 @@ runQueryTip (AnyConsensusModeParams cModeParams) network mOutFile = do
...
+            case eResult of
+              Left acqFail -> left (ShelleyQueryCmdGeneric (show acqFail))
+              Right eraHistory -> return $ getProgress slotNo eraHistory
+          MaryEraInCardanoMode -> do
+            let epochQuery = QueryEraHistory CardanoModeIsMultiEra  -- QueryInShelleyBasedEra sbe QueryEpoch
+            eResult <- liftIO $ queryNodeLocalState lNodeConnInfo Nothing epochQuery
+            case eResult of
+              Left acqFail -> left (ShelleyQueryCmdGeneric (show acqFail))
+              Right eraHistory -> return $ getProgress slotNo eraHistory
+          mode -> left (ShelleyQueryCmdGeneric ("Not cardano mode: " <> show mode))
```
newhoggy, line 240: Too much repetition above.  How to reduce?

newhoggy:
## System Start

* [`gdStartTime`][1]
* [`sgSystemStart`][2]


[1]: https://github.com/input-output-hk/cardano-ledger-specs/blob/7d55693a4b8104e6ac897b6ccbcf5dd51f320632/byron/ledger/impl/src/Cardano/Chain/Genesis/Data.hs#L60
[2]: https://github.com/input-output-hk/cardano-ledger-specs/blob/7d55693a4b8104e6ac897b6ccbcf5dd51f320632/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/Genesis.hs#L127

newhoggy:
Finally got this to work:

```
CARDANO_NODE_SOCKET_PATH=local-node/socket cardano-cli query tip --mainnet
{
    "epoch": 261,
    "now": 112875500,
    "tipTime": 112875497,
    "hash": "49719ff677b73f431a3f47e7b5469aaed2c3f62b2b33022a6d35dfafec693b3b",
    "slot": 27512297,
    "block": 5622849,
    "era": "Mary",
    "syncProgress": "100.00"
}
```

Here is an unsynced version:

```
CARDANO_NODE_SOCKET_PATH=local-node-2/socket cardano-cli query tip --mainnet
{
    "epoch": null,
    "now": 112875455,
    "tipTime": 8874060,
    "hash": "127c70fe1416b6aa053731270378442cd80b362732bec1971cf68ade23494557",
    "slot": 443703,
    "block": 443647,
    "era": "Byron",
    "syncProgress": "7.86"
}
```

johnalotoski:
Adding a `hydra-dont-build` label to this PR to avoid hydra repeatedly trying to re-eval the error in the eval.  Please remove this label once you'd like hydra to try building it again.

PR #2582, opened 2021-04-07
CAD-2765: trace-forward library prototype

denisshevchenko:
This PR introduces `trace-forward`: a new library for forwarding `LogObject`s to an external program (for example, logger or RTView). It is a **_replacement_** for the old plugins [`trace-forwarder`](https://github.com/input-output-hk/iohk-monitoring-framework/tree/master/plugins/backend-trace-forwarder) and [`trace-acceptor`](https://github.com/input-output-hk/iohk-monitoring-framework/tree/master/plugins/backend-trace-acceptor) from `iohk-monitoring-framework`.

This library is based on `typed-protocols`. There are two pairs of demo-programs: how to use this library [separately](https://github.com/input-output-hk/cardano-node/tree/cad-2765-trace-forward-prototype/trace-forward/demo) and how to combine it with another library [ekg-forward](https://github.com/input-output-hk/ekg-forward) using `Mux`. The second demo is a practical example of how to integrate these two libraries in `cardano-node` (because eventually the node should forward _both_ `LogObject`s and EKG-metrics).

Please note that currently it's implemented _without_ pipelining: our benchmarks show that a consumer can ask `LogObject`s and EKG-metrics quite fast.


trace-forward/README.md
```diff
@@ -0,0 +1,9 @@
+# trace-forward
+
+## What Is It
+
+`trace-forward` is a library allowing to forward `LogObject`s from one process to another one. The node is using this library to forward different `LogObject`s to exernal acceptors (for example, logger or [RTView](https://github.com/input-output-hk/cardano-rt-view)).
```
deepfire, line 5: Not just `LogObject`s, but also metrics -- we're using two protocols after all.

Jimbo4350 (changes requested):
Nice, but why is this not in `iohk-monitoring`? This doesn't belong in `cardano-node`.

erikd:
Totally agree with @Jimbo4350 . the `node` repo is not dumping ground for code that may or may not be maintained. Especially not anything marked as a "protoype".

trace-forward/src/Trace/Forward/Acceptor.hs
```diff
@@ -0,0 +1,37 @@
...
+runTraceAcceptor
+  :: (CBOR.Serialise a,
+      ShowProxy a,
+      Typeable a)
+  => AcceptorConfiguration a  -- ^ Acceptor configuration.
+  -> TBQueue (LogObject a)    -- ^ The queue all received 'LogObject's will be write in.
+  -> IO ()
+runTraceAcceptor config loQueue =
+  try (void $ listenToForwarder config loQueue) >>= \case
+    Left (_e :: SomeException) ->
```
coot, line 35: I don't think that catching `AsyncExceptions` here is what one would like here.

trace-forward/src/Trace/Forward/Forwarder.hs
```diff
@@ -0,0 +1,42 @@
...
+runTraceForwarder
+  :: (CBOR.Serialise a,
+      ShowProxy a,
+      Typeable a)
+  => ForwarderConfiguration a  -- ^ Forwarder configuration.
+  -> TBQueue (LogObject a)     -- ^ The queue the forwarder will take 'LogObject's from.
+  -> IO ()
+runTraceForwarder config@ForwarderConfiguration{..} loQueue =
+  try (connectToAcceptor config loQueue) >>= \case
+    Left (_e :: SomeException) -> do
```
coot, line 36: The same here.

trace-forward/src/Trace/Forward/Network/Acceptor.hs
```diff
@@ -0,0 +1,157 @@
...
+doListenToForwarder
+  :: Ord addr
+  => Snocket IO fd addr
+  -> addr
+  -> ProtocolTimeLimits (Handshake UnversionedProtocol Term)
+  -> OuroborosApplication 'ResponderMode addr LBS.ByteString IO Void ()
+  -> IO Void
+doListenToForwarder snocket address timeLimits app = do
+  networkState <- newNetworkMutableState
+  _ <- async $ cleanNetworkMutableState networkState
```
coot, line 83: Have you thought about async exceptions when using `async`? I don't think you need that spawned thread to run if the parent thread is killed.

coot (changes requested):

denisshevchenko:
> the node repo is not dumping ground for code that may or may not be maintained.

Why "may or may not be maintained"? This code _will_ be maintained by our team.

> Especially not anything marked as a "prototype".

It's a prototype just because it's the first version. But I can remove this word. :-)

trace-forward/src/Trace/Forward/Network/Acceptor.hs
denisshevchenko: > I don't think you need that spawned thread to run if the parent thread is killed.

@coot Of course, I don't need it. Should I use spawning with automatic cancelation instead?

trace-forward/src/Trace/Forward/Acceptor.hs
denisshevchenko: Sorry, I didn't understand. This code works as expected: when the network connection is broken, I catch the exception and try to re-establish the connection. Should I do it in some other way?

trace-forward/src/Trace/Forward/Network/Acceptor.hs
coot: You probably want `race` to avoid the situation when the `cleanNetworkMutableState` thread errored but the main thread continues.

coot:
`cardano-node` is an integration point, so the question is do you need anything from `cardano-node` to implement the `trace-forwarder`?

denisshevchenko:
> do you need anything from `cardano-node` to implement the `trace-forward`?

@coot What do you mean by `cardano-node` in this case? Anything from `cardano-node` _repository_ or anything from https://github.com/input-output-hk/cardano-node/tree/master/cardano-node?

coot:
Either :).

denisshevchenko:
@coot Well, right **_now_** - no, `trace-forward` doesn't have any dependencies from `cardano-node`.

trace-forward/src/Trace/Forward/Network/Acceptor.hs
denisshevchenko: Thanks. I did it using `waitAnyCancel` (solution with `race_` didn't pass tests).

trace-forward/src/Trace/Forward/Acceptor.hs
coot: Try to kill the program with `Ctrl-C`, and you'll see that the program does not exit, but keeps on going.  You could use something like this to avoid catching async exceptions:
```
tryJust (\e -> case fromException e of { SomeAsyncException {} -> Nothing; _ -> Just e) 
```

erikd:
> Why "may or may not be maintained"? This code will be maintained by our team.

That still does not explain why it should be in the `node` repo and not in `iohk-minitoring` repo.

erikd (changes requested):
This does not belong in the `node` repo.

denisshevchenko:
It will be merged in another PR.

PR #2587, opened 2021-04-08
Cad 2658 new stake script file option

newhoggy:
Implements remaining parts of https://github.com/input-output-hk/cardano-node/issues/2333

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
+++ b/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
@@ -22,8 +22,8 @@ import           Cardano.Api.Shelley
 import           Cardano.CLI.Mary.TxOutParser (parseTxOutAnyEra)
 import           Cardano.CLI.Mary.ValueParser (parseValue)
 import           Cardano.CLI.Shelley.Commands
-import           Cardano.CLI.Shelley.Key (InputFormat (..), PaymentSource (..),
-                   VerificationKeyOrFile (..), VerificationKeyOrHashOrFile (..),
+import           Cardano.CLI.Shelley.Key (InputFormat (..), PaymentVerifier (..),
+                   StakeVerifier (..), VerificationKeyOrFile (..), VerificationKeyOrHashOrFile (..),
                    VerificationKeyTextOrFile (..), deserialiseInput, renderInputDecodeError)
 import           Cardano.CLI.Types
 import           Control.Monad.Fail (fail)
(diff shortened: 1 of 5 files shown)


newhoggy:
The `address build` command now looks like this:

```
$ cabal run -- exe:cardano-cli address build
Usage: cardano-cli address build (--payment-verification-key STRING |
                                   --payment-verification-key-file FILE |
                                   --payment-script-file FILE)
                                 [--stake-verification-key STRING |
                                   --stake-verification-key-file FILE |
                                   --stake-script-file FILE]
                                 [--out-file FILE]
  Build a Shelley payment address, with optional delegation to a stake address.

Available options:
  --payment-verification-key STRING
                           Payment verification key (Bech32-encoded)
  --payment-verification-key-file FILE
                           Filepath of the payment verification key.
  --payment-script-file FILE
                           Filepath of the payment script.
  --stake-verification-key STRING
                           Stake verification key (Bech32 or hex-encoded).
  --stake-verification-key-file FILE
                           Filepath of the staking verification key.
  --stake-script-file FILE Filepath of the staking script.
```

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -203,6 +208,22 @@ pScript = ScriptFile <$> Opt.strOption
   <> Opt.completer (Opt.bashCompleter "file")
   )
 
+pPaymentScript :: Parser ScriptFile
+pPaymentScript = ScriptFile <$> Opt.strOption
+  (  Opt.long "payment-script-file"
```
Jimbo4350, line 213: Rather than having multiple parsers, lets turn it into a function that accepts two strings

Jimbo4350 (changes requested):
LGTM, just one comment

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
newhoggy: I made the requested changes, although looking back it it, I feel it makes the code less regular.

erikd (approved):

newhoggy:
bors merge

newhoggy:
bors merge

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
Jimbo4350: I agree, I held off for a while doing this but in my Alonzo PR I had to because the number of script file parsers with minor changes to the help text was becoming ridiculous.

PR #2595, opened 2021-04-09
Update keys_and_addresses.md

cgi1:
going through the tut, the `cardano-cli query utxo` yields an error: 

```
bash-4.4# cardano-cli query utxo --shelley-era --address $(cat payment.addr) --mainnet
Shelley command failed: query utxo  Error: A query from a certain era was applied to a ledger from a different era: EraMismatch {ledgerEraName = "Mary", otherEraName = "Shelley"}
```

This can be solved by explicitly mention the era (which has changed for weeks ago from shelley to mary)


newhoggy:
`--shelley-era` and `--mary-era` are no longer an option as of `1.26.1`.

cgi1:
It would have been great, if you would have taken the time into why the error pops up and how to prevent it from a user perspective.

I see this everywhere in the managing of error messages within Cardano... They are not helping the operators and there is very less information out there how to fix. So if the next person uses google to search for the metinioned errror message, he will not know what to do based on your closing reason.

PR #2599, opened 2021-04-12
Rename occurrences of liveview to simpleview

newhoggy:


newhoggy:
I renamed the few remaining cases of `liveview` to `simpleview` in the documentation and also notice `liveview` is mentioned in `scripts/benchmarking/issue-genesis-utxo-expenditure.sh` as well.

After the rename, I tried running it and noticed that it fails with:

```bash
$ scripts/benchmarking/issue-genesis-utxo-expenditure.sh
Invalid option `--real-pbft'

Usage: cardano-cli (Era based commands | Byron specific commands |
                     Miscellaneous commands)
```

I'm interested in knowing if this script is still in use and needs to be maintained or if it can be deleted.

newhoggy:
I've rollback the changes related to benchmarking scripts because they will be taken care of by the benchmarking team.

erikd (approved):

Jimbo4350 (approved):

newhoggy:
bors merge

PR #2600, opened 2021-04-12
Convert flake to use haskell.nix

craigem:
Convert the existing flake.nix to use the documented [example](https://input-output-hk.github.io/haskell.nix/tutorials/getting-started-flakes/) from [haskell.nix](https://github.com/input-output-hk/haskell.nix) as a basis.

Depends on https://github.com/input-output-hk/iohk-nix/pull/470

Should be import `./nix/haskell.nix` with some tweaks the `haskell.nix` to get the things it needs correctly.

- [X] `nix build .` will build `cardano-node` by default.
- [X] `nix build .#cardano-node` to be explicit
- [X] `nix build .#cardano-cli`
- [X] `nix build .#cardano-ping`
- [X] `nix build .#db-converter`

Specific system targets can be built, ie:
```
nix build .#packages.x86_64-darwin.cardano-node
```
A `nix repl` equivalent can be reached running: `nix run .#repl`

### ToDo:

- [x] Import custom `libsodium`
- [x] Import `nix/haskell.nix` 
- [x] Provide shortcuts so applications can be built as per the above examples, rather than as `"cardano-cli:exe:cardano-cli"`
- [x] Provide builds for all Cardano networks.
- [x] Fix `nix repl` (replaced - see above)
- [x] Remove legacy Nix files


craigem (commented):
Very happy with this PR :smiley: 

Can we remove my comment from line 6 in release.nix?

It looks like we've now set iohk-nix (line 11) to follow nixpkgs on line 7 - so my comment on line 5 is now redundant?

Otherwise everything looks good to go.

disassembler (approved):

newhoggy (approved):

newhoggy (approved):

PR #2603, opened 2021-04-13
CAD-2772 add cardano-tx-generator

MarcFontaine:
* Add tx-generator package.
  (copy https://github.com/input-output-hk/cardano-benchmarking/tree/master/cardano-tx-generator)
* Update default.nix and cabal.project.

### DIFF
--- a/bench/tx-generator/README.md
+++ b/bench/tx-generator/README.md
@@ -1,0 +1,22 @@
+# Transaction Generator
+
+The transaction generator stresses a cluster of Cardano nodes with a simmulated work load of transaction.
+It works in two phases:
+
+* Phase 1: Secure a genesis fund and prepare a set of initial UTxOs via the local protocol.
+* Phase 2: Connect to a set of nodes via the node-to-node protocol and offer/transmit a number of transaction at a certain rate limit.
+
+## Running the tx-generator
+Either provide the configuration via a number of CLI arguments or run a JSON based benchmarking script.
+An example script is in `bench/tx-generator/test/script.json`.
+
+```
+Usage: tx-generator COMMAND
+
+Available commands:
+  cliArguments             tx-generator with CLI arguments
+  eraTransition            tx-generator demo era transition
+  json                     tx-generator run JsonScript
+```
+
+Current maintainers: Marc and Serge
--- a/bench/tx-generator/src/Cardano/Benchmarking/CliArgsScript.hs
+++ b/bench/tx-generator/src/Cardano/Benchmarking/CliArgsScript.hs
@@ -1,0 +1,158 @@
+{- HLINT ignore "Use record patterns" -}
+{- HLINT ignore "Redundant bracket" -}
+{- HLINT ignore "Reduce duplication" -}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+module Cardano.Benchmarking.CliArgsScript
+  (
+    GeneratorCmd
+  , parseGeneratorCmd
+  , runPlainOldCliScript
+  , runEraTransitionTest
+  ) where
+
+import           Prelude (error)
+import           Data.Text (unpack)
+import qualified Data.List.NonEmpty as NE
+
+import           Control.Tracer (traceWith)
+
+import           Cardano.Api
+import           Cardano.Prelude hiding (option)
+
+import           Ouroboros.Network.NodeToClient (IOManager)
+
+import           Cardano.Benchmarking.GeneratorTx.Benchmark
+import           Cardano.Benchmarking.GeneratorTx (readSigningKey)
+import           Cardano.Benchmarking.DSL
+import           Cardano.Benchmarking.Tracer
+import           Cardano.Benchmarking.GeneratorTx.Error (TxGenError(..))
+import           Cardano.Benchmarking.GeneratorTx.LocalProtocolDefinition (CliError(..), runBenchmarkScriptWith)
+
+runPlainOldCliScript :: IOManager -> GeneratorCmd -> IO (Either CliError ())
+runPlainOldCliScript
+  iocp
+  (GenerateCmd
+     logConfigFile
+     socketFile
+     benchmarkEra
+     cliPartialBenchmark
+     fundOptions
+  )
+  = runExceptT $ runBenchmarkScriptWith iocp logConfigFile socketFile
+      $ plainOldCliScript cliPartialBenchmark benchmarkEra fundOptions
+
+runEraTransitionTest :: IOManager -> GeneratorCmd -> IO (Either CliError ())
+runEraTransitionTest
+  iocp
+  (GenerateCmd
+     logConfigFile
+     socketFile
+     _benchmarkEra
+     cliPartialBenchmark
+     fundOptions
+  )
+  = runExceptT $ runBenchmarkScriptWith iocp logConfigFile socketFile
+      $ eraTransitionTest cliPartialBenchmark fundOptions
+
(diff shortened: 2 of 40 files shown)


Jimbo4350 (changes requested):
Nice! However, I don't think this belong in cardano-node.

erikd:
Unless there is an incredibly good reason to do move it to the `node` repo it should stay in `iohk-monitoring` .

MarcFontaine:
The background of this PR is the plan to run benchmarks for a cluster of nodes as part of the `cardano-node` CI
(either nightly or even on a per-PR basis). The nix-setup and the scripts for these benchmarks in "some non trivial way" depend on the `tx-generator` being part of the node repository. The benchmarking team will write down a more detailed explanation in an external document. (I'll post the link.)

erikd:
The fact that these are needed to run the benchmarks does still not mean this should be in the node repo.

erikd (changes requested):
This does not belong in the `node` repo.

MarcFontaine:
@erikd @Jimbo4350 could you please have a look at this PR again.

cardano-node/cardano-node.cabal
```diff
@@ -1,7 +1,7 @@
 cabal-version: 2.4
 
 name:                   cardano-node
-version:                1.27.0
+version:                1.26.1
```
deepfire, line 4: @MarcFontaine, this has slipped by.

default.nix
```diff
@@ -25,9 +25,8 @@ let
   packages = {
     inherit haskellPackages
       cardano-node cardano-node-profiled cardano-node-eventlogged
-      cardano-cli db-converter cardano-ping
+      cardano-cli db-converter cardano-ping tx-generator
       scripts environments dockerImage submitApiDockerImage bech32;
-
```
deepfire, line 30: @MarcFontaine, this has slipped by.

release.nix
```diff
@@ -126,6 +126,8 @@ let
     [ "cardano-node-profiled" "components" "exes" ]
     [ "checks" "tests" "cardano-node-chairman" ]
     [ "workbench" ] [ "profiles" ]
+    [ "haskellPackages" "cardano-node-chairman" "checks" ] [ "checks" "tests" "cardano-node-chairman" ]
```
deepfire, line 129: @MarcFontaine, also slipped by.

release.nix
```diff
@@ -126,6 +126,8 @@ let
...
+    [ "haskellPackages" "tx-generator" ]
```
deepfire, line 130: @MarcFontaine, re `musl` issues -- try to change this line to 
`[ "tx-generator" "components" "exes" ]`

release.nix
MarcFontaine: This is unclear.
Did I add that line by mistake ?

deepfire: This is what this PR adds to the baseline, indeed -- and it shouldn't, since this line is completely unrelated to the generator.

deepfire:
Ok, the build had a propagated failure, restarted.

release.nix
deepfire: Ok, we found the right incantation.

deepfire (approved):
LGTM!

angerman:
The fix in 46a1ead is likely due to some bug in haskell.nix, specifically in
https://github.com/input-output-hk/haskell.nix/blob/621f52e29b1d9dce372ef7aee747bcddeb915c8a/builder/comp-builder.nix#L45-L48

once haskell.nix is fixed these kinds of shenanigans should not be needed anymore.

erikd (dismissed):
There should not be any commits comtaining text like "rebase to master".

Commits messages like this contain zero useful information. I cannot see why this is not a single commit.

bench/tx-generator/src/Cardano/Benchmarking/CliArgsScript.hs
```diff
@@ -62,6 +62,7 @@ plainOldCliScript _ _ (FundsUtxo _ _ _) _ = error "plainOldCliScript FundsUtxo n
 plainOldCliScript _ _ (FundsSplitUtxo _ _) _ = error "plainOldCliScript FundsSplitUtxo not supported"
 plainOldCliScript cliPartialBenchmark benchmarkEra (FundsGenesis keyFile) (tracers, dslSet) = do
   case benchmarkEra of
+      AnyCardanoEra AlonzoEra  -> error "AlonzoEra not supported"
```
erikd, line 65: Pretty sure the rest of the node (on `master`) already supports `Alonzo`, so we should fix this.

newhoggy: Yeah, or some explanation of why we don't support alonzo era and what impact that might have.

MarcFontaine: We do not want to lose all the history of the tx-generator. This PR is (feature wise) the latest version of tx-genrator
when we stopped development in the old repo.
After it is merged to `cardano-node` we plan to remove some of the features and `CliArgsScript.hs` will be removed all together.

erikd: > We do not want to lose all the history of the tx-generator. 

What history?

The `master` branch of the node currently supports `Alonzo`. If this PR is merged it will almost certainly *break* it .

> After it is merged to cardano-node we plan to remove some of the features and CliArgsScript.hs will be removed all together.

No, you need to fix it first, then it can be merged.

bench/tx-generator/README.md
```diff
@@ -0,0 +1 @@
+# Transaction Generator
```
Jimbo4350, line 1: Can we add an introduction as to how to use the transaction generator and list the team responsible for this library?

Jimbo4350 (approved):
Minor changes.

Jimbo4350 (changes requested):
Two minor changes.

bench/tx-generator/src/Cardano/Benchmarking/CliArgsScript.hs
deepfire: @erikd, that support is being worked on in an follow-up branch, and it's a non-trivial bit.

deepfire: Moreover, you are looking at the CLI-based configuration for the generator, which is only needed for a very short interim period, and will be removed quite soon.

deepfire: And yet still, the generator can be used with Alonzo clusters -- it's just that it can't generate Alonzo-specific transactions yet.

erikd: So when are the commits going to be squashed?

Jimbo4350:
@deepfire @MarcFontaine Guys the coding style appears inconsistent. Can you spend a little time cleaning up the formatting?

bench/tx-generator/src/Cardano/Benchmarking/CliArgsScript.hs
MarcFontaine: Serge did a first squash Yesterday.
We will squash it again, after the reviews, when it is ready to merge.

erikd: I would must rather review *after* the squash rather than before.

MarcFontaine:
OK, I changed the formatting of the imports..

Jimbo4350 (approved):
Be sure to get @erikd's approval.

Jimbo4350:
> OK, I changed the formatting of the imports..

The code is still somewhat inconsistent regarding indentation.

PR #2604, opened 2021-04-13
Exporting functions for accessing StakePoolMetadataReference.

ksaric:
Exporting functions for accessing StakePoolMetadataReference.

### DIFF
--- a/cardano-api/src/Cardano/Api/Tx.hs
+++ b/cardano-api/src/Cardano/Api/Tx.hs
@@ -23,6 +23,9 @@ module Cardano.Api.Tx (
     getTxBody,
     getTxWitnesses,
 
+    -- ** Fetching certificates
+    getTxCertificates,
+
     -- ** Signing in one go
     ShelleySigningKey(..),
     toShelleySigningKey,
@@ -54,6 +57,7 @@ import qualified Data.ByteString as BS
 import qualified Data.ByteString.Lazy as LBS
 
 import qualified Data.Map.Strict as Map
+import qualified Data.Sequence.Strict as Seq
 import qualified Data.Set as Set
 import qualified Data.Vector as Vector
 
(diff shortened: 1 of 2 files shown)


cardano-api/src/Cardano/Api/Tx.hs
```diff
@@ -108,6 +112,20 @@ import           Cardano.Api.SerialiseCBOR
...
 
+import qualified Cardano.Ledger.ShelleyMA.TxBody as Allegra
+
+-- | Get all the @Certificate@ from @Tx@.
+getTxCertificates :: forall era. Tx era -> Seq.StrictSeq Certificate
+getTxCertificates tx =
+    case getTxBody tx of
+        -- Byron doesn't have any additional certificates!
+        ByronTxBody {} -> Seq.Empty
+        ShelleyTxBody ShelleyBasedEraShelley (Shelley.TxBody _ _ certs _ _ _ _ _) _aux ->
```
Jimbo4350, line 123: Why not use the `TxBody` record accessor `_certs`?

cardano-api/src/Cardano/Api/Tx.hs
```diff
@@ -108,6 +112,20 @@ import           Cardano.Api.SerialiseCBOR
 import           Cardano.Api.SerialiseTextEnvelope
 import           Cardano.Api.TxBody
 
+import qualified Cardano.Ledger.ShelleyMA.TxBody as Allegra
+
+-- | Get all the @Certificate@ from @Tx@.
+getTxCertificates :: forall era. Tx era -> Seq.StrictSeq Certificate
```
DavidEichmann, line 118: We generally don't expose uses of `StrictSeq`. Is there a particular reason to use it here over `[]`?

DavidEichmann:
After talking with @dcoutts, we probably want to expose a function or pattern to gain access to a `TxBodyContent` which contains the certificates. I've got a WIP implementation of `getTxBodyContent :: TxBody -> TxBodyContent` on my branch https://github.com/input-output-hk/cardano-node/tree/davide/getTxBodyContent

cardano-api/src/Cardano/Api/Tx.hs
ksaric: A valid suggestion, Jordan.

cardano-api/src/Cardano/Api/Tx.hs
ksaric: Just thought of trying to keep the original data structure in place without conversions in case somebody needs it.

ksaric:
> After talking with @dcoutts, we probably want to expose a function or pattern to gain access to a `TxBodyContent` which contains the certificates. I've got a WIP implementation of `getTxBodyContent :: TxBody -> TxBodyContent` on my branch https://github.com/input-output-hk/cardano-node/tree/davide/getTxBodyContent

That's fine, as long as I can get to the `Certificate` and the `StakePoolMetadataReference`, that's good.
I'll close this PR for now, I have to start porting `smash` to `db-sync` anyway and for now, we won't be using this API. Hopefully we can use it as time progresses, I volunteer for that, from what I saw it's pretty mature and usable, so your addition for the pattern will be useful.

PR #2605, opened 2021-04-14
Remove Byron Bech32 support

newhoggy:
We use the prefix to distinguish keys, but according to the code, Byron and Shelley share the same prefix:

```bash
instance SerialiseAsBech32 (VerificationKey ByronKey) where
    bech32PrefixFor         _ =  "addr_xvk"
    bech32PrefixesPermitted _ = ["addr_xvk"]
```

```bash
instance SerialiseAsBech32 (VerificationKey PaymentExtendedKey) where
    bech32PrefixFor         _ =  "addr_xvk"
    bech32PrefixesPermitted _ = ["addr_xvk"]
```

We use the following list when deserialising:

```bash
    bech32Types =
      [ FromSomeType (AsVerificationKey AsByronKey)
                     AByronVerificationKey
      , FromSomeType (AsVerificationKey AsPaymentKey)
                     APaymentVerificationKey
      , FromSomeType (AsVerificationKey AsPaymentExtendedKey)
                     APaymentExtendedVerificationKey
      ]
```

So when deserialising a key like `addr_xvk18359zklg0x68ru6yhxyvsv3slt6wscs00mzsc45jrv9840dfutj5uxg6z6rp4w2pany4kjgg5tly8s5xv7fuyq5tf8kw9ehr7cy4lzcenyhck`, it thinks the keys are Byron because Byron appears first in the list and shares the same prefix as Shelley.

Having an instance for Byron declare that Byron keys have the `addr_xvk` prefix seems wrong.  This PR removes support for serialising and deserialising Byron keys with that prefix.

### DIFF
--- a/cardano-api/src/Cardano/Api/KeysByron.hs
+++ b/cardano-api/src/Cardano/Api/KeysByron.hs
@@ -52,11 +52,10 @@ import qualified Cardano.Crypto.Hashing as Byron
 import qualified Cardano.Crypto.Signing as Byron
 import qualified Cardano.Crypto.Wallet as Wallet
 
-import           Cardano.Api.Hash
 import           Cardano.Api.HasTypeProxy
+import           Cardano.Api.Hash
 import           Cardano.Api.Key
 import           Cardano.Api.KeysShelley
-import           Cardano.Api.SerialiseBech32
 import           Cardano.Api.SerialiseCBOR
 import           Cardano.Api.SerialiseRaw
 import           Cardano.Api.SerialiseTextEnvelope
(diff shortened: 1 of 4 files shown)


newhoggy:
This fixes https://github.com/input-output-hk/cardano-node/issues/2596

With this change the following works as expected:

```
$ cardano-cli address build --payment-verification-key addr_xvk18359zklg0x68ru6yhxyvsv3slt6wscs00mzsc45jrv9840dfutj5uxg6z6rp4w2pany4kjgg5tly8s5xv7fuyq5tf8kw9ehr7cy4lzcenyhck --testnet-magic 1097911063
addr_test1vptgn6nxuhfc0rt3s7xydxlczm2rncn7dhcqt4jp4d9edqqax2mf6%
$ cardano-cli address info --address addr_test1vptgn6nxuhfc0rt3s7xydxlczm2rncn7dhcqt4jp4d9edqqax2mf6
```

dcoutts (approved):
So now the only way to use Byron keys is via a file (using text envelope format) right?

Jimbo4350 (approved):

newhoggy:
> So now the only way to use Byron keys is via a file (using text envelope format) right?

Yep.

newhoggy:
bors merge

PR #49, opened 2019-07-11
ip subscription manager

coot:

PR #2608, opened 2021-04-14
[Part 5] - Provide high quality multiarch docker image

tdiesler:
This is the fivth installment of [#2360] Provide high quality multiarch docker image

Now we have ...

* Preserved functionality via -e NETWORK
* Node configuration through env vars
* Integrated configurable topology update
* Graceful shutdown on docker stop
* Improved access to the CLI
* Comprehensive [documentation](https://tdiesler.gitbook.io/cardano/v/iohk/)

There is a README about this [here](https://github.com/tdiesler/cardano-node/tree/ghi2606/nix/docker).

PS: Once this is merged we'd need to cleanup a few links and perhaps find a better home for the docs.

### DIFF
--- a/nix/docker/context/bin/run-node
+++ b/nix/docker/context/bin/run-node
@@ -1,0 +1,119 @@
+#!/bin/bash
+
+source /usr/local/bin/proc-sigterm
+
+echo "Running the cardano node ..."
+
+# Shift the first option by one index
+shift
+
+# Define a few defaults
+CARDANO_CONFIG_BASE="/opt/cardano/config"
+
+if [[ -z $CARDANO_CONFIG ]]; then
+  CARDANO_CONFIG="$CARDANO_CONFIG_BASE/mainnet-config.json"
+fi
+
+if [[ -z $CARDANO_TOPOLOGY ]]; then
+  CARDANO_TOPOLOGY="$CARDANO_CONFIG_BASE/mainnet-topology.json"
+fi
+
+if [[ -z $CARDANO_DATABASE_PATH ]]; then
+  CARDANO_DATABASE_PATH="/opt/cardano/data"
+fi
+
+if [[ -z $CARDANO_SOCKET_PATH ]]; then
+  CARDANO_SOCKET_PATH="/opt/cardano/ipc/socket"
+fi
+
+if [[ -z $CARDANO_BIND_ADDR ]]; then
+  CARDANO_BIND_ADDR="0.0.0.0"
+fi
+
+if [[ -z $CARDANO_PORT ]]; then
+  CARDANO_PORT=3001
+fi
+
+if [ -v $CARDANO_BLOCK_PRODUCER ]; then
+  CARDANO_BLOCK_PRODUCER=false
+fi
+
+#####################################################################
+#
+# Print run environment
+#
+printRunEnv () {
+
+  echo "CARDANO_BIND_ADDR=$CARDANO_BIND_ADDR"
+  echo "CARDANO_BLOCK_PRODUCER=$CARDANO_BLOCK_PRODUCER"
+  echo "CARDANO_CONFIG=$CARDANO_CONFIG"
+  echo "CARDANO_DATABASE_PATH=$CARDANO_DATABASE_PATH"
+  echo "CARDANO_PORT=$CARDANO_PORT"
+  echo "CARDANO_SOCKET_PATH=$CARDANO_SOCKET_PATH"
+  echo "CARDANO_TOPOLOGY=$CARDANO_TOPOLOGY"
+
+  if [ "$CARDANO_BLOCK_PRODUCER" = true ]; then
+
+    if [ -v $CARDANO_SHELLEY_KES_KEY ]; then
+      CARDANO_SHELLEY_KES_KEY="$CARDANO_CONFIG_BASE/keys/kes.skey"
+    fi
+
+    if [ -v $CARDANO_SHELLEY_VRF_KEY ]; then
+      CARDANO_SHELLEY_VRF_KEY="$CARDANO_CONFIG_BASE/keys/vrf.skey"
+    fi
+
+    if [ -v $CARDANO_SHELLEY_OPERATIONAL_CERTIFICATE ]; then
+      CARDANO_SHELLEY_OPERATIONAL_CERTIFICATE="$CARDANO_CONFIG_BASE/keys/node.cert"
+    fi
+
+    echo "CARDANO_SHELLEY_KES_KEY=$CARDANO_SHELLEY_KES_KEY"
+    echo "CARDANO_SHELLEY_VRF_KEY=$CARDANO_SHELLEY_VRF_KEY"
+    echo "CARDANO_SHELLEY_OPERATIONAL_CERTIFICATE=$CARDANO_SHELLEY_OPERATIONAL_CERTIFICATE"
+  fi
+}
+
+# Override default values with explicit options
+options=($@)
+for i in "${!options[@]}"
+do
+    j=$((i + 1))
+    key=${options[i]}
+    val=${options[j]}
+    found=false
+
+    # echo "$i/$j: ${key} ${val}"
+
+    case ${key} in
+      --config) CARDANO_CONFIG=${val}; found=true;;
+      --topology) CARDANO_TOPOLOGY=${val}; found=true;;
+      --database-path) CARDANO_DATABASE_PATH=${val}; found=true;;
+      --socket-path) CARDANO_SOCKET_PATH=${val}; found=true;;
+      --host-addr) CARDANO_BIND_ADDR=${val}; found=true;;
+      --port) CARDANO_PORT=${val}; found=true;;
+    esac
+
+    if [[ $found == true ]]; then
+      options[i]="";
+      options[j]="";
+    fi
+done
+
+printRunEnv
+
+effopts=(--config $CARDANO_CONFIG \
+  --topology $CARDANO_TOPOLOGY \
+  --database-path $CARDANO_DATABASE_PATH \
+  --socket-path $CARDANO_SOCKET_PATH \
+  --host-addr $CARDANO_BIND_ADDR \
+  --port $CARDANO_PORT)
+
+effopts+=(${options[@]})
+
+echo "cardano-node run ${effopts[@]}"
+
+# The IPC socket dir is not created on demand
+mkdir -p `dirname $CARDANO_SOCKET_PATH`
+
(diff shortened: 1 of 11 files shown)


tdiesler:
@disassembler I've now rebased this work on 1.26.1 and (temporarily) removed the k8s stuff, because I want to give it a little more incubation time with Alonzo. Folks, who are interested in k8s can always go to https://github.com/tdiesler/nessus-cardano and see what we've got so far. When ready, we'll create another PR for that.

This should also make it a little easier for you and your team to digest this PR.

nix/docker/context/bin/functions
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+
+#####################################################################
+#
+# Trap the SIGTERM signal
+#
+trapSIGTERM () {
+
+  PID="$!"
+  trap "sigtermHandler $PID" SIGTERM
```
nyetwurk, line 10: Why not just set `STOPSIGNAL` in the docker image, or 
```
config = {
      EntryPoint = [ "${entry-point}/bin/entry-point" ];
      StopSignal = "INT";
    };
```
if using nix?

tdiesler: I looked into this and it would work. However, if we do this we loose

```
Signalling for shutdown ...
```

which comes from the wrapper script that runs cardano-node as a sub-process in the background. I personally would prefer to make this change dependent on 

* [#2648] Indicate graceful shutdown with a log message

Here is the issues in the incubator project, where we can follow up on this: 

* [#49](https://github.com/tdiesler/nessus-cardano/issues/49) Simplify stop signal handling by using SIGINT directly

tdiesler:
Rebased on latest master

Asafkbalink:
The provided external documentation is nice, but lacks a lot of info - how do I generate the [secret keys](https://tdiesler.gitbook.io/cardano/v/iohk/k8s/cardano-k8s#block-producer-configuration) in the first place?
I'm looking for a step-by-step documentation on how to run a stake pool using Kubernetes, with 0 knowledge on how to start a Cardano's stake pool, but with enough knowledge of Kubernetes to make this work.

tdiesler:
Rebased on latest master

jbgi:
LGTM, except maybe be more explicit in the doc about the content of CARDANO_CUSTOM_PEERS being sent to a thrid-party external API (api.clio.one). That might be a security issue for some users. Ideally the merge could be done locally.

tdiesler:
@jbgi Thanks for giving your LGTM. Could you perhaps be more specific about what you like to have changed? Currently, we have the topology update built in [like this](https://github.com/tdiesler/cardano-node/blob/ghi2606/nix/docker/context/bin/topologyUpdate#L132) because it (unfortunately) is still needed until we have the p2p manager in place. Isn't it true that currently (almost) all nodes use this API to have their topology updated (i.e. all of those send their custom peers as url param to api.clio.one)?

There is [lots of docs over here](https://tdiesler.gitbook.io/cardano/v/iohk/plain-docker/running-the-nodes), which we can then gradually bring over as well, so that all of the Docker stuff gets documented really nicely.

On DockerHub, I have [these docs](https://hub.docker.com/r/nessusio/cardano-node).

nyetwurk:
> LGTM, except maybe be more explicit in the doc about the content of CARDANO_CUSTOM_PEERS being sent to a third-party API (api.clio.one). That might be a security issue for some users. Ideally the merge could be done locally.

Can you be more specific? I have not looked at the code here, but my own implementation of topologyUpdater does not require sending CUSTOM_PEERS off node

nyetwurk:
Ok, I see. We do something like this and merge custom peers locally
```bash
  curl -s -f -4 -o "${TOPOLOGY}.orig" "https://api.clio.one/htopology/v1/fetch/?max=${MAX_PEERS}&magic=${NWMAGIC}"
  echo "Fetched $(jq '.Producers | length' ${TOPOLOGY}.orig) peers"
  if [[ -r "${CUSTOM_PEERS}" ]]; then
    echo "Merging ${TOPOLOGY}.orig and ${CUSTOM_PEERS}"
    jq -s '.[0].Producers += .[1] | .[0]' ${TOPOLOGY}.orig ${CUSTOM_PEERS} > ${TOPOLOGY}
  fi
```

jbgi:
@nyetwurk Thanks for provided exactly the implementation I was looking for!
@tdiesler could you use it in the PR?

nyetwurk:
> @nyetwurk Thanks for providing exactly the implementation I was looking for!
> @tdiesler could you use it in this PR?

Note that if your CUSTOM_PEERS is a string, not a file, it will be different :) I can add a suggested bashism time permitting

rdlrt:
I guess it's a late comment - as I am just re-iterating, but the [original source for topologyUpdater](https://github.com/cardano-community/guild-operators/blob/alpha/scripts/cnode-helper-scripts/topologyUpdater.sh#L123-L134) was also updated months ago to remove any references/possibility to send custom peers remotely to API, and instead uses local peers but does not use `jq`, the example formats for CUSTOM_PEERS was updated back then (tho the code does handle both).

tdiesler:
@rdlrt Thanks for that - very helpful. I'll try to incorporate what you do.

seriru:
It would also be great if the docker image could look at the /etc/hosts file to resolve locally configured hosts. #2660

tdiesler:
@seriru I don't think this a valid request - see [here](https://github.com/input-output-hk/cardano-node/issues/2660#issuecomment-843107620)

jbgi (approved):

nix/docker/context/bin/run-node
```diff
@@ -0,0 +1,276 @@
...
+rm -rf $CARDANO_DATABASE_PATH/lost+found
+
+if [[ $CARDANO_BLOCK_PRODUCER == true ]]; then
+
+  runBlockProducerNode
+
+else
+
+  if [[ $CARDANO_UPDATE_TOPOLOGY == true ]]; then
+    topologyUpdate -l &
```
jbgi, line 270: why is this run in the background? The relay node could start before topology is fully updated.

jbgi, line 270: so the topology update is done every hour in the background. But currently the node need to be restarted to pickup any change to the topology file (since will change with p2p support).

Should a synchronous update be done instead (or in addition to)?

jbgi (changes requested):

nix/docker/context/bin/run-node
tdiesler: Yes, we run the push/fetch for the topology update once per hour in the background. We used to do this in a cron job that ran inside the docker container (just like almost every other node). This however created an unnecessary dependency on base image specifics (i.e. on the OS) which unfortunately are not the same across supported architectures (i.e. its different for arm64).

How does a non-docker node pick up changes to the topology file when they are done async from a cron job? To be honest, I never questioned this because I assumed that the node somehow "watches" its topology updates. The async cron job approach suggested this and the the above has always worked (for many months so far).

jbgi: Currently we restart the node every time the topology is updated.

tdiesler: Who restarts the node, is the cron job supposed to do that? A restart is not instantaneous, especially not if graceful shutdown is not done properly. Even when it is, it may take minutes for the node to communicate with the BP again, which may lead to block loss. 

Perhaps it is sufficient to publish the node's IP/DNS to be connected well and it is not actually necessary to read the updated topology. Restarting the node every hour seems quite wrong and I somehow can't believe that many SPOs actually do that.

tdiesler: We could also remove the topology update from this PR. It isn't part of the existing image and would be taken out with Alonzo anyway.

nix/docker/compose/docker-compose.yaml
```diff
@@ -0,0 +1,80 @@
...
+            }
+          ]
+        }'
+      CARDANO_BLOCK_PRODUCER: "true"
+    command: "run"
+    depends_on: [ "relay" ]
+    volumes:
+      - bprod-keys:/var/cardano/config/keys
+      - bprod-data:/opt/cardano/data
+
```
coot, line 73: Please remove the docker compose part, we don't encourage to run block producing node and a relay on the same host.  We also advice on running at least two relays per BP node (for redundancy, in future, with the p2p release, one might need to run more relays to support all client nodes (wallets)).

nix/docker/context/bin/run-node
```diff
@@ -0,0 +1,276 @@
...
+  CARDANO_SOCKET_PATH="/opt/cardano/ipc/socket"
+fi
+
+if [[ -z $CARDANO_LOG_DIR ]]; then
+  CARDANO_LOG_DIR="/opt/cardano/logs"
+fi
+
+if [[ -z $CARDANO_BIND_ADDR ]]; then
+  CARDANO_BIND_ADDR="0.0.0.0"
+fi
```
coot, line 36: It would be nice to also add optional support for `ipv6` addresses.

coot (changes requested):
The p2p nodes will make it possible to reload configuration on `SIGHUP` signal; will this require additional support on the `docker` side?

nix/docker/compose/docker-compose.yaml
tdiesler: Removed, as agreed

nix/docker/context/bin/run-node
tdiesler: Yes, noted down as a feature request in the incubator.
https://github.com/tdiesler/nessus-cardano/issues/58

tdiesler:
> The p2p nodes will make it possible to reload configuration on `SIGHUP` signal; will this require additional support on the `docker` side?

No, this would come out of the box through the (unfortunately rather ill-named) `docker kill` command.

tdiesler:
I rebased this work on the last stable tag (i.e. 1.27.0) because the current node config in master has some issues with a missing Alonzo genesis file. Also, `${scripts.${script}}` now seems to point to a directory rather than an executable, which breaks the existing docker image.

I could look at the latter and see if I find something. The Alonzo stuff I don't dare to touch.

tdiesler:
Rebased on latest master and tested. 

${scripts.${script}} now seems to point to a directory rather than an executable, which breaks `-e NETWORK=mainnet`.
The images consumes its configurations from [configuration/cardano](https://github.com/input-output-hk/cardano-node/tree/master/configuration/cardano). To make it work, I had to modify the main-config and add the Alozo genisis. Is that correct?

tdiesler:
Rebased on PR #2796

PR #2611, opened 2021-04-15
Remove GH work around

newhoggy:


erikd (changes requested):
Seems fine, but please retittle the commit so its obvious that its the GH actions where a wrokaround was removed.

newhoggy:
This particular `apt-get` repository was previously causing issues due to flakiness and was previous removed at the beginning of the build.  Github Actions have removed this from the initial repository list, so builds are now failing and the the work around is no longer needed.

Jimbo4350 (approved):

newhoggy:
bors merge

PR #2633, opened 2021-04-20
Support for LedgerState and folding over all blocks

DavidEichmann:
The `foldBlocks` function was originally used by @JaredCorduan to implement an audit tool. In my first commit I remove the existing cardano-api `LedgerState` type and corresponding queries. This is to make room for the new `LedgerState` type that I introduce in the second commit.

I eventually plan to add a query for the full ledger state. Once that is merged, I'll be able to reintroduce the queries that I've removed here.

### DIFF
--- a/cardano-api/src/Cardano/Api/LedgerState.hs
+++ b/cardano-api/src/Cardano/Api/LedgerState.hs
@@ -1,0 +1,846 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Cardano.Api.LedgerState
+  ( -- * Initialization / Accumulation
+    Env(..)
+  , envSecurityParam
+  , LedgerState
+      ( ..
+      , LedgerStateByron
+      , LedgerStateShelley
+      , LedgerStateAllegra
+      , LedgerStateMary
+      )
+  , initialLedgerState
+  , applyBlock
+
+    -- * Traversing the block chain
+  , foldBlocks
+
+   -- * Errors
+  , FoldBlocksError(..)
+  , GenesisConfigError(..)
+  , InitialLedgerStateError(..)
+  , renderFoldBlocksError
+  , renderGenesisConfigError
+  , renderInitialLedgerStateError
+  )
+  where
+
+import           Prelude
+
+import           Control.Exception
+import           Control.Monad.Trans.Class
+import           Control.Monad.Trans.Except
+import           Control.Monad.Trans.Except.Extra
+import           Data.Aeson as Aeson
+import qualified Data.Aeson.Types as Data.Aeson.Types.Internal
+import           Data.ByteArray (ByteArrayAccess)
+import qualified Data.ByteArray
+import           Data.ByteString as BS
+import qualified Data.ByteString.Base16 as Base16
+import           Data.ByteString.Short as BSS
+import           Data.Foldable
+import           Data.IORef
+import           Data.Sequence (Seq)
+import qualified Data.Sequence as Seq
+import           Data.Text (Text)
+import qualified Data.Text as Text
+import qualified Data.Text.Encoding as Text
+import           Data.Word
+import qualified Data.Yaml as Yaml
+import           System.FilePath
+
+import           Cardano.Api.Block
+import           Cardano.Api.Eras
+import           Cardano.Api.IPC (ConsensusModeParams (CardanoModeParams), EpochSlots (..),
+                   LocalChainSyncClient (LocalChainSyncClientPipelined),
+                   LocalNodeClientProtocols (..), LocalNodeClientProtocolsInMode,
+                   LocalNodeConnectInfo (..), connectToLocalNode)
+import           Cardano.Api.Modes (CardanoMode)
+import           Cardano.Api.NetworkId (NetworkId)
+import qualified Cardano.Chain.Genesis
+import qualified Cardano.Chain.Update
+import qualified Cardano.Crypto
+import qualified Cardano.Crypto.Hash.Blake2b
+import qualified Cardano.Crypto.Hash.Class
+import qualified Cardano.Crypto.Hashing
+import qualified Cardano.Crypto.ProtocolMagic
+import           Cardano.Slotting.Slot (WithOrigin (At, Origin))
+import qualified Cardano.Slotting.Slot as Slot
+import           Network.TypedProtocol.Pipelined (Nat (..))
+import qualified Ouroboros.Consensus.Block.Abstract as Consensus
+import qualified Ouroboros.Consensus.Byron.Ledger.Block as Byron
+import qualified Ouroboros.Consensus.Cardano as Consensus
+import qualified Ouroboros.Consensus.Cardano.Block as Consensus
+import qualified Ouroboros.Consensus.Cardano.CanHardFork as Consensus
+import qualified Ouroboros.Consensus.Cardano.Node as Consensus
+import qualified Ouroboros.Consensus.Config as Consensus
+import qualified Ouroboros.Consensus.HardFork.Combinator.AcrossEras as HFC
+import qualified Ouroboros.Consensus.HardFork.Combinator.Basics as HFC
+import qualified Ouroboros.Consensus.Ledger.Abstract as Ledger
+import qualified Ouroboros.Consensus.Ledger.Extended as Ledger
+import qualified Ouroboros.Consensus.Node.ProtocolInfo as Consensus
+import qualified Ouroboros.Consensus.Shelley.Eras as Shelley
+import qualified Ouroboros.Consensus.Shelley.Ledger.Block as Shelley
+import qualified Ouroboros.Consensus.Shelley.Ledger.Ledger as Shelley
+import qualified Ouroboros.Consensus.Shelley.Protocol as Shelley
+import qualified Ouroboros.Network.Block
+import           Ouroboros.Network.Protocol.ChainSync.ClientPipelined
+                   (ChainSyncClientPipelined (ChainSyncClientPipelined),
+                   ClientPipelinedStIdle (CollectResponse, SendMsgDone, SendMsgRequestNextPipelined),
+                   ClientStNext (..))
+import           Ouroboros.Network.Protocol.ChainSync.PipelineDecision
+import qualified Shelley.Spec.Ledger.BaseTypes as Shelley.Spec
+import qualified Shelley.Spec.Ledger.Credential as Shelley.Spec
+import qualified Shelley.Spec.Ledger.Genesis as Shelley.Spec
+import qualified Shelley.Spec.Ledger.Keys as Shelley.Spec
+import qualified Shelley.Spec.Ledger.PParams as Shelley.Spec
+
+data InitialLedgerStateError
+  = ILSEConfigFile Text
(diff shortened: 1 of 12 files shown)


cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -0,0 +1,913 @@
...
+                  { envLedgerConfig = Consensus.topLevelConfigLedger topLevelConfig
+                  , envProtocolConfig = Consensus.topLevelConfigProtocol topLevelConfig
+                  }
+
+readDbSyncNodeConfig :: NodeConfigFile -> IO DbSyncNodeConfig
+readDbSyncNodeConfig (NodeConfigFile ncf) = do
+    ncfg <- parseNodeConfig <$> readByteString ncf "node"
+    coalesceConfig ncfg (mkAdjustPath ncf)
+
+data NodeConfig = NodeConfig
```
Jimbo4350, line 369: There is `NodeConfiguration` in `cardano-node` that has this information and can be extended if there are fields missing.

Jimbo4350 (commented):
Some initial comments. This looks like it was pretty hairy to sort out :sweat_smile:

cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -0,0 +1,913 @@
...
+
+-- | Get the environment and initial ledger state.
+initialLedgerState
+  :: FilePath
+  -- ^ Path to the cardano-node config file (e.g. <path to cardano-node project>/configuration/cardano/mainnet-config.json)
+  -> IO (Env, LedgerState)
+  -- ^ The environment and initial ledger state
+initialLedgerState dbSyncConfFilePath = do
+  dbSyncConf <- readDbSyncNodeConfig (NodeConfigFile dbSyncConfFilePath)
+  genConf <- fmap (either (error . Text.unpack . renderDbSyncNodeError) id) $ runExceptT (readCardanoGenesisConfig dbSyncConf)
```
Jimbo4350, line 116: If this is eventually intended for `cardano-cli` let's return an `Either` so we can handle errors there.

cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -0,0 +1,913 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE EmptyCase #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE RecordWildCards #-}
```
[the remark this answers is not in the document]
DavidEichmann: That and other extensions are not needed. I'll remove them.

cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -0,0 +1,913 @@
...
+    connectInfo
+    (protocols stateIORef env ledgerState)
+
+  readIORef stateIORef
+  where
+  connectInfo :: LocalNodeConnectInfo CardanoMode
+  connectInfo =
+      LocalNodeConnectInfo {
+        localConsensusModeParams = CardanoModeParams (EpochSlots 21600),
+        localNodeNetworkId       = Mainnet,
```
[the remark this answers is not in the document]
DavidEichmann: I'll turn it into a parameter

cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -0,0 +1,913 @@
...
+    = ChainSyncClientPipelined $ pure $ clientIdle_RequestMoreN Origin Origin Zero initialLedgerStateHistory
+    where
+        initialLedgerStateHistory = Seq.singleton (0, ledgerState0, Origin) -- TODO is the initial ledger state at slot 0?
+
+        pushLedgerState' = pushLedgerState env
+
+        clientIdle_RequestMoreN
+          :: WithOrigin BlockNo
+          -> WithOrigin BlockNo
+          -> Nat n
```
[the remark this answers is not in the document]
DavidEichmann: "Number of requests inflight." will add a comment

cardano-api/src/Cardano/Api/LedgerState.hs
DavidEichmann: Yes... actually I used `error` in a few places and need to turn these into propper errors

cardano-api/src/Cardano/Api/LedgerState.hs
DavidEichmann: I think I misunderstood this yesterday. Seeing as I'll be able to remove this once the query for ledger config gets merged, Perhaps we're ok with having `NodeConfig` for now? I'll add a TODO in the code.

DavidEichmann:
Changes:
* Use `ExceptT` instead of `error`
* Remove `DbSyncNodeConfig` and use `NodeConfig` instead. There were isomorphic any way (just an artifact of copying from db-sync)
* Actually check the Shelley hash from the config file.
* Take the network as a parameter to `foldBlocks`

cardano-api/src/Cardano/Api/LedgerState.hs
DavidEichmann: @Jimbo4350 I'm using `ExceptT` now. Is that ok with you?

cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -0,0 +1,849 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# OPTIONS_GHC -freduction-depth=0 #-}
```
Jimbo4350, line 9: I can seem to find any docs on this. Do you have a source?

cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -0,0 +1,849 @@
...
+
+renderInitialLedgerStateError :: InitialLedgerStateError -> Text
+renderInitialLedgerStateError ilse = case ilse of
+  ILSEConfigFile err ->
+    "Failed to read or parse the network config file: " <> err
+  ILSEGenesisFile err ->
+    "Failed to read or parse a genesis file linked from the network config file: "
+    <> renderGenesisConfigError err
+  ILSELedgerConsensusConfig err ->
+    "Failed to derived the Ledger or Consensus config: "
```
Jimbo4350, line 124: should be derive

cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -0,0 +1,849 @@
...
+import qualified Shelley.Spec.Ledger.Keys as Shelley.Spec
+import qualified Shelley.Spec.Ledger.PParams as Shelley.Spec
+
+data InitialLedgerStateError
+  = ILSEConfigFile Text
+  -- ^ Failed to read or parse the network config file.
+  | ILSEGenesisFile GenesisConfigError
+  -- ^ Failed to read or parse a genesis file linked from the network config file.
+  | ILSELedgerConsensusConfig GenesisConfigError
+  -- ^ Failed to derived the Ledger or Consensus config.
```
Jimbo4350, line 114: should be derive

cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -0,0 +1,849 @@
...
+  -- Connect to the node.
+  lift $ connectToLocalNode
+    connectInfo
+    (protocols stateIORef errorIORef env ledgerState)
+
+  (lift $ readIORef errorIORef) >>= \case
+    Just err -> throwE (FoldBlocksApplyBlockError err)
+    Nothing -> lift $ readIORef stateIORef
+  where
+  connectInfo :: LocalNodeConnectInfo CardanoMode
```
Jimbo4350, line 253: Lets put an indent here to align with the coding style of the repo. This will make the `where` clause stand out more.

cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -0,0 +1,849 @@
...
+  protocols :: IORef a -> IORef (Maybe Text) -> Env -> LedgerState -> LocalNodeClientProtocolsInMode CardanoMode
+  protocols stateIORef errorIORef env ledgerState =
+      LocalNodeClientProtocols {
+        localChainSyncClient    = LocalChainSyncClientPipelined (chainSyncClient 50 stateIORef errorIORef env ledgerState),
+        localTxSubmissionClient = Nothing,
+        localStateQueryClient   = Nothing
+      }
+
+  -- | Add a new ledger state to the history
+  pushLedgerState :: Env -> LedgerStateHistory -> SlotNo -> LedgerState -> BlockInMode CardanoMode
```
Jimbo4350, line 270: Annotations on these types would be useful as well.

cardano-api/src/Cardano/Api/LedgerState.hs
Jimbo4350: No worries. I think the best solution would be to move the node's configuration stuff to the api. This is good for now.

cardano-api/src/Cardano/Api/Query.hs
```diff
@@ -148,9 +144,6 @@ data QueryInShelleyBasedEra era result where
      -- QueryPoolRanking
      --   :: QueryInShelleyBasedEra era RewardProvenance
 
-     QueryLedgerState
```
Jimbo4350, line 151: Why are we removing the `QueryLedgerState`? Can't we leave it for now until you're ready to replace it?

Jimbo4350 (commented):
Good work, the code is a lot clearer now. A couple comments but nothing major.

cardano-api/src/Cardano/Api/LedgerState.hs
DavidEichmann: Wow.. not sure how that got there. I've removed it.

DavidEichmann:
Update:

* The first commit simply renames some things instead of deleting the cli ledger state query
* Resolved PR comments

Jimbo4350 (approved):
LGTM! Is there a TODO regarding potentially moving the `NodeConfiguration` machinery to `cardano-api`? You should also update the PR description as you are no longer removing a query.

DavidEichmann:
> Is there a TODO regarding potentially moving the NodeConfiguration machinery to cardano-api?
I added this: https://github.com/input-output-hk/cardano-node/pull/2633/files#diff-b415c60c9e90bf9d54569de2fb6117807512a0f98488005a076913cafa7c947dR133

Update:
* Removed an old TODO and cleaned up some imports

DavidEichmann:
Update:
* Rebase
* ouroboros-network (version bumped after the rebase) no longer exports `Protocol` which this code used transitively. Removing use of `Protocol` turned out to be pretty trivial (after a bit of fumbling).

DavidEichmann:
bors merge

cardano-api/src/Cardano/Api/LedgerState.hs
Jimbo4350: Update: The genesis file will eventually not be needed so we don't need to move the node configuration stuff to cardano-api

DavidEichmann:
Update:
* Fix hlint errors

DavidEichmann:
bors merge

JaredCorduan:
:tada: `foldBlocks` fills me with joy, thanks again @DavidEichmann !

PR #2644, opened 2021-04-22
cardano-api: Add fromShelleyScript, fromAllegraScript, fromMaryScript

cblp:


cblp:
@dcoutts please check that V2 is always ok

Jimbo4350 (changes requested):
Right direction :+1:  with one minor adjustment

cardano-api/src/Cardano/Api/Script.hs
```diff
@@ -749,6 +752,27 @@ toShelleyScript (ScriptInEra langInEra (SimpleScript _ script)) =
       SimpleScriptV2InAllegra -> toAllegraTimelock script
       SimpleScriptV2InMary    -> toAllegraTimelock script
 
+fromShelleyScript :: Ledger.Script (ShelleyLedgerEra ShelleyEra)
```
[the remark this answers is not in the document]
cblp: done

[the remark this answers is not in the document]
cblp: but with a bit simpler interface

cardano-api/src/Cardano/Api/Script.hs
```diff
@@ -749,6 +751,25 @@ toShelleyScript (ScriptInEra langInEra (SimpleScript _ script)) =
       SimpleScriptV2InAllegra -> toAllegraTimelock script
       SimpleScriptV2InMary    -> toAllegraTimelock script
 
+fromShelleyBasedScript  ::  forall era
+                        .   IsShelleyBasedEra era
+                        =>  Ledger.Script (ShelleyLedgerEra era)
+                        ->  ScriptInEra era
+fromShelleyBasedScript script =
+  case shelleyBasedEra @era of
```
dcoutts, line 759: We've tended to conclude that for most functions like this things work out better in the end if we pass the era explicitly rather than implicitly via a type class argument.

i.e. like this (untested)
```suggestion
fromShelleyBasedScript  :: ShelleyBasedEra era
                        -> Ledger.Script (ShelleyLedgerEra era)
                        -> ScriptInEra era
fromShelleyBasedScript era script =
  case era of
```

cardano-api/src/Cardano/Api/Script.hs
```diff
@@ -3,6 +3,7 @@
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TypeApplications #-}
```
dcoutts, line 6: If we use the suggestion below we will not need this extension.

dcoutts (approved):
LGTM, with the minor tweak below.

Yes I think always using the max version for the era is ok.

If you're happy with the suggestion, squash and then feel free to merge.

cardano-api/src/Cardano/Api/Script.hs
Jimbo4350: Yep the majority of the API we expose expects the era explicitly.

cblp: done

cblp:
bors merge

cblp:
@Jimbo4350 re-review please

Jimbo4350 (commented):
LGTM :+1:

Jimbo4350 (approved):

PR #2650, opened 2021-04-23
CAD-2839 Workbench fixes

deepfire:
- prebuild executables explicitly
- work around a Nix linker purity issue due to `NIX_ENFORCE_PURITY` in `shellHook`
- automatically deal with `state-cluster` being partially built, while it's safely default

### DIFF
--- a/nix/supervisord-cluster/default.nix
+++ b/nix/supervisord-cluster/default.nix
@@ -190,6 +190,8 @@ let
 
     ln -s ${profile.topology.files} "${stateDir}"/topology
 
+    prebuild-executables
+
     genesis_args+=(
         ## Positionals:
         ${profile.JSON}
--- a/nix/supervisord-cluster/default.nix
+++ b/nix/supervisord-cluster/default.nix
@@ -1,15 +1,21 @@
+let
+  basePortDefault    = 30000;
+  cacheDirDefault    = "${__getEnv "HOME"}/.cache";
+  stateDirDefault    = "state-cluster";
+  profileNameDefault = "default-mary";
+in
 { pkgs
 , lib
 , bech32
-, basePort ? 30000
-, stateDir ? "state-cluster"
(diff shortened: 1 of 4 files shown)


nix/supervisord-cluster/default.nix
```diff
@@ -1,15 +1,21 @@
...
+in
 { pkgs
 , lib
 , bech32
-, basePort ? 30000
-, stateDir ? "state-cluster"
-, cacheDir ? "${__getEnv "HOME"}/.cache"
+, basePort              ? basePortDefault
+, stateDir              ? stateDirDefault
+, cacheDir              ? cacheDirDefault
```
MarcFontaine, line 12: I see that pattern in nix all the time and I often ask me the question,
if that function is ever called with arguments that override the defaults.
If indeed it is, would it be better to make these arguments just regular/ mandatory arguments.
(or use partial application to specialize the function).
Somehow it seems to be common in nix to wrap every argument in a `Maybe`.
Maybe here is a case for `convention over configuration` and it is better to just decide for good that `state-dir=="state-cluster"`

MarcFontaine (approved):
LGTM

nix/supervisord-cluster/default.nix
MarcFontaine: An author of a library may wish to provide a generic and reusable API (because the user does not want to not modify the source code of the library).
But this is not a library. It is just a helper function that is not meant to be reused by third party.

MarcFontaine: Is this `"${__getEnv "HOME"}/.cache"` similar to `unsafePerformIO` ?

deepfire: Yes,  `cacheDir` computation pierces the abstraction, for the sake of providing a convenient default people will use in like 99.9% of time : -)

PR #2656, opened 2021-04-26
Add snapshot interval to the configuration of the node

EncodePanda:
Closes: https://github.com/input-output-hk/cardano-node/issues/2585

### DIFF
--- a/cardano-node/src/Cardano/Node/Configuration/POM.hs
+++ b/cardano-node/src/Cardano/Node/Configuration/POM.hs
@@ -1,7 +1,6 @@
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE NamedFieldPuns #-}
-{-# LANGUAGE StandaloneDeriving #-}
 
 {-# OPTIONS_GHC -Wno-noncanonical-monoid-instances #-}
 {-# OPTIONS_GHC -Wno-orphans            #-}
@@ -38,9 +37,6 @@ import           Cardano.Tracing.Config
 import           Ouroboros.Network.Block (MaxSlotNo (..))
 import           Ouroboros.Network.NodeToNode (DiffusionMode (..))
 import           Ouroboros.Consensus.Storage.LedgerDB.DiskPolicy (SnapshotInterval (..))
-
-deriving instance Eq SnapshotInterval
-deriving instance Generic SnapshotInterval
 
 data NodeConfiguration
   = NodeConfiguration
--- a/cardano-node/src/Cardano/Node/Configuration/POM.hs
+++ b/cardano-node/src/Cardano/Node/Configuration/POM.hs
@@ -1,8 +1,10 @@
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE StandaloneDeriving #-}
 
 {-# OPTIONS_GHC -Wno-noncanonical-monoid-instances #-}
+{-# OPTIONS_GHC -Wno-orphans            #-}
 
 module Cardano.Node.Configuration.POM
   ( NodeConfiguration (..)
@@ -35,6 +37,10 @@ import           Cardano.Node.Types
 import           Cardano.Tracing.Config
 import           Ouroboros.Network.Block (MaxSlotNo (..))
 import           Ouroboros.Network.NodeToNode (DiffusionMode (..))
+import           Ouroboros.Consensus.Storage.LedgerDB.DiskPolicy (SnapshotInterval (..))
+
+deriving instance Eq SnapshotInterval
+deriving instance Generic SnapshotInterval
 
 data NodeConfiguration
   = NodeConfiguration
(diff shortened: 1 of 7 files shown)


cardano-node/src/Cardano/Node/Configuration/POM.hs
```diff
@@ -250,6 +261,7 @@ defaultPartialNodeConfiguration =
     , pncLoggingSwitch = Last $ Just True
     , pncSocketPath = mempty
     , pncDiffusionMode = Last $ Just InitiatorAndResponderDiffusionMode
+    , pncSnapshotInterval = mempty
```
Jimbo4350, line 264: So instead of `fromMaybe DefaultSnapshotInterval $ getLast $ pncSnapshotInterval pnc` in the FromJSON instance, put `Last $ Just DefaultSnapshotInterval` here instead. This allows us to keep all our default configuration options in one place.

You can then change:
`fromMaybe DefaultSnapshotInterval $ getLast $ pncSnapshotInterval pnc`
to
`getLast $ pncSnapshotInterval pnc` 

`defaultPartialNodeConfiguration` will specify `DefaultSnapshotInterval` if nothing is specified in the node configuration yaml file.

Reference: https://medium.com/@jonathangfischoff/the-partial-options-monoid-pattern-31914a71fc67

cardano-node/src/Cardano/Node/Configuration/POM.hs
```diff
@@ -299,6 +312,7 @@ makeNodeConfiguration pnc = do
              , ncProtocolConfig = protocolConfig
              , ncSocketPath = getLast $ pncSocketPath pnc
              , ncDiffusionMode = diffusionMode
+             , ncSnapshotInterval = fromMaybe DefaultSnapshotInterval $ getLast $ pncSnapshotInterval pnc
```
Jimbo4350, line 315: See above comment

cardano-node/src/Cardano/Node/Parsers.hs
```diff
@@ -253,6 +258,15 @@ parseVrfKeyFilePath =
         <> completer (bashCompleter "file")
     )
 
+-- TODO revisit because it sucks
+parseSnapshotInterval :: Parser SnapshotInterval
```
Jimbo4350, line 262: Nice comment :smile: . I would hijack this: https://github.com/input-output-hk/cardano-node/blob/e32ca3408aa8f77b5d1bdbbd42634e1a048f8a97/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs#L2443
and this: https://github.com/input-output-hk/cardano-node/blob/e32ca3408aa8f77b5d1bdbbd42634e1a048f8a97/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs#L2535

then you can just use `Opt.option` and all is right in the world.

Jimbo4350 (commented):
Looking good! Just a couple comments.

cardano-node/src/Cardano/Node/Configuration/POM.hs
```diff
@@ -1,8 +1,10 @@
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE StandaloneDeriving #-}
 
 {-# OPTIONS_GHC -Wno-noncanonical-monoid-instances #-}
+{-# OPTIONS_GHC -Wno-orphans            #-}
```
Jimbo4350, line 7: Is the `{-# OPTIONS_GHC -Wno-orphans  #-}` pragma still necessary?

newhoggy: Looks like we don't actually use the `Generic` instance.

newhoggy: I created a PR to add these instances upstream.

EncodePanda: `SnapshotInterval` is missing (as you saw) two instances: `Generic` and `Eq`. I saw that @newhoggy already made a PR with needed changes thank you for that :) It is a change which we however can **NOT** use yet. 'Why??' one might wonder. Let me explain:
One important thing to understand is the fact that this PR is based on `ouroboros-consensus` merge commit `ec79b8725 Merge #2966` not the `master` branch. Why is that? Because the `master` has updated dependencies to `ledger-spec`, `cardano-prelude` and `cardano-base`. Those had breaking changes on the `cardano-node` and thus I've decided to update node's consensus dependency only up to relevant change (`ec79b8725 Merge #2966`).

What we can do about it:
1. I will make changes proposed by @Jimbo4350 and move the instances two `Cardano.Node.Orphans`
2. I will merge @newhoggy PR today to master(
3. Once node catches up with latest changes in consensus (that will probably happen during the Alonzo integration) we will be able to remove those orphans. Compiler will most likely tell us that this is needed.

Does that sounds like a plan?

cardano-node/src/Cardano/Node/Parsers.hs
EncodePanda: Omg, I forgot about that one 🤦  :) 
Thanks for the hint, I will have a look right away.

cardano-node/src/Cardano/Node/Configuration/POM.hs
newhoggy: Sounds great!  Thanks!

newhoggy (approved):

Jimbo4350 (changes requested):
See request re: node configuration

EncodePanda:
Closed in https://github.com/input-output-hk/cardano-node/commit/565486051ba67188fd9c3ed736e3f7f2b4ad2873

PR #2675, opened 2021-05-05
Hotfix: nix-shell clusters

deepfire:
This fixes a couple of bugs in `start-cluster` inside nix-shell.


nc6 (approved):

nc6 (changes requested):
Looks good, but I noticed the hydra eval fails...

deepfire:
@nc6, thanks fixed!

The remaining eval error is the same as in `master`: https://hydra.iohk.io/jobset/Cardano/cardano-node#tabs-errors

nc6 (approved):

PR #2676, opened 2021-05-06
CAD-2894 add wallet

MarcFontaine:
* Use wallet abstraction in benchmarking scripts.
* Refactor SubmissionClient.


deepfire (approved):
Looks good, thank you Marc!

PR #2678, opened 2021-05-06
CAD-2907 trace:  AddedToCurrentChain/SwitchedToAFork: log elongation as headBlockNo Δ

deepfire:
This adds an `chainLengthDelta` field to the `AddedToCurrentChain` and `SwitchedToAFork` ChainDB traces.

### DIFF
--- a/cardano-node/src/Cardano/Tracing/OrphanInstances/Consensus.hs
+++ b/cardano-node/src/Cardano/Tracing/OrphanInstances/Consensus.hs
@@ -26,6 +26,7 @@ import           Cardano.Tracing.Render (renderChainHash, renderChunkNo, renderH
                    renderHeaderHashForVerbosity, renderPoint, renderPointAsPhrase,
                    renderPointForVerbosity, renderRealPoint, renderRealPointAsPhrase,
                    renderTipBlockNo, renderTipForVerbosity, renderTipHash, renderWithOrigin)
+import           Cardano.Slotting.Slot (fromWithOrigin)
 
 import           Ouroboros.Consensus.Block (BlockProtocol, CannotForge, ConvertRawHash (..),
                    ForgeStateUpdateError, Header, RealPoint, getHeader, headerPoint, realPointHash,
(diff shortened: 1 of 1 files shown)


cardano-node/src/Cardano/Tracing/OrphanInstances/Consensus.hs
```diff
@@ -708,6 +711,8 @@ instance ( ConvertRawHash blk
          Just (_, _, _, s2 :: AF.AnchoredFragment (Header blk)) ->
            AF.toOldestFirst s2
          Nothing -> [] -- No sense to do validation here.
+     chainLengthΔ :: AF.AnchoredFragment (Header blk) -> AF.AnchoredFragment (Header blk) -> Int
+     chainLengthΔ = on (-) (fromWithOrigin (-1) . fmap (fromIntegral . unBlockNo) . AF.headBlockNo)
```
nfrisby, line 715: I'm unsure about this `-1`, but it probably doesn't matter or at least only in the very minor scenario of only immediately starting a fresh node.

deepfire: @nfrisby, I think that since `BlockNo 0` is right after Genesis, it follows that:

`chainLengthΔ (..AF @ BlockNo 0) (..AF @Genesis)` -> `1` -- which _seems_ to make sense.

Maybe.

nc6:
The term "elongation" can be a little confusing here, I might just stick with "chainLengthIncrease"

deepfire:
> The term "elongation" can be a little confusing here, I might just stick with "chainLengthIncrease"

Maybe `chainLengthDelta` then?

newhoggy:
`growth`?

deepfire:
@newhoggy, `growth` is similarly confusing -- it suggests increase -- whereas `lengthDelta` has no bias.

newhoggy:
`chainLengthDelta` sounds good.  😁

nc6:
Also fine with `chainLengthDelta`

nc6 (approved):

cardano-node/src/Cardano/Tracing/OrphanInstances/Consensus.hs
nfrisby: Ah; OK. Since we're applying this function to both arguments of `-`, I think this is good. And your reasoning seems right to me for Cardano, which is what's relevant here.

So: LGTM!

Jimbo4350 (approved):

PR #2688, opened 2021-05-10
In query tip command, make epoch number available on Byron

newhoggy:
This is done by using the history interpreter to convert slot number to epoch number.


newhoggy:
Resolves https://github.com/input-output-hk/cardano-node/issues/2568

newhoggy:
On `mainnet`:

```
CARDANO_NODE_SOCKET_PATH=local-node/socket cardano-cli query tip --mainnet
{
    "epoch": 264,
    "hash": "34197ddb795fdb50d6933f7490f79d77101392f9e66dc3f6ee358d218505124c",
    "slot": 29062271,
    "block": 5698710,
    "era": "Mary"
}
```

newhoggy:
On testnet:

```
CARDANO_NODE_SOCKET_PATH=example/node-bft2/node.sock cardano-cli query tip --testnet-magic 42
{
    "epoch": 11,
    "hash": "85d1f515ccede00557e28ff3fece650b509108806dfb255dee4e539bb3fe316a",
    "slot": 1112,
    "block": 1070,
    "era": "Byron"
}
```

cardano-api/src/Cardano/Api/Query.hs
```diff
@@ -81,26 +87,40 @@ import           Cardano.Api.ProtocolParameters
...
 
---     QueryEraHistory :: QueryInMode mode EraHistory
+data EraHistory mode where
+  EraHistory
+    :: ConsensusBlockForMode mode ~ Consensus.HardForkBlock xs
+    => ConsensusMode mode
+    -> History.Interpreter xs
+    -> EraHistory mode
+
+slotToEpoch :: SlotNo -> EraHistory mode -> Either Qry.PastHorizonException (EpochNo, Word64, Word64)
```
Jimbo4350, line 122: What are the two `Word64`s? A comment here would help.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -164,35 +169,43 @@ runQueryTip (AnyConsensusModeParams cModeParams) network mOutFile = do
...
+      -> ExceptT ShelleyQueryCmdError IO (Either Qry.PastHorizonException (EpochNo, Word64, Word64))
+    mSlotToEpoch cMode lNodeConnInfo slotNo = case cMode of
+      CardanoMode -> do
+        let epochQuery = QueryEraHistory CardanoModeIsMultiEra
+        eResult <- liftIO $ queryNodeLocalState lNodeConnInfo Nothing epochQuery
+        case eResult of
+          Left acqFail -> left (ShelleyQueryCmdAcquireFailure acqFail)
+          Right eraHistory -> return $ slotToEpoch slotNo eraHistory
+
+      mode -> left (ShelleyQueryCmdUnsupportedMode (show mode))
```
Jimbo4350, line 208: Rather than showing `ConsensusMode` we should wrap it in `AnyConsensusMode` and propagate that type through the errors. `AnyConsensusMode` can be rendered with `renderMode`

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -201,6 +214,13 @@ runQueryTip (AnyConsensusModeParams cModeParams) network mOutFile = do
       Aeson.Object $ obj <> HMS.fromList [name .= Aeson.Null]
     toObject _ _ _ = Aeson.Null
 
+    toJsonPastHorizonException :: Qry.PastHorizonException -> Aeson.Value
```
Jimbo4350, line 217: We should propagate this through the existing error type `ShelleyQueryCmdError`. I think it would be strange to render some errors via `ShelleyQueryCmdError` and some via JSON output.

Jimbo4350 (commented):
Nice work! This was a lot simpler than I thought it would be. A few comments below.

cardano-api/src/Cardano/Api/Query.hs
newhoggy: Added type aliases to help with that.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: It depends.  If we go via `ShelleyQueryCmdError` if something wrong happens we get no values.  If we do this instead, we will get partial data even when some queries fail.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: Done.  Thanks!

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: I've taken the approach where if the epoch query fails, the JSON will be printed without the epoch and an error is logged to `stdout`.

cardano-api/src/Cardano/Api/Query.hs
erikd: Where are the type aliases? Would it not be better the be a `newtype` ?

erikd (approved):

Jimbo4350 (approved):
Looking good! One minor comment

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -93,6 +99,13 @@ renderShelleyQueryCmdError err =
     ShelleyQueryCmdEraMismatch (EraMismatch ledgerEra queryEra) ->
       "\nAn error mismatch occured." <> "\nSpecified query era: " <> queryEra <>
       "\nCurrent ledger era: " <> ledgerEra
+    ShelleyQueryCmdUnsupportedMode mode -> "Unsupported mode: " <> renderMode mode
+    ShelleyQueryCmdPastHorizon e -> "Past horizon: " <> show e
+
+renderMode :: AnyConsensusMode -> Text
```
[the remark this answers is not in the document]
newhoggy: I've moved the function to a common module.

newhoggy:
bors merge

PR #2692, opened 2021-05-11
Sync progress 3

newhoggy:


cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -192,14 +223,31 @@ runQueryTip (AnyConsensusModeParams cModeParams) network mOutFile = do
     & fmap tuple3Fst
     & logExceptContinue renderShelleyQueryCmdError
 
+  tipTimeResult <- mProgressQuery anyEra consensusMode localNodeConnInfo tipSlotNo
+    <&> bimap toJsonPastHorizonException fst
+
+  let jsonTipTime :: Aeson.Value = either identity (toJSON . relativeTimeSeconds) tipTimeResult
+
+  systemStart <- mStartTime consensusMode localNodeConnInfo
```
newhoggy, line 231: The system start is obtained here.

PR #2693, opened 2021-05-11
Add more detail about #2547

nc6:
Introduced some breaking changes which are visible from the CLI and API
consumers.


Jimbo4350 (approved):

nc6:
bors merge

PR #2703, opened 2021-05-17
Fix typo in README

misli:


newhoggy:
Fixed in https://github.com/input-output-hk/cardano-node/pull/2860.

Thanks!

PR #2708, opened 2021-05-18
Add p2p support to nixos service

jbgi:
this switch iohk-nix to this branch: https://github.com/input-output-hk/iohk-nix/pull/474/files

### DIFF
--- a/nix/nixos/cardano-node-service.nix
+++ b/nix/nixos/cardano-node-service.nix
@@ -28,13 +28,45 @@ let
           (era: epoch:
             { "Test${era}HardForkAtEpoch" = epoch;
             })
-          cfg.forceHardForks));
+          cfg.forceHardForks))
+      // {
+        TargetNumberOfRootPeers = cfg.targetNumberOfRootPeers;
+        TargetNumberOfKnownPeers = cfg.targetNumberOfKnownPeers;
+        TargetNumberOfEstablishedPeers = cfg.targetNumberOfEstablishedPeers;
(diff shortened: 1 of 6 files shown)


coot (approved):

disassembler (approved):

nix/nixos/cardano-node-service.nix
```diff
@@ -404,6 +458,38 @@ in {
         description = ''Internal representation of the config.'';
       };
 
+      targetNumberOfRootPeers = mkOption {
+        type = types.int;
+        default = cfg.nodeConfig.TargetNumberOfRootPeers or 30;
```
karknu, line 463: I think that a default value of 30 is too few root peers. I propose that it is increased to 60.
Given your formulas below it means that by default the node will have 20 active peers which is what we have assumed.

nix/nixos/cardano-node-service.nix
```diff
@@ -14,13 +14,34 @@ let
...
-      Producers = cfg.producers ++ (cfg.instanceProducers i);
-    });
+    topology = if cfg.topology != null then cfg.topology else toFile "topology.yaml" (toJSON ({
+      LocalRoots = {
+        groups = map (g: {
+          localRoots = {
+            inherit (g) addrs;
+            advertise = g.advertise or false;
+          };
+          valency = g.valency or 1;
```
jbgi, line 33:
```suggestion
          valency = g.valency or (length g.addrs);
```

PR #2712, opened 2021-05-19
Avoid eval time dependency on cardano-cli and cardano-topology

hamishmack:
This avoids building all of `cardano-cli` and `cardano-topology` on the hydra eval server at `eval` time.

The job that causes the eval time issue is `linux.native.shell.x86_64-linux` and you can reproduce the problem on a linux machine by running:

```
nix-collect-garbage
nix-instantiate release.nix -A linux.native.shell.x86_64-linux
```

It should not be necessary to download or build `cardano-cli` or `cardano-topology` for this to work.


rvl (approved):
Looks good.
So now it doesn't need a large set of dependencies just to generate the json file which is loaded at eval time.

PR #2728, opened 2021-05-21
Added error message in case Topology fails to parse in SIGHUP feature

bolt12:
Failing to parse in SIGHUP feature does not kill the node but the error is silent as it was not logged. Now it is.

### DIFF
--- a/cardano-node/src/Cardano/Node/Run.hs
+++ b/cardano-node/src/Cardano/Node/Run.hs
@@ -242,6 +242,10 @@ handleSimpleNode scp runP trace nodeTracers nc onKernel = do
 
   nt <- readTopologyFileOrError nc
 
+  traceNamedObject
+          (appendName "topology-file" trace)
+          (meta, LogMessage (Text.pack "Successfully read topology configuration file"))
+
   let (localRoots, publicRoots) = producerAddresses nt
 
   localRootsVar <- newTVarIO localRoots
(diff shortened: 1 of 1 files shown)


cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -344,24 +348,34 @@ handleSimpleNode p trace nodeTracers nc onKernel = do
...
-      traceNamedObject
-        (appendName "public-roots" trace)
-        (meta, LogMessage . Text.pack . show $ publicRoots)
-      traceNamedObject
-        (appendName "use-ledger-after-slot" trace)
-        (meta, LogMessage . Text.pack . show $ useLedgerAfterSlot nt)
+      result <- try $ readTopologyFileOrError nc
+
+      case result of
+        Left (FatalError err) ->
```
coot, line 354: what is `FatalError`, is this some specaill exception thrown by `readTopologyFileOrError`?  Will this handle `IOException`s?

bolt12: FatalError is thrown by `readTopologyFileOrError` (https://github.com/input-output-hk/cardano-node/blob/p2p-master/cardano-node/src/Cardano/Node/Configuration/TopologyP2P.hs#L182)

coot (approved):
LGTM

PR #2746, opened 2021-05-26
CAD-2907 tracing:  improve usability of the ChainSync server traces

deepfire:
Instead of going through the Tip/Point serialisations
this allows a more direct access to things that matter:

1. block hash
2. slot number
3. block number

We only convert immediately necessary traces, but the portal has been opened and the call is out.

Release-relevant information (cc @LaurenceIO): this PR changes structure of some of the traces configured by `TraceChainSyncHeaderServer` and `TraceChainSyncClient` config options.


karknu (approved):
LGTM

Pornprasert:
Leave

PR #2751, opened 2021-05-27
cardano-api:test: Add roundtrip tests for getTransactionBodyContent/makeTransactionBody

cblp:
Also add txix bounds check in `makeTransactionBody`.


cblp:
Tests have found this
```diff
      , txAuxScripts =
          TxAuxScripts
            AuxScriptsInAllegraEra [
              ScriptInEra
-               SimpleScriptV1InAllegra
+               SimpleScriptV2InAllegra
                SimpleScript
-                 SimpleScriptV1
+                 SimpleScriptV2
                  RequireSignature
                    "ebe9de78a37f84cc819c0669791aa0474d4f0a764e54b9f90cfe2137"
            ]
```

@dcoutts how to deal with this?

cardano-api/test/Test/Cardano/Api/TxBody.hs
```diff
@@ -0,0 +1,221 @@
...
+normalizeOriginal :: TxBodyContent ViewTx era -> TxBodyContent ViewTx era
+normalizeOriginal content =
+  content
+    { txIns          = sortOn fst            $ txIns          content
+    , txMetadata     = normalizeMetadata     $ txMetadata     content
+    , txAuxScripts   = normalizeAuxScripts   $ txAuxScripts   content
+    , txWithdrawals  = normalizeWithdrawals  $ txWithdrawals  content
+    , txCertificates = normalizeCertificates $ txCertificates content
+    , txMintValue    = normalizeMintValue    $ txMintValue    content
+    }
```
newhoggy, line 60: Is it documented anywhere what it means to "normalize"?

cardano-api/test/Test/Cardano/Api/TxBody.hs
```diff
@@ -0,0 +1,221 @@
...
+        TxBody content -> content
+        _              -> panic "impossible")
+      (makeTransactionBody . rebuildBodyContent)
+
+
+rebuildBodyContent :: TxBodyContent build era -> TxBodyContent BuildTx era
+rebuildBodyContent TxBodyContent{..} =
+  TxBodyContent
+    { txIns            = map rebuildTxIn     txIns
+    , txProtocolParams = panic "rebuildBodyContent: txProtocolParams"
```
newhoggy, line 186: What's the reason for the `panic` here?

cardano-api/test/Test/Cardano/Api/TxBody.hs
```diff
@@ -0,0 +1,221 @@
...
+  (TxIn, BuildTxWith BuildTx (Witness WitCtxTxIn era))
+rebuildTxIn = second $ const $ BuildTxWith $ KeyWitness KeyWitnessForSpending
+
+rebuildWithdrawals :: TxWithdrawals build era -> TxWithdrawals BuildTx era
+rebuildWithdrawals = \case
+  TxWithdrawalsNone                 -> TxWithdrawalsNone
+  TxWithdrawals support withdrawals ->
+    TxWithdrawals
+      support
+      [ (address, amount, panic "rebuildWithdrawals")
```
newhoggy, line 204: What's the reason for the `panic` here?

cardano-api/test/Test/Cardano/Api/TxBody.hs
```diff
@@ -0,0 +1,221 @@
...
+      | (address, amount, _) <- withdrawals
+      ]
+
+rebuildCertificates :: TxCertificates build era -> TxCertificates BuildTx era
+rebuildCertificates = \case
+  TxCertificatesNone                    -> TxCertificatesNone
+  TxCertificates support certificates _ ->
+    TxCertificates support certificates $ panic "rebuildCertificates"
```
newhoggy, line 212: What's the reason for the `panic` here?

cardano-api/test/Test/Cardano/Api/TxBody.hs
cblp: Added a line before. Is it OK or needs more docs?

cardano-api/test/Test/Cardano/Api/TxBody.hs
cblp: Added more descriptive message.

cardano-api/test/Test/Cardano/Api/TxBody.hs
cblp: Added more descriptive message.

cardano-api/test/Test/Cardano/Api/TxBody.hs
cblp: Added more descriptive message.

cardano-api/test/Test/Cardano/Api/TxBody.hs
```diff
@@ -0,0 +1,233 @@
...
+    content <- forAll $ genTxBodyContent era
+    tripping
+      (normalizeOriginal $ review content)
+      (\_ -> makeTransactionBody content)
+      (<&> \(TxBody content') -> normalizeRoundtrip content')
+
+
+-- * normalize = strip unnecessary details
+
+normalizeOriginal :: TxBodyContent ViewTx era -> TxBodyContent ViewTx era
```
Jimbo4350, line 51: Are these normalizations necessary? Do the roundtrip tests fail without them?

cardano-api/test/Test/Cardano/Api/TxBody.hs
```diff
@@ -0,0 +1,233 @@
...
+normalizeMetadata :: TxMetadataInEra era -> TxMetadataInEra era
+normalizeMetadata = \case
+  TxMetadataInEra _ (TxMetadata m) | null m -> TxMetadataNone
+  other                                     -> other
+
+normalizeAuxScripts :: TxAuxScripts era -> TxAuxScripts era
+normalizeAuxScripts = \case
+  TxAuxScripts _       []      -> TxAuxScriptsNone
+  TxAuxScripts support scripts ->
+    TxAuxScripts support $ map upgradeScriptInEra scripts
```
Jimbo4350, line 81: Why are we upgrading the scripts?

Jimbo4350 (changes requested):
A couple of comments

cardano-api/test/Test/Cardano/Api/TxBody.hs
cblp: Yes, after roundtrip, `Just mempty` may become `Nothing` or vice versa.

We either normalize them to the same value or write equality alternative equality functions.

cardano-api/test/Test/Cardano/Api/TxBody.hs
cblp: The instruction set from V1 may be used as V2, and we can't determine the language version from the transaction. When the user uses V1 instructions, we don't know if they used the V1 tag or the V2 tag. So it's safe to drop this information.

Jimbo4350 (changes requested):
Looking good but a few comments. I will finish the review tomorrow.

cardano-api/gen/Gen/Cardano/Api/Typed.hs
```diff
@@ -667,50 +674,50 @@ genProtocolParameters =
...
+  protocolUpdatePrices              <- ifAlonzoBased era genExecutionUnitPrices
+  protocolUpdateMaxTxExUnits        <- ifAlonzoBased era genExecutionUnits
+  protocolUpdateMaxBlockExUnits     <- ifAlonzoBased era genExecutionUnits
+  protocolUpdateMaxValueSize        <- ifAlonzoBased era genNat
+  protocolUpdateCollateralPercent   <- ifAlonzoBased era genNat
+  protocolUpdateMaxCollateralInputs <- ifAlonzoBased era genNat
+  pure ProtocolParametersUpdate{..}
+
+
+genUpdateProposal :: CardanoEra era -> Gen UpdateProposal
```
dcoutts, line 715: For example, I doubt this needs to be era-dependent now. Isn't it just the ProtocolParameters where we have a new property that needs a value of the parameters with certain params set?

cardano-api/src/Cardano/Api/ProtocolParameters.hs
```diff
@@ -973,22 +972,33 @@ toAlonzoPParamsUpdate
     , Alonzo._minPoolCost     = toShelleyLovelace <$>
                                   maybeToStrictMaybe protocolUpdateMinPoolCost
     , Alonzo._coinsPerUTxOWord  = toShelleyLovelace <$>
-                                  maybeToStrictMaybe protocolUpdateUTxOCostPerWord
+                                  maybeToStrictMaybe
+                                    (protocolUpdateUTxOCostPerWord >>= non 0)
```
dcoutts, line 976: I don't understand what is going on here, or why. Zero is a perfectly acceptable value of the protocol param. Setting the param to zero is not the same as not changing the param it at all.

cardano-api/src/Cardano/Api/ProtocolParameters.hs
```diff
@@ -973,22 +972,33 @@ toAlonzoPParamsUpdate
...
+                                        protocolUpdateCollateralPercent
+                                        >>= non 0
+    , Alonzo._maxCollateralInputs   = maybeToStrictMaybe $
+                                        protocolUpdateMaxCollateralInputs
+                                        >>= non 0
     }
 
+-- | Check if value is not special.
+non :: (Alternative f, Eq a) => a -> a -> f a
+non z n = n <$ guard (n /= z)
```
dcoutts, line 1001: All the uses of this are deeply suspicious. There are not special values like this.

cardano-api/src/Cardano/Api/ProtocolParameters.hs
```diff
@@ -426,6 +424,7 @@ data ProtocolParametersUpdate =
        -- | The minimum permitted value for new UTxO entries, ie for
        -- transaction outputs.
        --
+       -- /Removed in Alonzo/
```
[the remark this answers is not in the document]
dcoutts: As Jordan says, or perhaps "No longer used in Alonzo, replaced by ...

Since it's not removed, it's plainly still there :-)

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -1458,18 +1463,17 @@ fromLedgerTxInsCollateral
...
-               -> Set (Shelley.TxIn StandardCrypto)
-    collateral ShelleyBasedEraShelley = const Set.empty
-    collateral ShelleyBasedEraAllegra = const Set.empty
-    collateral ShelleyBasedEraMary    = const Set.empty
-    collateral ShelleyBasedEraAlonzo  = Alonzo.collateral'
+    collateral = case era of
+      ShelleyBasedEraShelley -> Set.empty
+      ShelleyBasedEraAllegra -> Set.empty
+      ShelleyBasedEraMary    -> Set.empty
+      ShelleyBasedEraAlonzo  -> Alonzo.collateral' body
```
dcoutts, line 1476: Any reason for this change?

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -1618,11 +1622,14 @@ fromLedgerTxExtraKeyWitnesses sbe body =
     ShelleyBasedEraShelley -> TxExtraKeyWitnessesNone
     ShelleyBasedEraAllegra -> TxExtraKeyWitnessesNone
     ShelleyBasedEraMary    -> TxExtraKeyWitnessesNone
-    ShelleyBasedEraAlonzo  -> TxExtraKeyWitnesses
+    ShelleyBasedEraAlonzo
+      | null keyhashes     -> TxExtraKeyWitnessesNone
```
dcoutts, line 1626: Can we use `Set.null` here just to be explicit. The `Foldable` null overloading is a bit sketchy when it comes to the asymptotic performance. Yes, the set instance overrides it, but it's easy to make a representation change and not notice you've suddenly changed the asymptotic performance.

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -1767,8 +1774,10 @@ fromLedgerTxMintValue era body =
 makeByronTransactionBody :: TxBodyContent BuildTx ByronEra
                          -> Either (TxBodyError ByronEra) (TxBody ByronEra)
 makeByronTransactionBody TxBodyContent { txIns, txOuts } = do
-    ins'  <- NonEmpty.nonEmpty txIns      ?! TxBodyEmptyTxIns
-    let ins'' = NonEmpty.map (toByronTxIn . fst) ins'
+    ins' <- NonEmpty.nonEmpty (map fst txIns) ?! TxBodyEmptyTxIns
+    for_ ins' $ \txin@(TxIn _ (TxIx txix)) ->
+      guard (txix <= maxByronTxInIx) ?! TxBodyInIxOverflow txin
+    let ins'' = fmap toByronTxIn ins'
```
dcoutts, line 1780: If this is fixing a bug, it'd be better to pull it out into its own commit and say so.

cardano-api/test/Test/Cardano/Api/Typed/CBOR.hs
```diff
@@ -5,62 +5,35 @@ module Test.Cardano.Api.Typed.CBOR
...
-  roundtrip_CBOR (AsTx AsMaryEra) (genTx MaryEra)
-
-prop_roundtrip_tx_alonzo_CBOR :: Property
-prop_roundtrip_tx_alonzo_CBOR =
-  roundtrip_CBOR (AsTx AsAlonzoEra) (genTx AlonzoEra)
+test_roundtrip_txbody_CBOR :: [TestTree]
+test_roundtrip_txbody_CBOR =
+  [ testProperty (show era) $
+    roundtrip_CBOR (proxyToAsType Proxy) (genTxBody era)
+  | AnyCardanoEra era <- allCardanoEras
```
dcoutts, line 29: Yes, indeed better and shorter. Use the `[minBound..maxBound]` idiom here.

cardano-api/gen/Gen/Cardano/Api.hs
```diff
@@ -22,10 +22,10 @@ import qualified Hedgehog.Range as Range
...
 genMetadata :: MonadGen m => m (Metadata era)
 genMetadata = do
-  numberOfIndicies <- Gen.integral (Range.linear 1 15)
-  let indexes = map (\i -> fromIntegral i :: Word64) [1..numberOfIndicies]
-  mDatums <- Gen.list (Range.singleton numberOfIndicies) genMetadatum
-  return . Metadata . Map.fromList $ zip indexes mDatums
+  numberOfIndices <- Gen.integral (Range.linear 1 15)
+  let indices = map (\i -> fromIntegral i :: Word64) [1..numberOfIndices]
+  mData <- Gen.list (Range.singleton numberOfIndices) genMetadatum
+  return . Metadata . Map.fromList $ zip indices mData
```
dcoutts, line 28: For example this should be it's own (micro) commit that just says it's fixing some spelling, so that we don't have to peer at it to see what is being changed or why.

cardano-api-test/test/Test/Cardano/Api/TxBody.hs
```diff
@@ -0,0 +1,263 @@
...
+normalizeMetadata = \case
+  TxMetadataInEra _ (TxMetadata m) | null m -> TxMetadataNone
+  other                                     -> other
+
+normalizeAuxScripts :: TxAuxScripts era -> TxAuxScripts era
+normalizeAuxScripts = \case
+  TxAuxScripts _       []      -> TxAuxScriptsNone
+  TxAuxScripts support scripts ->
+    -- sorting uses script versions, hence sort after upgrade
+    TxAuxScripts support $ sort $ map upgradeScriptInEra scripts
```
dcoutts: So this is where you are using the new `Ord` instance on the `ScriptInEra` type. If this just needs to be sorted by the script language/version then you don't need that `Ord` instance (which as you found, is quite complicated and needs lots of other `Ord` instances). I think all you would need here is to use `sortBy` and to use a comparison function that uses the `AnyScriptLanguage` type which already has an `Ord` instance.

cardano-api/src/Cardano/Api/Script.hs
```diff
@@ -390,6 +390,11 @@ data Script lang where
 deriving instance (Eq   (Script lang))
 deriving instance (Show (Script lang))
 
+instance Ord (Script lang) where
+  -- versions must be the same
+  compare (SimpleScript _ sA) (SimpleScript _ sB) = compare sA sB
+  compare (PlutusScript _ sA) (PlutusScript _ sB) = compare sA sB
```
dcoutts, line 396: Double check that you _really_ need all these new tricky `Ord` instances. I'm moderately confident that you could do it all with just `AnyScriptLanguage` which already has an `Ord` instance (where it naturally makes sense since that type covers all the languages and their versions).

cardano-api/src/Cardano/Api/Script.hs
```diff
@@ -879,6 +894,24 @@ data SimpleScript lang where
 deriving instance Eq   (SimpleScript lang)
 deriving instance Show (SimpleScript lang)
 
+instance Ord (SimpleScript lang) where
```
dcoutts, line 897: I'm extremely sceptical that we need to order actual scripts by their value. At most I can imagine a round trip via the ledger types putting things into order based on hash, but not the value. This must be unnecessary, surely?

cardano-api-test/test/Test/Cardano/Api/TxBody.hs
```diff
@@ -0,0 +1,263 @@
...
+    zeroIsNothing :: (Eq a, Num a) => Maybe a -> Maybe a
+    zeroIsNothing = \case
+      Just 0 -> Nothing
+      x      -> x
+    zeroPricesIsNothing = \case
+      Just (ExecutionUnitPrices 0 0) -> Nothing
+      x                              -> x
+    zeroExecutionIsNothing = \case
+      Just (ExecutionUnits 0 0) -> Nothing
+      x                         -> x
```
dcoutts: These seem very suspicious. There should be a difference between setting a param to zero and not changing it at all. This seems wrong, or working around a problem elsewhere. What was going wrong here? Perhaps it's a real bug.

cardano-api-test/test/Test/Cardano/Api/TxBody.hs
```diff
@@ -0,0 +1,263 @@
...
+
+
+-- * Ugrading scripts
+--
+-- The instruction set from V1 may be used as V2,
+-- and we can't determine the language version from the transaction.
+-- When the user uses V1 instructions,
+-- we don't know if they used the V1 tag or the V2 tag.
+-- So it's safe to drop this information.
```
dcoutts: :+1: yes, good explanation, and useful to have here since it's not obvious.

cardano-api-test/test/Test/Cardano/Api/TxBody.hs
```diff
@@ -0,0 +1,263 @@
...
+review :: TxBodyContent BuildTx era -> TxBodyContent ViewTx era
+review TxBodyContent{..} =
+  TxBodyContent
+    { txCertificates    = reviewCertificates txCertificates
+    , txExtraScriptData = ViewTx
+    , txIns             = map reviewTxIn     txIns
+    , txMintValue       = reviewMintValue    txMintValue
+    , txProtocolParams  = ViewTx
+    , txWithdrawals     = reviewWithdrawals  txWithdrawals
+    , ..
```
dcoutts: :+1: I think this is a legitimate use of record wildcards, so keep it, despite the typical style guide advice.

cardano-api-test/test/Test/Cardano/Api/TxBody.hs
```diff
@@ -0,0 +1,263 @@
...
+import           Test.Tasty.Hedgehog (testProperty)
+import           Test.Tasty.TH (testGroupGenerator)
+
+import           Cardano.Api
+
+import           Gen.Cardano.Api.Typed (genTxBody, genTxBodyContent)
+
+
+test_roundtrip_TxBody_make_get :: [TestTree]
+test_roundtrip_TxBody_make_get =
```
dcoutts: I think this is sufficiently interesting and tricky to justify a decent comment on how this round trip test works. What are we going from and to? Why is there normalisation involved, and what normalisation are we doing? What is `review` about etc. Perhaps draw out the diagram of the conversions we're checking are equal.

It might be good to put the other property `test_roundtrip_TxBody_get_make` immediately after and document that too, to show the other direction we're checking. Seeing both together will make it clearer what we're checking overall. And presumably that one will be the same diagram, but checking different paths through the diagram, right?

cardano-api-test/test/Test/Cardano/Api/TxBody.hs
```diff
@@ -0,0 +1,263 @@
...
+
+test_roundtrip_TxBody_get_make :: [TestTree]
+test_roundtrip_TxBody_get_make =
+  [ testProperty (show era) $
+    property $ do
+      txbody <- forAll $ genTxBody era
+      tripping
+        txbody
+        (\(TxBody content) -> content)
+        (makeTransactionBody . rebuildBodyContent)
```
dcoutts: When you document this property, make it clear why we need `genTxBody` and `rebuildBodyContent` rather than `genTxBody`. It's not clear on first reading, and `rebuildBodyContent` of course adds quite a bit of code, so there must be a good reason!

dcoutts (changes requested):
This is great, and clearly a lot of work, and important useful work. It's mostly ready.

There's a number of comments below, but the main meta comment is that there is a huge amount going on in a single commit here. Please take the approach of optimising PRs for review by your peers by splitting logical changes out into their own patches in the PR. This gives you the opportunity in the commit message to say what you are doing any why, rather than making the reviewer guess, and also try to disentangle all the different changes that are going on at once.

This single commit contains isolated bug fixes, multiple independent changes to API code, changes to general purpose generator infrastructure, nice simplifications of existing property tests, and of course the main thing of a new test for the round trip property. I'm pretty sure most of these things are logically independent, and that everything would compile after each change.

This would probably end up at around 10 commits I'd guess. There's a lot here.

cardano-api/gen/Gen/Cardano/Api/Typed.hs
```diff
@@ -1,6 +1,8 @@
 {-# LANGUAGE GADTs #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards #-}
```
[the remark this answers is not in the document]
cblp: done

cardano-api/gen/Gen/Cardano/Api/Typed.hs
```diff
@@ -733,6 +744,27 @@ genExecutionUnits :: Gen ExecutionUnits
 genExecutionUnits = ExecutionUnits <$> Gen.integral (Range.constant 0 1000)
                                    <*> Gen.integral (Range.constant 0 1000)
 
+-- | 'Gen.maybe' but with condition if era is Alonzo-based
+ifAlonzoBased :: CardanoEra era -> Gen a -> Gen (Maybe a)
+ifAlonzoBased era gen
+  | isAlonzoBased era = Gen.maybe gen
```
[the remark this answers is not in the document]
cblp: This is not future-proof. What will we do in the next era?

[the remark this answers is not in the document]
cblp: No, we don't want to have all fields of ProtocolParametersUpdate always filled, we want some of them filled, some empty.

cardano-api/gen/Gen/Cardano/Api/Typed.hs
```diff
@@ -646,8 +653,8 @@ genPraosNonce = makePraosNonce <$> Gen.bytes (Range.linear 0 32)
 genMaybePraosNonce :: Gen (Maybe PraosNonce)
 genMaybePraosNonce = Gen.maybe genPraosNonce
 
-genProtocolParameters :: Gen ProtocolParameters
-genProtocolParameters =
+genProtocolParameters :: CardanoEra era -> Gen ProtocolParameters
```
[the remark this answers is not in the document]
cblp: It is needed specifically for `protocolParamCostModels` which must be empty before Alonzo and must be eventually non-empty for Alonzo.

Yes, for simple round-trip things we cannot omit an era, but we can walk through all eras, and we do it.

cardano-api/gen/Gen/Cardano/Api/Typed.hs
cblp: genProtocolParametersUpdate has 9 era-dependent fields.

I'm not sure I fully understand the question.

Do you want to generate all parameters regardless of the era? And just ignore unknown parameters in the end?

cardano-api/src/Cardano/Api/ProtocolParameters.hs
cblp: ok, removed

cardano-api/src/Cardano/Api/ProtocolParameters.hs
cblp: Well, it's strange, I found that implementation treats Just 0 and Nothing the same way, so I added some unification here. Now tests say it is wrong. I removed this.

cardano-api/src/Cardano/Api/TxBody.hs
cblp: While I was reading this code, I reformatted it to help myself understand it. I think it's simpler and clear now. Do you think it has some performance downsides?

cardano-api/src/Cardano/Api/TxBody.hs
cblp: Do you know containers with `null` slower than O(1)?

cardano-api/test/Test/Cardano/Api/Typed/CBOR.hs
cblp: I've done this mainly to keep it working when new eras are added.

cardano-api-test/test/Test/Cardano/Api/TxBody.hs
cblp: Oops, I rewrote it without RWC before I found this comment. Should I return to RWC?

cardano-api/src/Cardano/Api/Script.hs
cblp: we need to unify
```hs
, txAuxScripts =
  TxAuxScripts AuxScriptsInAlonzoEra
    [ ScriptInEra
        PlutusScriptV1InAlonzo
        (PlutusScript PlutusScriptV1 (PlutusScriptSerialised ""))
    , ScriptInEra
        SimpleScriptV2InAlonzo
        (SimpleScript SimpleScriptV2 (RequireSignature "ebe9de78a37f84cc819c0669791aa0474d4f0a764e54b9f90cfe2137"))
    ]
```
and
```hs
, txAuxScripts =
  TxAuxScripts AuxScriptsInAlonzoEra
    [ ScriptInEra
        SimpleScriptV2InAlonzo
        (SimpleScript SimpleScriptV2 (RequireSignature "ebe9de78a37f84cc819c0669791aa0474d4f0a764e54b9f90cfe2137"))
    , ScriptInEra
        PlutusScriptV1InAlonzo
        (PlutusScript PlutusScriptV1 (PlutusScriptSerialised ""))
    ]
```
What is a proper way to do it?

cblp: nevermind, done

cardano-api/gen/Gen/Cardano/Api/Typed.hs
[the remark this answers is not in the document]
Jimbo4350: >No, we don't want to have all fields of ProtocolParametersUpdate always filled, we want some of them filled, some empty.
We want all the Alonzo relevant protocol parameter fields filled in the Alonzo era, and the pre-Alonzo era protocol parameter fields to default to `Nothing`.

cblp:
@dcoutts @Jimbo4350 I've split it into sensible commits, please review. (also, while re-reading my own code, fixed a freshly introduced bug)

cardano-api/test/Test/Cardano/Api/TxBody.hs
```diff
@@ -0,0 +1,285 @@
...
+-- | Original data: Unify empty and None.
+--
+-- Roundtrip data: Sort transaction input ids.
+normalizeInsCollateral :: TxInsCollateral era -> TxInsCollateral era
+normalizeInsCollateral = \case
+  TxInsCollateralNone -> TxInsCollateralNone
+  -- for original
+  TxInsCollateral _ ins | null ins -> TxInsCollateralNone
+  -- for roundtrip
+  TxInsCollateral support ins -> TxInsCollateral support $ sort ins
```
Jimbo4350, line 105: We could refactor as follows:
```
  TxInsCollateral support ins
    | null ins -> TxInsCollateralNone
    | otherwise -> TxInsCollateral support $ sort ins
 ```

File: cardano-api/test/Test/Cardano/Api/TxBody.hs [hunk 28]
```diff
@@ -0,0 +1,285 @@
...
+normalizeMintValue :: TxMintValue ViewTx era -> TxMintValue ViewTx era
+normalizeMintValue = \case
+  TxMintValue _ v _ | v == mempty -> TxMintNone
+  other                           -> other
+
+
+-- * Ugrading scripts
+--
+-- The instruction set from V1 may be used as V2,
+-- and we can't determine the language version from the transaction.
```
Jimbo4350, line 151: We should change this comment to: "We can't determine the script language version from the ledger transaction. Therefore we default to the highest version for the time being. TODO: Revisit this"

cardano-api/test/Test/Cardano/Api/TxBody.hs
```diff
@@ -0,0 +1,285 @@
...
+  RequireAllOf scripts  -> RequireAllOf $ map upgradeSimpleScript scripts
+  RequireAnyOf scripts  -> RequireAnyOf $ map upgradeSimpleScript scripts
+  RequireMOf n scripts  -> RequireMOf n $ map upgradeSimpleScript scripts
+
+
+-- * Review: Change 'TxBodyContent' “phase” to 'ViewTx'
+--
+-- Basically, it's just dropping witness information.
+
+review :: TxBodyContent BuildTx era -> TxBodyContent ViewTx era
```
Jimbo4350, line 183: `review` isn't a very indicative name. We should rename this to something like "toTxBodyContentView".

cardano-api/test/Test/Cardano/Api/TxBody.hs
```diff
@@ -0,0 +1,285 @@
...
+    , txMetadata        =                      txMetadata       body
+    , txMintValue       = reviewMintValue    $ txMintValue      body
+    , txOuts            =                      txOuts           body
+    , txProtocolParams  = ViewTx
+    , txUpdateProposal  =                      txUpdateProposal body
+    , txValidityRange   =                      txValidityRange  body
+    , txWithdrawals     = reviewWithdrawals  $ txWithdrawals    body
+    }
+
+reviewTxIn ::
```
Jimbo4350, line 202: Same for all of these `reviewX` functions. We should rename them to something like "toTxInView".

Jimbo4350 (commented):
This is a lot better and easier to look at. Good job, just a few comments.

cardano-api-test/test/Test/Cardano/Api/TxBody.hs
cblp: This is a roundtrip path. On the way back, we start with `TxBodyContent ViewTx`, and to get back to `TxBody` we have to build it again.

Should I put this in the code?

cardano-api/gen/Gen/Cardano/Api/Typed.hs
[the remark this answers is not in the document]
cblp: all the Alonzo relevant protocol parameter fields filled in the Alonzo era _simultaneously_?

cardano-api/test/Test/Cardano/Api/Typed/CBOR.hs
```diff
@@ -22,45 +24,18 @@ import           Test.Cardano.Api.Typed.Orphans ()
...
-
-prop_roundtrip_tx_mary_CBOR :: Property
-prop_roundtrip_tx_mary_CBOR =
-  roundtrip_CBOR (AsTx AsMaryEra) (genTx MaryEra)
-
-prop_roundtrip_tx_alonzo_CBOR :: Property
-prop_roundtrip_tx_alonzo_CBOR =
-  roundtrip_CBOR (AsTx AsAlonzoEra) (genTx AlonzoEra)
+test_roundtrip_txbody_CBOR :: [TestTree]
+test_roundtrip_txbody_CBOR =
```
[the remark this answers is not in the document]
cblp: No, tasty-th uses prefixes to determine the type of a test, and test groups allow `test_` only.

cardano-api/test/Test/Cardano/Api/TxBody.hs
cblp: I was looking for a prefix for a group of functions. What about "view" and "build"?
```
viewBodyContent
viewTxIn
viewWithdrawals
viewCertificates
viewMintValue
buildBodyContent
buildTxIn
buildWithdrawals
buildCertificates
buildMintValue
```

cblp: Because I think prefixes serve to group the functions better than suffixes.

cardano-api/src/Cardano/Api/TxBody.hs
Jimbo4350: I think we should leave a type signature

cardano-api/test/Test/Cardano/Api/TxBody.hs
Jimbo4350: "view" is better imo :+1:

cardano-api/src/Cardano/Api/Eras.hs
```diff
@@ -200,6 +199,30 @@ instance Eq AnyCardanoEra where
...
+      AnyCardanoEra MaryEra    -> 3
+      AnyCardanoEra AlonzoEra  -> 4
+
+   toEnum = \case
+      0 -> AnyCardanoEra ByronEra
+      1 -> AnyCardanoEra ShelleyEra
+      2 -> AnyCardanoEra AllegraEra
+      3 -> AnyCardanoEra MaryEra
+      4 -> AnyCardanoEra AlonzoEra
+      n -> error $ "AnyCardanoEra.toEnum: bad argument: " <> show n
```
Jimbo4350, line 224: A better error message would be: `"AnyCardanoEra.toEnum: " <> show n <> " does not correspond to any known enumerated era." `

cardano-api/gen/Gen/Cardano/Api/Typed.hs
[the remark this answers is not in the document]
Jimbo4350: > all the Alonzo relevant protocol parameter fields filled in the Alonzo era simultaneously?
Yes they have to be. All of the Alonzo related protocol parameters are required in the Alonzo era whilst setting the minUTxO parameter to `Nothing`.

cardano-api/test/Test/Cardano/Api/TxBody.hs
```diff
@@ -0,0 +1,285 @@
...
+-- * Build: Change 'TxBodyContent' “phase” to 'BuildTx'
+--
+-- Here we make up the most trivial witnesses or plug holes with 'panic'
+-- to make sure the fields are not touched.
+
+buildBodyContent :: TxBodyContent ViewTx era -> TxBodyContent BuildTx era
+buildBodyContent body =
+  TxBodyContent
+    { txAuxScripts      =                     txAuxScripts     body
+    , txCertificates    = buildCertificates $ txCertificates   body
```
Jimbo4350, line 237: Alignment

cardano-api/test/Test/Cardano/Api/TxBody.hs
```diff
@@ -194,6 +194,7 @@ viewBodyContent body =
     , txMintValue       = viewMintValue    $ txMintValue      body
     , txOuts            =                    txOuts           body
     , txProtocolParams  = ViewTx
+    , txScriptValidity  =                    txScriptValidity body
     , txUpdateProposal  =                    txUpdateProposal body
     , txValidityRange   =                    txValidityRange  body
```
Jimbo4350, line 199: This can be merged with the previous commit

Jimbo4350 (commented):
Couple more comments but almost there.

cardano-api/test/Test/Cardano/Api/TxBody.hs
cblp: Do you think misalignment will make it more maintainable?

cardano-api/gen/Gen/Cardano/Api/Typed.hs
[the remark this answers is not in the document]
cblp: so, in Alonzo, `{prices = Just _, costModels = Nothing}` is an invalid update?

cardano-api/test/Test/Cardano/Api/TxBody.hs
cblp: done

cardano-api/gen/Gen/Cardano/Api/Typed.hs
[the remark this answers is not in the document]
Jimbo4350: > so, in Alonzo, {prices = Just _, costModels = Nothing} is an invalid update?
Yes this is an invalid update

cardano-api/gen/Gen/Cardano/Api/Typed.hs
```diff
@@ -772,11 +770,13 @@ genExecutionUnits :: Gen ExecutionUnits
...
-ifAlonzoBased :: CardanoEra era -> Gen a -> Gen (Maybe a)
-ifAlonzoBased era gen
-  | isAlonzoBased era = Gen.maybe gen
-  | otherwise         = pure Nothing
+-- | Gen for Alonzo-specific parameters
+alonzoParam :: CardanoEra era -> Gen a -> Gen (Maybe a)
+alonzoParam era gen =
+  case era of
+    AlonzoEra             -> Just <$> gen
+    _ | isAlonzoBased era -> Gen.maybe gen
```
Jimbo4350, line 803: Why this extra case with `isAlonzoBased`?  Why not pattern match on all the eras instead?

cardano-api/gen/Gen/Cardano/Api/Typed.hs
```diff
@@ -522,22 +522,36 @@ genTxBodyContent era = do
...
+    , txExtraKeyWits
+    , txProtocolParams
+    , txWithdrawals
+    , txCertificates
+    , txUpdateProposal
+    , txMintValue
+    , txScriptValidity
     }
 
+fixDatumHashCollisions :: [TxOut CtxTx era] -> [TxOut CtxTx era]
```
Jimbo4350, line 541: Was this causing test failures?

cardano-api/test/Test/Cardano/Api/TxBody.hs
```diff
@@ -85,9 +125,10 @@ normalizeOriginal content =
     , txWithdrawals = normalizeWithdrawals $ txWithdrawals content
     }
 
--- | Normalizations applied to roundtrip result data only
-normalizeRoundtrip :: TxBodyContent ViewTx era -> TxBodyContent ViewTx era
-normalizeRoundtrip content@TxBodyContent{txAuxScripts, txIns, txInsCollateral} =
+-- | Normalizations applied to roundtrip result data
+normalizeContentRoundtrip ::
+  TxBodyContent ViewTx era -> TxBodyContent ViewTx era
```
Jimbo4350, line 136: Why this indent here?

cardano-api/test/Test/Cardano/Api/TxBody.hs
```diff
@@ -284,3 +372,30 @@ buildMintValue = \case
 
 tests :: TestTree
 tests = $testGroupGenerator
+
+
+-- | hack to work around https://github.com/input-output-hk/cardano-ledger-specs/pull/2529
+newtype Show' a = Show' a
+  deriving newtype Eq
+
+instance Show a => Show (Show' a) where
```
Jimbo4350, line 388: I don't understand. Why do we need this?

Jimbo4350 (changes requested):
Still some more comments to be addressed.

cardano-api/gen/Gen/Cardano/Api/Typed.hs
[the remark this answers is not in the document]
cblp: In eras after Alonzo, all these parameters must be set at all protocol updates, too?

cardano-api/gen/Gen/Cardano/Api/Typed.hs
cblp: added docs

cardano-api/test/Test/Cardano/Api/TxBody.hs
cblp: This was needed for debugging. Removed.

cardano-api/test/Test/Cardano/Api/TxBody.hs
cblp: This doesn't fit in 80 columns.

cardano-api/gen/Gen/Cardano/Api/Typed.hs
cblp: I have rewritten this. Please take a look.

cardano-api/test/Test/Cardano/Api/TxBody.hs
Jimbo4350: The max is 100 columns

Jimbo4350 (changes requested):
There is still a lot of unnecessary diff in this PR. I recommend staging all of your changes and re-commiting them.

cardano-api/gen/Gen/Cardano/Api/Typed.hs
```diff
@@ -560,6 +561,16 @@ genTxBody era = do
...
     Right txBody -> pure txBody
 
+genTxBody' ::
+  IsCardanoEra era =>
+  CardanoEra era ->
+  Gen (BuildTxWith BuildTx (Maybe ProtocolParameters), TxBody era)
+genTxBody' era = do
+  content <- genTxBodyContent era
+  case makeTransactionBody content of
+    Left err -> fail (displayError err)
```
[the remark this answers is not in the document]
cblp: MonadTest is not available in this context. Only MonadFail.

cardano-api/gen/Gen/Cardano/Api/Typed.hs
```diff
@@ -767,6 +771,18 @@ genExecutionUnits :: Gen ExecutionUnits
...
+-- | Gen for Alonzo-specific parameters
+alonzoParam :: CardanoEra era -> Gen a -> Gen (Maybe a)
+alonzoParam era gen
+  | anyCardanoEra era >= anyCardanoEra AlonzoEra = Just <$> gen
+  | otherwise = pure Nothing
+
+-- | 'Gen.maybe' but with condition if era is not Alonzo-based
+untilAlonzo :: CardanoEra era -> Gen a -> Gen (Maybe a)
+untilAlonzo era gen
+  | anyCardanoEra era >= anyCardanoEra AlonzoEra = pure Nothing
```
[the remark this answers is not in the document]
cblp: sure!

cardano-api/gen/Gen/Cardano/Api/Typed.hs
```diff
@@ -560,6 +561,16 @@ genTxBody era = do
     Left err -> fail (displayError err)
     Right txBody -> pure txBody
 
+genTxBody' ::
+  IsCardanoEra era =>
+  CardanoEra era ->
+  Gen (BuildTxWith BuildTx (Maybe ProtocolParameters), TxBody era)
```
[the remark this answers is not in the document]
cblp: done

cardano-api/test/Test/Cardano/Api/TxBody.hs
```diff
@@ -0,0 +1,395 @@
...
+
+      -- make another TxBody -- roundtrip
+      body2 <- evalEither $ makeTransactionBody content2
+
+      assertEqBodies body1 body2
+
+  | AnyCardanoEra era <- [minBound..]
+  ]
+
+
```
Jimbo4350, line 94: A comment like: "The auxiliary data hashes are different (and therefore must be ignored) due to the representational differences between the ledger's tx body and the api's txbody" would be useful

Jimbo4350 (approved):
LGTM! Just two comments to fix.

cardano-api/test/Test/Cardano/Api/TxBody.hs
cblp: Extended docs a bit.

cardano-api/gen/Gen/Cardano/Api/Typed.hs
```diff
@@ -536,6 +540,33 @@ genTxBodyContent era = do
...
     }
 
+
+-- | Ensure that all script data with the same hash are
+-- either all presented as values or all presented as hashes.
+--
+-- It's possible to have a hash without its datum, and also with its datum,
+-- and these two possibilities are semantically equivalent.
+fixDatumHashCollisions :: forall era. [TxOut CtxTx era] -> [TxOut CtxTx era]
+fixDatumHashCollisions outs = map replaceOutHashWithItsDatum outs where
```
Jimbo4350, line 551: the `where` clause needs to be on a newline and indented

cardano-api/test/Test/Cardano/Api/TxBody.hs
Jimbo4350: If you do indent a type signature then it is indented as follows:
```
normalizeContentRoundtrip 
  :: TxBodyContent ViewTx era -> TxBodyContent ViewTx era
```

cardano-api/test/Test/Cardano/Api/TxBody.hs
```diff
@@ -0,0 +1,395 @@
...
+  (ScriptInEra slang1 script1)
+  (ScriptInEra slang2 script2) =
+    case testEquality lang1 lang2 of
+      Nothing -> compare (AnyScriptLanguage lang1) (AnyScriptLanguage lang2)
+      Just Refl -> normalizeScriptOrder script1 script2
+    where
+      lang1 = languageOfScriptLanguageInEra slang1
+      lang2 = languageOfScriptLanguageInEra slang2
+
+normalizeScriptOrder :: Script lang -> Script lang -> Ordering
```
Jimbo4350, line 200: Why not implement an `Ordering` instance for `Script lang`? This is hacky.

Jimbo4350 (changes requested):

cardano-api/test/Test/Cardano/Api/TxBody.hs
cblp: I don't think this instance is natural enough to expose it in the API. It is used here only to compare script sets, not to order them.

cblp: So this is why it is named "normalize order", not "sort".

I'll add it to its doc.

cblp:
@dcoutts @Jimbo4350 please update your review

Jimbo4350 (changes requested):
There are `cardano-api` test failures and also `roundtrip tx CBOR` doesn't appear to be terminating.

carbolymer (approved):

cardano-api/test/Test/Cardano/Api/TxBody.hs
```diff
@@ -0,0 +1,377 @@
...
+  let content2 = buildBodyContent txProtocolParams content1
+
+  -- make another TxBody -- roundtrip
+  body2 <- evalEither $ createAndValidateTransactionBody content2
+
+  assertEqBodies body1 body2
+
+
+assertEqBodies :: TxBody era -> TxBody era -> PropertyT IO ()
+assertEqBodies _x _y = True === False -- TODO
```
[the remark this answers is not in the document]
carbolymer: There's `Eq` instance: https://github.com/input-output-hk/cardano-node/blob/4c99b0cf55a97b7413379c8960bd4fbd57df83b8/cardano-api/src/Cardano/Api/TxBody.hs#L1902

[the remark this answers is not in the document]
newhoggy: That worked!  Thanks!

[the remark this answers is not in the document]
newhoggy: It works inso far it compiles.  Looking at the code, the assertion functions want to compare for equality in a way that ignores script ordering.  Is that what the equality gives me?

PR #2756, opened 2021-05-28
CAD-2907 Tracing: SendFetchRequest

deepfire:
This adds a more detailed rendering of the `SendFetchRequest` trace.

### DIFF
--- a/cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
+++ b/cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
@@ -31,6 +31,7 @@ import           Ouroboros.Consensus.Block (ConvertRawHash (..), getHeader)
 import           Ouroboros.Consensus.Ledger.SupportsMempool (GenTx, HasTxs (..), txForgetValidated,
                    txId)
 import           Ouroboros.Consensus.Node.Run (RunNode, estimateBlockSize)
+import qualified Ouroboros.Network.AnchoredFragment as AF
 import           Ouroboros.Network.Block
 import           Ouroboros.Network.BlockFetch.ClientState (TraceFetchClientState,
                    TraceLabelPeer (..))
@@ -43,6 +44,7 @@ import qualified Ouroboros.Network.NodeToClient as NtC
 import           Ouroboros.Network.NodeToNode (ErrorPolicyTrace (..), TraceSendRecv (..),
                    WithAddr (..))
 import qualified Ouroboros.Network.NodeToNode as NtN
+import           Ouroboros.Network.Point (fromWithOrigin)
 import           Ouroboros.Network.Protocol.BlockFetch.Type (BlockFetch, Message (..))
 import           Ouroboros.Network.Protocol.ChainSync.Type (ChainSync)
 import qualified Ouroboros.Network.Protocol.ChainSync.Type as ChainSync
@@ -713,14 +715,26 @@ instance ToObject SlotNo where
     mkObject [ "kind" .= String "SlotNo"
              , "slot" .= toJSON (unSlotNo slot) ]
 
-
-instance ConvertRawHash header => ToObject (TraceFetchClientState header) where
+instance (HasHeader header, ConvertRawHash header)
+  => ToObject (TraceFetchClientState header) where
   toObject _verb BlockFetch.AddedFetchRequest {} =
     mkObject [ "kind" .= String "AddedFetchRequest" ]
-  toObject _verb BlockFetch.SendFetchRequest {} =
-    mkObject [ "kind" .= String "SendFetchRequest" ]
   toObject _verb BlockFetch.AcknowledgedFetchRequest {} =
     mkObject [ "kind" .= String "AcknowledgedFetchRequest" ]
+  toObject _verb (BlockFetch.SendFetchRequest af) =
+    mkObject [ "kind" .= String "SendFetchRequest"
+             , "head" .= String (renderChainHash
+                                  (renderHeaderHash (Proxy @header))
+                                  (AF.headHash af))
+             , "length" .= toJSON (fragmentLength af)]
+   where
+     fragmentLength :: AF.AnchoredFragment header -> Int
+     fragmentLength f = fromIntegral blockLength
+      where
+        headNo = unBlockNo (fromWithOrigin (BlockNo 0) (AF.headBlockNo f))
+        blockLength = case unBlockNo . blockNo <$> AF.last f of
+                        Left _  -> headNo
+                        Right b -> headNo - b + 1
   toObject _verb (BlockFetch.CompletedBlockFetch pt _ _ _ _) =
     mkObject [ "kind"  .= String "CompletedBlockFetch"
              , "block" .= String
--- a/cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
+++ b/cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
@@ -32,6 +32,7 @@ import           Ouroboros.Consensus.Ledger.SupportsMempool (GenTx, HasTxs (..),
                    txId)
 import           Ouroboros.Consensus.Node.Run (RunNode, estimateBlockSize)
 import qualified Ouroboros.Network.AnchoredFragment as AF
+import qualified Ouroboros.Network.AnchoredSeq as AS
 import           Ouroboros.Network.Block
 import           Ouroboros.Network.BlockFetch.ClientState (TraceFetchClientState,
                    TraceLabelPeer (..))
@@ -44,7 +45,6 @@ import qualified Ouroboros.Network.NodeToClient as NtC
 import           Ouroboros.Network.NodeToNode (ErrorPolicyTrace (..), TraceSendRecv (..),
                    WithAddr (..))
 import qualified Ouroboros.Network.NodeToNode as NtN
-import           Ouroboros.Network.Point (fromWithOrigin)
 import           Ouroboros.Network.Protocol.BlockFetch.Type (BlockFetch, Message (..))
 import           Ouroboros.Network.Protocol.ChainSync.Type (ChainSync)
 import qualified Ouroboros.Network.Protocol.ChainSync.Type as ChainSync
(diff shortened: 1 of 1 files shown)


cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
```diff
@@ -712,12 +709,16 @@ instance ToObject SlotNo where
     mkObject [ "kind" .= String "SlotNo"
              , "slot" .= toJSON (unSlotNo slot) ]
 
-
-instance ConvertRawHash header => ToObject (TraceFetchClientState header) where
+instance (HasHeader header, ConvertRawHash header)
+ => ToObject (TraceFetchClientState header) where
```
coot, line 713: Please don't mix refactoring and essential changes; this makes reviews / rebases harder.

cardano-node/src/Cardano/Tracing/Util.hs
```diff
@@ -0,0 +1,37 @@
...
+  where
+    -- Block numbers start at 1. We ignore the genesis EBB, which has block number 0.
+    blockN = unBlockNo $ fromWithOrigin (BlockNo 1) (AF.headBlockNo af)
+    firstBlock = case unBlockNo . blockNo <$> AF.last af of
+      -- Empty fragment, no blocks. We have that @blocks = 1 - 1 = 0@
+      Left _  -> 1
+      -- The oldest block is the genesis EBB with block number 0,
+      -- don't let it contribute to the number of blocks
+      Right 0 -> 1
+      Right b -> b
```
coot, line 37: if you use:
```
blockN = unBlockNo $ fromWithOrigin (BlockNo 0) (AF.headBlockNo af)
```
then it will simplify the calculation, you will not need a special case for `Right 0`, but you will need to add `1` which is expected: when one wants to get the number of integers `x` such that `low <= x <= high` then its `high - low + 1`.

I would renamed `blockN` as `headNo` and `firstBlock` as `lastNo`.

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
```diff
@@ -712,12 +709,16 @@ instance ToObject SlotNo where
...
   toObject _verb BlockFetch.AddedFetchRequest {} =
     mkObject [ "kind" .= String "AddedFetchRequest" ]
   toObject _verb BlockFetch.AcknowledgedFetchRequest {} =
     mkObject [ "kind" .= String "AcknowledgedFetchRequest" ]
+  toObject _verb (BlockFetch.SendFetchRequest af) =
+    mkObject [ "kind" .= String "SendFetchRequest"
+             , "head"   .= String (renderChainHash (renderHeaderHash (Proxy @header)) $ AF.headHash af)
```
coot, line 720: Could you try to fit this into 80 columns?

coot (changes requested):

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
deepfire: It was either that, or code duplication.

cardano-node/src/Cardano/Tracing/Util.hs
deepfire: This is a change in semantics for this computation, which is not something I was prepared to do.

cardano-node/src/Cardano/Tracing/Util.hs
```diff
@@ -0,0 +1,37 @@
...
+{-# LANGUAGE GeneralisedNewtypeDeriving #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+module Cardano.Tracing.Util
```
Jimbo4350, line 16: If we're creating a new module for utilities, lets move them all over e.g `chainInformation`, `fragmentChainDensity` etc

cardano-node/src/Cardano/Tracing/Util.hs
```diff
@@ -0,0 +1,37 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE DisambiguateRecordFields #-}
+{-# LANGUAGE EmptyCase #-}
```
Jimbo4350, line 4: I suspect a lot of these pragmas are unnecessary e.g `EmptyCase`

Jimbo4350 (changes requested):
Just a couple of minor changes. Looks good otherwise :smile:

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
Jimbo4350: There are also some unnecessary spaces  here: `, "head"   .= String `

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
deepfire: @coot, addressed this.

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
deepfire: Fixed.

cardano-node/src/Cardano/Tracing/Util.hs
deepfire: Fixed.

cardano-node/src/Cardano/Tracing/Util.hs
deepfire: Fixed.

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
coot: Thanks

cardano-node/src/Cardano/Tracing/Util.hs
coot: I see, you don't want to count the ebb, what about this:
```
headNo = unblockNo (fromOrigin (BlockNo 0) (AF.headBlockNo af)
blockLength = case unBlockNo . blockNo <$> AF.last af of
  Left _  -> headNo
  Right 0 -> headNo
  Right b -> b - headNo + 1
```
This is equivalent to your formula, isn't it?

deepfire: I don't think it's equivalent -- the minimum value produced by the existing code is 1, whereas in the above varient the minimum is 0.

I also think this discussion is what I wanted to avoid -- I just aimed to get the same logic we already have, with minimum friction.

coot: > I don't think it's equivalent -- the minimum value produced by the existing code is 1, whereas in the above variant the minimum is 0.
It can also be `0` because `blockNo` can be `1`. All I did I subtracted `1` from both `blockN` and `firstBlock` (which does not change the result) and inlined the subtraction, though I made a mistake it should be `headNo - b + 1` in the last case, it should be:

```
headNo = unblockNo (fromOrigin (BlockNo 0) (AF.headBlockNo af)
blockLength = case unBlockNo . blockNo <$> AF.last af of
  Left _  -> headNo
  Right 0 -> headNo -- do not count genesis EBB
  Right b -> headNo - b + 1
```

I think this is much simpler than the original.

> I also think this discussion is what I wanted to avoid -- I just aimed to get the same logic we already have, with minimum friction.
I wasn't aware of that, the diff does not show that.   If you don't want to include this in this PR, could you create another PR for that?

deepfire: I've rewritten the PR to only include minimum changes and no refactoring now.

coot: > I've rewritten the PR to only include minimum changes and no refactoring now. I also don't introduce semantic changes in the calculation.

I don't suggest making semantic changes, just making this cleaner whether in this PR, which makes sense, since we are adding this to `OrphanedInstances/Network`; or another one.

deepfire: @coot, I don't entirely understand your version of the code, which is why I was saying "semantic changes".

I trust you're correct, so I now took your version.

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
```diff
@@ -713,14 +715,27 @@ instance ToObject SlotNo where
     mkObject [ "kind" .= String "SlotNo"
              , "slot" .= toJSON (unSlotNo slot) ]
 
-
-instance ConvertRawHash header => ToObject (TraceFetchClientState header) where
+instance (HasHeader header, ConvertRawHash header)
+ => ToObject (TraceFetchClientState header) where
```
Jimbo4350, line 719: This should be 2 spaces

Jimbo4350 (approved):
LGTM! Please make the formatting change before merging.

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
deepfire: Done.

cardano-node/src/Cardano/Tracing/Util.hs
deepfire: Ok, so after further discussion with Marcin and Nicholas Frisby, it became clear that the question is clear as mud.. even if we ignore the EBB (which is a bit unfaithful in general, but is an acceptable simplification if 1) we only care about tracing and 2) in Shelley+).

deepfire: As a summary -- the semantic for the intended length should be:

- by how many blocks will be the chain elongated, as a result of adoption of the particular AF.

..or alternatively (equivalently), "how many non-anchor blocks in the AF".

Nicolas suggested going for the `AnchoredSeq` API to answer this properly.

coot (approved):
LGTM

nfrisby (approved):
The `fragmentLength` logic looks good to me, given the Slack discussion we had about what the `length` member's semantics. Glad to see GHC's somewhat-recent improvements to the pattern checker are kicking in here; Serge sees no warnings on that two-alternative case expression 🙌 

Approving.

PR #2757, opened 2021-05-28
Network tracers should be owned by the networking team

coot:


CODEOWNERS
```diff
@@ -7,8 +7,8 @@ cardano-cli/README.md                                         @fsancheziohk @nah
 cabal.project                                                 @LaurenceIO
 
 # General reviewers per PR
-#                       David          Duncan   Jordan     Erik   John
-*                       @DavidEichmann @dcoutts @Jimbo4350 @erikd @newhoggy
+#                       David          Duncan   Jordan     Erik   John      Marcin Karl
+*                       @DavidEichmann @dcoutts @Jimbo4350 @erikd @newhoggy @coot  @karknu
```
Jimbo4350, line 11: I'm happy to give you merge permissions on `cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs` but if something needs to be merged urgently otherwise you can ping one of the codeowners

Jimbo4350 (changes requested):

CODEOWNERS
coot: My intentions here isn't about merging things quickly; I'd prefer if the node team also reviews our PRs (the more eyes on it the better), but rather making sure that we are notified and asked to provide a review when other teams are making changes.

coot: @Jimbo4350 ping

Jimbo4350: Sorry, we've have problems in the past with people merging things in the node without node teams approval and we want to enforce this.

coot: > You can subscribe to the node to be notified about PRs and you don't need this specific change to be automatically asked to review a PR that makes changes to cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs

This does not work, because the updates from `cardano-node` are overwhelming. But we can relay on your judgment to require a review from us if things are related to the networking (tracers, network configuration options, the topology file, setting options in example config files - this went bad a few times!).

PR #2763, opened 2021-05-30
Fix broken links

kdchaires:
While reading the guide in docs.cardano.org, I noticed these links were broken.


newhoggy:
Fixed in https://github.com/input-output-hk/cardano-node/pull/2860.

Thanks!

PR #2765, opened 2021-05-31
Round trip JSON test on AlonzoGenesis

newhoggy:


cardano-api/test/Gen/Genesis.hs
```diff
@@ -0,0 +1,79 @@
...
+import qualified Cardano.Ledger.Alonzo.Language as Alonzo
+import qualified Cardano.Ledger.Alonzo.Scripts as Alonzo
+import qualified Cardano.Ledger.Coin as Ledger
+import qualified Hedgehog.Gen as Gen
+import qualified Hedgehog.Range as Range
+
+coin :: MonadGen m => Range Integer -> m Ledger.Coin
+coin r = do
+  unCoin' <- Gen.integral r
+  return $ Ledger.Coin unCoin'
```
newhoggy, line 23: I think we should follow `hedgehog` style for naming generators:

* Generators should always go in a `Gen` module
* Don't use the `gen` prefix
* Always import generators `qualified as Gen`
* Always use `Gen.` to prefix at use sites.

cardano-api/cardano-api.cabal
```diff
@@ -171,9 +172,11 @@ test-suite cardano-api-test
                       , text
                       , time
 
-  other-modules:        Test.Cardano.Api.Crypto
+  other-modules:        Gen.Genesis
```
Jimbo4350, line 175: Let's put these generators in the existing `Gen` module.

cardano-api/test/Gen/Genesis.hs
Jimbo4350: I don't see what's wrong with the existing style we use and I don't think it's worth it to change it all for the sake of following the `hedgehog` style.

cardano-api/cardano-api.cabal
newhoggy: Moved to `Test.Cardano.Api.Gen`

cardano-api/test/Gen/Genesis.hs
newhoggy: I've reverted to the `gen` prefix style.

You're right that it's not worth changing it for now.  I think in future we should consider that generators ought to live in the same project that declares the type and they should be packaged in a library so it can be re-used.

If we go there, then I would prefer to use the standard style.

cardano-api/test/Test/Cardano/Api/Gen.hs
```diff
@@ -5,38 +5,98 @@
...
+  let mDatumList = int ++ bytes ++ str
+
+  singleMetadatum <- Gen.element mDatumList
+
+  Gen.choice
+    [ return $ List mDatumList
+    , return $ Map [(singleMetadatum, singleMetadatum)]
+    , return $ Map [(List mDatumList, singleMetadatum)]
+    , return $ Map [(singleMetadatum, List mDatumList)]
+    ]
```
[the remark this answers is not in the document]
newhoggy: Fixed.  Thanks!

newhoggy:
bors merge

newhoggy:
bors merge

PR #2766, opened 2021-05-31
Fix nix update scripts to work without using a system-wide nixFlake

jbgi:


rvl (approved):
Documentation is good, thanks @jbgi.

PR #2768, opened 2021-06-01
Bring back Windows and macOS builds on Hydra

rvl:
The Windows and macOS builds were disabled in #2657 so that at least something could be merged.

- [x] Revert commit 8c0e2e9a152c3e240ce5957618e2ca88ae8aab0b.
- [ ] Fix failed build of `plutus-core:library:plutus-core` when cross-compiling to windows.

See [Hydra jobset for PR #2768](https://hydra.iohk.io/jobset/Cardano/cardano-node-pr-2768#tabs-jobs)

cc: @Jimbo4350 @disassembler @jbgi

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,7 +2,7 @@
   "nodes": {
     "customConfig": {
       "locked": {
-        "narHash": "sha256-t+2eKBUm6IZfWfgPUBcTEP64FJorFcY30NgmqAjLrjI=",
+        "narHash": "sha256-NdBgwn6xJOM5j+T4ViTq+7FToInEbzvKcdOBpjQvILw=",
         "path": "./custom-config",
         "type": "path"
       },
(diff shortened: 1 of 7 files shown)


michaelpj:
We have no cross-compilation builds for the plutus repo so things are probably broken there. Moritz is working on that atm.

hamishmack:
I think everything builds now except `linux.native.nixosTests.cardanoNodeEdge.x86_64-linux.x86_64-linux`.

I updated to `ghc8105` in the hope it would avoid ghc recompiles (since that is the version haskell.nix uses).  It actually did not (I think there must be an overlay to nixpkgs that prevents sharing with the `haskell.nix` CI builds).  Also because `iohk-nix-utils` still uses `ghc8104` it still built the native version of that anyway.

rvl:
@hamishmack, it's possible to set the ghc version for `iohk-nix-utils`. I recently added a `mkProject` function for that. Here's an example: https://github.com/input-output-hk/cardano-wallet/blob/bd95b3a470bf754a380eeffbe8e920ead003aa65/nix/build-tools-overlay.nix#L81-L90

jbgi (approved):

PR #2769, opened 2021-06-01
Aeson instances

newhoggy:


newhoggy:
I am getting `null` in my result though:

```
CARDANO_NODE_SOCKET_PATH=example/node-bft1/node.sock cardano-cli query pool-params --stake-pool-id 5560380b51e1717a751739a0be8f00a7a91bbcb2e8e209dc51b5becb --testnet-magic 42 --cardano-mode
{
    "poolParams": null,
    "futurePoolParams": null,
    "retiring": null
}
```

newhoggy:
The `null` happens because the values here are all `Nothing`: https://github.com/input-output-hk/cardano-node/blob/75e2ef21134615dcc88d471ae9b7d11d801b25e2/cardano-cli/src/Cardano/CLI/Types.hs#L120

newhoggy:
Currently depends on this https://github.com/input-output-hk/cardano-ledger-specs/pull/2308

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -796,6 +802,7 @@ obtainLedgerEraClassConstraints
 obtainLedgerEraClassConstraints ShelleyBasedEraShelley f = f
 obtainLedgerEraClassConstraints ShelleyBasedEraAllegra f = f
 obtainLedgerEraClassConstraints ShelleyBasedEraMary    f = f
-obtainLedgerEraClassConstraints ShelleyBasedEraAlonzo  _ =
-  panic "obtainLedgerEraClassConstraints: Alonzo era not implemented yet"
+obtainLedgerEraClassConstraints ShelleyBasedEraAlonzo  f = f
 
+deriving instance ToJSON (Alonzo.TxOut (Alonzo.AlonzoEra StandardCrypto))
+deriving newtype instance ToJSON (Shelley.CompactAddr StandardCrypto)
```
newhoggy, line 808: These to lines are required by line `805` above.

PR #2770, opened 2021-06-01
CAD-2998: trace-forward and trace-forward-demo.

denisshevchenko:


trace-dispatcher/src/Cardano/Logging/Tracer/Forward.hs
```diff
@@ -12,20 +15,20 @@ module Cardano.Logging.Tracer.Forward
...
                      writeTBQueue)
 import           Control.Exception (SomeException, try)
-import           Control.Monad (forever)
 import           Control.Monad.IO.Class
 import           Control.Monad.STM (atomically)
-import           Control.Tracer (contramap, stdoutTracer)
+import           GHC.Generics (Generic)
+
+-- Temporary solution, to avoid conflicts with trace-dispatcher.
+import "contra-tracer" Control.Tracer (contramap, stdoutTracer)
```
jutaro, line 27: Why is this necessary? Can't trace dispatcher use the arrow based lib? Do we need (yet another) tracer here anyway?

jutaro (approved):
I generally think all is fine. That the interface to trace dispatcher uses the old contra-tracer is the only part I would like to discuss about. Do we need a tracer based interface at all at this place?

PR #2772, opened 2021-06-01
Add comment about local development

weebl2000:


README.rst
```diff
@@ -408,6 +408,8 @@ Development
 
 run *ghcid* with: ``ghcid -c "cabal repl exe:cardano-node --reorder-goals"``
 
+Note: It might be required to set ``ghc-options: -Wwarn`` in `cabal.project` when developing on modules. Otherwise GHC complains about packages specified that are not required.
```
dcoutts, line 411: It'd be better to recommend setting the existing `development` flags in the `cabal.project` file, e.g.
```
package shelley-spec-ledger
  flags: +development
```
This is what the top level `cardano-haskell` `cabal.project` file does for example.

https://github.com/input-output-hk/cardano-haskell/

weebl2000: I will test if this works and change the comment.

weebl2000: image

I don't think it's sufficient.

weebl2000: @dcoutts I've amended the readme to be more specific and added the ``development`` flag as you suggested.

weebl2000:
@Jimbo4350 Sorry for tag, but not sure if Duncan is still involved in the project. Small comment about developing locally that might help people. Suggestions have been processed.

weebl2000:
@coot Sorry for shameless tagging. But this was basically approved. Can we merge it?

olgahryniuk (approved):

coot:
I prefer the other way around, remove `-Werror` and only use it in CI and for those who want to have it in their dev cycle, set `-Werror` in `cabal.project.local` file.

newhoggy (approved):

PR #2773, opened 2021-06-01
If only one of IPv4 / IPv6 is configured, try the other protocol silently

### LINKED ISSUE #2732: [BUG] - IPv6 addresses in mainnet-toplogy.json do not work
**Internal/External**
*External*

**Area**
*Other*  

**Summary**
If one is to place an IPv6 address in Producers section of topology.json, this does not work. 

**Steps to reproduce**
1. Configure your topology.json that you start cardano-node with
2. Add IPv6 address like so:

```
{
  "Producers": [
    {
      "addr": "2001:984:5a5:1:aaaa:aada:f00d:cafe",
      "port": 3001,
      "valency": 1
    }
  ]
}
```

3. Start cardano-node
4. See error in logs

```
cardano.node.IpSubscription:Error:621
Unsupported remote target address [2001:984:5a5:1:aaaa:aada:f00d:cafe]:3001
```

You can see that it parses it as an IPv6 address, hence the [] block tags. But still it does not connect. This machine is fully IPv6 connected.

**Expected behavior**
I expect IPv6 to work, future of the internet.

- OS: Ubuntu
- Version 20.04
- Node version 1.27.0

- Seems clear enough

**Additional context**

telnet works fine:

```
$ telnet -6 2001:984:5a5:1:aaaa:aada:f00d:cafe 3001
Trying 2001:984:5a5:1:aaaa:aada:f00d:cafe...
Connected to 2001:984:5a5:1:aaaa:aada:f00d:cafe.
Escape character is '^]'.
```

weebl2000:
Do not fail if it is missing, unlike when both IPv4 and IPv6 are missing.

should fix #2732

I am not sure if this is the best way to fix it, but currently the problem is that even if a host is IPv6-enabled but only has an IPv4 host address specified when starting the node, it cannot connect to IPv6 addresses. When the P2P protcol is going live, making sure both IPv4 and IPv6 work is even more important I reckon.

In the current situation IOHK relay DNS names do not return AAAA records. Even if AAAA records would be added now, not a single relay that has **only** an IPv4 host address specified will be able to connect to any of the IPv6 addresses, even though they might be fully IPv6 capable. Unless DNS entries are handled differently that is.

If neither IPv4 or IPv6 is specified, node can connect to IPv6 just fine.

### DIFF
--- a/cardano-node/src/Cardano/Node/Configuration/Socket.hs
+++ b/cardano-node/src/Cardano/Node/Configuration/Socket.hs
@@ -169,6 +169,21 @@ gatherConfiguredSockets NodeConfiguration { ncNodeIPv4Addr,
 
               pure (ipv4', ipv6')
 
+            -- When IPv4 host is specified, try IPv6, but do not fail if missing.
+            -- Very relevant if operator is using RFC4941 on their block producing-node.
+            (_, Nothing) -> do
+              info <- nodeAddressInfo Nothing ncNodePortNumber
+              let ipv6' = SocketInfo <$> find ((== AF_INET6) . addrFamily) info
+              
+              pure (ipv4, ipv6')
+
+            -- When IPv6 host is specified, try IPv4, but do not fail if missing.
+            (Nothing, _) -> do
+              info <- nodeAddressInfo Nothing ncNodePortNumber
+              let ipv4' = SocketInfo <$> find ((== AF_INET)  . addrFamily) info
+              
+              pure (ipv4', ipv6)
+
             _ -> pure (ipv4, ipv6)
 
 


weebl2000:
@coot I heard you were involved with networking side of things. I was just wondering if this is something that is desirable behavior or not. This is my first PR to cardano-node, I'm not sure if there is some other process I should go through in the development process. If so, please let me know. :)

coot (changes requested):
This looks good, thanks for your contribution; Just a minor change request.

coot:
However this removes the ability of using `ipv4` only on a `ipv6` capable host so I am not sure if this change is desirable.  If you want to use `ipv6` you can be explicit about it.

coot:
After a discussion we decided against this change because it limits the way one can configure a node.  It turns out that there many ipv6 capable hosts which are connected to ipv6 networks that are not connected to the internet.

weebl2000:
> After a discussion we decided against this change because it limits the way one can configure a node. It turns out that there many ipv6 capable hosts which are connected to ipv6 networks that are not connected to the internet.

I agree you should be able to configure without IPv6, but currently I feel a lot of operator will not know they are limiting IPv6 operation if they only specify IPv4 host address. Would it be a good addition to do add this behavior if an IPv6 address is specified in the topology file? @coot

weebl2000:
> After a discussion we decided against this change because it limits the way one can configure a node. It turns out that there many ipv6 capable hosts which are connected to ipv6 networks that are not connected to the internet.

I think the primary concern would be IPv6 hosts that *are* connected to the internet that do not want to connect to IPv6 cardan nodes.

PR #2775, opened 2021-06-02
Implement the necessary aeson instances for pool-params query for Alonzo

newhoggy:


cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -796,6 +802,11 @@ obtainLedgerEraClassConstraints
...
-obtainLedgerEraClassConstraints ShelleyBasedEraAlonzo  _ =
-  panic "obtainLedgerEraClassConstraints: Alonzo era not implemented yet"
-
+obtainLedgerEraClassConstraints ShelleyBasedEraAlonzo  f = f
+
+instance ToJSON (Alonzo.TxOut (Alonzo.AlonzoEra StandardCrypto)) where
+  toJSON (Alonzo.TxOut addr value dataHash) = Aeson.object
+    [ "addr" .= toJSON addr
+    , "value" .= toJSON value
+    , "dataHash" .= toJSON dataHash
```
newhoggy, line 811: `TxOut` field names.

cardano-api/src/Cardano/Api/Orphans.hs
```diff
@@ -369,3 +372,65 @@ instance ToJSON Alonzo.AlonzoGenesis where
       , "collateralPercentage" .= Alonzo.collateralPercentage v
       , "maxCollateralInputs" .= Alonzo.maxCollateralInputs v
       ]
+
+instance ToJSON (Alonzo.PParams era) where
```
erikd, line 376: I really hate orphan instances, but at least these are in an `Orphans` module.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
erikd: Sorry, what is that comment about?

erikd (approved):
LGTM!

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: I added to comment so I can make a permalink to share with the ledger team.

newhoggy:
bors merge

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
Jimbo4350: We already have this instance in `Cardano.Tracing.OrphanInstances.Shelley`. What we should do is move the instance in `Cardano.Tracing.OrphanInstances.Shelley` to `Cardano.Api.Orphans` and import it instead.

cardano-api/src/Cardano/Api/Orphans.hs
```diff
@@ -369,4 +375,76 @@ instance ToJSON Alonzo.AlonzoGenesis where
...
+        <*> obj .: "prices"
+        <*> obj .: "maxTxExUnits"
+        <*> obj .: "maxBlockExUnits"
+        <*> obj .: "maxValSize"
+        <*> obj .: "collateralPercentage"
+        <*> obj .: "maxCollateralInputs"
+
+deriving instance ToJSON (Alonzo.PParamsUpdate (Alonzo.AlonzoEra StandardCrypto))
+
+instance (Ledger.Era era, Show (Ledger.Value era), ToJSON (Ledger.Value era))
```
Jimbo4350, line 440: :100:

Jimbo4350 (approved):
LGTM!

newhoggy:
bors merge

PR #2612, opened 2021-04-16
Calculate minimum ADA value

newhoggy:

PR #2779, opened 2021-06-02
Wire up calculate-min-fee and calculate-min-value

Jimbo4350:


cardano-api/src/Cardano/Api/ProtocolParameters.hs
```diff
@@ -1323,3 +1331,36 @@ fromAlonzoPParams
...
+-- mininimum UTxO value derived from the 'ProtocolParameters'
+calcMinimumDeposit :: ProtocolParameters -> Value -> Lovelace
+calcMinimumDeposit pp v  =
+  case protocolParamMinUTxOValue pp of
+    Nothing ->
+      case protocolParamUTxOCostPerWord pp of
+        Nothing ->
+          error "calcMinimumDeposit: uTxOCostPerWord and minUTxOValue protocol parameters were not found"
+        Just (Lovelace cPerWord) ->
+          -- TODO: Alonzo - ledger should expose a function to calculate the required deposit
```
newhoggy, line 1344: TODO item

cardano-api/src/Cardano/Api/ProtocolParameters.hs
```diff
@@ -1323,3 +1331,36 @@ fromAlonzoPParams
...
+          Lovelace $ max adaOnlyUTxOSize (utxoEntrySizeWithoutVal + Val.size (toMaryValue v)) * cPerWord
+    Just minUTxOValue ->
+      fromShelleyLovelace $ Shelley.scaledMinDeposit (toMaryValue v) (toShelleyLovelace minUTxOValue)
+
+ where
+    -- TODO: Copied from ledger. Ledger should expose a function for this
```
newhoggy, line 1350: TODO item for discussion

JaredCorduan: this is a part of `utxoEntrySize` mentioned above

cardano-api/src/Cardano/Api/ProtocolParameters.hs
JaredCorduan: The size is calculated here:

https://github.com/input-output-hk/cardano-ledger-specs/blob/ecb7275b2de624cede629fec2c4e9237639425e3/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs#L104

and is multiplied by `adaPerUTxOWord` here:

https://github.com/input-output-hk/cardano-ledger-specs/blob/ecb7275b2de624cede629fec2c4e9237639425e3/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs#L394

Jimbo4350: @JaredCorduan That calculates the size of the tx and then the cost (via multiplication by `adaPerUTxOWord`).  We want to only calculate the minimum deposit for a multi asset value.

JaredCorduan: just let me know, we could certainly add something for you. did you want a function that takes a value and a maybe hash? or are you good now?

Jimbo4350:
Subsumed by #2612

PR #2780, opened 2021-06-03
Debugging slow compile times

newhoggy:


newhoggy:
https://gitlab.haskell.org/ghc/ghc/-/issues/20036

PR #2792, opened 2021-06-06
Use the 8.10.2 version of GHC

saikatdas0790:
Since that is the version being installed and setup from line 85

Otherwise the cabal build does not work from line 153


newhoggy:
Closing because we should be using `ghc-8.10.4`.  Documentation will be updated accordingly.

PR #2802, opened 2021-06-08
Provide high quality multiarch docker image

tdiesler:
We now have the first part of the planed docker changes available in master (see #2796). Kudos to @jbgi for that.
This PR is about the remainder that we agreed on so far ...

There is a good chance that we will indeed have a much improved docker image with Alonzo. :-)


tdiesler:
We decided to postpone the script based topology update for now and revisit if Alonzo p2p should get delayed for some reason

PR #2805, opened 2021-06-09
Public generator library

newhoggy:
Previously a `cardano-api-test` and `cardano-api:test` were both sharing modules in the `cardano-api/test` directory structure which lent some confusion as to where test modules were supposed to go.  The `cardano-api-test` module exported generators that were used by other projects in the same repository and existed for that purpose.

This PR replaces that with a new sub-library `cardano-api:gen` inside `cardano-api.cabal`.  ("Public sub-library" is a `cabal-3.x` feature)

In doing so we can define all our generators in one place have ensure they are re-usable throughout all the `cardano-node` packages.  `cardano-api:gen` also serves as a place where we can place common code aside from generators.  For example test setup code.

All the generators are located under `Gen.*` where `*` is the module where the type was defined.  For example `Cardano.Api.TxMetadata` defines `TxMetadataJsonSchema` so the generator for `TxMetadataJsonSchema` lives in `Gen.Cardano.Api.TxMetadata`.

This replaces early scheme where sometimes generators are found in a special `Gen` submodule and sometimes they are mixed in with property tests.  The change means generators are found in a predictable/consistent place.

After this I can add new generators without feeling conflicted about where I should put them.


newhoggy:
NOTE: `haskell.nix` does not support public sub-libraries so I've added a commit to switch to use `cardano-api-gen` and `cardano-api-test` projects to achieve the same thing.  The hope is once `haskell.nix` gets this support we can revert to using public sub-libraries.

Issue #2801, opened 2021-06-08
[FR] - Allow docker container configuration through env vars


tdiesler:
**Internal/External**
*External*  

**Area**
*Other*   Docker 

**Describe the feature you'd like**

The standard way to configure docker containers is through env vars (rather than through cmd line args)
This allows higher order orchestration frameworks like docker compose, k8s to take advantage of a unified configuration mechanism.

**Additional context / screenshots**

For example the existing cardano-db-sync compose script would benefit from this

tdiesler:
Merged

PR #2812, opened 2021-06-10
[#2801] Allow docker container configuration through env vars

tdiesler:


jbgi (approved):

jbgi:
Thanks a lot! (and for bearing with all the requested changes!)

tdiesler:
You're welcome. We could now start thinking about doc updates [like so](https://hub.docker.com/r/nessusio/cardano-node) and slowly start thinking about multiarch support, which is also [already available here](https://github.com/tdiesler/nessus-cardano).

PR #2813, opened 2021-06-10
CAD 2907: Alonzo support in the workbench

deepfire:
1. Add Alonzo profiles into the workbench and
2. Fix a small bug that prevented profile names from being properly passed through

Test with:

```make cls cluster-shell CLUSTER_PROFILE=default-alzo```


denisshevchenko (approved):

MarcFontaine (approved):
Looks good to me

PR #2831, opened 2021-06-15
CAD-2907 workbench: Alonzo support

deepfire:
This makes the `scripts/plutus/example-txin-locking-plutus-script.sh` script work with the workbench:

1. Alonzo-specific updates in the workbench genesis generation
2. Adaptations in the script
3. Small quality-of-life improvement: set `CARDANO_NODE_SOCKET_PATH` automatically inside the shell

Test with:

1. `make cls cluster-shell-dev CLUSTER_ARGS_EXTRA='--arg withHoogle false'`, and once inside the shell:
2. `./scripts/plutus/example-txin-locking-plutus-script.sh`
3. `cardano-cli transaction submit --tx-file example/work/alonzo.tx --testnet-magic 42`

The last command should output: `Transaction successfully submitted.`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ profiles:
 profile-names:
 	@./nix/workbench/wb profile-names
 
-CLUSTER_PROFILE     = default-mary
+CLUSTER_PROFILE     = default-alzo
 CLUSTER_ARGS_EXTRA ?=
 
 cluster-shell:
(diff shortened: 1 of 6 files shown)


MarcFontaine:
The modified script will no longer work in the original context.
I.e. it will not work with the second cluster config.
It would be better to rename the new script and keep the old version.

MarcFontaine (approved):
looks good to me

PR #2836, opened 2021-06-17
New script to sync hlint rules

newhoggy:


newhoggy:
This introduces a new script `scripts/reconfigure-hlint.sh`.

When run, this script will find extract all the `HLINT` rules in the project's `*.hs` files and insert them into `.hlint.yaml` in between the sections marked `# BEGIN-GENERATED` and `# END-GENERATED`.  Any existing content will be replaced.

This script is needed because the currently Haskell Language Server is unable to parse `HLINT` annotations in Haskell modules and can only read rules in the `.hlint.yaml` file.  This is despite `HLINT` annotations being the _recommended_ way to describe hlint rules.

.hlint.yaml
```diff
@@ -53,7 +53,7 @@
 # Ignore some builtin hints
```
erikd, line 53: How often is this file likely to change? Its basiclay (mostly?) autogenerated,.If it changes all the time the it maybe should not be stored in git.

newhoggy: I've renamed `.hlint.yaml` to `.hlint.template.yaml` and added `.hlint.yaml` to `.gitignore`.  The script will now generate a `.hlint.yaml` using `.hlint.template.yaml` as the template.

README.rst
```diff
@@ -406,8 +406,20 @@ Byron vote submission:
...
 
+Haskell Language Server
+-----------------------
+
+When using Haskell Langague Server with Visual Studio Code, you may find that
+`HLINT annotations are ignored<https://github.com/haskell/haskell-language-server/issues/638>`.
+
+To work around this, you may run the script `./scripts/reconfigure-hlint.sh` to generate a `.hlint.yaml`
+file with HLINT ignore rules derived from the source code.
+
```
newhoggy, line 422: Added some documentation.

erikd (approved):

newhoggy:
bors merge

PR #2850, opened 2021-06-22
One cabal.project file

newhoggy:


dcoutts (changes requested):
We cannot have any dependencies on `plutus-tx-plugin` or we will not be able to build on Windows.

PR #2853, opened 2021-06-22
Add plutus support to tx-generator

MarcFontaine:
* Add proof-of-concept Plutus example.
* Various cleanups.
* Switch to V7 network protocol.
* Support TxAdditionSize (only non-Plutus case).


deepfire (approved):
LGTM!

PR #2862, opened 2021-06-24
p2p-master cleanup

bolt12:
All commits compile
Simplified a lot of commits and squashed/fixup them together


coot:
This branch needs to be updated.

PR #2863, opened 2021-06-24
Improve cli help for script data options.

james-iohk:
Clarify what is supported when using script data `file` or `value` options.

I decided it is worth listing the data types because bool and null are not supported.


dcoutts (approved):

james-iohk:
bors merge

PR #2870, opened 2021-06-27
Add comment about IPv4/IPv6 connectivity

weebl2000:
Since https://github.com/input-output-hk/cardano-node/pull/2773 was rejected.

### DIFF
--- a/doc/stake-pool-operations/getConfigFiles_AND_Connect.md
+++ b/doc/stake-pool-operations/getConfigFiles_AND_Connect.md
@@ -44,6 +44,7 @@ Get the complete list of available options with `cardano-node run --help`
   	                                specified slot
     -h,--help                       Show this help text
 
+**Note**: If you do not specify `--host-addr` nor `--host-ipv6-addr`, node will use the default IPv4 and IPv6 addresses (depending which addresses are available).  If one specifies one of them only that address will be used, in particular if you only provide an IPv4 address, the node will not connect over IPv6.
 To start a passive node:
 
      cardano-node run \
--- a/doc/stake-pool-operations/core_relay.md
+++ b/doc/stake-pool-operations/core_relay.md
@@ -44,3 +44,7 @@ Make your __relay node__ `talk` to your __block-producing__ node and __other rel
         }
       ]
     }
+
+**Note**: If you want to connect to IPv4 and IPv6 relays, you must either not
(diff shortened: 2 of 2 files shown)


weebl2000:
image

As an addition to this, explicitly state the behaviour for the host address parameters.

weebl2000:
@coot @dcoutts Small addition to documentation explicitly stating IPv4/IPv6 connectivity behaviour. This might save people some time.

doc/stake-pool-operations/getConfigFiles_AND_Connect.md
```diff
@@ -44,6 +44,9 @@ Get the complete list of available options with `cardano-node run --help`
   	                                specified slot
     -h,--help                       Show this help text
 
+**Note**: If you specify a host address for IPv4, but do not specify one for IPv6,
+your node will **NOT** be able to connect over IPv6, and vice versa.
+
```
coot, line 49:
```suggestion
**Note**: If you don't specify `--host-addr` nor `--host-ipv6-addr`, node will use the default ipv4 and ipv6 addresses (depending which addresses are available).  If one specifies one of them only that address will be used, in particular if you only provide an IPv4 address, the node will not connect over IPv6.
```

I think it's better to be a little bit more descriptive and phrase it to suggest why the api works the way it does. It's not about not being able to connect over ipv6 but rather that you can precisely express which addresses the node can use.

coot (changes requested):
Thanks for your contribution; a suggestion with a different phrasing is given.

doc/stake-pool-operations/getConfigFiles_AND_Connect.md
weebl2000: I've committed your suggestion. It still leaves open the question of specifying an IPv4 host address but relying on IPv6 privacy extension (RFC4941), where ideally you would not want to specify an IPv6 address, since they can change all the time.

coot (approved):
LGTM

coot:
bors merge

weebl2000:
@coot Do I have to fix anything or is the CI broken?

coot:
let's retry, but could you first squash all the commits.

weebl2000:
> let's retry, but could you first squash all the commits.

Squashed.

coot:
bors merge

PR #2874, opened 2021-06-28
Plutus spending script docs

Jimbo4350:


doc/reference/plutus/plutus-spending-script-example.md
```diff
@@ -0,0 +1,119 @@
...
+
+Follow the instructions displayed in the terminal to start your Alonzo cluster.
+
+#### Sending ADA to the script address
+
+In order to require a Plutus script to validate the spending of a tx ouput, we must put the tx output at the script address of the said Plutus script. However, before we do that, we must create a datum hash:
+
+```bash
+> cardano-cli transaction hash-script-data --script-data-value 42
+> 9e1199a988ba72ffd6e9c269cadb3b53b5f360ff99f112d9b2ee30c4d74ad88b
```
newhoggy, line 60: I'm wondering if we're already using `42` for testnet magic, if we should use another number for datum hash?

newhoggy (approved):

doc/reference/plutus/plutus-spending-script-example.md
```diff
@@ -0,0 +1,119 @@
...
+#### Creating the `AlwaysSucceeds` Plutus spending script
+
+The plutus-example executable will automagically generate several Plutus scripts in the CLI-compatiable text envelope format.
+
+Run the following commands:
+
+```bash
+cd plutus-example
+
+cabal run exe:plutus-example
```
MarcFontaine, line 35: I miserably failed to reproduce this step.
The catch seems to be that this is not a nix-based build.
Do you have any idea how to build the Plutus examples on NixOS ?

Jimbo4350: Sorry I don't use nix in my workflow. You'll need to bug somebody within IOHK that does.

doc/reference/plutus/plutus-spending-script-example.md
Jimbo4350: We can but I don't think its an issue.

PR #2880, opened 2021-06-29
Merge query tip queries into the same connection

newhoggy:


cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -255,27 +265,59 @@ runQueryTip (AnyConsensusModeParams cModeParams) network mOutFile = do
...
+      , localStateQueryClient   = Just (singleQuery waitResult mpoint resultVarEraHistoryAndSystemStart)
       , localTxSubmissionClient = Nothing
       }
-    atomically (takeTMVar resultVar)
+
+    atomically waitResult
   where
+    chainSyncGetCurrentTip
+      :: forall mode a
+      .  STM a
```
coot, line 292: You can ignore this suggestion:
```suggestion
      :: forall mode
      .  (forall a. STM a)
```
I think this will encode that you don't need the value at all.

newhoggy: I like this suggestion.  Thanks!

newhoggy: This is odd.  The suggested change doesn't compile:

```

src/Cardano/CLI/Shelley/Run/Query.hs:283:81: error:
    • Couldn't match type ‘a’
                     with ‘(ChainTip,
                            Either AcquireFailure (EraHistory CardanoMode, SystemStart))’
      ‘a’ is a rigid type variable bound by
        a type expected by the context:
          forall a. STM a
        at src/Cardano/CLI/Shelley/Run/Query.hs:283:81-90
      Expected type: STM a
        Actual type: STM
                       (ChainTip,
                        Either AcquireFailure (EraHistory CardanoMode, SystemStart))
    • In the first argument of ‘chainSyncGetCurrentTip’, namely
        ‘waitResult’
      In the second argument of ‘($)’, namely
        ‘chainSyncGetCurrentTip waitResult resultVarChainTip’
      In the ‘localChainSyncClient’ field of a record
    |
283 |       { localChainSyncClient    = LocalChainSyncClient $ chainSyncGetCurrentTip waitResult resultVarChainTip
```

newhoggy: I'll keep the code as it is for now.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -201,20 +203,28 @@ runQueryTip
...
-        case eResult of
-          Left acqFail -> left (ShelleyQueryCmdAcquireFailure acqFail)
+        (chainTip, result) <- liftIO $ queryChainTipAndEraHistoryAndSystemStart localNodeConnInfo Nothing
+
+        let tipSlotNo = case chainTip of
+              ChainTipAtGenesis -> 0
+              ChainTip slotNo _ _ -> slotNo
+
+        case result of
+          Left _ -> do
```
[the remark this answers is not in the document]
newhoggy: Fixed!  The error is rendered into the JSON output.

erikd (approved):

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -201,20 +203,34 @@ runQueryTip
...
-          Left acqFail -> left (ShelleyQueryCmdAcquireFailure acqFail)
+        (chainTip, result) <- liftIO $ queryChainTipAndEraHistoryAndSystemStart localNodeConnInfo Nothing
+
+        let tipSlotNo = case chainTip of
+              ChainTipAtGenesis -> 0
+              ChainTip slotNo _ _ -> slotNo
+
+        case result of
+          Left acqFail -> do
+            let output = encodePretty $ O.QueryTipOutput
```
Jimbo4350, line 220: Do we really need to qualify this? `Cardano.CLI.Shelley.Output` lives in the same library.

cardano-cli/src/Cardano/CLI/Shelley/Output.hs
```diff
@@ -32,7 +34,8 @@ instance ToJSON QueryTipOutput where
         , "block" .= bNum
         , "era" .= era a
         , "epoch" .= epoch a
-        , "syncProgress" .= either id toJSON (syncProgress a)
+        , "syncProgress" .= syncProgress a
+        , "errors" .= if L.null (errors a) then Nothing else Just (errors a)
```
Jimbo4350, line 38: Why not:
```        
, "errors" .= if L.null (errors a) then Null else toJSON (errors a)
```

cardano-cli/src/Cardano/CLI/Shelley/Output.hs
```diff
@@ -43,5 +46,6 @@ instance ToJSON QueryTipOutput where
...
```
Jimbo4350, line 50: Why not:
```        
, "errors" .= if L.null (errors a) then Null else toJSON (errors a)
```

Jimbo4350 (changes requested):
Looking good. A couple of comments.

PR #2884, opened 2021-06-29
Update retire_stakepool.md

CarlosLopezDeLara:
Laying out the process for retiring a stake pool. 

If the stake key is deregistered before the pool is retired or before the pool deposit is paid back, the deposit goes to treasury.

### DIFF
--- a/doc/stake-pool-operations/retire_stakepool.md
+++ b/doc/stake-pool-operations/retire_stakepool.md
@@ -2,8 +2,19 @@
 
 To retire a pool we need to:
 
-* Create a **deregistration certificate** and
-* Submit the certificate to the blockchain with a **transaction**
+0. Make sure that any pool owners are aware of your retirement intention and will not remove their Pledge until the pool has retired. 
+1. Create a **deregistration certificate** and specify an epoch when your pool will retire. There is a limit on how far in the future this can be (given by a protocol parameter - see below). Make sure you give sufficient time for delegators to switch pools
+2. Submit the certificate to the blockchain with a **transaction**
+3. Notify your delegators of your intention to retire the pool using your usual communication mechanisms. Wallets and other tools should highlight that the pool is retiring and discourage new delegations.
+4. Keep your system working normally until we reach the epoch you chose on the certificate for the pool to expire. Retirement takes place when the first block of the selected epoch is processed (before the transactions for that block are processed.) 
+5. Move your funds out from your pledge address to a different wallet.  
+6. When the pool is effectively retired, the pool deposit is sent back to the rewards account.
+7. **Wait for any pending rewards to be paid.**
+8. Withdraw any rewards and pool deposit from the rewards account.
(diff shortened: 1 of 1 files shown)


kevinhammond (dismissed):
0. Make sure that any pool owners are aware of your retirement intention and will not remove their Pledge until the pool has retired.

1. Create a deregistration certificate and specify an epoch when your pool will retire.  There is a limit on how far in the future this can be (given by a protocol parameter - see below).  Make sure you give sufficient time for delegators to switch pools

2.  Submit ...

3.  Notify your delegators of your intention to retire the pool using your usual communication mechanisms.  Wallets and other tools should highlight that the pool is retiring and discourage new delegation/encourage new delegations.

4. Keep your system working...

Reach the epoch => reach the end of/start of the epoch?  Check this, please.

5. When the pool has retired move ..., and advise other owners that they may remove their own pledge.  Note that if you are setting up a new pool, you will need to have registered it in advance and timed the pledge move so that both pools are correctly pledged

I think the pledge can be removed after your step 3 so move this step here so people don't lock pledge longer than they need to.

send back => sent back

5. **Wait for any pending rewards and deposit to be paid into the rewards account**

**Bold this!**

7.  **Once you have successfullly withdrawn your rewards and deposit, de-register the stake key**

**Bold this too!**

doc/stake-pool-operations/retire_stakepool.md
```diff
@@ -2,8 +2,17 @@
 
 To retire a pool we need to:
 
-* Create a **deregistration certificate** and
-* Submit the certificate to the blockchain with a **transaction**
+1. Create a **deregistration certificate**
```
JaredCorduan, line 5: I would call this a retirement certificate.

JaredCorduan:
> Reach the epoch => reach the end of/start of the epoch? Check this, please.

If the pool is set to retire in epoch `e`, then the the retirement happens when the first block of epoch `e` is processed (and before the transactions in the block).

JaredCorduan (approved):
Looks good to me!

doc/stake-pool-operations/retire_stakepool.md
CarlosLopezDeLara: Makes sense to me, however this is how it is called on the CLI, using a different name might lead to confusion.  

```
cardano-cli stake-pool
Usage: cardano-cli stake-pool (registration-certificate | 
                                deregistration-certificate | id | metadata-hash)
  Stake pool commands

Available options:

Available commands:
  registration-certificate Create a stake pool registration certificate
  deregistration-certificate
                           Create a stake pool deregistration certificate
  id                       Build pool id from the offline key
  metadata-hash            Print the hash of pool metadata.
```

JaredCorduan: ah, consistency with the CLI does seem better!

CarlosLopezDeLara:
@kevinhammond  Can you take a look to the changes.

Jimbo4350 (approved):

Issue #2896, opened 2021-07-05
[BUG] - Docker run fails with  invalid/wrong Shelley genesis file


tdiesler:
*External*  

**Summary**

The basic docker run command ...

```
docker run --rm -it \
  -p 3001:3001 \
  -v node-data:/opt/cardano/data \
  inputoutput/cardano-node:dev run
```

fails with ...

```
Running the cardano node ...
CARDANO_BIND_ADDR=0.0.0.0
CARDANO_BLOCK_PRODUCER=false
CARDANO_CONFIG=/opt/cardano/config/mainnet-config.json
CARDANO_DATABASE_PATH=/opt/cardano/data
CARDANO_LOG_DIR=/opt/cardano/logs
CARDANO_PORT=3001
CARDANO_SOCKET_PATH=/opt/cardano/ipc/socket
CARDANO_TOPOLOGY=/opt/cardano/config/mainnet-topology.json
cardano-node run --config /opt/cardano/config/mainnet-config.json --topology /opt/cardano/config/mainnet-topology.json --database-path /opt/cardano/data --socket-path /opt/cardano/ipc/socket --host-addr 0.0.0.0 --port 3001
Wrong Shelley genesis file: the actual hash is "02846121d1b8b8ecf3ce9739d46352c2b504148b4f93f1937dea66b6f0f681c7", but the expected Shelley genesis hash given in the node configuration file is "06cc024b823b6d20f5dde2faf8de2d895f47983ab584db38ea62111b61038e35"
```

There are two issues here AFAICS.

1. The expected hash is the AlonzoGenesisHash in the config
2. There may be a problem with mainnet-alonzo-genesis instead of mainnet-shelley-genesis

**Additional context**

I'd take a look at this

tdiesler:
The error message refers to "Shelley genesis file", but the hash mismatch is with the Alonzo genesis file. I fixed the mainnet-alozo-genisis.json and changed the error message to read "Wrong Shelley/Alonzo genesis ..."

I'm not sure if this is right. Perhaps Shelley.hs should never have been called with the Alonzo genesis hash.

tdiesler:
There is a confusion in the codebase. 

[...]

PR #2897, opened 2021-07-05
[#2896]  Docker run fails with invalid/wrong Shelley genesis file

tdiesler:


tdiesler:
The error message refers to "Shelley genesis file", but the hash mismatch is with the Alonzo genesis file. I fixed the mainnet-alozo-genisis.json and changed the error message to read "Wrong Shelley/Alonzo genesis ..."

Not sure if this is right. Perhaps Shelley.hs should never have been called with the Alonzo genesis hash.

tdiesler:
There is a confusion in the codebase. 

@dcoutts changed from [adaPerUTxOWord => lovelacePerUTxOWord](https://github.com/input-output-hk/cardano-node/commit/a6478b06f436a566ce9b195d283f46d58aa3368d#diff-22d9d289c003162bcf5cda338d6c3609828081713f6f6184b11e850f5b8f7678)
@deepfire changed from [lovelacePerUTxOWord => adaPerUTxOWord](https://github.com/input-output-hk/cardano-node/commit/39ca0acbcd7a5b9d324788701ee94242b4812301#diff-a95fc30e1c452160ff83acd7282a03acf9db021277cfff9edeb84a62810916e7)

Which one should it be? 

If its going to be `lovelacePerUTxOWord` the nix/workbench/genesis.sh thing would need to get changed and applied to the cache (probably).

DON'T MERGE. Going back to adaPerUTxOWord is probably wrong.

tdiesler:
Done

PR #2904, opened 2021-07-07
Update register_stakepool.md

soyfrien:
I think this can be avoided if the limitations are mentioned
```
cardano-cli stake-pool metadata-hash --pool-metadata-file poolMetadata.json
Command failed: stake-pool metadata-hash  Error: Error validating stake pool metadata: Stake pool metadata must consist of at most 512 bytes, but it consists of 556 bytes
```


Jimbo4350 (approved):
Thanks :+1:

PR #2905, opened 2021-07-07
Fixed chairman tests for P2P and NonP2P Modes

bolt12:
- Updated config file and topology file
- Fixed valency in Run.hs
- Added P2P Switch flag to chairman config files
- Added mkTopology function to switch accordingly as well
- Disable P2P Mode for Byron and Shelley tests
- Enable P2P Mode for ByronShelley tests
- General chairman-tests refactoring to support P2P Switch


cardano-node-chairman/src/Testnet/Byron.hs
```diff
@@ -70,22 +73,61 @@ defaultTestnetOptions = TestnetOptions
...
+  J.encode
+  $ J.object
+    [ ( "Producers"
+      , J.toJSON (flip fmap ([0 .. numBftNodes - 1] L.\\ [i]) (\j -> J.object
+        [ ("addr", "127.0.0.1")
+        , ("valency", J.toJSON @Int 1)
+        , ("port", J.toJSON (allPorts L.!! j))
+        ]
+                                                             )
+                 )
```
newhoggy, line 106: Please fix indentation here to be similar to what you've done below.

cardano-node-chairman/src/Testnet/Byron.hs
```diff
@@ -70,22 +73,59 @@ defaultTestnetOptions = TestnetOptions
...
+rewriteConfiguration True _ "EnableP2P: False"                 = "EnableP2P: True"
+rewriteConfiguration False _ "EnableP2P: True"                 = "EnableP2P: False"
+rewriteConfiguration _ _ s                                     = s
 
 rewriteParams :: TestnetOptions -> Value -> Value
 rewriteParams testnetOptions = rewriteObject
   $ HM.insert "slotDuration" (J.toJSON @String (show @Int (slotDuration testnetOptions)))
 
+mkTopologyConfig :: Int -> Int -> [Int] -> Bool -> ByteString
+mkTopologyConfig i numBftNodes allPorts False =
```
coot, line 96: @newhoggy I have a comment outside of this PR, I think writing `json` object in this way is a bit strange.  I think it would be nicer to have a `ToJSON` instance for topology files and use the data structure and turn it into json.

bolt12: The JSON instance for the Topology format already exists. I also found this style very weird, especially when there are template files that we can read, parse, manipulate and turn into json again, much more simpler!

newhoggy: If there's already an accessible type, then by all means 😁

coot: Do we have a tests that checks that `ToJSON` and `FromJSON` are compatible?

bolt12: Yes: https://github.com/input-output-hk/cardano-node/blob/p2p-master/cardano-node/test/Test/Cardano/Node/Json.hs#L51

cardano-node-chairman/src/Testnet/ByronShelley.hs
```diff
@@ -166,7 +215,7 @@ testnet testnetOptions H.Conf {..} = do
     AtEpoch n -> ["TestShelleyHardForkAtEpoch: " <> show @Int n]
 
   H.readFile (base </> "configuration/chairman/byron-shelley/configuration.yaml")
-    <&> L.unlines . (<> forkMethod) . L.lines
+    <&> L.unlines . fmap (rewriteConfiguration (enableP2P testnetOptions)) . (<> forkMethod) . L.lines
```
coot, line 218: This line is too long.

coot (approved):
LGTM

cardano-node-chairman/src/Testnet/ByronShelley.hs
bolt12: Will fix!

PR #2910, opened 2021-07-08
Submit api testing

newhoggy:
Integration test for `cardano-submit-api`.  This is run in CI.

### DIFF
--- a/cardano-testnet/src/Testnet/SubmitApi.hs
+++ b/cardano-testnet/src/Testnet/SubmitApi.hs
@@ -1,0 +1,111 @@
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Testnet.SubmitApi
+  ( submitApi
+  , submitApiSubmitTransaction
+  ) where
+
+import           Control.Monad
+import           Data.Aeson (FromJSON (..), (.:))
+import           Data.Eq
+import           Data.Function
+import           Data.Functor
+import           Data.Int
+import           Data.Maybe
+import           Data.Semigroup
+import           Data.Text (Text)
+import           GHC.Generics
+import           System.IO (FilePath)
+import           Text.Show
+
+import qualified Data.Aeson as J
+import qualified Data.ByteString.Base16 as Base16
+import qualified Data.ByteString.Lazy as LBS
+import qualified Data.Text as Text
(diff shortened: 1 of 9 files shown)


cardano-testnet/src/Testnet/SubmitApi.hs
```diff
@@ -0,0 +1,140 @@
...
+import           Text.Show
+
+import qualified Data.Aeson as J
+import qualified Data.ByteString.Base16 as Base16
+import qualified Data.ByteString as BSS
+import qualified Data.ByteString.Lazy as LBS
+import qualified Data.ByteString.Char8 as BS8
+import qualified Data.HashMap.Lazy as HM
+import qualified Data.List as L
+import qualified Data.Text as T
```
erikd, line 47: There are only 26 letters in the alphabet, so using single letters for qualified imports may mean we run out of letters. I would suggest `List` and `Text` for these.

cardano-testnet/src/Testnet/SubmitApi.hs
```diff
@@ -0,0 +1,140 @@
...
+import qualified Data.Text.Encoding as T
+import qualified Data.Time.Clock as DTC
```
erikd, line 49: And `Time` for this.

cardano-testnet/src/Testnet/SubmitApi.hs
```diff
@@ -0,0 +1,140 @@
...
+import qualified Hedgehog as H
+import qualified Hedgehog.Extras.Stock.IO.File as IO
+import qualified Hedgehog.Extras.Stock.IO.Network.Socket as IO
+import qualified Hedgehog.Extras.Stock.IO.Network.Sprocket as IO
+import qualified Hedgehog.Extras.Stock.String as S
+import qualified Hedgehog.Extras.Test.Base as H
+import qualified Hedgehog.Extras.Test.Concurrent as H
+import qualified Hedgehog.Extras.Test.File as H
+import qualified Hedgehog.Extras.Test.Network as H
+import qualified Hedgehog.Extras.Test.Process as H
```
erikd, line 60: Is there any name collision when importing all `Hedgehog.Extra` modules as `HExtra` ?

erikd (approved):
Not sure about how you are naming qualified imports, but otherwise :+1:

cardano-testnet/src/Testnet/SubmitApi.hs
newhoggy: Fixed!

cardano-testnet/src/Testnet/SubmitApi.hs
newhoggy: No know name collisions as yet.

newhoggy:
bors merge

PR #2925, opened 2021-07-14
CAD-3098: integrate generator into the workbench

deepfire:
This implements the initial transaction generation integration into the workbench.

The transaction generator is started once the local cluster comes online (there is no sophisticated run sense/control as in `cardano-ops` yet).

In addition, this also optionally provides the R Studio inside `nix-shell` (only when `withR` is `true`).

### DIFF
--- a/nix/nixos/tx-generator-service.nix
+++ b/nix/nixos/tx-generator-service.nix
@@ -13,6 +13,8 @@
       exeName = "tx-generator";
 
       extraOptionDecls = {
+        scriptMode      = boolOpt true       "Whether to use the modern script parametrisation mode of the generator.";
+
         ## TODO: the defaults should be externalised to a file.
         ##
         tx_count        =  intOpt 1000       "How many Txs to send, total.";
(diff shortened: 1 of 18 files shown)


bench/tx-generator/configuration-default.json
```diff
@@ -0,0 +1,72 @@
+{
```
MarcFontaine, line 1: `bench/tx-generator` is basically the root directory of the cabal package.
Lets just move that file to `bench/` or somewhere else.

nix/nixos/tx-generator-service.nix
```diff
@@ -0,0 +1,94 @@
+(import ../. {}).commonLib.defServiceModule
```
MarcFontaine, line 1: Wow that looks very polished.(maybe too much).
I guess this is more like a stub.
Basically it is adding yet an other layer to the configuration of the tx-generator.
The `tx-generator` itself will change and all the changes will propagate here.
OK but now we have it done anyway..

MarcFontaine: Many nix-services do indeed have an option to just pass a custom configuration file (for example JSON) as configuration.
Would that not makes things simpler ?

MarcFontaine (approved):
Could you please move `bench/tx-generator/configuration-default.json` to a better location,
before merging this PR.

nix/nixos/tx-generator-service.nix
deepfire: > Basically it is adding yet an other layer to the configuration of the tx-generator.

Are you familiar with NixOS service definitions?  This is one and is necessary for deployment.

deepfire: > Many nix-services do indeed have an option to just pass a custom configuration file (for example JSON) as configuration.
> Would that not makes things simpler ?

This is already implemented, indeed.

Jimbo4350 (approved):

PR #2931, opened 2021-07-16
Update docs for Alonzo

newhoggy:


Jimbo4350 (changes requested):
We also need to add Alonzo genesis related fields with brief explanations. `understanding-the-config-files.md` also needs to be updated appropriately.

Jimbo4350 (approved):

PR #2932, opened 2021-07-16
Merge query tip queries

newhoggy:
Merge query tip queries.  Check for version 8 compatibility.

### DIFF
--- a/cardano-api/cardano-api.cabal
+++ b/cardano-api/cardano-api.cabal
@@ -61,6 +61,7 @@ library
                         Cardano.Api.Hash
                         Cardano.Api.HasTypeProxy
                         Cardano.Api.IPC
+                        Cardano.Api.IPC.Monad
                         Cardano.Api.Key
                         Cardano.Api.KeysByron
                         Cardano.Api.KeysPraos
--- a/cardano-api/src/Cardano/Api/IPC/Monad.hs
+++ b/cardano-api/src/Cardano/Api/IPC/Monad.hs
@@ -1,0 +1,70 @@
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE GeneralisedNewtypeDeriving #-}
+
+module Cardano.Api.IPC.Monad
+  ( LocalStateQueryScript
+  , sendMsgQuery
+  , setupLocalStateQueryScript
+  ) where
+
+import Cardano.Api.Block
+import Cardano.Api.IPC
+import Control.Applicative
+import Control.Concurrent.STM
+import Control.Monad
+import Control.Monad.IO.Class
+import Control.Monad.Trans.Cont
+import Data.Either
+import Data.Function
+import Data.Maybe
+import Data.Ord
+import Shelley.Spec.Ledger.Scripts ()
+import System.IO
+
+import qualified Ouroboros.Network.Protocol.LocalStateQuery.Client as Net.Query
+import qualified Ouroboros.Network.Protocol.LocalStateQuery.Type as Net.Query
+
+-- | Monadic type for constructing local state queries.
+--
+-- Use 'sendMsgQuery' in a do block to construct queries of this type and convert
+-- the expression to a 'Net.Query.LocalStateQueryClient' with 'setupLocalStateQueryScript'.
+newtype LocalStateQueryScript block point query r m a = LocalStateQueryScript
+  { runLocalStateQueryScript :: ContT (Net.Query.ClientStAcquired block point query m r) m a
+  } deriving (Functor, Applicative, Monad, MonadIO)
+
+-- | Use 'sendMsgQuery' in a do block to construct monadic local state queries.
+sendMsgQuery :: Monad m => query a -> LocalStateQueryScript block point query r m a
+sendMsgQuery q = LocalStateQueryScript . ContT $ \f -> pure $
+  Net.Query.SendMsgQuery q $
+    Net.Query.ClientStQuerying
+    { Net.Query.recvMsgResult = f
+    }
+
+-- | Use 'sendMsgQuery' in a do block to construct monadic local state queries.
+setupLocalStateQueryScript ::
+     STM x
+  -> Maybe ChainPoint
+  -> NodeToClientVersion
+  -> TMVar (Maybe (Either Net.Query.AcquireFailure a))
+  -> LocalStateQueryScript (BlockInMode CardanoMode) ChainPoint (QueryInMode CardanoMode) () IO a
+  -> Net.Query.LocalStateQueryClient (BlockInMode CardanoMode) ChainPoint (QueryInMode CardanoMode) IO ()
+setupLocalStateQueryScript waitDone mPointVar' ntcVersion resultVar' f =
+  LocalStateQueryClient $
+    if ntcVersion >= NodeToClientV_8
+      then do
+        pure . Net.Query.SendMsgAcquire mPointVar' $
+          Net.Query.ClientStAcquiring
+          { Net.Query.recvMsgAcquired = runContT (runLocalStateQueryScript f) $ \result -> do
+              atomically $ putTMVar resultVar' (Just (Right result))
+              void $ atomically waitDone
+              pure $ Net.Query.SendMsgRelease $ pure $ Net.Query.SendMsgDone ()
+
+          , Net.Query.recvMsgFailure = \failure -> do
+              atomically $ putTMVar resultVar' (Just (Left failure))
+              void $ atomically waitDone
+              pure $ Net.Query.SendMsgDone ()
+          }
+      else do
+        atomically $ putTMVar resultVar' Nothing
+        void $ atomically waitDone
+        pure $ Net.Query.SendMsgDone ()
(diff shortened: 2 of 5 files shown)


newhoggy:
These are examples of possible outputs:

Version 9:

```
{
    "epoch": 0,
    "hash": "d2ce9ffa65e9a7a056e18e46be79c4f1390885ab5194afbe64abba7612850681",
    "slot": 914,
    "block": 64,
    "era": "Alonzo",
    "syncProgress": "100.00"
}
```

Version 8:

```
{
    "epoch": 0,
    "hash": "d2ce9ffa65e9a7a056e18e46be79c4f1390885ab5194afbe64abba7612850681",
    "slot": 914,
    "block": 64,
    "era": "Mary",
}
```

Version 7 and earlier:

```
{
    "hash": "899573caae74be7410fee52f1b65d46f82bd542796d3dca42b94031242dc44d8",
    "slot": 480,
    "block": 24
}
```

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -208,26 +210,35 @@ runQueryTip
...
   let localNodeConnInfo = LocalNodeConnectInfo cModeParams network sockPath
-      consensusMode = consensusModeOnly cModeParams
-
-  anyEra <- determineEra cModeParams localNodeConnInfo
-  tip <- liftIO $ getLocalChainTip localNodeConnInfo
-
-  let tipSlotNo = case tip of
-        ChainTipAtGenesis -> 0
-        ChainTip slotNo _ _ -> slotNo
+  let consensusMode = consensusModeOnly cModeParams
```
Jimbo4350, line 213: No need for two `let`s here

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -208,26 +210,35 @@ runQueryTip
...
-        case eResult of
-          Left acqFail -> left (ShelleyQueryCmdAcquireFailure acqFail)
-          Right (eraHistory, mSystemStart) ->
+        (chainTip, result) <- liftIO $ queryChainTipAndEraHistoryAndSystemStart localNodeConnInfo Nothing
+
+        let tipSlotNo = case chainTip of
+              ChainTipAtGenesis -> 0
+              ChainTip slotNo _ _ -> slotNo
+
+        case join (either (const Nothing) Just <$> result) of
```
Jimbo4350, line 223: What we should do to avoid the gymnastics here is use `ShelleyQueryCmdError` ( i.e `ShelleyQueryCmdAcquireFailure`) instead of the `Maybe` in `singleQuery`.  This will help simplify the code and we can deal with the case (more clearly imo) where the node to client version < 9.

Jimbo4350 (dismissed):
A couple thoughts.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: These cases are not failures though.  Does `ShelleyQueryCmdError` still make sense?

cardano-api/cardano-api.cabal
```diff
@@ -61,6 +61,7 @@ library
                         Cardano.Api.Hash
```
erikd, line 61: I would suggest that the commit subject line should be "Merge query tip queries" and the rest could go in the message body.

erikd (approved):
Please fix the commit message, but otherwise :+1:

cardano-api/cardano-api.cabal
newhoggy: Pushed the new commit message.

newhoggy:
bors merge

newhoggy:
bors merge

cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -0,0 +1,65 @@
...
+import Data.Function
+import Data.Maybe
+import Data.Ord
+import Shelley.Spec.Ledger.Scripts ()
+import System.IO
+
+import qualified Ouroboros.Network.Protocol.LocalStateQuery.Client as Net.Query
+import qualified Ouroboros.Network.Protocol.LocalStateQuery.Type as Net.Query
+
+newtype LocalStateQueryScript block point query r m a = LocalStateQueryScript
```
Jimbo4350, line 26: Can we add a comment here with a short description? We should indent at `via` as well.

cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -0,0 +1,65 @@
...
+
+sendMsgQuery :: Monad m => query a -> LocalStateQueryScript block point query r m a
+sendMsgQuery q = LocalStateQueryScript $ \f -> pure $
+  Net.Query.SendMsgQuery q $
+    Net.Query.ClientStQuerying
+    { Net.Query.recvMsgResult = f
+    }
+
+setupLocalStateQueryScript ::
+     STM x
```
Jimbo4350, line 40: What is `x`?

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -207,97 +199,116 @@ runQueryTip
...
-                      , O.syncProgress = mSyncProgress
-                      }
-
-                case mOutFile of
-                  Just (OutputFile fpath) -> liftIO $ LBS.writeFile fpath output
-                  Nothing                 -> liftIO $ LBS.putStrLn        output
     mode -> left (ShelleyQueryCmdUnsupportedMode (AnyConsensusMode mode))
 
-queryEraHistoryAndSystemStart
+queryQueryTip
```
Jimbo4350, line 253: Let's move this to the api

Jimbo4350 (changes requested):

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -207,97 +199,116 @@ runQueryTip
...
+              systemStart <- fmap getSystemStart (O.mSystemStart localState) & hoistMaybe ShelleyQueryCmdSystemStartUnavailable
+              nowSeconds <- toRelativeTime (SystemStart systemStart) <$> liftIO getCurrentTime
+              tipTimeResult <- getProgress tipSlotNo (O.eraHistory localState) & bimap ShelleyQueryCmdPastHorizon fst & except
+
+              let tolerance = RelativeTime (secondsToNominalDiffTime 600)
+
+              return $ flip (percentage tolerance) nowSeconds tipTimeResult
+
+            mSyncProgress <- hushM syncProgressResult $ \e -> do
+              liftIO . T.hPutStrLn IO.stderr $ "Warning: Sync progress unavailable: " <> renderShelleyQueryCmdError e
```
Jimbo4350, line 234: Same thing here, why the addition `hushM` vs rendering this in the `ToJSON` instance?

Jimbo4350:
Minor changes. We should revisit this later and have a think on how to make it more concise.

cardano-api/src/Cardano/Api/IPC/Monad.hs
newhoggy: The `x` is unknown.  I don't care about the type of the value.  See `void $ atomically waitDone` later in the function.  All I need is to know that the value exists, but I don't care what type it actually is.  The fact that I can read the value indicates everything is complete and I can exit.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: `getLocalChainTip` is still used elsewhere.

`queryQueryTip` is specific to `query tip`.  Maybe it should stay here?

cardano-api/src/Cardano/Api/IPC/Monad.hs
newhoggy: Added comments.

cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -0,0 +1,72 @@
...
+-- | Monadic type for constructing local state queries.
+--
+-- Use 'sendMsgQuery' in a do block to construct queries of this type and convert
+-- the expression to a 'Net.Query.LocalStateQueryClient' with 'setupLocalStateQueryScript'.
+newtype LocalStateQueryScript block point query r m a = LocalStateQueryScript
+  { runLocalStateQueryScript
+      :: (a -> m (Net.Query.ClientStAcquired block point query m r))
+      -> m (Net.Query.ClientStAcquired block point query m r)
+  } deriving (Functor, Applicative, Monad, MonadIO)
+      via ContT (Net.Query.ClientStAcquired block point query m r) m
```
newhoggy, line 35: The `Monad` instances are defined here.

newhoggy:
bors merge

newhoggy:
bors merge

newhoggy:
bors merge

PR #2943, opened 2021-07-19
Workbench: optional generator startup

deepfire:
Workbench changes:

1. add `--no-generator` to `start-cluster`, to skip the automatic `tx-generator` startup
2. add `--help` to `start-cluster`

### DIFF
--- a/nix/supervisord-cluster/default.nix
+++ b/nix/supervisord-cluster/default.nix
@@ -137,16 +137,19 @@ let
   start = pkgs.writeScriptBin "start-cluster" ''
     set -euo pipefail
 
-    workbench-prebuild-executables
-
     batch_name=${batchName}
+    run_start_flags=()
 
     while test $# -gt 0
     do case "$1" in
-        --batch-name ) batch_name=$2; shift;;
-        --trace | --debug ) set -x;;
+        --batch-name )                   batch_name=$2; shift;;
+        --no-generator | --no-gen )      run_start_flags+=($1);;
+
+        --trace | --debug )              set -x;;
         --trace-wb | --trace-workbench ) export WORKBENCH_EXTRA_FLAGS=--trace;;
         * ) break;; esac; shift; done
+
+    workbench-prebuild-executables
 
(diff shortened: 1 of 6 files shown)


nix/supervisord-cluster/default.nix
```diff
@@ -134,20 +134,41 @@ let
...
+  startClusterUsage = ''
+Usage:
+   start-cluster [FLAGS..]
+
+   Flags:
+
+      --batch-name NAME               Override the batch name (default: ${batchName})
+      --no-generator | --no-gen       Don't auto-start the tx-generator
+
+      --trace | --debug               Trace the start-cluster script
```
MarcFontaine, line 146: is there is difference between `--trace` and `--debug` ?

nix/workbench/run.sh
```diff
@@ -289,7 +298,12 @@ EOF
         )
         genesis prepare "${genesis_args[@]}"
 
-        backend start-run "$dir" "$@"
+        ## Record genesis.
+        cp "$dir"/genesis/genesis.json "$dir"/genesis.json
```
MarcFontaine, line 302: Why do you need to do that ?
Why is this `cp` at this point in the script ?
IMHO having the same config twice in two files is confusing.

nix/supervisord-cluster/default.nix
deepfire: No, none.

nix/workbench/run.sh
deepfire: It's necessary due to how genesis caching works.

nix/supervisord-cluster/default.nix
MarcFontaine: This confusing. There should not be two different flags that actually do the same thing.
Please choose one of the two.

MarcFontaine (approved):
Just merge it as it is

deepfire:
We should work in increasing familiarity with the workbench inside the team -- including its concepts and principles.

It's easy to misinterpret things otherwise.

PR #2946, opened 2021-07-20
Deploy Haddock for merge to master only

james-iohk:
The readme says "The documentation is built with each push, but is only published from master branch" yet it is configured to publish on every push. This changes it to publish on every push to master.


james-iohk:
Changed target branch with #3012. Closing.

PR #2950, opened 2021-07-20
Updating node's p2p-master branch

coot:
Accommodate changes introduced in https://github.com/input-output-hk/ouroboros-network/pull/3274

- Updated p2p integration
- Tracers: parametrise using addresses rather than connection ids.


bolt12 (approved):
LGTM!

PR #2960, opened 2021-07-25
CAD-3243: trace-forward library.

denisshevchenko:
This is `trace-forward` library: `typed-protocols`-based library allowing to forward tracing items from one process to another one. This is the first step toward a new tracing infrastructure (please see https://github.com/input-output-hk/cardano-node/pull/2847). The `trace-dispatcher` will use this library to forward `TraceObject`s to external acceptors (for example, `cardano-tracer`).

You can think of this library as a **_replacement_** for old plugins [trace-forwarder](https://github.com/input-output-hk/iohk-monitoring-framework/tree/master/plugins/backend-trace-forwarder) and [trace-acceptor](https://github.com/input-output-hk/iohk-monitoring-framework/tree/master/plugins/backend-trace-acceptor) from `iohk-monitoring-framework`.


deepfire (approved):
LGTM!

It's great to finally have things moving towards the final merge!

cabal.project
```diff
@@ -60,6 +61,9 @@ package cardano-node-chairman
 package cardano-testnet
   tests: True
 
+package trace-forward
+  tests: True
```
newhoggy, line 65: Is it necessary to enable tests?  We usually disable dependency tests.

trace-forward/src/Trace/Forward/Acceptor.hs
```diff
@@ -0,0 +1,35 @@
...
+import           Trace.Forward.Configuration (AcceptorConfiguration (..))
+import           Trace.Forward.Protocol.Type (NodeInfoStore)
+
+runTraceAcceptor
+  :: (CBOR.Serialise lo,
+      ShowProxy lo,
+      Typeable lo)
+  => AcceptorConfiguration lo -- ^ Acceptor configuration.
+  -> TBQueue lo               -- ^ The queue received 'TraceObject's will be written in.
+  -> NodeInfoStore            -- ^ The store node's basic info will be written in.
```
newhoggy, line 27: I'm a little confused by the English of the comments.

Is the basic info being written to trace?

trace-forward/src/Trace/Forward/Configuration.hs
```diff
@@ -0,0 +1,55 @@
...
+
+type Host = Text
+type Port = Word16
+
+-- | Specifies how to connect to the peer.
+data HowToConnect
+  = LocalPipe    !FilePath    -- ^ Local pipe (UNIX or Windows).
+  | RemoteSocket !Host !Port  -- ^ Remote socket (host and port).
+
+-- | Acceptor configuration, parametrized by trace item's type.
```
newhoggy, line 25: Typo.  Should be `parameterized`.

trace-forward/src/Trace/Forward/Configuration.hs
```diff
@@ -0,0 +1,55 @@
...
+data AcceptorConfiguration lo = AcceptorConfiguration
+  { -- | The tracer that will be used by the acceptor in its network layer.
+    acceptorTracer    :: !(Tracer IO (TraceSendRecv (TraceForward lo)))
+    -- | The endpoint that will be used to listen to the forwarder.
+  , forwarderEndpoint :: !HowToConnect
+    -- | The request specifies how many 'TraceObject's will be requested.
+  , whatToRequest     :: !Request
+    -- | Additional action that will be performed every time the acceptor will
+    -- receive the reply from the forwarder.
```
newhoggy, line 34: I think this should be "Additional action that will be performed every time the acceptor receives a reply from the forwarder"

trace-forward/src/Trace/Forward/Configuration.hs
```diff
@@ -0,0 +1,55 @@
...
+  , actionOnReply     :: !([lo] -> IO ())
+    -- | 'IORef' that can be used as a brake: if an external thread will set it to
+    -- 'True', the acceptor will send 'MsgDone' message to the forwarder and their
+    -- session will be closed.
```
newhoggy, line 38: I think this should read: "'IORef' that can be used as a brake: if an external thread sets it to 'True', the acceptor will send 'MsgDone' message to the forwarder and their session will be closed."

trace-forward/src/Trace/Forward/Configuration.hs
```diff
@@ -0,0 +1,55 @@
...
+  , shouldWeStop      :: !(IORef Bool)
+    -- | An action that will be performed before sending 'MsgDone' message.
+  , actionOnDone      :: !(IO ())
+  }
+
+-- | Forwarder configuration, parametrized by trace item's type.
```
newhoggy, line 44: ---> parameterized

trace-forward/src/Trace/Forward/Forwarder.hs
```diff
@@ -0,0 +1,34 @@
...
+
+import           Trace.Forward.Configuration (ForwarderConfiguration (..))
+import           Trace.Forward.Network.Forwarder (connectToAcceptor)
+
+runTraceForwarder
+  :: (CBOR.Serialise lo,
+      ShowProxy lo,
+      Typeable lo)
+  => ForwarderConfiguration lo -- ^ Forwarder configuration.
+  -> TBQueue lo                -- ^ The queue the forwarder will take 'TraceObject's from.
```
newhoggy, line 26: This is okay, but I think "The queue from which the forwarder will take 'TraceObject's" sounds a bit better.

cabal.project
deepfire: It's not an _external_ dependency, so here would be the only place where the tests are run.

denisshevchenko:
@newhoggy Thanks! Fixed.

cabal.project
newhoggy: Oh, yeah, I was also wondering about it being an internal dependency.  But if you've had that discussion that's fine. 😁

trace-forward/trace-forward.cabal
```diff
@@ -0,0 +1,99 @@
...
+                       -Widentities
+                       -Wincomplete-uni-patterns
+                       -Wincomplete-record-updates
+  if impl(ghc >= 8.0)
+    ghc-options:       -Wredundant-constraints
+  if impl(ghc >= 8.2)
+    ghc-options:       -fhide-source-paths
+  if impl(ghc >= 8.4)
+    ghc-options:       -Wmissing-export-lists
+                       -Wpartial-fields
```
newhoggy, line 29: There's a lot of IFs for versions of GHC that `cardano-node` doesn't support anymore?

What's the story here?

denisshevchenko: Well, I just copied this section from another Haskell project (to be precise, from `iohk-monitoring-framework`).

denisshevchenko: Should I remove it?

coot: Do we need to support `8.0`, `8.2` and `8.4`? On CI we only build with `8.10` right now.

I wouldn't use `-fhide-source-path`.

trace-forward/trace-forward.cabal
```diff
@@ -0,0 +1,99 @@
...
+                       , typed-protocols-examples
+                       , text
+                       , time
+
+  default-language:    Haskell2010
+  ghc-options:         -rtsopts
+                       -Wall
+                       -Wno-unticked-promoted-constructors
+                       -Wno-orphans
+                       -threaded
```
coot, line 99: Why not the other useful options as in the `common-options` library?

trace-forward/trace-forward.cabal
deepfire: @denisshevchenko, I would have used the `cardano-node` cabal definitions as baseline, instead of `iohk-monitoring-framework`.

denisshevchenko: Done.

trace-forward/trace-forward.cabal
denisshevchenko: Fixed.

trace-forward/README.md
```diff
@@ -0,0 +1,5 @@
+# trace-forward
+
+`trace-forward` is a library allowing to forward tracing items from one process to another one. It is built upon [`typed-protocols`](https://github.com/input-output-hk/ouroboros-network/tree/master/typed-protocols).
+
+The `trace-dispatcher` is using `trace-forward` to forward `TraceObject`s from the node to exernal acceptors (for example, `cardano-tracer`).
```
Jimbo4350, line 5: Can we add the devs who are responsible for this library here?

Jimbo4350 (changes requested):
One minor addition and then I'm happy to approve

trace-forward/README.md
denisshevchenko: Yes, we can. But should I add them here? Or it should be added in `CODEOWNERS` file?

denisshevchenko: Done.

denisshevchenko:
@Jimbo4350 Done.

trace-forward/README.md
Jimbo4350: You should add them here.

denisshevchenko: Ok.

Jimbo4350 (approved):
LGTM

deepfire:
bors merge

trace-forward/src/Trace/Forward/Configuration.hs
```diff
@@ -0,0 +1,55 @@
...
+data ForwarderConfiguration lo = ForwarderConfiguration
+  { -- | The tracer that will be used by the forwarder in its network layer.
+    forwarderTracer    :: !(Tracer IO (TraceSendRecv (TraceForward lo)))
+    -- | The endpoint that will be used to connect to the acceptor.
+  , acceptorEndpoint   :: !HowToConnect
+    -- | An action that returns node's basic information.
+  , nodeBasicInfo      :: !(IO NodeInfo)
+    -- | Additional action that will be performed every time the forwarder will
+    --   receive the request from the acceptor.
+  , actionOnRequest    :: !(Request -> IO ())
```
coot, line 54: How this is used? Is it just for testing purposes?  The same question about `actionDone` in `AcceptorConfiguration`.

trace-forward/test/Test/Trace/Forward/Demo/Configs.hs
```diff
@@ -0,0 +1,36 @@
...
+  :: HowToConnect
+  -> IORef Bool
+  -> AcceptorConfiguration TraceItem
+mkAcceptorConfig ep weAreDone = AcceptorConfiguration
+  { acceptorTracer    = nullTracer
+  , forwarderEndpoint = ep
+  , whatToRequest     = GetTraceObjects 10
+  , actionOnReply     = const $ pure ()
+  , shouldWeStop      = weAreDone
+  , actionOnDone      = putStrLn "Acceptor: we are done!"
```
coot, line 24: Using `putStrLn` will just pollute the command line (which is annoying for tests). It's better to use logging system for such messages.

trace-forward/src/Trace/Forward/Network/Forwarder.hs
```diff
@@ -0,0 +1,151 @@
...
+          (forwarderTracer config)
+          (Forwarder.codecTraceForward CBOR.encode CBOR.decode
+                                       CBOR.encode CBOR.decode
+                                       CBOR.encode CBOR.decode)
+          (byteLimitsTraceForward (fromIntegral . LBS.length))
+          timeLimitsTraceForward
+          channel
+          (Forwarder.traceForwarderPeer $ readItems config loQueue)
+      atomically $ putTMVar cv r
+      waitSibling siblingVar
```
coot, line 118: I think this just blocks forever.  The only place `siblingVar` is subtracted is in `waitSibling` which is called once for the given `siblingVar` created in the same monadic context.

What is the purpose of `siblingVar`?

trace-forward/src/Trace/Forward/Acceptor.hs
```diff
@@ -0,0 +1,35 @@
...
+import           Trace.Forward.Network.Acceptor (listenToForwarder)
+import           Trace.Forward.Configuration (AcceptorConfiguration (..))
+import           Trace.Forward.Protocol.Type (NodeInfoStore)
+
+runTraceAcceptor
+  :: (CBOR.Serialise lo,
+      ShowProxy lo,
+      Typeable lo)
+  => AcceptorConfiguration lo -- ^ Acceptor configuration.
+  -> TBQueue lo               -- ^ All 'TraceObject's received from the node will be written in this queue.
```
coot, line 26: This means a pull based protocol is translated into a push based application, though the boundedness of the queue might help a bit with back-pressure. Is this necessary tough?  Why cannot we process the data in the protocol handlers?

coot (changes requested):

trace-forward/src/Trace/Forward/Configuration.hs
denisshevchenko: Yes, for testing purposes only.

trace-forward/src/Trace/Forward/Network/Forwarder.hs
denisshevchenko: Honestly, I don't know. I just copied this snippet from `ouroboros-network-framework/test/Test/Ouroboros/Network/Subscription.hs`.

trace-forward/src/Trace/Forward/Protocol/Type.hs
```diff
@@ -0,0 +1,214 @@
...
+  show MsgRequest{}         = "MsgRequest"
+  show MsgReply{}           = "MsgReply"
+  show MsgDone{}            = "MsgDone"
+
+instance Show (ClientHasAgency (st :: TraceForward lo)) where
+  show TokIdle = "TokIdle"
+
+instance Show (ServerHasAgency (st :: TraceForward lo)) where
+  show TokNodeInfoBusy = "TokNodeInfoBusy"
+  show TokBusy{}       = "TokBusy"
```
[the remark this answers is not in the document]
denisshevchenko: I already provide `instance ShowProxy (TraceForward lo)`, line 87.

trace-forward/test/Test/Trace/Forward/Demo/Tests.hs
```diff
@@ -0,0 +1,101 @@
...
+          Local  -> LocalPipe <$> mkLocalPipePath
+
+  forwarderQueue <- newTBQueueIO $ fromIntegral n
+  acceptorQueue  <- newTBQueueIO $ fromIntegral n
+  nodeInfoStore  <- newIORef []
+  weAreDone      <- newIORef False
+  -- Run the forwarder. It will take 'TraceItem's from 'forwarderQueue'.
+  void . forkIO $ runTraceForwarder (mkForwarderConfig ep (return nodeInfo))
+                                    forwarderQueue
+  threadDelay 3000000
```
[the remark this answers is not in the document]
denisshevchenko: Fixed.

trace-forward/src/Trace/Forward/Configuration.hs
coot: Then we should avoid that if that's possible.

trace-forward/src/Trace/Forward/Network/Forwarder.hs
coot: But there we have two threads writing updating the `siblingVar`, so it has a chance to terminate.

trace-forward/src/Trace/Forward/Protocol/Limits.hs
```diff
@@ -0,0 +1,53 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE PolyKinds #-}
+{-# LANGUAGE RankNTypes #-}
+
+module Trace.Forward.Protocol.Limits
+  ( byteLimitsTraceForward
+  , timeLimitsTraceForward
+  ) where
```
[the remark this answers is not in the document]
denisshevchenko: Both pipe and host:port are supported: https://github.com/input-output-hk/cardano-node/blob/a8d2ad0eb51064bba7e6061149f70c5021c8fa00/trace-forward/src/Trace/Forward/Network/Acceptor.hs#L70-L78

trace-forward/src/Trace/Forward/Network/Acceptor.hs
```diff
@@ -0,0 +1,225 @@
...
+
+listenToForwarder
+  :: (CBOR.Serialise lo,
+      ShowProxy lo,
+      Typeable lo)
+  => AcceptorConfiguration lo
+  -> TBQueue lo
+  -> NodeInfoStore
+  -> IO ()
+listenToForwarder config@AcceptorConfiguration{..} loQueue niStore = withIOManager $ \iocp ->
```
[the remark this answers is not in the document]
denisshevchenko: Yes, it's a top level `withIOManager`.

> Can you re-use the `iocp` that was started by `cardano-node`?

Should I accept `iocp` as an argument from the node?

trace-forward/src/Trace/Forward/Configuration.hs
denisshevchenko: Ok, I'll remove it.

trace-forward/src/Trace/Forward/Acceptor.hs
denisshevchenko: @coot Sorry, I didn't understand... All `TraceObject`s received from the forwarder should be stored somewhere, right? We think that `TBQueue` is a good solution.

trace-forward/src/Trace/Forward/Network/Forwarder.hs
denisshevchenko: I removed it.

trace-forward/src/Trace/Forward/Acceptor.hs
coot: You don't need to store them, you just need to process them and if you do that inside protocols handlers (without a queue) you will get better back-pressure, i.e. if the application is slow to process the data it will slow down the pace of requesting more data.

trace-forward/src/Trace/Forward/Protocol/Limits.hs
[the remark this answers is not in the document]
coot: Have you considered different protocol limits when using unix sockets / named pipes?

trace-forward/src/Trace/Forward/Network/Acceptor.hs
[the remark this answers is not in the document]
coot: > Should I accept iocp as an argument from the node?

Yes, that would be better.

trace-forward/src/Trace/Forward/Acceptor.hs
denisshevchenko: Ok, I removed the queue on the acceptor's side, now the handler is provided.

trace-forward/src/Trace/Forward/Protocol/Limits.hs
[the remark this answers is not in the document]
denisshevchenko: @coot I saw such limits here: https://github.com/input-output-hk/ouroboros-network/blob/c82309f403e99d916a76bb4d96d6812fb0a9db81/ouroboros-network/test/Test/Ouroboros/Network/KeepAlive.hs#L93-L94

trace-forward/src/Trace/Forward/Network/Acceptor.hs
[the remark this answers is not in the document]
denisshevchenko: Done.

trace-forward/src/Trace/Forward/Protocol/Limits.hs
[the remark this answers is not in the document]
coot: I was asking if you considered to use no limits for unix sockets / named pipes. In that case, consensus is using `runPeer` rather than `runPeerWithLimits`, it also sets `MiniProtocolLimits` (which is a mux ingress queue limit), to [maximumMiniProtocolLimits](https://input-output-hk.github.io/ouroboros-network/ouroboros-network/src/Ouroboros.Network.NodeToClient.html#maximumMiniProtocolLimits).

trace-forward/src/Trace/Forward/Protocol/Forwarder.hs
```diff
@@ -0,0 +1,63 @@
...
+import           Trace.Forward.Protocol.Type
+
+data TraceForwarder lo m a = TraceForwarder
+  { -- | The acceptor sent us a request for node's info.
+    recvMsgNodeInfoRequest
+      :: m (NodeInfo, TraceForwarder lo m a)
+
+    -- | The acceptor sent us a request for new 'TraceObject's.
+  , recvMsgRequest
+      :: forall blocking. TokBlockingStyle blocking
```
coot, line 24:
```suggestion
      :: forall blocking.
         TokBlockingStyle blocking
```

trace-forward/src/Trace/Forward/Protocol/Forwarder.hs
```diff
@@ -0,0 +1,63 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE RecordWildCards #-}
```
coot, line 5: I'd rather avoid `RecordWildCards` extendion; `NamedFieldPuns` is a better choice.

trace-forward/src/Trace/Forward/Forwarder.hs
```diff
@@ -0,0 +1,41 @@
...
+import           Ouroboros.Network.Util.ShowProxy (ShowProxy(..))
+
+import           Trace.Forward.Configuration (ForwarderConfiguration (..))
+import           Trace.Forward.Network.Forwarder (connectToAcceptor)
+
+runTraceForwarder
+  :: (CBOR.Serialise lo,
+      ShowProxy lo,
+      Typeable lo)
+    => IOManager
```
coot, line 26: Indentation

trace-forward/src/Trace/Forward/Forwarder.hs
```diff
@@ -0,0 +1,41 @@
...
+  -> ForwarderConfiguration lo -- ^ Forwarder configuration.
+  -> TBQueue lo                -- ^ The queue from which the forwarder will take 'TraceObject's.
+  -> IO ()
+runTraceForwarder iomgr config loQueue =
+  tryJust excludeAsyncExceptions (connectToAcceptor iomgr config loQueue) >>= \case
+    Left e -> do
+      putStrLn $ "trace-forward, forwarder has a problem: " <> show e
```
coot, line 33: What about something like this:
```
      putStrLn $ "trace-forward couldn't connect to " <> show address <> ": " <> show e
```
It would be nicer to use a tracer rather than `putStrLn` directly, even if you just trace it to `stdout`.

trace-forward/src/Trace/Forward/Network/Forwarder.hs
```diff
@@ -0,0 +1,138 @@
...
+forwardTraceObjects
+  :: (CBOR.Serialise lo,
+      ShowProxy lo)
+  => ForwarderConfiguration lo
+  -> TBQueue lo
+  -> RunMiniProtocol 'InitiatorMode LBS.ByteString IO () Void
+forwardTraceObjects config loQueue =
+  InitiatorProtocolOnly $
+    MuxPeerRaw $ \channel -> do
+      cv <- newEmptyTMVarIO
```
coot, line 103: What `cv` is used for? It seems useless: it's created here, later updated, and not passed anywhere else (there's no thread that is blocked on it).

trace-forward/src/Trace/Forward/Network/Forwarder.hs
```diff
@@ -0,0 +1,138 @@
...
+forwardTraceObjectsResp
+  :: (CBOR.Serialise lo,
+      ShowProxy lo)
+  => ForwarderConfiguration lo
+  -> TBQueue lo
+  -> RunMiniProtocol 'ResponderMode LBS.ByteString IO Void ()
+forwardTraceObjectsResp config loQueue =
+  ResponderProtocolOnly $
+    MuxPeerRaw $ \channel -> do
+      cv <- newEmptyTMVarIO
```
coot, line 126: The same here.

trace-forward/src/Trace/Forward/Network/Forwarder.hs
```diff
@@ -0,0 +1,138 @@
...
+      (r, trailing) <-
+        runPeerWithLimits
+          (forwarderTracer config)
+          (Forwarder.codecTraceForward CBOR.encode CBOR.decode
+                                       CBOR.encode CBOR.decode
+                                       CBOR.encode CBOR.decode)
+          (byteLimitsTraceForward (fromIntegral . LBS.length))
```
coot, line 133: What is the type of `fromIntegral` that is used here? `LBS.length` returns `Int64` if I am not mistaken.  I am worried about an overflow error.

trace-forward/src/Trace/Forward/Queue.hs
```diff
@@ -0,0 +1,58 @@
...
+-- | Returns at most N 'TraceObject's from the queue.
+getNTraceObjects
+  :: Word16
+  -> TBQueue lo
+  -> STM [lo]
+getNTraceObjects 0 _ = return []
+getNTraceObjects n loQueue =
+  tryReadTBQueue loQueue >>= \case
+    Just lo' -> (:) lo' <$> getNTraceObjects (n - 1) loQueue
+    Nothing  -> getNTraceObjects 0 loQueue
```
coot, line 52: I think this will be clearer:
```suggestion
    Nothing  -> return []
```

trace-forward/test/Test/Trace/Forward/Demo/Tests.hs
```diff
@@ -0,0 +1,112 @@
...
+  weAreDone <- newIORef False
+  -- Run the forwarder. It will take 'TraceItem's from 'forwarderQueue'.
+  void . forkIO . withIOManager $ \iomgr ->
+    runTraceForwarder iomgr
+                      (mkForwarderConfig ep (return nodeInfo))
+                      forwarderQueue
+  threadDelay 1000000
+  -- Run the acceptor. It will ask 'TraceItem's from the forwarder
+  -- and store them in 'acceptorQueue'.
+  void . forkIO . withIOManager $ \iomgr ->
```
coot, line 63: Move `withIOManager` to the top, i.e. `ioProperty $ withIOManager $ \iomgr -> ...`

coot: with `forkIO` one can leak threads, please use `async` package and make sure the thread is killed when the test returns.

trace-forward/src/Trace/Forward/Network/Acceptor.hs
```diff
@@ -0,0 +1,193 @@
...
+  => Snocket IO fd addr
+  -> addr
+  -> ProtocolTimeLimits (Handshake UnversionedProtocol Term)
+  -> OuroborosApplication 'ResponderMode addr LBS.ByteString IO Void ()
+  -> IO ()
+doListenToForwarder snocket address timeLimits app = do
+  networkState <- newNetworkMutableState
+  nsAsync <- async $ cleanNetworkMutableState networkState
+  clAsync <- async . void $
+    withServerNode
```
coot, line 97: The acceptor (run by an external process), was supposed to connect to the forwarder (run by the node), not the other way around, isn't it?

trace-forward/src/Trace/Forward/Network/Acceptor.hs
```diff
@@ -0,0 +1,193 @@
...
+      ShowProxy lo,
+      Typeable lo)
+  => AcceptorConfiguration lo
+  -> ([lo] -> IO ())
+  -> (NodeInfo -> IO ())
+  -> RunMiniProtocol 'ResponderMode LBS.ByteString IO Void ()
+acceptTraceObjects config loHandler niHandler =
+  ResponderProtocolOnly $
+    MuxPeerRaw $ \channel -> do
+      sv <- newEmptyTMVarIO
```
coot, line 126: This `sv` is not needed.

trace-forward/src/Trace/Forward/Network/Acceptor.hs
```diff
@@ -0,0 +1,193 @@
...
+      ShowProxy lo,
+      Typeable lo)
+  => AcceptorConfiguration lo
+  -> ([lo] -> IO ())
+  -> (NodeInfo -> IO ())
+  -> RunMiniProtocol 'InitiatorMode LBS.ByteString IO () Void
+acceptTraceObjectsInit config loHandler niHandler =
+  InitiatorProtocolOnly $
+    MuxPeerRaw $ \channel -> do
+      sv <- newEmptyTMVarIO
```
coot, line 152: This one as well.

trace-forward/src/Trace/Forward/Network/Acceptor.hs
```diff
@@ -0,0 +1,193 @@
...
+
+acceptorActions
+  :: (CBOR.Serialise lo,
+      ShowProxy lo,
+      Typeable lo)
+  => AcceptorConfiguration lo
+  -> ([lo] -> IO ())
+  -> (NodeInfo -> IO ())
+  -> Bool
+  -> Bool
```
coot, line 175: Please include haddocks.

coot (changes requested):

trace-forward/src/Trace/Forward/Protocol/Type.hs
```diff
@@ -0,0 +1,218 @@
...
+  , niVersion         :: !Text
+  , niCommit          :: !Text
+  , niStartTime       :: !UTCTime
+  , niSystemStartTime :: !UTCTime
+  } deriving (Eq, Generic, Show)
+
+instance ShowProxy NodeInfo
+instance Serialise NodeInfo
+
+instance Protocol (TraceForward lo) where
```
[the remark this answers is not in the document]
deepfire: We're abstracting over the type of the log object we forward.

trace-forward/src/Trace/Forward/Queue.hs
coot: Anther implementation is `catMaybes <$> replicateM n (tryReadTBQueue q)`. It's also non blocking but it does not shortcut if it's the queue is shallow. I would refactor your implementation into:
```
getNTraceObjects n0 q = go n0
  where
    go 0 = return  []
    go n = tryReadTBQueue q >>= \case
       Just lo   -> (lo :) <$> go (pred n)
       Nothing -> return []
```

trace-forward/src/Trace/Forward/Network/Acceptor.hs
denisshevchenko: Removed.

trace-forward/src/Trace/Forward/Network/Acceptor.hs
denisshevchenko: Removed.

trace-forward/test/Test/Trace/Forward/Demo/Tests.hs
denisshevchenko: There is no need to kill these threads explicitly: after it's done, I set `weAreDone :: IORef Bool` to `True`, so the session will be correctly finished.

trace-forward/src/Trace/Forward/Network/Acceptor.hs
denisshevchenko: Done.

trace-forward/src/Trace/Forward/Network/Acceptor.hs
denisshevchenko: By default - yes, the acceptor is connecting to the forwarder.

denisshevchenko: By default - yes. But technically, `cardano-tracer` can work as `Initiator` (by default) as well as `Responder`: https://github.com/input-output-hk/cardano-node/blob/tracing-master/cardano-tracer/src/Cardano/Tracer/Acceptors.hs#L147-L159

trace-forward/src/Trace/Forward/Protocol/Type.hs
```diff
@@ -0,0 +1,218 @@
...
+    MsgRequest
+      :: TokBlockingStyle blocking
+      -> Request
+      -> Message (TraceForward lo) 'StIdle ('StBusy blocking)
+
+    -- | Reply with a list of 'TraceObject's for the acceptor.
+    -- State: Busy -> Idle.
+    MsgReply
+      :: BlockingReplyList blocking lo
+      -> Message (TraceForward lo) ('StBusy blocking) 'StIdle
```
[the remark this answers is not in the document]
denisshevchenko: Done.

trace-forward/src/Trace/Forward/Network/Forwarder.hs
denisshevchenko: I took it from `KeepAlive` protocol.

trace-forward/src/Trace/Forward/Forwarder.hs
denisshevchenko: Done.

trace-forward/src/Trace/Forward/Protocol/Type.hs
```diff
@@ -0,0 +1,218 @@
...
+--    them to the acceptor by request.
+-- 3. The acceptor is an application that receives 'TraceObject's from the
+--    forwarder.
+-- 4. You can think of the acceptor as a client, and the forwarder as a server.
+--    After the connection is established, the acceptor asks for 'TraceObject's,
+--    the forwarder replies to it.
+
+-- | The request for N 'TraceObject's.
+-- The acceptor will send this request to the forwarder.
+newtype Request = GetTraceObjects Word16
```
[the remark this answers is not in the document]
denisshevchenko: Done.

trace-forward/src/Trace/Forward/Protocol/Forwarder.hs
denisshevchenko: Done.

trace-forward/src/Trace/Forward/Acceptor.hs
```diff
@@ -0,0 +1,45 @@
...
+  -> AcceptorConfiguration lo -- ^ Acceptor configuration.
+  -> ([lo] -> IO ())          -- ^ The handler for 'TraceObject's received from the node.
+  -> (NodeInfo -> IO ())      -- ^ The handler for node's info received from the node.
+  -> IO ()
+runTraceAcceptor iomgr config@AcceptorConfiguration{forwarderEndpoint} loHandler niHandler =
+  tryJust excludeAsyncExcs (listenToForwarder iomgr config loHandler niHandler) >>= \case
+    Left e -> do
+      logTrace $ "trace-forward, connection with "
+                 <> show forwarderEndpoint <> " failed: " <> show e
+      sleep 1.0
```
coot, line 37: Let's make this more robust and use some progression, e.g. `1, 2, 4, 8, 16` seconds.

trace-forward/src/Trace/Forward/Configuration.hs
```diff
@@ -0,0 +1,48 @@
...
+data AcceptorConfiguration lo = AcceptorConfiguration
+  { -- | The tracer that will be used by the acceptor in its network layer.
+    acceptorTracer    :: !(Tracer IO (TraceSendRecv (TraceForward lo)))
+    -- | The endpoint that will be used to listen to the forwarder.
+  , forwarderEndpoint :: !HowToConnect
+    -- | The request specifies how many 'TraceObject's will be requested.
+  , whatToRequest     :: !NumberOfTraceObjects
+    -- | 'IORef' that can be used as a brake: if an external thread sets
+    --   it to 'True', the acceptor will send 'MsgDone' message to the
+    --   forwarder and their session will be closed.
```
coot, line 36: What happens if the client is in a wrong state which does not allow to send `MsgDone`.  Will it timeout? Will the owning thread (that set's the `IORef`) will be notified (does it need to?).

trace-forward/src/Trace/Forward/Forwarder.hs
```diff
@@ -0,0 +1,45 @@
...
+  => IOManager                 -- ^ 'IOManager' from the external application.
+  -> ForwarderConfiguration lo -- ^ Forwarder configuration.
+  -> TBQueue lo                -- ^ The queue from which the forwarder will take 'TraceObject's.
+  -> IO ()
+runTraceForwarder iomgr config@ForwarderConfiguration{acceptorEndpoint} loQueue =
+  tryJust excludeAsyncExceptions (connectToAcceptor iomgr config loQueue) >>= \case
+    Left e -> do
+      logTrace $ "trace-forward, connection with "
+                 <> show acceptorEndpoint <> " failed: " <> show e
+      sleep 1.0
```
coot, line 37: Let's use some progression to make it more robust, e.g. `1,2,4,8,16` seconds.

trace-forward/src/Trace/Forward/Network/Acceptor.hs
coot: Do we need this flexibility? This could be a source of misconfiguration or integration errors.

trace-forward/src/Trace/Forward/Protocol/Limits.hs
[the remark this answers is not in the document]
coot: ping

trace-forward/test/Test/Trace/Forward/Demo/Tests.hs
```diff
@@ -0,0 +1,114 @@
...
+  -- and store them in 'acceptorQueue'.
+  void . forkIO $
+    runTraceAcceptor
+      iomgr
+      (mkAcceptorConfig ep weAreDone)
+      (traceItemsHandler acceptedItems)
+      nodeInfoHandler
+  threadDelay 1000000
+  -- Run the forwarder. It will take 'TraceItem's from 'forwarderQueue'.
+  void . forkIO $
```
coot, line 65: We're leaking a thread here.  You won't notice this in `IO` but when you'll start using `IOSim` for testing you could be hit by this.

coot (changes requested):

coot:
This PR is mostly using `IO`.  However, the consensus and networking code (p2p) can be run in `IOSim`.  This can enhance how things are tested.  I am not sure if the `cardano-node` team has a somewhere in their plans to be able to run the node in `IOSim` as well, even if not using `IOSim` for _tracer-forwarder_ would be worth it.  This does not need to be done in this PR, but I'd hope there will be a subsequent PR for that.

trace-forward/src/Trace/Forward/Network/Acceptor.hs
deepfire: Yes, unfortunately we do @coot -- since we need the node to be the responder always.

So in the local case it is unavoidable.

trace-forward/src/Trace/Forward/Forwarder.hs
denisshevchenko: Done.

trace-forward/src/Trace/Forward/Acceptor.hs
denisshevchenko: Done.

trace-forward/src/Trace/Forward/Network/Acceptor.hs
denisshevchenko: Yes, we need it. This idea was from Neil, and Serge can explain it better.

trace-forward/test/Test/Trace/Forward/Demo/Tests.hs
denisshevchenko: Fixed.

trace-forward/src/Trace/Forward/Configuration.hs
denisshevchenko: > What happens if the client is in a wrong state which does not allow to send `MsgDone`.

Yes, the client won't send `MsgDone` **_immediately_** after we set this `IORef` to `True`. But it will send it during the next request.

trace-forward/src/Trace/Forward/Protocol/Limits.hs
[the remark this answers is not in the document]
denisshevchenko: Removed.

trace-forward/src/Trace/Forward/Configuration.hs
coot: Will this external thread block until the protocol exists? Using `IORef` suggests that it does not, will it then terminate the protocol abruptly without giving it a chance to send `MsgDone`?

trace-forward/src/Trace/Forward/Network/Acceptor.hs
coot: So the node can run the acceptor as well? Is that the intention?

trace-forward/test/Test/Trace/Forward/Protocol/Tests.hs
```diff
@@ -0,0 +1,29 @@
...
+import           Test.Trace.Forward.Protocol.Codec ()
+import           Test.Trace.Forward.Protocol.TraceItem
+
+tests :: TestTree
+tests = testGroup "Trace.Forward.Protocol"
+  [ testProperty "codec" prop_codec_TraceForward
+  ]
+
+prop_codec_TraceForward :: AnyMessageAndAgency (TraceForward TraceItem) -> Bool
+prop_codec_TraceForward msg =
```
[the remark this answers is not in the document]
coot: ping

trace-forward/src/Trace/Forward/Utils.hs
```diff
@@ -0,0 +1,39 @@
...
+runActionInLoop
+  :: IO ()
+  -> HowToConnect
+  -> Word
+  -> IO ()
+runActionInLoop action endpoint prevDelay =
+  tryJust excludeAsyncExceptions action >>= \case
+    Left e -> do
+      logTrace $ "trace-forward, connection with " <> show endpoint <> " failed: " <> show e
+      let currentDelay = mkDelayInSecs
```
coot, line 24: This `let` is unnecesary :); I'd just rename `mkDelayInSecs` to `currentDelay` and use that instead.

trace-forward/src/Trace/Forward/Utils.hs
```diff
@@ -0,0 +1,39 @@
...
+      sleep $ fromIntegral currentDelay
```
coot, line 25: In other parts of the code you're using `threadDelay`.  I think it's better to be consistent and use it all the way through.  This makes things harder to confuse seconds vs miliseconds. From which package does `sleep` comes from?

denisshevchenko: From `extra` package.

denisshevchenko: So yes, I could use `sleep` everywhere.

trace-forward/src/Trace/Forward/Network/Acceptor.hs
```diff
@@ -0,0 +1,171 @@
...
+  :: Ord addr
+  => Snocket IO fd addr
+  -> addr
+  -> ProtocolTimeLimits (Handshake UnversionedProtocol Term)
+  -> OuroborosApplication 'ResponderMode addr LBS.ByteString IO Void ()
+  -> IO ()
+doListenToForwarder snocket address timeLimits app = do
+  networkState <- newNetworkMutableState
+  nsAsync <- async $ cleanNetworkMutableState networkState
+  clAsync <- async . void $
```
coot, line 86: It will be cleaner (and safer) to use `race_` to run the two threads.

trace-forward/src/Trace/Forward/Utils.hs
denisshevchenko: Done.

trace-forward/src/Trace/Forward/Network/Acceptor.hs
coot: Another remark, I don't think you actually need to run `withServerNode` in a separate thread. You could just use:
```
waitAnyCancel [nsAsync, serverAsync]
```
inside the call back to `withServerNode`.


If an async exception hits, you will leak both threads (that's why `race_` is safer).  What might want here is
```
withAsync (cleanNetworkMutableState networkState)
  $ \nsAsync ->
    withServerNode ... $ \serverAsync ->
      waitAny [nsAsync, serverAsync] -- no need for `waitAnyCancel`, `withAsync` and `withServerNode` will handle that

Comment (denisshevchenko) [reply trace-forward/src/Trace/Forward/Network/Acceptor.hs:86]: Thanks, done.

File: trace-forward/src/Trace/Forward/Utils.hs (hunk 41 above)
Comment (denisshevchenko) [reply trace-forward/src/Trace/Forward/Utils.hs:25]: Now `sleep` is everywhere.

File: trace-forward/src/Trace/Forward/Network/Acceptor.hs [hunk 43]
```diff
@@ -0,0 +1,171 @@
...
+import           Trace.Forward.Configuration (AcceptorConfiguration (..), HowToConnect (..))
+
+listenToForwarder
+  :: (CBOR.Serialise lo,
+      ShowProxy lo,
+      Typeable lo)
+  =>  IOManager
+  -> AcceptorConfiguration lo
+  -> ([lo] -> IO ())
+  -> (NodeInfo -> IO ())
```
coot, line 59: I cannot find where production actions are defined.   Is the acceptor process developed in a separate branch?

trace-forward/src/Trace/Forward/Utils.hs
denisshevchenko: Thanks, removed.

trace-forward/src/Trace/Forward/Network/Acceptor.hs
denisshevchenko: Production actions are defined in `cardano-tracer` (a new application that uses _acceptor's side_ of `trace-forward` library). But `cardano-tracer` is still in another branch.

trace-forward/src/Trace/Forward/Configuration.hs
denisshevchenko: > Will this external thread block until the protocol exists?

No, this external thread just sets `IORef` to `True`, nothing more. And then, during the next attempt to send request, the library checks this `IORef` and handles it.

trace-forward/src/Trace/Forward/Network/Acceptor.hs
coot: I see, thanks for clarification. Could you share a link to them?

coot (approved):
LGTM, @denisshevchenko thanks for cooperation!

denisshevchenko:
bors merge

trace-forward/src/Trace/Forward/Configuration.hs
coot: I missed your reply.  I was asking about the thread that tries to shut down the acceptor not the acceptor it self.  How will it wait for it to finish?

deepfire:
Marcin has one last question.

trace-forward/src/Trace/Forward/Configuration.hs
denisshevchenko: > I was asking about the thread that tries to shut down the acceptor not the acceptor it self. How will it wait for it to finish?

It doesn't wait for the finish of the acceptor. This thread only sets `IORef` to `True`, it's like he said to the acceptor: "Hey, you should stop, please send `MsgDone` to the forwarder as soon as you can".

coot: Does it mean it will leak a thread (the protocol handlers) if the acceptor will not finish quickly?

denisshevchenko: Yes, theoretically it's possible that, if the handler (https://github.com/input-output-hk/cardano-node/blob/9f82616efe1f2836e7435f4b3c0f4d48d28c83d0/trace-forward/src/Trace/Forward/Network/Acceptor.hs#L164) takes 2 hours, the `IORef` will be checked (and `MsgDone` will be sent) only after 2 hours. :-)

coot: I see, so I'd suggest when you set it, you monitor the thread that runs the protocol handlers and kill it after let say 30 seconds (or maybe quicker) if it did not return by it self.  You can use `Async` in combination with `registerDelay` to achieve this.

denisshevchenko: Done.

trace-forward/src/Trace/Forward/Network/Acceptor.hs
```diff
@@ -0,0 +1,200 @@
...
+runActionWithTimer action timerInSecs = do
+  flag <- registerDelay delayInMicroSecs
+  race_ action $ timer flag
+ where
+  delayInMicroSecs = timerInSecs * 1000000
+
+  timer flag' =
+    unlessM (readTVarIO flag') $ do
+      sleep 0.01
+      timer flag'
```
coot, line 200:
```suggestion
runActionWithTimer action timerInSecs = do
    v <- registerDelay (delayInSecs * 1_000_000)
    withAsync action $ \a -> atomically $
       waitSTM a `orElse` readTVar v >>= check
```

coot (changes requested):

trace-forward/src/Trace/Forward/Network/Acceptor.hs
deepfire: Thank you a lot for the suggestions @coot !

denisshevchenko: IMHO, it's much harder to read/understand...

coot: Good question. First of all it a declarative description what it does rather than a procedural. So I'd argue that it is easier to understand, although that might depend how familiar one is with stm. One could also wrap it with [`FirstToFinish`](https://input-output-hk.github.io/ouroboros-network/monoidal-synchronisation/Data-Monoid-Synchronisation.html#t:FirstToFinish) and use monoid instance insead  of `orElse`; or use `<|>` intead of `orElse`.

Beside it being declarative, its also more efficient. There's no semi-busy loop which needs to be handled by cpu every 0.1sec. The stm transaction will wake up only once either if the thread is done or when the timeout expires.

trace-forward/src/Trace/Forward/Network/Acceptor.hs
```diff
@@ -0,0 +1,200 @@
...
+acceptorActions config@AcceptorConfiguration{whatToRequest, shouldWeStop} loHandler niHandler askForNI False =
+  -- We are able to send request for:
+  -- 1. node's info,
+  -- 2. new 'TraceObject's.
+  -- But request for node's info should be sent only once (in the beginning of session).
+  if askForNI
+    then
+      Acceptor.SendMsgNodeInfoRequest $ \replyWithNI -> do
+        runActionWithTimer (niHandler replyWithNI) limitInSecs
+        readIORef shouldWeStop <&> acceptorActions config loHandler niHandler False
```
coot, line 164: This is not do what I asked for.  I wanted the whole `acceptActions` to terminate if the `IORef` is set to true and does not terminate within a timeout, now it will timeout the `niHandler` action but it has nothing to do with the `IORef`.

coot (changes requested):

trace-forward/src/Trace/Forward/Network/Acceptor.hs
coot: I'd suggest to turn the `IORef` into a `TVar` and use it in `runActionWithTimer`.  It will require two stm transactions first would block on on this tvar. Then one can create a timerVar and block on it, something like this:
```
timeoutWhenStopped :: TVar Bool -> Int ->  IO a -> IO (Maybe a)
timeoutWhenStoppepd stopVar delay io =
  fmap (either (\_ -> Nothing) Just) $
  race io
         $ do atomically (readTVar stopVar >>= check)
                 v <- registerDealay delay    
             atomically (readTVar v >>= check)       
```
You cannot apply it to `TraceAcceptor` type directly, since it's not an `IO` action, but you can plug it where you run the acceptor using `runPeer`.

deepfire: @coot, what you are alluding to is quite involved -- can you suggest a concrete change -- we'll be happy to accept.

coot: The biggest change is `IORef -> TVar`, but it's a mechanical change - so quite easy to do.  Other than that it's just wrapping `runPeer` inside `timeoutWhenStopped`.

denisshevchenko: `TVar` instead of `IORef` is already done.

denisshevchenko: `TVar` instead of `IORef` is already done.

denisshevchenko: `TVar` instead of `IORef` is already done: https://github.com/input-output-hk/cardano-node/blob/c6f8f7555993cf31c8a63906ef42ce4dad6a6118/trace-forward/src/Trace/Forward/Network/Acceptor.hs#L168

denisshevchenko: > wrapping `runPeer` inside `timeoutWhenStopped`

Honestly, I cannot understand why?? I do understand why we need stop timer - to avoid too long (theoretically - infinite) time for performing of external handlers. And now it's STM-transactions with `orElse` - ok, I understand. But why do we need to wrap `runPeer` inside `timeoutWhenStopped`? `runPeer` calls `acceptorActions`, `acceptorActions` calls `runActionWithTimer`...

coot: `Peer`, which is a result of calling `acceptorActions` is just a data type, you cannot effectively limit how long it will execute after the stop var was set to `True`, or any other resource handling.  You can only do that around `runPeer` which returns an `IO` action, `IO` actions can be timed out.

denisshevchenko: Ok. But I still don't understand how to do it exactly. Are there any existing typed-protocols in `ouroboros-network` where I can find an example of it? And does any from existing protocols have such a `TVar`-based stopping mechanism?

coot (approved):

PR #2963, opened 2021-07-26
Improved monadic queries

newhoggy:


cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -32,6 +42,21 @@ newtype LocalStateQueryScript block point query r m a = LocalStateQueryScript
   { runLocalStateQueryScript :: ContT (Net.Query.ClientStAcquired block point query m r) m a
   } deriving (Functor, Applicative, Monad, MonadIO)
 
+data BlockInModeWrapper where
+  BlockInModeWrapper :: BlockInMode mode -> BlockInModeWrapper
+
+data Program query next
+  = LocalChainSync (Either (ChainPoint, ChainTip) (BlockInModeWrapper, ChainTip) -> next)
+  | forall result. QueryLocalState (query result) (result -> next)
```
newhoggy, line 50: Can do `LocalChainSync` and `QueryLocalState` from the same monadic expression.

cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -32,6 +42,21 @@ newtype LocalStateQueryScript block point query r m a = LocalStateQueryScript
...
+
+deriving instance Functor (Program query)
+
+localChainSync :: MonadIO m => FreeT (Program query) m (Either (ChainPoint, ChainTip) (BlockInModeWrapper, ChainTip))
+localChainSync = liftF (LocalChainSync id)
+
+queryLocalState :: MonadIO m => query result -> FreeT (Program query) m result
+queryLocalState query = liftF (QueryLocalState query id)
```
newhoggy, line 58: So in the same expression I can query both `localChainSync` and `queryLocalState`.

PR #2967, opened 2021-07-26
Add Plutus example that computes sum [1..n]

MarcFontaine:
The benchmarking team uses this plutus script to test smart contracts with changing running times.


erikd (commented):
Why is there a `plutus-example` directory inside a `plutus-example` directory? Sure on of them is redundant.

MarcFontaine:
> Why is there a `plutus-example` directory inside a `plutus-example` directory? Sure one of them is redundant.

Building the Plutus examples is more tricky than it seems.
(Circular dependencies, pinning dependencies to the right versions etc).
That is also why the Plutus examples have their own `cabal.project` file.
I do not know all the details, I'm just building on existing work by Jordan.

Jimbo4350 (approved):
:+1:

PR #2976, opened 2021-07-27
CAD-3092 bench:  move locli and cardano-topology

deepfire:
This moves the `locli` and `cardano-topology` binaries to `bench/`.


denisshevchenko (approved):

Jimbo4350 (approved):
:+1:

newhoggy (approved):

PR #2977, opened 2021-07-28
Make monadic queries easier to access

newhoggy:
Hides away the STM details so callers don't have to deal with them anymore.  Reduces the layers of errors types and streamlines error handling.  Renamed Script to Expr to avoid confusion with things like Plutus Scripts.

### DIFF
--- a/cardano-api/src/Cardano/Api/IPC/Monad.hs
+++ b/cardano-api/src/Cardano/Api/IPC/Monad.hs
@@ -2,9 +2,10 @@
 {-# LANGUAGE GeneralisedNewtypeDeriving #-}
 
 module Cardano.Api.IPC.Monad
-  ( LocalStateQueryScript
+  ( LocalStateQueryExpr
   , sendMsgQuery
-  , setupLocalStateQueryScript
+  , executeQueryLocalState
+  , executeQueryLocalStateWithChainSync
   ) where
 
 import Cardano.Api.Block
@@ -21,40 +22,113 @@ import Data.Ord
 import Shelley.Spec.Ledger.Scripts ()
 import System.IO
 
+import qualified Ouroboros.Network.Protocol.ChainSync.Client as Net.Sync
 import qualified Ouroboros.Network.Protocol.LocalStateQuery.Client as Net.Query
 import qualified Ouroboros.Network.Protocol.LocalStateQuery.Type as Net.Query
 
 -- | Monadic type for constructing local state queries.
 --
 -- Use 'sendMsgQuery' in a do block to construct queries of this type and convert
--- the expression to a 'Net.Query.LocalStateQueryClient' with 'setupLocalStateQueryScript'.
-newtype LocalStateQueryScript block point query r m a = LocalStateQueryScript
-  { runLocalStateQueryScript :: ContT (Net.Query.ClientStAcquired block point query m r) m a
+-- the expression to a 'Net.Query.LocalStateQueryClient' with 'setupLocalStateQueryExpr'.
+newtype LocalStateQueryExpr block point query r m a = LocalStateQueryExpr
+  { runLocalStateQueryExpr :: ContT (Net.Query.ClientStAcquired block point query m r) m a
   } deriving (Functor, Applicative, Monad, MonadIO)
 
 -- | Use 'sendMsgQuery' in a do block to construct monadic local state queries.
-sendMsgQuery :: Monad m => query a -> LocalStateQueryScript block point query r m a
-sendMsgQuery q = LocalStateQueryScript . ContT $ \f -> pure $
+sendMsgQuery :: Monad m => query a -> LocalStateQueryExpr block point query r m a
+sendMsgQuery q = LocalStateQueryExpr . ContT $ \f -> pure $
   Net.Query.SendMsgQuery q $
     Net.Query.ClientStQuerying
     { Net.Query.recvMsgResult = f
     }
 
+-- | Execute a local state query expression.
+executeQueryLocalState
+  :: LocalNodeConnectInfo CardanoMode
+  -> Maybe ChainPoint
+  -> (NodeToClientVersion -> LocalStateQueryExpr (BlockInMode CardanoMode) ChainPoint (QueryInMode CardanoMode) () IO a)
+  -> IO (Either AcquireFailure (Maybe a))
+executeQueryLocalState connectInfo mpoint f = do
+  resultVarQueryTipLocalState <- newEmptyTMVarIO
+  waitResult <- pure $ sequence <$> readTMVar resultVarQueryTipLocalState
+
+  connectToLocalNodeWithVersion
+    connectInfo
+    (\ntcVersion ->
+      LocalNodeClientProtocols
+      { localChainSyncClient    = NoLocalChainSyncClient
+      , localStateQueryClient   = Just $ setupLocalStateQueryExpr waitResult mpoint ntcVersion resultVarQueryTipLocalState (f ntcVersion)
+      , localTxSubmissionClient = Nothing
+      }
+    )
+
+  atomically waitResult
+
+-- | Execute a local state query expression concurrently with a chain sync.
+executeQueryLocalStateWithChainSync
+  :: LocalNodeConnectInfo CardanoMode
+  -> Maybe ChainPoint
+  -> (NodeToClientVersion -> LocalStateQueryExpr (BlockInMode CardanoMode) ChainPoint (QueryInMode CardanoMode) () IO a)
+  -> IO (ChainTip, Either AcquireFailure (Maybe a))
+executeQueryLocalStateWithChainSync connectInfo mpoint f = do
+  resultVarQueryTipLocalState <- newEmptyTMVarIO
+  resultVarChainTip <- newEmptyTMVarIO
+
+  waitResult <- pure $ (,)
+    <$> readTMVar resultVarChainTip
+    <*> (sequence <$> readTMVar resultVarQueryTipLocalState)
+
+  connectToLocalNodeWithVersion
+    connectInfo
+    (\ntcVersion ->
+      LocalNodeClientProtocols
+      { localChainSyncClient    = LocalChainSyncClient $ chainSyncGetCurrentTip waitResult resultVarChainTip
+      , localStateQueryClient   = Just $ setupLocalStateQueryExpr waitResult mpoint ntcVersion resultVarQueryTipLocalState (f ntcVersion)
+      , localTxSubmissionClient = Nothing
+      }
+    )
+
+  atomically waitResult
+
+  where
+    chainSyncGetCurrentTip
+      :: STM b
+      -> TMVar  ChainTip
+      -> ChainSyncClient (BlockInMode mode) ChainPoint ChainTip IO ()
+    chainSyncGetCurrentTip waitDone tipVar =
+      ChainSyncClient $ pure clientStIdle
+      where
+        clientStIdle :: Net.Sync.ClientStIdle (BlockInMode mode) ChainPoint ChainTip IO ()
+        clientStIdle =
+          Net.Sync.SendMsgRequestNext clientStNext (pure clientStNext)
+
+        clientStNext :: Net.Sync.ClientStNext (BlockInMode mode) ChainPoint ChainTip IO ()
+        clientStNext = Net.Sync.ClientStNext
+          { Net.Sync.recvMsgRollForward = \_block tip -> ChainSyncClient $ do
+              void . atomically $ putTMVar tipVar tip
+              void $ atomically waitDone
+              pure $ Net.Sync.SendMsgDone ()
+          , Net.Sync.recvMsgRollBackward = \_point tip -> ChainSyncClient $ do
+              void . atomically $ putTMVar tipVar tip
+              void $ atomically waitDone
+              pure $ Net.Sync.SendMsgDone ()
+          }
+
 -- | Use 'sendMsgQuery' in a do block to construct monadic local state queries.
-setupLocalStateQueryScript ::
+setupLocalStateQueryExpr ::
      STM x
   -> Maybe ChainPoint
   -> NodeToClientVersion
   -> TMVar (Maybe (Either Net.Query.AcquireFailure a))
-  -> LocalStateQueryScript (BlockInMode CardanoMode) ChainPoint (QueryInMode CardanoMode) () IO a
+  -> LocalStateQueryExpr (BlockInMode CardanoMode) ChainPoint (QueryInMode CardanoMode) () IO a
   -> Net.Query.LocalStateQueryClient (BlockInMode CardanoMode) ChainPoint (QueryInMode CardanoMode) IO ()
-setupLocalStateQueryScript waitDone mPointVar' ntcVersion resultVar' f =
+setupLocalStateQueryExpr waitDone mPointVar' ntcVersion resultVar' f =
   LocalStateQueryClient $
     if ntcVersion >= NodeToClientV_8
       then do
         pure . Net.Query.SendMsgAcquire mPointVar' $
           Net.Query.ClientStAcquiring
-          { Net.Query.recvMsgAcquired = runContT (runLocalStateQueryScript f) $ \result -> do
+          { Net.Query.recvMsgAcquired = runContT (runLocalStateQueryExpr f) $ \result -> do
               atomically $ putTMVar resultVar' (Just (Right result))
               void $ atomically waitDone
               pure $ Net.Query.SendMsgRelease $ pure $ Net.Query.SendMsgDone ()
--- a/cardano-api/src/Cardano/Api/IPC/Monad.hs
+++ b/cardano-api/src/Cardano/Api/IPC/Monad.hs
@@ -20,7 +20,6 @@ import Control.Monad
 import Control.Monad.IO.Class
 import Control.Monad.Trans.Cont
 import Control.Monad.Trans.Except
-import Data.Bifunctor
 import Data.Either
 import Data.Function
 import Data.Maybe
@@ -52,10 +51,9 @@ newtype LocalStateQueryExpr block point query r m a = LocalStateQueryExpr
 executeQueryLocalState
   :: LocalNodeConnectInfo mode
   -> Maybe ChainPoint
-  -> (AcquireFailure -> e)
   -> (NodeToClientVersion -> ExceptT e (LocalStateQueryExpr (BlockInMode mode) ChainPoint (QueryInMode mode) () IO) a)
-  -> IO (Either e a)
-executeQueryLocalState connectInfo mpoint mapError f = do
+  -> IO (Either AcquireFailure (Either e a))
+executeQueryLocalState connectInfo mpoint f = do
   tmvResultLocalState <- newEmptyTMVarIO
   let waitResult = readTMVar tmvResultLocalState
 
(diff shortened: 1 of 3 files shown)


cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -143,21 +143,27 @@ runQueryProtocolParameters (AnyConsensusModeParams cModeParams) network mOutFile
...
-                    $ QueryInShelleyBasedEra sbe QueryProtocolParameters
-      result <- executeQuery
-                  era
-                  cModeParams
-                  localNodeConnInfo
-                  query
-      writeProtocolParameters mOutFile result
-    Nothing -> left $ ShelleyQueryCmdEraConsensusModeMismatch (AnyConsensusMode cMode) anyE
+  result <- liftIO $ executeQueryLocalState localNodeConnInfo Nothing $ \_ntcVersion -> do
+    anyE@(AnyCardanoEra era) <- determineEra2 cModeParams
```
newhoggy, line 147: Determine era is a query that yields a result that determines if follow up queries are made.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -829,6 +778,15 @@ determineEra cModeParams localNodeConnInfo =
         Left acqFail -> left $ ShelleyQueryCmdAcquireFailure acqFail
         Right anyCarEra -> return anyCarEra
 
+determineEra2 :: Monad m
+  => ConsensusModeParams mode
+  -> ExceptT e (LocalStateQueryExpr block point (QueryInMode mode) r m) AnyCardanoEra
+determineEra2 cModeParams =
```
DavidEichmann, line 784: Maybe rename this to `determineEraExpr`?

DavidEichmann (changes requested):

cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -14,57 +15,136 @@ import Control.Concurrent.STM
...
-        atomically $ putTMVar resultVar' Nothing
+
+-- | Use 'sendMsgQuery' in a do block to construct monadic local state queries.
+setupLocalStateQueryExpr ::
+     STM x
+  -> Maybe ChainPoint
+  -> TMVar (Either Net.Query.AcquireFailure a)
+  -> LocalStateQueryExpr (BlockInMode mode) ChainPoint (QueryInMode mode) () IO a
+  -> Net.Query.LocalStateQueryClient (BlockInMode mode) ChainPoint (QueryInMode mode) IO ()
+setupLocalStateQueryExpr waitDone mPointVar' resultVar' f =
```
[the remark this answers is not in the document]
newhoggy: `waitDone` is need still because it's what keeps the protocols alive.  `waitDone` only returns when all the component `TVar`s are populated, indicating that it is okay to exit.

Specifically, it is needed by `executeQueryLocalStateWithChainSync`, but not `executeQueryLocalState`, but they share the `setupLocalStateQueryExpr` implementation, so it is retained in both cases.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: I was inspired by your suggestion to also rename `sendMsgQuery` to `queryExpr`

newhoggy: Also moved `determineEraExpr` to the `Monad` module because it seems pretty general.

cardano-api/src/Cardano/Api/IPC/Monad.hs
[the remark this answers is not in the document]
DavidEichmann: Why do we need to keep the protocols alive? Aren't the mini protocols allowed to exit at different times? I would assume that we don't need to keep the local-state-query protocol going in order to still use the chain-sync protocol.

cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -1,70 +1,164 @@
...
+
+    , Net.Query.recvMsgFailure = \failure -> do
+        atomically $ putTMVar resultVar' (Left failure)
         void $ atomically waitDone
         pure $ Net.Query.SendMsgDone ()
+    }
+
+-- | Use 'queryExpr' in a do block to construct monadic local state queries.
+queryExpr :: Monad m => query a -> ExceptT e (LocalStateQueryExpr block point query r m) a
+queryExpr q = ExceptT $ do
```
DavidEichmann, line 148: Why do we wrap in `ExceptT`? It looks like we never fail. Similarly in `determineEraExpr`

cardano-api/src/Cardano/Api/IPC/Monad.hs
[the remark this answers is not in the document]
newhoggy: I believe if you get to the `Done` state, all miniprotocols terminate.  This was causing an STM deadlock previously.

cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -1,70 +1,164 @@
...
-    Net.Query.ClientStQuerying
-    { Net.Query.recvMsgResult = f
-    }
+-- | Execute a local state query expression.
+executeQueryLocalState
+  :: LocalNodeConnectInfo mode
+  -> Maybe ChainPoint
+  -> (AcquireFailure -> e)
+  -> (NodeToClientVersion -> ExceptT e (LocalStateQueryExpr (BlockInMode mode) ChainPoint (QueryInMode mode) () IO) a)
+  -> IO (Either e a)
```
DavidEichmann, line 57: Please add documentation for the `Maybe ChainPoint` argument.

This seems overly complicated. Why not just:

```haskell
executeQueryLocalState
  :: LocalNodeConnectInfo mode
  -> Maybe ChainPoint
  -> (NodeToClientVersion -> LocalStateQueryExpr (BlockInMode mode) ChainPoint (QueryInMode mode) () IO a)
  -> IO (Either AcquireFailure a)
```

cardano-api/src/Cardano/Api/IPC/Monad.hs
newhoggy: This is mostly for convenience because the expression that needs to be provided must also have `ExceptT`.  To not this be the case, I need a combinator that can easily lift `Monad m => m a` to `Monad m => ExceptT e m a`.

cardano-api/src/Cardano/Api/IPC/Monad.hs
newhoggy: I'm trying to find the right ergonomics for the use site.  A lot of the time these monadic expressions will want error handling and I'm trying to avoid multilayered Eithers and Maybes as much as possible.

cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -21,40 +22,113 @@ import Data.Ord
...
-  { runLocalStateQueryScript :: ContT (Net.Query.ClientStAcquired block point query m r) m a
+-- the expression to a 'Net.Query.LocalStateQueryClient' with 'setupLocalStateQueryExpr'.
+newtype LocalStateQueryExpr block point query r m a = LocalStateQueryExpr
+  { runLocalStateQueryExpr :: ContT (Net.Query.ClientStAcquired block point query m r) m a
   } deriving (Functor, Applicative, Monad, MonadIO)
 
 -- | Use 'sendMsgQuery' in a do block to construct monadic local state queries.
-sendMsgQuery :: Monad m => query a -> LocalStateQueryScript block point query r m a
-sendMsgQuery q = LocalStateQueryScript . ContT $ \f -> pure $
+sendMsgQuery :: Monad m => query a -> LocalStateQueryExpr block point query r m a
```
Jimbo4350: Typically we will want to use the IO monad right? If that's the case, lets concretize that and make things less generic.

cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -21,40 +22,113 @@ import Data.Ord
...
+      where
+        clientStIdle :: Net.Sync.ClientStIdle (BlockInMode mode) ChainPoint ChainTip IO ()
+        clientStIdle =
+          Net.Sync.SendMsgRequestNext clientStNext (pure clientStNext)
+
+        clientStNext :: Net.Sync.ClientStNext (BlockInMode mode) ChainPoint ChainTip IO ()
+        clientStNext = Net.Sync.ClientStNext
+          { Net.Sync.recvMsgRollForward = \_block tip -> ChainSyncClient $ do
+              void . atomically $ putTMVar tipVar tip
+              void $ atomically waitDone
```
Jimbo4350, line 117: What is `void $ atomically waitDone` doing here? Is it necessary?

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -207,7 +205,18 @@ runQueryTip (AnyConsensusModeParams cModeParams) network mOutFile = do
     CardanoMode -> do
       let localNodeConnInfo = LocalNodeConnectInfo cModeParams network sockPath
 
-      (chainTip, emLocalState) <- liftIO $ queryQueryTip localNodeConnInfo Nothing
+      (chainTip, emLocalState) <- liftIO $
+        executeQueryLocalStateWithChainSync localNodeConnInfo Nothing $ \ntcVersion -> do
+          era <- sendMsgQuery (QueryCurrentEra CardanoModeIsMultiEra)
```
Jimbo4350: Can we factor this into `cardano-api`?

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -141,21 +143,27 @@ runQueryProtocolParameters (AnyConsensusModeParams cModeParams) network mOutFile
...
+      LegacyByronEra -> return $ Left ShelleyQueryCmdByronEra
+      ShelleyBasedEra sbe -> do
+        let cMode = consensusModeOnly cModeParams
+        case toEraInMode era cMode of
+          Just eInMode -> do
+            ppResult :: Either EraMismatch ProtocolParameters <- sendMsgQuery $
+              QueryInEra eInMode $ QueryInShelleyBasedEra sbe QueryProtocolParameters
+            case ppResult of
+              Right pp -> return (Right pp)
+              Left e -> return (Left (ShelleyQueryCmdEraMismatch e))
```
Jimbo4350: Lots of brackets (minor)

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
Jimbo4350: We should move this to `cardano-api` and call it `eraQueryExpression` or something like that.

cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -47,17 +46,17 @@ executeQueryLocalState
   :: LocalNodeConnectInfo mode
   -> Maybe ChainPoint
   -> (NodeToClientVersion -> LocalStateQueryExpr (BlockInMode mode) ChainPoint (QueryInMode mode) () IO a)
-  -> IO (Either AcquireFailure (Maybe a))
+  -> IO (Either AcquireFailure a)
```
Jimbo4350: :100:

cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -40,61 +48,55 @@ newtype LocalStateQueryExpr block point query r m a = LocalStateQueryExpr
...
 -- | Execute a local state query expression.
 executeQueryLocalState
   :: LocalNodeConnectInfo mode
   -> Maybe ChainPoint
-  -> (NodeToClientVersion -> LocalStateQueryExpr (BlockInMode mode) ChainPoint (QueryInMode mode) () IO a)
-  -> IO (Either AcquireFailure a)
-executeQueryLocalState connectInfo mpoint f = do
-  resultVarQueryTipLocalState <- newEmptyTMVarIO
-  waitResult <- pure $ readTMVar resultVarQueryTipLocalState
+  -> (AcquireFailure -> e)
```
Jimbo4350, line 55: Can't we map the error with `firstExceptT` if we wanted to?

Jimbo4350 (commented):
Looking good! A few comments but I see a couple of them are out dated. It would be worth re-splitting up this PR again to include changes in later commits.

cardano-api/src/Cardano/Api/IPC/Monad.hs
newhoggy: Yep we typically do.

cardano-api/src/Cardano/Api/IPC/Monad.hs
newhoggy: It's waiting for all the queries in all the protocols to be "done".  If we exit when in one protocol before another protocol is done we get an STM deadlock because the other protocol is aborted and never finishes.

cardano-api/src/Cardano/Api/IPC/Monad.hs
newhoggy: Yes, we can.  But in this case the function will need to return an extra layer of `Either` in order to convey both the `AcquireFailure` which this function generates, or the `e` which the monadic expression generates.

cardano-api/src/Cardano/Api/IPC/Monad.hs
newhoggy: Okay, I was able to use `lift` to remove some `ExceptT`s.  Not sure why it was so hard the first time.

cardano-api/src/Cardano/Api/IPC/Monad.hs
newhoggy: `Maybe ChainPoint` was pre-existing.  I don't actually know what it is for.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: I think this line doesn't exist anymore.

cardano-api/src/Cardano/Api/IPC/Monad.hs
newhoggy: Okay, this is sorted.

cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -17,54 +22,138 @@ import Control.Monad.Trans.Cont
...
+  -> Maybe ChainPoint
+  -> TMVar (Either Net.Query.AcquireFailure a)
+  -> LocalStateQueryExpr (BlockInMode mode) ChainPoint (QueryInMode mode) () IO a
+  -> Net.Query.LocalStateQueryClient (BlockInMode mode) ChainPoint (QueryInMode mode) IO ()
+setupLocalStateQueryExpr waitDone mPointVar' resultVar' f =
+  LocalStateQueryClient . pure . Net.Query.SendMsgAcquire mPointVar' $
+    Net.Query.ClientStAcquiring
+    { Net.Query.recvMsgAcquired = runContT (runLocalStateQueryExpr f) $ \result -> do
+        atomically $ putTMVar resultVar' (Right result)
+        void $ atomically waitDone
```
Jimbo4350, line 133: Likewise here, a comment explaining the function of `waitDone` would be useful.

cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -17,54 +22,138 @@ import Control.Monad.Trans.Cont
...
+queryExpr :: Monad m => query a -> LocalStateQueryExpr block point query r m a
+queryExpr q =
+  LocalStateQueryExpr . ContT $ \f -> pure $
+    Net.Query.SendMsgQuery q $
+      Net.Query.ClientStQuerying
+      { Net.Query.recvMsgResult = f
+      }
+
+-- | A monad expresion that determines what era the node is in.
+determineEraExpr :: Monad m
```
Jimbo4350, line 152: :+1:

cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -17,54 +22,138 @@ import Control.Monad.Trans.Cont
...
+        pure $ Net.Query.SendMsgRelease $ pure $ Net.Query.SendMsgDone ()
+
+    , Net.Query.recvMsgFailure = \failure -> do
+        atomically $ putTMVar resultVar' (Left failure)
         void $ atomically waitDone
         pure $ Net.Query.SendMsgDone ()
+    }
+
+-- | Use 'queryExpr' in a do block to construct monadic local state queries.
+queryExpr :: Monad m => query a -> LocalStateQueryExpr block point query r m a
```
Jimbo4350, line 143: Let's also concretize `query a` to `QueryInMode mode result` and m to `IO` in order to make the usage of `queryExpr` a bit more obvious.

Jimbo4350 (commented):
Couple more comments but happy to greenlight after.

cardano-api/src/Cardano/Api/IPC/Monad.hs
newhoggy: Done!

cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -17,54 +22,148 @@ import Control.Monad.Trans.Cont
...
+        pure $ Net.Query.SendMsgRelease $ pure $ Net.Query.SendMsgDone ()
+
+    , Net.Query.recvMsgFailure = \failure -> do
+        atomically $ putTMVar resultVar' (Left failure)
+        void $ atomically waitDone -- Wait for all protocols to complete before exiting.
         pure $ Net.Query.SendMsgDone ()
+    }
+
+-- | Use 'queryExpr' in a do block to construct monadic local state queries.
+queryExpr :: query a -> LocalStateQueryExpr block point query r IO a
```
Jimbo4350, line 153: `query a` -> `QueryInMode mode result` ? We should be able to do that.

cardano-api/src/Cardano/Api/IPC/Monad.hs
newhoggy: Done!

cardano-api/src/Cardano/Api/IPC/Monad.hs
newhoggy: Done!

cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -17,54 +22,144 @@ import Control.Monad.Trans.Cont
...
 
--- | Use 'sendMsgQuery' in a do block to construct monadic local state queries.
-sendMsgQuery :: Monad m => query a -> LocalStateQueryScript block point query r m a
-sendMsgQuery q = LocalStateQueryScript . ContT $ \f -> pure $
-  Net.Query.SendMsgQuery q $
-    Net.Query.ClientStQuerying
-    { Net.Query.recvMsgResult = f
-    }
+-- | Execute a local state query expression.
+executeQueryLocalState
```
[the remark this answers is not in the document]
newhoggy: Done!

cardano-api/src/Cardano/Api/IPC/Monad.hs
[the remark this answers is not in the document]
coot: > Why do we need to keep the protocols alive? Aren't the mini protocols allowed to exit at different times? 

This indeed is a consequence of old mux api, we haven't yet updated the `ouroboros-network` client api for its new interface which would allow that.  We have an issue for this here: https://github.com/input-output-hk/ouroboros-network/issues/3231

Jimbo4350 (approved):

DavidEichmann (approved):

newhoggy:
bors merge

angerman:
bors try

newhoggy:
bors merge

newhoggy:
bors merge

PR #2983, opened 2021-07-29
Run all tests in nightly build

newhoggy:
This sets up a new scheduled "Nightly Trigger" workflow which triggers the `Haskell CI` workflow with a parameter that enables all tests including that of dependencies.


erikd (approved):

newhoggy:
bors merge

PR #2988, opened 2021-07-29
Fix nightly build trigger

newhoggy:
Previously the build trigger did not work because two workflows had the name `Haskell CI` and the trigger picked the wrong one.

I've forked the Github Action to support supplying a path to the workflow file instead of a name which removes the ambiguity.


erikd (approved):

newhoggy:
bors merge

newhoggy:
bors merge

newhoggy:
bors merge

PR #3000, opened 2021-08-01
Fix mkfiles.sh script for Windows

newhoggy:


scripts/byron-to-alonzo/mkfiles.sh
```diff
@@ -59,12 +59,23 @@ FUNDS_PER_BYRON_ADDRESS=$((${FUNDS_PER_GENESIS_ADDRESS} - 1000000))
...
+UNAME=$(uname -s) DATE=
+case $UNAME in
+  Darwin )      DATE="gdate";;
+  Linux )       DATE="date";;
+  MINGW64_NT* ) UNAME="Windows_NT"
+                DATE="date";;
 esac
 
+sprocket() {
+  if [ "$OS" == "Windows_NT" ]; then
```
mkoura, line 71: Shouldn't this use `"$UNAME"`?

mkoura (approved):
LGTM. I haven't ran it on Windows though.

newhoggy:
bors merge

Jimbo4350 (approved):

scripts/byron-to-alonzo/mkfiles.sh
```diff
@@ -59,12 +59,23 @@ FUNDS_PER_BYRON_ADDRESS=$((${FUNDS_PER_GENESIS_ADDRESS} - 1000000))
...
+case $UNAME in
+  Darwin )      DATE="gdate";;
+  Linux )       DATE="date";;
+  MINGW64_NT* ) UNAME="Windows_NT"
+                DATE="date";;
 esac
 
+sprocket() {
+  if [ "$UNAME" == "Windows_NT" ]; then
+    echo -n "\\\\.\\pipe\\"
```
DavidEichmann, line 72: Can also use single quotes

```bash
echo -n '\\.\pipe\'
```

scripts/byron-to-alonzo/mkfiles.sh
```diff
@@ -476,13 +487,14 @@ mkdir -p run
 
 for NODE in ${BFT_NODES}; do
   (
+    SPROCKET="$(sprocket "${ROOT}/${NODE}/node.sock")"
```
DavidEichmann, line 490: Do you need to define a variable? Can you not use this inline. Likewise in the next hunk.

scripts/byron-to-alonzo/mkfiles.sh
```diff
@@ -59,12 +59,23 @@ FUNDS_PER_BYRON_ADDRESS=$((${FUNDS_PER_GENESIS_ADDRESS} - 1000000))
...
-  * )            DATE="date";;
+UNAME=$(uname -s) DATE=
+case $UNAME in
+  Darwin )      DATE="gdate";;
+  Linux )       DATE="date";;
+  MINGW64_NT* ) UNAME="Windows_NT"
+                DATE="date";;
 esac
 
+sprocket() {
```
DavidEichmann, line 70: I'd like to see some documentation explaining why we need to do this. Based on your changes and the name of the PR, it looks like we cant just use a file path for the socket? I have no idea what that path `\\.\pipe\` refers to on windows. Is this some special path for pipes (or socket equivalent for windows)? I assume forward slashes really don't work?

DavidEichmann (changes requested):

DavidEichmann:
@newhoggy Have you run this on windows?

scripts/byron-to-alonzo/mkfiles.sh
newhoggy: Windows is weird.

So on Windows, we're normally used to paths looking like this: `C:\MyDirectory\MyFile`, but this is only what the user sees.

Filenames actually look like this to the operating system filesystem: `\\?\C:\MyDirectory\MyFile` and `\\.\C:\MyDirectory\MyFile`

This way of writing filenames is called the UNC Path or the device path and allows you to refer to things that are not files.  For example named pipes.

So long story short, Windows does not follow the Unix philosophy where everything is a file.  Instead the devices and non-file resources are referenced via a schema that embeds the filesystem but reserves other namespaces for non-files.

newhoggy:
> Have you run this on windows?

This PR was tested in a Windows VM 😁

scripts/byron-to-alonzo/mkfiles.sh
newhoggy: Inlined!

scripts/byron-to-alonzo/mkfiles.sh
newhoggy: Fixed!

scripts/byron-to-alonzo/mkfiles.sh
newhoggy: I've added a short explanation as a comment to the script.

DavidEichmann: Amazing thanks!

DavidEichmann (approved):

newhoggy:
bors merge

PR #3001, opened 2021-08-01
Export shelleyBasedToCardanoEra from Cardano.Api

rvl:
This little function is handy to have.

Actually, it would be extremely helpful if every Cardano.Api.* module were exposed from the cardano-api package, unless there's a really good reason for it to be private.


newhoggy (approved):

rvl:
Thanks

bors r+

PR #3004, opened 2021-08-02
Fix and enable Windows tests

newhoggy:


newhoggy:
Previously some windows tests were disabled.  We desired to run all tests in nightly builds so those remaining tests should be enabled.

Unfortunately, some of those tests fail because they invoke bash to run a script that then calls the CLI.  The failure arises because the Haskell test code is unable to locate the bash executable.

This PR fixes that by allowing CI to defined `BASH_PATH` before invoking the tests so that the can look it up to find the bash executable.

erikd (approved):

newhoggy:
bors merge

PR #3005, opened 2021-08-02
tx-generator: reduce tracing output

MarcFontaine:
Reduce the tracing output in the tx-generator/splitting phase.
Only trace summary message.

### DIFF
--- a/bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
+++ b/bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
@@ -306,12 +306,11 @@ localSubmitTx tx = do
   submitTracer <- btTxSubmit_ <$> get BenchTracers
   submit <- getLocalSubmitTx
   ret <- liftIO $ submit tx
-  let
-    msg = case ret of
-      SubmitSuccess -> "local submit success."
-      SubmitFail e -> mconcat
-        [ "local submit failed: " , show e , " (" , show tx , ")"]
-  liftIO $ traceWith submitTracer $ TraceBenchTxSubDebug msg
+  case ret of
+    SubmitSuccess -> return ()
+    SubmitFail e -> liftIO $ traceWith submitTracer $
+                      TraceBenchTxSubDebug $ mconcat
+                        [ "local submit failed: " , show e , " (" , show tx , ")"]
   return ret
 
 makeMetadata :: forall era. IsShelleyBasedEra era => ActionM (TxMetadataInEra era)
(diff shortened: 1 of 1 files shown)


bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
```diff
@@ -306,12 +306,11 @@ localSubmitTx tx = do
...
-    msg = case ret of
-      SubmitSuccess -> "local submit success."
-      SubmitFail e -> mconcat
-        [ "local submit failed: " , show e , " (" , show tx , ")"]
-  liftIO $ traceWith submitTracer $ TraceBenchTxSubDebug msg
+  case ret of
+    SubmitSuccess -> return ()
+    SubmitFail e -> do
+      let msg = mconcat [ "local submit failed: " , show e , " (" , show tx , ")"]
+      liftIO $ traceWith submitTracer $ TraceBenchTxSubDebug msg
```
deepfire, line 313: Our typically accepted style here would be to avoid the `let`:
```
liftIO $ traceWith submitTracer $
  TraceBenchTxSubDebug $ mconcat
    [ "local submit failed: " , show e , " (" , show tx , ")"]
```

bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
```diff
@@ -477,15 +476,22 @@ createChangeInEra value count _proxy = do
...
 createChangeGeneric :: ([Lovelace] -> ActionM (Either String (TxInMode CardanoMode))) -> Lovelace -> Int -> ActionM ()
 createChangeGeneric createCoins value count = do
+  submitTracer <- btTxSubmit_ <$> get BenchTracers
   let
     coinsList = replicate count value
     maxTxSize = 30
     chunks = chunkList maxTxSize coinsList
+    msg = mconcat [ "createChangeGeneric: outputs: ", show count
+                  , " value: ", show value
+                  , " number of txs: ", show $ (count + maxTxSize -1) `div` maxTxSize
```
deepfire, line 486: Hmm, but why do we have to re-compute the implicitly available data?

Isn't the `nTxs == length chunks`?

deepfire (approved):
LGTM -- but small changes.

bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
MarcFontaine: Ahh hmm yes OFC it is

Issue #2941, opened 2021-07-19
[BUG] - cardano-cli cannot parse script-data-value with spaces


tdiesler:
**Internal/External**
*External*   otherwise.

**Area**
*Plutus*    Related to Plutus Scripts (Alonzo).

**Summary**

This works ...

```
cardano-cli transaction hash-script-data --script-data-value '"HelloWorld"'
```

but this doesn't ...

```
cardano-cli transaction hash-script-data --script-data-value '"Hello World"'
option --script-data-value: Failed reading: string without end

Usage: cardano-cli transaction hash-script-data (--script-data-file FILE | 
                                                  --script-data-value JSON VALUE)
```

james-iohk:
@tdiesler This works for me on Ubuntu.

Can you share more details about your environment (at least console and os etc)

tdiesler:
Oh dear, this is a Docker thing. You can leave it with me for now.

tdiesler:
```
$ alias cardano-cli="docker run --rm -it \
  -v node-ipc:/opt/cardano/ipc \
  inputoutput/cardano-node:dev cli"

$ cardano-cli transaction hash-script-data --script-data-value '"Hello World"'
10506b1214d64bb7898d045946aeb4305ca3184ae10e95eaec9517db250a4184
```

Jimbo4350:
Closing this. If this is still relevant please reopen.

PR #3009, opened 2021-08-03
[#2941] cardano-cli cannot parse script-data-value with spaces

tdiesler:


Jimbo4350:
I've asked @LudvikGalois to look into this.

Issue #2908, opened 2021-07-08
[BUG] - Support for RTS options in docker run limited to -e NETWORK


tdiesler:
*External* 

**Area**
*Docker*

**Summary**

A docker container started with NETWORK ...

```
docker run --rm -it \
  -e NETWORK=mainnet \
  inputoutput/cardano-node:dev
```

... gets default RTS options appended to the command line. 

```
+RTS
-N2
-A16m
-qg
-qb
--disable-delayed-os-memory-return
-RTS
```

This isn't the case when started without -e NETWORK.

**Additional context**

I would take a look at this.

Jimbo4350:
Closing this. If this is still relevant please reopen.

PR #3010, opened 2021-08-03
[#2908] Support for RTS options in docker run limited to -e NETWORK

tdiesler:


nix/docker/context/bin/run-node
```diff
@@ -36,6 +36,11 @@ if [[ -z $CARDANO_PORT ]]; then
   CARDANO_PORT=3001
 fi
 
+if [[ -z $CARDANO_RTS_OPTS ]]; then
+  # See: nix/nixos/cardano-node-service.nix#L401
```
drobakowski, line 40:
```suggestion
  # See: nix/nixos/cardano-node-service.nix#L512
```
The position was updated in the meantime...

drobakowski:
Any news on this? #2908

tdiesler:
You can use [nessus-cardano](https://github.com/tdiesler/nessus-cardano) while we're waiting for this to show up upstream.

disassembler:
we pass `$@` so RTS options or any other flags can be passed at runtime. Is there a reason that doesn't work for your use case?

PR #3011, opened 2021-08-03
Add ghc-options to make cabal-repl work for plutus-example

james-iohk:


james-iohk:
... and these options are already applied there

PR #3013, opened 2021-08-03
Bump haskell.nix for caching fix for flake. Lighter devops shell.

jbgi:
include cardano-ping in devops shell, remove haskell librairies.


deepfire (approved):
LGTM!

PR #3015, opened 2021-08-04
Tx generator: BUGFIX: do not ignore errors in `createChange`

MarcFontaine:
BUGFIX:Do not ignore errors in createChange


deepfire (approved):
Thank you!

bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
```diff
@@ -484,7 +484,7 @@ createChangeGeneric createCoins value count = do
   forM_ chunks $ \coins -> do
```
deepfire, line 484: By the way, why is `maxTxSize` hard-coded to 30 above?

PR #3020, opened 2021-08-05
Use forked optparse-applicative fork with different package name

newhoggy:
So that other projects can use the original


erikd (approved):

newhoggy:
bors merge

PR #3031, opened 2021-08-05
Remove http server running on a public port.

MarcFontaine:
A webserver was started by the workbench.
It is unclear why it would be needed.
It could be a possible security risk.

### DIFF
--- a/nix/supervisord-cluster/supervisor-conf.nix
+++ b/nix/supervisord-cluster/supervisor-conf.nix
@@ -43,10 +43,6 @@ let
         stderr_logfile = "${stateDir}/generator/stderr";
         autostart      = false;
       };
-      "program:webserver" = {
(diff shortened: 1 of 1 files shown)


deepfire (changes requested):
@MarcFontaine, since you remove that instead of disabling by default -- can I ask you to negotiate this change with @mkoura ?

mkoura:
In QA we don't use that supervisor cluster anymore, so we are not affected by the change.
However I don't see anything wrong with running that http server. The http server serves pools metadata on address:port that each pool announced during pool registration, so it's desirable to keep it there.

MarcFontaine:
@deepfire The PR has already been prepared and we have a confirmation from Martin that the http server is not used by the QA team any longer.
I suggest that we merge this PR and remove it.

Jimbo4350:
Closing this. Please re-open this PR if it is still relevant.

PR #3050, opened 2021-08-10
New --script-valid and --script-invalid flags

newhoggy:
So that transactions containing Plutus scripts that are expected to fail validation can still be submitted.

### DIFF
--- a/bench/tx-generator/src/Cardano/Benchmarking/Wallet.hs
+++ b/bench/tx-generator/src/Cardano/Benchmarking/Wallet.hs
@@ -119,6 +119,7 @@ genTx key networkId fee metadata inFunds outValues validity
     , txCertificates = TxCertificatesNone
     , txUpdateProposal = TxUpdateProposalNone
     , txMintValue = TxMintNone
+    , txScriptValidity = BuildTxWith TxScriptValidityNone
     }
 
   mkTxOut v = TxOut (Tx.keyAddress @ era networkId key) (mkTxOutValueAdaOnly v) TxOutDatumHashNone
--- a/cardano-api/src/Cardano/Api/Fees.hs
+++ b/cardano-api/src/Cardano/Api/Fees.hs
@@ -548,7 +548,7 @@ evaluateTransactionBalance _ _ _ (ByronTxBody _) =
     --TODO: we could actually support Byron here, it'd be different but simpler
 
 evaluateTransactionBalance pparams poolids utxo
-                           (ShelleyTxBody era txbody _ _ _) =
+                           (ShelleyTxBody era txbody _ _ _ _) =
     withLedgerConstraints era evalAdaOnly evalMultiAsset
   where
     isNewPool :: Ledger.KeyHash Ledger.StakePool Ledger.StandardCrypto -> Bool
@@ -746,6 +746,24 @@ instance Error TxBodyErrorAutoBalance where
   displayError (TxBodyErrorNonAdaAssetsUnbalanced val) =
       "Non-Ada assets are unbalanced: " <> Text.unpack (renderValue val)
 
+handleExUnitsErrors ::
+     ScriptValidity -- ^ Mark script as expected to pass or fail validation
+  -> Map ScriptWitnessIndex ScriptExecutionError
+  -> Map ScriptWitnessIndex ExecutionUnits
+  -> Either TxBodyErrorAutoBalance (Map ScriptWitnessIndex ExecutionUnits)
+handleExUnitsErrors scriptValidity failures exUnitsMap =
+    if null relevantFailures
+      then Right exUnitsMap
+      else Left (TxBodyScriptExecutionError relevantFailures)
+  where relevantFailures :: [(ScriptWitnessIndex, ScriptExecutionError)]
+        relevantFailures = filter byScriptValidity (Map.toList failures)
+        byScriptValidity :: (ScriptWitnessIndex, ScriptExecutionError) -> Bool
+        byScriptValidity (_, e) = case scriptValidity of
+          ScriptValid -> True
+          ScriptInvalid -> case e of
+            ScriptErrorEvaluationFailed _ -> False
+            _ -> True
+
 -- | This is much like 'makeTransactionBody' but with greater automation to
 -- calculate suitable values for several things.
 --
@@ -790,28 +808,29 @@ makeTransactionBodyAutoBalance eraInMode systemstart history pparams
     -- 3. update tx with fees
     -- 4. balance the transaction and update tx change output
 
-    txbody0 <- first TxBodyError $
-               makeTransactionBody txbodycontent {
-                 txOuts = TxOut changeaddr
-                                (lovelaceToTxOutValue 0)
-                                TxOutDatumHashNone
-                        : txOuts txbodycontent
-                                --TODO: think about the size of the change output
-                                -- 1,2,4 or 8 bytes?
-               }
+    txbody0 <-
+      first TxBodyError $ makeTransactionBody txbodycontent
+        { txOuts =
+              TxOut changeaddr (lovelaceToTxOutValue 0) TxOutDatumHashNone
+            : txOuts txbodycontent
+            --TODO: think about the size of the change output
+            -- 1,2,4 or 8 bytes?
+        }
 
     exUnitsMap <- first TxBodyErrorValidityInterval $
                     evaluateTransactionExecutionUnits
                       eraInMode
                       systemstart history
-                      pparams utxo
+                      pparams
+                      utxo
                       txbody0
 
-    exUnitsMap' <- case Map.mapEither id exUnitsMap of
-                     (failures, exUnitsMap')
-                       | Map.null failures -> return exUnitsMap'
-                       | otherwise         -> Left (TxBodyScriptExecutionError
-                                                     (Map.toList failures))
+    let aScriptValidity = unBuildTxWith (txScriptValidity txbodycontent)
+
+    exUnitsMap' <-
+      case Map.mapEither id exUnitsMap of
+        (failures, exUnitsMap') ->
+          handleExUnitsErrors (txScriptValidityToScriptValidity aScriptValidity) failures exUnitsMap'
 
     let txbodycontent1 = substituteExecutionUnits exUnitsMap' txbodycontent
 
@@ -869,12 +888,12 @@ makeTransactionBodyAutoBalance eraInMode systemstart history pparams
     -- fit within the encoding size we picked above when calculating the fee.
     -- Yes this could be an over-estimate by a few bytes if the fee or change
     -- would fit within 2^16-1. That's a possible optimisation.
-    txbody3 <- first TxBodyError $ -- TODO: impossible to fail now
-               makeTransactionBody txbodycontent1 {
-                 txFee  = TxFeeExplicit explicitTxFees fee,
-                 txOuts = TxOut changeaddr balance TxOutDatumHashNone
-                        : txOuts txbodycontent
-               }
+    txbody3 <-
+      first TxBodyError $ -- TODO: impossible to fail now
+        makeTransactionBody txbodycontent1 {
+          txFee  = TxFeeExplicit explicitTxFees fee,
+          txOuts = TxOut changeaddr balance TxOutDatumHashNone : txOuts txbodycontent
+        }
     return txbody3
  where
    era :: ShelleyBasedEra era
--- a/cardano-api/src/Cardano/Api/Tx.hs
+++ b/cardano-api/src/Cardano/Api/Tx.hs
@@ -22,6 +22,7 @@ module Cardano.Api.Tx (
     Tx(.., Tx),
     getTxBody,
     getTxWitnesses,
+    ScriptValidity(..),
 
     -- ** Signing in one go
     ShelleySigningKey(..),
@@ -109,7 +110,6 @@ import           Cardano.Api.NetworkId
 import           Cardano.Api.SerialiseCBOR
 import           Cardano.Api.SerialiseTextEnvelope
 import           Cardano.Api.TxBody
-
 
 -- ----------------------------------------------------------------------------
 -- Signed transactions
@@ -394,7 +394,6 @@ instance IsCardanoEra era => HasTextEnvelope (KeyWitness era) where
         MaryEra    -> "TxWitness MaryEra"
         AlonzoEra  -> "TxWitness AlonzoEra"
 
-
 pattern Tx :: TxBody era -> [KeyWitness era] -> Tx era
 pattern Tx txbody ws <- (getTxBodyAndWitnesses -> (txbody, ws))
   where
@@ -413,7 +412,7 @@ getTxBody (ShelleyTx era tx) =
       ShelleyBasedEraShelley -> getShelleyTxBody tx
       ShelleyBasedEraAllegra -> getShelleyTxBody tx
       ShelleyBasedEraMary    -> getShelleyTxBody tx
-      ShelleyBasedEraAlonzo  -> getAlonzoTxBody ScriptDataInAlonzoEra tx
+      ShelleyBasedEraAlonzo  -> getAlonzoTxBody ScriptDataInAlonzoEra TxScriptValiditySupportedInAlonzoEra tx
   where
     getShelleyTxBody :: forall ledgerera.
                         ShelleyLedgerEra era ~ ledgerera
@@ -433,13 +432,15 @@ getTxBody (ShelleyTx era tx) =
                     (Map.elems msigWits)
                     TxBodyNoScriptData
                     (strictMaybeToMaybe txAuxiliaryData)
+                    TxScriptValidityNone
 
     getAlonzoTxBody :: forall ledgerera.
                        ShelleyLedgerEra era ~ ledgerera
                     => ScriptDataSupportedInEra era
+                    -> TxScriptValiditySupportedInEra era
                     -> Alonzo.ValidatedTx ledgerera
                     -> TxBody era
-    getAlonzoTxBody scriptDataInEra
+    getAlonzoTxBody scriptDataInEra txScriptValidityInEra
                     Alonzo.ValidatedTx {
                       Alonzo.body = txbody,
                       Alonzo.wits = Alonzo.TxWitness'
@@ -448,12 +449,14 @@ getTxBody (ShelleyTx era tx) =
                                      txscripts
                                      txdats
                                      redeemers,
-                      Alonzo.auxiliaryData = auxiliaryData
+                      Alonzo.auxiliaryData = auxiliaryData,
+                      Alonzo.isValid = isValid
                     } =
       ShelleyTxBody era txbody
                     (Map.elems txscripts)
                     (TxBodyScriptData scriptDataInEra txdats redeemers)
                     (strictMaybeToMaybe auxiliaryData)
+                    (TxScriptValidity txScriptValidityInEra (isValidToScriptValidity isValid))
 
 
 getTxWitnesses :: forall era. Tx era -> [KeyWitness era]
@@ -503,11 +506,10 @@ getTxWitnesses (ShelleyTx era tx) =
         map (ShelleyBootstrapWitness era) (Set.elems bootWits)
      ++ map (ShelleyKeyWitness       era) (Set.elems addrWits)
 
-
 makeSignedTransaction :: forall era.
-                         [KeyWitness era]
-                      -> TxBody era
-                      -> Tx era
+     [KeyWitness era]
+  -> TxBody era
+  -> Tx era
 makeSignedTransaction witnesses (ByronTxBody txbody) =
     ByronTx
   . Byron.annotateTxAux
@@ -518,7 +520,9 @@ makeSignedTransaction witnesses (ByronTxBody txbody) =
 makeSignedTransaction witnesses (ShelleyTxBody era txbody
                                                txscripts
                                                txscriptdata
-                                               txmetadata) =
+                                               txmetadata
+                                               scriptValidity
+                                               ) =
     case era of
       ShelleyBasedEraShelley -> makeShelleySignedTransaction txbody
       ShelleyBasedEraAllegra -> makeShelleySignedTransaction txbody
@@ -568,7 +572,7 @@ makeSignedTransaction witnesses (ShelleyTxBody era txbody
                           | sw <- txscripts ])
             datums
             redeemers)
-          (Alonzo.IsValid True) --TODO: expose this in the API
+          (txScriptValidityToIsValid scriptValidity)
           (maybeToStrictMaybe txmetadata)
       where
         (datums, redeemers) =
@@ -582,7 +586,7 @@ makeByronKeyWitness :: forall key.
                     -> TxBody ByronEra
                     -> SigningKey key
                     -> KeyWitness ByronEra
-makeByronKeyWitness _ (ShelleyTxBody era _ _ _ _) = case era of {}
+makeByronKeyWitness _ (ShelleyTxBody era _ _ _ _ _) = case era of {}
 makeByronKeyWitness nw (ByronTxBody txbody) =
     let txhash :: Byron.Hash Byron.Tx
         txhash = Byron.hashDecoded txbody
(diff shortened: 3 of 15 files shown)


cardano-cli/src/Cardano/CLI/Byron/Parsers.hs
```diff
@@ -430,6 +432,12 @@ parseByronVote =
 -- CLI Parsers
 --------------------------------------------------------------------------------
 
+pIsValid :: Parser IsValid
+pIsValid = fmap (IsValid . not) . switch $ mconcat
```
Jimbo4350: Why not use `flag` that defaults to `True`? And then just `fmap` `IsValid`?

See slack conversation for bike shedding of names :smile:

cardano-api/src/Cardano/Api/Tx.hs
```diff
@@ -503,19 +504,21 @@ getTxWitnesses (ShelleyTx era tx) =
         map (ShelleyBootstrapWitness era) (Set.elems bootWits)
      ++ map (ShelleyKeyWitness       era) (Set.elems addrWits)
 
-
 makeSignedTransaction :: forall era.
-                         [KeyWitness era]
-                      -> TxBody era
-                      -> Tx era
-makeSignedTransaction witnesses (ByronTxBody txbody) =
+      Alonzo.IsValid
```
Jimbo4350: So I know this is where the spec specifies the `IsValid` flag but if we specify this here, we then have to provide an `IsValid` value for Byron txs that doesn't get used. What we should do is add the `IsValid` flag in `ShelleyTxBody` instead to avoid this.

Jimbo4350 (changes requested):
Right direction but we really should provide this value via `ShelleyTxBody`

cardano-api/src/Cardano/Api/Fees.hs
```diff
@@ -222,15 +222,16 @@ estimateTransactionFee _ _ _ (ByronTx _) =
...
                        -> Word  -- ^ The number of Byron key witnesses
                        -> Lovelace
-evaluateTransactionFee _ _ _ byronwitcount | byronwitcount > 0 =
+evaluateTransactionFee _ _ _ _ byronwitcount | byronwitcount > 0 =
   error "evaluateTransactionFee: TODO support Byron key witnesses"
 
-evaluateTransactionFee pparams txbody keywitcount _byronwitcount =
-    case makeSignedTransaction [] txbody of
+evaluateTransactionFee pparams isValid txbody keywitcount _byronwitcount =
+    case makeSignedTransaction isValid [] txbody of
```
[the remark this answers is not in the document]
JaredCorduan: yep, you can totally make that a constant.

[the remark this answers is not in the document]
newhoggy: This argument is gone now.

cardano-api/src/Cardano/Api/Tx.hs
newhoggy: The flag has been moved to `ShelleyTxBody`.

cardano-cli/src/Cardano/CLI/Byron/Parsers.hs
newhoggy: The ADT is now called `ScriptValidity`. The option is `--script-valid` and `--script-invalid`

cardano-api/src/Cardano/Api/Fees.hs
```diff
@@ -777,12 +777,13 @@ makeTransactionBodyAutoBalance
   -> ProtocolParameters
   -> Set PoolId       -- ^ The set of registered stake pools
   -> UTxO era         -- ^ Just the transaction inputs, not the entire 'UTxO'.
+  -> ScriptValidity
```
Jimbo4350, line 780: We should put a comment here similar to: "Mark script as expected to pass or fail validation"

cardano-api/src/Cardano/Api/Tx.hs
```diff
@@ -433,6 +432,7 @@ getTxBody (ShelleyTx era tx) =
                     (Map.elems msigWits)
                     TxBodyNoScriptData
                     (strictMaybeToMaybe txAuxiliaryData)
+                    ScriptValid
```
Jimbo4350, line 435: Hmm we should probably implement a GADT similar to `TxBodyScriptData era`.

cardano-cli/src/Cardano/CLI/Byron/Commands.hs
```diff
@@ -87,6 +88,7 @@ data ByronCommand =
   | SpendUTxO
         NetworkId
         ByronKeyFormat
+        ScriptValidity
```
Jimbo4350, line 91: Why are we modifying the Byron era? There are no scripts in the Byron era.

Jimbo4350 (changes requested):
See comment re: GADT

cardano-api/src/Cardano/Api/Fees.hs
newhoggy: Done

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -609,11 +609,29 @@ pTransaction =
     Opt.subparser
       $ Opt.command "sign-witness" assembleInfo <> Opt.internal
 
+  pScriptValidity :: Parser ScriptValidity
+  pScriptValidity = asum
+    [ Opt.flag' ScriptValid $ mconcat
+      [ Opt.long "script-valid"
+      , Opt.help "Assertion that the script is valid."
```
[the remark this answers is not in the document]
newhoggy: Done

newhoggy:
> Tested working on alonzo-purple testnet. Do we need a `--script-valid` as it is applied by default?

Removing it would simplify the CLI.  Thoughts @Jimbo4350 ?

cardano-cli/src/Cardano/CLI/Byron/Commands.hs
newhoggy: Fixed

newhoggy:
The following manual test shows the collateral was taken:

```
CARDANO_CLI="$(./scripts/bin-path.sh cardano-cli)" CARDANO_NODE_SOCKET_PATH=example/node-bft1/node.sock scripts/plutus/always-fails.sh
Socket path: example/node-bft1/node.sock
Socket path: /Users/jky/wrk/iohk/cardano-node
srwxr-xr-x  1 jky  staff  0 12 Aug 15:15 example/node-bft1/node.sock
Always succeeds Plutus script in use. Any datum and redeemer combination will succeed.
Script at: ./scripts/plutus/scripts/always-fails.plutus
{
    "2806950d520e0a31ec0920506fef5b5433ba646d9f0b20efdedc92efc4d5db7d#0": {
        "address": "addr_test1vq2q3rtj666wlscq5787rx70xvvnwgvf58mv4zev5kzp35q7ea29d",
        "value": {
            "lovelace": 1000000000000
        }
    }
}Transaction successfully submitted.
Pausing for 5 seconds...
{
    "31614b4592f9baaa5bd3eb29a34456aad780f6e239147482351d2811547cea8e#0": {
        "address": "addr_test1vq2q3rtj666wlscq5787rx70xvvnwgvf58mv4zev5kzp35q7ea29d",
        "value": {
            "lovelace": 666666666394
        }
    }
}Plutus txin
31614b4592f9baaa5bd3eb29a34456aad780f6e239147482351d2811547cea8e#1
Collateral
31614b4592f9baaa5bd3eb29a34456aad780f6e239147482351d2811547cea8e#0
UTxO for collateral address before submission
                           TxHash                                 TxIx        Amount
--------------------------------------------------------------------------------------
31614b4592f9baaa5bd3eb29a34456aad780f6e239147482351d2811547cea8e     0        666666666394 lovelace + TxOutDatumHashNone
Submit the tx with plutus script and wait 5 seconds...
Transaction successfully submitted.

UTxO for collateral address after submission
                           TxHash                                 TxIx        Amount
--------------------------------------------------------------------------------------


Querying UTxO at addr_test1vpqgspvmh6m2m5pwangvdg499srfzre2dd96qq57nlnw6yctpasy4. If there is ADA at the address the Plutus script successfully executed!

                           TxHash                                 TxIx        Amount
--------------------------------------------------------------------------------------
```

cardano-api/src/Cardano/Api/Tx.hs
newhoggy: Done!

bench/tx-generator/src/Cardano/Benchmarking/Wallet.hs
```diff
@@ -98,7 +98,7 @@ genTx :: forall era. IsShelleyBasedEra era
   -> TxMetadataInEra era
   -> TxGenerator era
 genTx key networkId fee metadata inFunds outValues validity
-  = case makeTransactionBody txBodyContent of
+  = case makeTransactionBody TxBodyScriptValidityNone txBodyContent of
```
Jimbo4350, line 101: We need the benchmarking folks to check their stuff still works.

cardano-api/src/Cardano/Api/Fees.hs
```diff
@@ -746,6 +746,24 @@ instance Error TxBodyErrorAutoBalance where
...
 
+handleExUnitsErrors ::
+     ScriptValidity -- ^ Mark script as expected to pass or fail validation
+  -> Map ScriptWitnessIndex ScriptExecutionError
+  -> Map ScriptWitnessIndex ExecutionUnits
+  -> Either TxBodyErrorAutoBalance (Map ScriptWitnessIndex ExecutionUnits)
+handleExUnitsErrors scriptValidity failures exUnitsMap =
+    if null relevantFailures
+      then Right exUnitsMap
+      else Left (TxBodyScriptExecutionError relevantFailures)
```
Jimbo4350, line 757: :+1:

cardano-api/src/Cardano/Api/Tx.hs
```diff
@@ -433,13 +432,15 @@ getTxBody (ShelleyTx era tx) =
                     (Map.elems msigWits)
                     TxBodyNoScriptData
                     (strictMaybeToMaybe txAuxiliaryData)
+                    TxBodyScriptValidityNone
 
     getAlonzoTxBody :: forall ledgerera.
                        ShelleyLedgerEra era ~ ledgerera
                     => ScriptDataSupportedInEra era
+                    -> TxBodyScriptValiditySupportedInEra era
```
Jimbo4350, line 440: We don't need to parameterize this. We can hardcode the value on line 459 as we know we are in the Alonzo era.

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -216,6 +227,75 @@ import           Cardano.Ledger.Crypto (StandardCrypto)
...
+deriving instance Show (TxBodyScriptValidity era)
+
+txBodyScriptValiditySupportedInEra :: CardanoEra era -> Maybe (TxBodyScriptValiditySupportedInEra era)
+txBodyScriptValiditySupportedInEra ByronEra   = Nothing
+txBodyScriptValiditySupportedInEra ShelleyEra = Nothing
+txBodyScriptValiditySupportedInEra AllegraEra = Nothing
+txBodyScriptValiditySupportedInEra MaryEra    = Nothing
+txBodyScriptValiditySupportedInEra AlonzoEra  = Just TxBodyScriptValiditySupportedInAlonzoEra
+
+txBodyScriptValidityToScriptValidity :: TxBodyScriptValidity era -> ScriptValidity
```
Jimbo4350, line 285: This should be parameterised with `ShelleyBasedEra era` so we can utilize the `EmptyCase` extension to disallow the Byron era.

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -232,18 +240,62 @@ data ScriptValidity
...
+isValidToScriptValidity (Alonzo.IsValid True) = ScriptValid
+
+-- | A representation of whether the era supports tx script validity.
+--
+-- The Mary and subsequent eras support script validity.
+--
+data TxBodyScriptValidity era where
+  TxBodyScriptValidityNone :: TxBodyScriptValidity era
+
+  -- | Tx script validity is supported in transactions in the 'Alonzo' era.
```
Jimbo4350, line 263: Should be "from the 'Alonzo' era onwards."

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -1479,9 +1534,9 @@ instance Error TxBodyError where
       "parameters to hash"
 
 
-makeTransactionBody :: forall era. ()
-  => IsCardanoEra era
-  => ScriptValidity -- ^ Mark script as expected to pass or fail validation
+makeTransactionBody :: forall era.
+     IsCardanoEra era
+  => TxBodyScriptValidity era -- ^ Mark script as expected to pass or fail validation
```
Jimbo4350, line 1539: So we shouldn't infect `makeTransactionBody` with a new parameter. We should add `TxBodyScriptValidity era` to `TxBodyContent` and feed the value in that way.

Jimbo4350 (changes requested):
Right direction, should be one last change required.

cardano-api/src/Cardano/Api/Tx.hs
newhoggy: I think it's still needed.  If we hardcode the value, we still have to prove the hardcoded value matches the `era` parameter.  Unless we deparameterize the entire function.

For example `ScriptDataSupportedInEra era` argument pretty much does the same thing.

cardano-api/src/Cardano/Api/TxBody.hs
newhoggy: If we do this, we can no longer implement:

```
makeTransactionBody :: forall era.
     IsCardanoEra era
  => TxBodyScriptValidity era -- ^ Mark script as expected to pass or fail validation
  -> TxBodyContent BuildTx era
  -> Either TxBodyError (TxBody era)
```

Which must accept `TxBodyScriptValidity era`.  Otherwise there will be no means to thread through the script validity.

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -216,6 +227,75 @@ import           Cardano.Ledger.Crypto (StandardCrypto)
...
+
+deriving instance Eq   (TxBodyScriptValidity era)
+deriving instance Show (TxBodyScriptValidity era)
+
+txBodyScriptValiditySupportedInEra :: CardanoEra era -> Maybe (TxBodyScriptValiditySupportedInEra era)
+txBodyScriptValiditySupportedInEra ByronEra   = Nothing
+txBodyScriptValiditySupportedInEra ShelleyEra = Nothing
+txBodyScriptValiditySupportedInEra AllegraEra = Nothing
+txBodyScriptValiditySupportedInEra MaryEra    = Nothing
+txBodyScriptValiditySupportedInEra AlonzoEra  = Just TxBodyScriptValiditySupportedInAlonzoEra
```
deepfire, line 283: Can we optimise this to:

```
txBodyScriptValiditySupportedInEra AlonzoEra  = Just TxBodyScriptValiditySupportedInAlonzoEra
txBodyScriptValiditySupportedInEra _ = Nothing
```
?

IIRC GHC compiles case matching alternative by alternative, in user-supplied order, so having a series of `Nothing` match failures would be quite inefficient?

cardano-api/src/Cardano/Api/TxBody.hs
newhoggy: This change is now in.  I'm liking this approach.

cardano-api/src/Cardano/Api/TxBody.hs
newhoggy: I'm wary of doing that because later if we add additional eras, the compiler won't catch that this function needs to change.

cardano-api/src/Cardano/Api/TxBody.hs
newhoggy: Modifying `TxBodyContent` to include script validity solve this problem for me.

newhoggy: I made the change and discovered I actually need the `CardanoEra era` argument in some cases.  For example the generators and the CLI parsers.

So I created a separate function for the `ShelleyBasedEra era` case.  Turns out the use case for `ShelleyBasedEra era` are all in the same module.

cardano-api/src/Cardano/Api/TxBody.hs
Jimbo4350: >the compiler won't catch that this function needs to change.
This is more important because a lot of functions need to be updated when we add a new era and GHC tells us.

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -609,11 +609,28 @@ pTransaction =
...
+    [ Opt.flag' ScriptValid $ mconcat
+      [ Opt.long "script-valid"
+      , Opt.help "Assertion that the script is valid. (default)"
+      ]
+    , Opt.flag' ScriptInvalid $ mconcat
+      [ Opt.long "script-invalid"
+      , Opt.help $ mconcat
+        [ "Assertion that the script is invalid.  "
+        , "If a transaction is submitted with such a script, "
+        , "the script will fail and the collateral taken"
```
Jimbo4350, line 623: We should have it default to `ScriptValid` as users will want to know if their script validates without losing their collateral.

txbody
```diff
@@ -0,0 +1,5 @@
+{
+    "type": "TxBodyAlonzo",
```
Jimbo4350, line 2: left over from testing?

Jimbo4350 (changes requested):
Nearly there!

MarcFontaine (approved):
The changes to the tx-generator look good to me.

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
newhoggy: It defaults to `TxScriptValidityNone`, which ultimately means `ScriptValid` here: https://github.com/input-output-hk/cardano-node/blob/1222dba100d0cd49eaa2634cc9d1380023185577/cardano-api/src/Cardano/Api/TxBody.hs#L292

txbody
newhoggy: Deleted.  Thanks!

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -1351,22 +1448,30 @@ deserialiseShelleyBasedTxBody era bs =
       len <- CBOR.decodeListLen
       txbody     <- fromCBOR
       txscripts  <- fromCBOR
-      txscriptdata <-
+      (txscriptdata, txScriptValidity) <-
```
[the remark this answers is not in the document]
newhoggy: Separate monadic action implemented.

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -1351,22 +1448,30 @@ deserialiseShelleyBasedTxBody era bs =
...
+                    return
+                      ( CBOR.Annotator $ \fbs ->
+                          TxBodyScriptData supported
+                            (flip CBOR.runAnnotator fbs datums)
+                            (flip CBOR.runAnnotator fbs redeemers)
+                      , case txScriptValiditySupportedInShelleyBasedEra era of
+                          Nothing -> return TxScriptValidityNone
+                          Just txScriptValiditySupportedInShelleyBasedEraWitness ->
+                            return $ TxScriptValidity txScriptValiditySupportedInShelleyBasedEraWitness scriptValidity
+                      )
```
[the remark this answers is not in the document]
newhoggy: Fixed!

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -1473,6 +1579,7 @@ fromLedgerTxBody era body mAux =
       , txMetadata
       , txAuxScripts
       , txExtraScriptData = ViewTx
+      , txScriptValidity = TxScriptValidityNone -- TODO I can't get this from the body.
```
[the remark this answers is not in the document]
newhoggy: Done!

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -2035,7 +2038,7 @@ makeShelleyTransactionBody era@ShelleyBasedEraShelley
         scripts
         TxBodyNoScriptData
         txAuxData
-        txScriptValidity
+        (unBuildTxWith txScriptValidity)
```
Jimbo4350, line 2041: This should be hardcoded to `TxScriptValidityNone` as era before Alonzo have no notion of script validity.

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -2111,7 +2114,7 @@ makeShelleyTransactionBody era@ShelleyBasedEraAllegra
...
```
Jimbo4350, line 2117: Same here and for all eras except Alonzo.

cardano-cli/src/Cardano/CLI/Shelley/Run/Transaction.hs
```diff
@@ -780,14 +780,14 @@ validateTxUpdateProposal era (Just (UpdateProposalFile file)) =
          return (TxUpdateProposal supported prop)
 
 validateTxScriptValidity :: forall era.
-                       CardanoEra era
-                    -> Maybe ScriptValidity
-                    -> ExceptT ShelleyTxCmdError IO (TxScriptValidity era)
-validateTxScriptValidity _ Nothing = return TxScriptValidityNone
+     CardanoEra era
+  -> Maybe ScriptValidity
+  -> ExceptT ShelleyTxCmdError IO (BuildTxWith BuildTx (TxScriptValidity era))
```
Jimbo4350, line 785: :+1:

Jimbo4350 (changes requested):
A minor change and I'm happy to green light.

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -1308,9 +1406,10 @@ serialiseShelleyBasedTxBody
   -> [Ledger.Script ledgerera]
   -> TxBodyScriptData era
   -> Maybe (Ledger.AuxiliaryData ledgerera)
+  -> TxScriptValidity era -- ^ Mark script as expected to pass or fail validation
   -> ByteString
 serialiseShelleyBasedTxBody _era txbody txscripts
-                            TxBodyNoScriptData txmetadata =
+                            TxBodyNoScriptData txmetadata _scriptValidity =
```
Jimbo4350, line 1412: Shouldn't we also CBOR encode `scriptValidity` here?

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
Jimbo4350: That seems semantically incorrect. `TxScriptValidityNone` is used within the context of not being available for a given era.

cardano-api/src/Cardano/Api/TxBody.hs
newhoggy: Done!

cardano-api/src/Cardano/Api/TxBody.hs
newhoggy: Done!

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -1320,13 +1419,14 @@ serialiseShelleyBasedTxBody _era txbody txscripts
...
-                            txmetadata =
+                            txmetadata txBodycriptValidity =
     CBOR.serializeEncoding' $
-        CBOR.encodeListLen 5
+        CBOR.encodeListLen 6
      <> CBOR.toCBOR txbody
      <> CBOR.toCBOR txscripts
      <> CBOR.toCBOR datums
      <> CBOR.toCBOR redeemers
+     <> CBOR.toCBOR (txScriptValidityToScriptValidity txBodycriptValidity)
```
Jimbo4350, line 1429: What will happen here is in a non-Alonzo era we will default to `ScriptValid` which is incorrect.

Jimbo4350 (changes requested):
Still a couple of things.

Jimbo4350 (approved):
LGTM!

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -216,6 +228,78 @@ import           Cardano.Ledger.Crypto (StandardCrypto)
...
+scriptValidityToIsValid ScriptInvalid = Alonzo.IsValid False
+scriptValidityToIsValid ScriptValid = Alonzo.IsValid True
+
+isValidToScriptValidity :: Alonzo.IsValid -> ScriptValidity
+isValidToScriptValidity (Alonzo.IsValid False) = ScriptInvalid
+isValidToScriptValidity (Alonzo.IsValid True) = ScriptValid
+
+-- | A representation of whether the era supports tx script validity.
+--
+-- The Mary and subsequent eras support script validity.
```
newhoggy, line 259: Should that be Alonzo?

PR #3053, opened 2021-08-10
CAD-3317 nix:  tx-generator-profiled attribute

deepfire:
This supplies a profiled build of `tx-generator`.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -31,7 +31,9 @@ let
   packages = {
     inherit haskellPackages shell
       cardano-node cardano-node-profiled cardano-node-eventlogged
-      cardano-cli db-converter cardano-ping tx-generator
+      cardano-cli db-converter cardano-ping
+      locli locli-profiled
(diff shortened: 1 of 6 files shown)


nix/nixos/cardano-node-service.nix
```diff
@@ -6,7 +6,7 @@
 with lib; with builtins;
 let
   cfg = config.services.cardano-node;
-  inherit (cfg.cardanoNodePkgs) commonLib cardano-node cardano-node-profiled cardano-node-eventlogged cardano-node-asserted;
+  inherit (cfg.cardanoNodePkgs) commonLib cardano-node cardano-node-profiled cardano-node-eventlogged cardano-node-asserted tx-generator-profile;
```
jbgi, line 9: probably should not be added there since this not used by the service.

jbgi (approved):

PR #3056, opened 2021-08-10
benchmarking:  generator and analysis fixes

deepfire:
1. generator run script: min UTxO parameter, change count and amount fixes, non-continuous submission mode, tree-based splitting
2. generator:  non-continuous mode
3. locli:  block-size filtering and CDF-of-CDFs for propagation


MarcFontaine (approved):
LGTM

PR #3062, opened 2021-08-11
benchmarking: generator and analysis fixes (try 2)

deepfire:
This is a re-submission of previously approved but mis-based https://github.com/input-output-hk/cardano-node/pull/3056


deepfire:
Since all key checks passed, and the same PR was previously approved -- merging manually.

PR #3073, opened 2021-08-13
CAD-608 trace-dispatcher library

jutaro:
This is `trace-dispatcher`, a library for tracing, which replaces the `iohk-monitoring-framework`.

* It is build completely on the base of contravariant functors
* Every aspect of tracing is controllable via the configuration, down to every individual message
* It has the ability to document all tracing messages
* It can be used beside the current tracing system in a first step
* When the old tracing infrastucture will be removed it can be dynamically reconfigured in a running node
* It is build for efficiency from the ground up, so that messages will never be stored

Documentation is included in docs/trace-dispatcher.md


trace-dispatcher/examples/Examples/Aggregation.hs
```diff
@@ -0,0 +1,60 @@
...
+  asMetrics BaseStats {..} =
+    [ DoubleM ["measure"] bsMeasure
+    , DoubleM ["sum"] bsSum]
+
+baseStatsDocumented :: Documented Double
+baseStatsDocumented =
+  Documented [
+    DocMsg 0.0 [] "This is the value of the measurement"
+  , DocMsg 0.0 [] "This is the sum of all measurments so far"
+  ]
```
denisshevchenko, line 39: Please choose some style for `[` and `]` placing and use it everywhere. I mean the position of `[`and `]`:

```
instance LogFormatting BaseStats where
  forMachine = mempty
  asMetrics BaseStats {..} =
     [ DoubleM ["measure"] bsMeasure
     , DoubleM ["sum"] bsSum
     ]

baseStatsDocumented :: Documented Double
baseStatsDocumented =
  Documented
    [ DocMsg 0.0 [] "This is the value of the measurement"
    , DocMsg 0.0 [] "This is the sum of all measurments so far"
    ]
```

trace-dispatcher/examples/Examples/Aggregation.hs
```diff
@@ -0,0 +1,60 @@
...
+    } deriving (Eq, Ord, Show, Generic)
+
+instance A.ToJSON BaseStats where
+    toEncoding = A.genericToEncoding A.defaultOptions
+
+instance LogFormatting BaseStats where
+  forMachine = mempty
+  asMetrics BaseStats {..} =
+    [ DoubleM ["measure"] bsMeasure
+    , DoubleM ["sum"] bsSum]
```
denisshevchenko, line 32: Please correct an indentation: 2 spaces everywhere. :-)

trace-dispatcher/examples/Examples/Aggregation.hs
jutaro: Resolved by commit `Unifying mutiline list layout`

trace-dispatcher/examples/Examples/Aggregation.hs
jutaro: Resolved by commit `Alignment fix`

trace-resources/README.md
```diff
@@ -0,0 +1,3 @@
+trace-resources
+
+Query resource statistics for various platforms for tracing.
```
Jimbo4350, line 3: Please put the devs/team responsible for this library here

Jimbo4350 (changes requested):

trace-dispatcher/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Revision history for trace-dispatcher
+
+## 0.1.0.0 -- YYYY-mm-dd
```
DavidEichmann, line 3: Perhaps clear this template entry.

trace-dispatcher/doc/trace-dispatcher.md
```diff
@@ -0,0 +1,798 @@
...
+
+`trace-dispatcher` is an attempt to enable design and implementation of simple, efficient __tracing systems__, one that has a reduced footprint in the executed program, has a more pleasant API, and provides self-documenting features.
+
+## Design decisions
+
+Key design decisions were:
+
+1. Retaining the separation of concerns in the frontend side, as provided by the `contra-tracer` library.  The client code should not need to concern itself with any details beyond passing the traces down to the system.
+2. Rely on __trace combinators__ primarily, as opposed to opting for a typeclass heavy API.
+3. Separation of data plane and control plane:  high-frequency events incur minimal processing on the data-plane, whereas complicated configuration logic only happens on the control plane, and that is proportional to infrequent reconfiguration events.
```
DavidEichmann, line 78: Reading this for the first time, I don't know what you mean by "data plane" and "control plane". Is the data plane something like the program producing the traces and the control plane is something like the (external?) process that consumes the trace events?

style: in one place use hyphenate "data-plane" in another you don't "data plane".

trace-dispatcher/doc/trace-dispatcher.md
```diff
@@ -0,0 +1,798 @@
...
+4. A tougher stance on separation of concerns in the backend side:  we choose to move expensive trace processing to an external process.
+5. A measure of backward compatibility with the previous logging system.
+6. Retaining the global namespace for all traces.
```
DavidEichmann, line 81: I think "the" should be "a" here? You're saying that all traces are keyed on some global key type?

trace-dispatcher/doc/trace-dispatcher.md
```diff
@@ -0,0 +1,798 @@
...
+
+## Overview and terminology
+
+The emitted __program traces__ (streams of __messages__ of arbitrary data types, where each data type defines a number of different __messages__) are collected across all program components, and undergo __trace interpretation__ by the __dispatcher__ into __metrics__ and __messages__, which are afterwards externalised.
+
+Therefore, we can conceptually decompose the __tracing system__ into three components:
+
+* __frontend__, the entry point for __program trace__ collection, which is just a single function `traceWith`;  Program locations that invoke this frontend (thereby injecting messages into the tracing system) is called __trace-ins__.
+* __dispatcher__, is a structured, namespaced set of contravariantly-composed transformations, triggered by the entry point.  Its role is specifically __trace interpretation__;
+* __backend__, externalises results of the interpretation ( __metrics__ and __messages__) outside the system, through __trace-outs__.
```
DavidEichmann, line 91: By "system" do you mean "process"? If yes, perhaps stick with "process"  since you've used that wording already.

trace-dispatcher/doc/trace-dispatcher.md
```diff
@@ -0,0 +1,798 @@
...
+data TraceAddBlockEvent blk =
+    IgnoreBlockOlderThanK (RealPoint blk)
+  | IgnoreBlockAlreadyInVolatileDB (RealPoint blk)
+  ...
+```
+
+__Traces__ cannot be entered into the tracing system, unless they are accompanied by a matching __tracer__ -- a monadic callback, that expresses the action of tracing of values of that particular type:
+
+```haskell
+trAddBlock :: Trace IO (TraceAddBlockEvent blk)
```
DavidEichmann, line 134: Typo? Is this meant to say `Tracer` i.e. missing an "r"?

DavidEichmann: I see we're adding a new `Trace` type and a new `Tracer` type and keeping around `Tracer` from `contra-tracer`. Is this a temporary naming scheme until we remove `contra-tracer`?

trace-dispatcher/doc/trace-dispatcher.md
```diff
@@ -0,0 +1,798 @@
...
+
+1. __trace filtering__ -- whether a trace reaches particular __trace-outs__ or not,
+2. __trace presentation__ -- which detail level is used during transformation of the __trace__ into __messages__.
+
+The __logging context__ of the trace is defined as follows:
+
+1. __trace filtering__ -- by __privacy__, __severity__ and __namespace__ context,
+2. __trace presentation__ -- by __detail level__ context.
+
+Severity an detail level can be configured.
```
DavidEichmann, line 195: "an" -> "and"

trace-dispatcher/doc/trace-dispatcher.md
```diff
@@ -0,0 +1,798 @@
...
+tracer   = setSeverity Notice trAddBlock
+
+-- Conditional annotation:
+tracer'' = withSeverity (\case
+                           IgnoreBlockOlderThanK{}          -> Warning
+                           IgnoreBlockAlreadyInVolatileDB{} -> Notice)
+                        trAddBlock
+```
+
+If the combinators are applied multiple times to a single trace, only the outermost application affects it -- the rest of them is ignored.
```
DavidEichmann, line 225: I wonder if in practice you even want to only partially override the severity i.e. keep the severity unchanged:

```haskell
tracer'' = withSeverity (\case
                           IgnoreBlockOlderThanK{}          -> Warning
                           IgnoreBlockAlreadyInVolatileDB{} -> Notice
                           InternalError                    -> error "I want to leave the severity unchanged")
                        trAddBlock
```

trace-dispatcher/doc/trace-dispatcher.md
```diff
@@ -0,0 +1,798 @@
...
+```haskell
+privately :: Trace m a -> Trace m a
+withPrivacy :: Monad m => (a -> Privacy) -> Trace m a -> Trace m a
+```
+
+`Public` is the default __privacy__, in the absence of privacy annotations.
+
+Trace privacy cannot be configured.
+
+See [Confidentiality and privacy filtering implementation](#Confidentiality-and-privacy -filtering-implementation) for a more full discussion of semantics.
```
DavidEichmann, line 259: I think you have an accidental space in `Confidentiality-and-privacy -filtering-implementation`

trace-dispatcher/doc/trace-dispatcher.md
```diff
@@ -0,0 +1,798 @@
...
+
+To further prevent occasional leaks of `Confidential` traces, all output from those traces is tagged with the `CONFIDENTIAL` keyword.
```
DavidEichmann, line 261: Ok great, but what if we do `withPrivacy (const Public) (privately someTracer)`. Have I now made all confidential messages public? If yes, this seems prone to programmer error.

trace-dispatcher/doc/trace-dispatcher.md
```diff
@@ -0,0 +1,798 @@
...
+
+### Frequency
+
+__Frequency filtering__ is yet another part of __trace filtering__, and represents an optional limit on the observable frequency of individual trace messages.
+
+Semantically, this is corresponds to a randomly-fair suppression of messages within a particular trace, when their moving-average frequency exceeds a given threshold parameter.
+
+The __frequency limiter__ itself emits a __suppression summary__ message under the following conditions:
+
+* when it message suppression begins, and
```
DavidEichmann, line 271: remove "it"

trace-dispatcher/doc/trace-dispatcher.md
```diff
@@ -0,0 +1,798 @@
...
+```
+
+The frequency filtering is intended to be applied to a subset of traces (those known to be noisy).  For this subset of traces the frequency limit can be configured.
+
+## Presentation
+### Formatting
+
+The `LogFormatting` typeclass is used to describe __trace presentation__ -- mapping __traces__ to __metrics__ and __messages__.
+
+* The `forMachine` method is used for a machine readable representation, which can varied through detail level.
```
DavidEichmann, line 297: "can be varied through"? or "can vary in"?

trace-dispatcher/doc/trace-dispatcher.md
```diff
@@ -0,0 +1,798 @@
...
+  It requires an implementation to be provided by the trace author.
+
+* the `forHuman` method shall represent the message in human readable form.
+  It's default implementation defers to `forMachine`.
+
+* the `asMetrics` method shall represent the message as `0` to `n` metrics.
+  It's default implementation assumes no metrics. If a text is given it is
+  appended as last element to the namespace.
```
DavidEichmann, line 305: I find the last sentence confusing. Perhaps something like:

```
Each metric can optionally specify a name space as a `Text`.
```

Though I wonder if there is a `type NameSapace = Text` somewhere :thinking:

trace-dispatcher/doc/trace-dispatcher.md
```diff
@@ -0,0 +1,798 @@
...
+The standard formatters transform a stream of messages of `a`, where `a` is an instance of `LogFormatter` to a stream of `FormattedMessages`.
+
+```haskell
+data FormattedMessage
+    = Human Text
+    | Machine Text
+    | Metrics [Metric]
+```
+
+`humanFormatter` takes a `Bool` argument, which tells if color codes for the standard output __trace-out__ shall be inserted, and an argument which is the app name, which gets prepended to the namespace, while the `machineFormatter` has as arguments the desired detail level and as well the application name.  `metricsFormatter` takes no extra arguments:
```
DavidEichmann, line 332: Style: you could probably move most of this paragraph into haddocks for the arguments in the code snippet bellow.

trace-dispatcher/src/Cardano/Logging/Formatter.hs
```diff
@@ -0,0 +1,262 @@
...
+                    , toHostname  = hn
+                    , toThreadId  = (pack . show) thid
+                  }
+          T.traceWith tr ( nlc
+                         , Just ctrl
+                         , FormattedForwarder to)
+
+-- | Format this trace for human readability
+-- The boolean value tells, if this representation is for the console and should be colored
+-- The text argument gives the application name which is prepended to the namespace
```
DavidEichmann, line 113: I think it's better style in general to put the comments on the actual arguments rather than in the function description: https://haskell-haddock.readthedocs.io/en/latest/markup.html#function-arguments

trace-resources/README.md
jutaro: Okay

trace-dispatcher/CHANGELOG.md
jutaro: Okay

trace-dispatcher/doc/trace-dispatcher.md
jutaro: Yes, all traces are keyed by a list of strings/text.

trace-dispatcher/doc/trace-dispatcher.md
jutaro: No, clarifying that to be tracing system, in the sense of a library, infrastructure component.

trace-dispatcher/doc/trace-dispatcher.md
jutaro: We build on the arrow based contra-tracer library. Since the iohk-monitoring framework depends on the non-arrow based contra-tracer framework, we integrated contra-tracer-0.1.. into the source tree.
This should become a dependency later.

trace-dispatcher/doc/trace-dispatcher.md
jutaro: In new binding to the node there has been no need for a partial binding. If needed you can use setSeverity for individual messages.

trace-dispatcher/doc/trace-dispatcher.md
jutaro: Yes you have, but that seems to me to fall more in the category of "bad will", then to be in any way prone to an error.

trace-dispatcher/doc/trace-dispatcher.md
jutaro: Yes: `type Namespace = [Text]`

trace-dispatcher/doc/trace-dispatcher.md
jutaro: Okay

trace-dispatcher/src/Cardano/Logging/Formatter.hs
jutaro: I've not used this style so far, but I will consider it for the future.

trace-dispatcher/doc/trace-dispatcher.md
```diff
@@ -0,0 +1,798 @@
...
+
+> We could have used the (`Type` * `Constructor`) pair, which is a more technical approach.  Problems with that:
+> 1. __synthetic traces__ exist.
+> 2. Developer-provided type/constructor names are not necessarily ideal from user standpoint.
+
+DECISION: there is value in maintaining a user-friendly trace message namespace.
+
+### Decide inline trace type annotation with trace function 2
+
+DECISION: we use `traceWith` in the library code and `traceNamed` in th dispatcher.
```
DavidEichmann, line 734: typo: "th" -> "the"

DavidEichmann (changes requested):

DavidEichmann:
I've gone through `trace-dispatcher.md` in detail now. I'll come back and look at the implementation soon.

trace-dispatcher/doc/trace-dispatcher.md
```diff
@@ -0,0 +1,798 @@
...
+```
+
+To route one trace to multiple tracers simultaneously we use the fact that Tracer is a `Semigroup` and then use `<>`, or `mconcat` for lists of tracers:
+
+```haskell
+(<>) :: Monoid m => m -> m -> m
+mconcat :: Monoid m => [m] -> m
+```
+
+In the third example we unite two traces to one tracer, for which we trivially use the same tracer on the right side.
```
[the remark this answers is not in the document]
jutaro: It means the example below, hope it then becomes clear.

trace-dispatcher/doc/trace-dispatcher.md
```diff
@@ -0,0 +1,798 @@
...
+> I would like to find a function `foldTrace`, that omits the MVar and can thus be called pure.  Help is appreciated.
+
+## Dispatcher routing toolkit
+
+During definition of the __trace dispatcher__, it is sometimes useful to have a number of functions to route them.
+
+To send the message of a trace to different tracers depending on some criteria use the following function:
+
+```haskell
+routingTrace :: Monad m => (a -> Trace m a) -> Trace m a
```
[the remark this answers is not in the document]
jutaro: The link is not working for me, but you are right. This document has two authors and the trace tracer duality section has been introduced by Serge, and I don't really get the meaning.
Technically Tracer come from trace-dispatcher, and Trace is the version with added functionality trace-dispatcher provides.
In my prosa (more or less) trace is a stream of messages, and tracer outputs this stream to some medium.
I will discuss this point with Serge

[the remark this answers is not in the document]
jutaro: The link is not working for me, but you are right. This document has two authors and the trace tracer duality section has been introduced by Serge, and I don't really get the meaning.
Technically Tracer come from trace-dispatcher, and Trace is the version with added functionality trace-dispatcher provides.
In my prosa (more or less) trace is a stream of messages, and tracer outputs this stream to some medium.
I will discuss this point with Serge

trace-dispatcher/doc/trace-dispatcher.md
```diff
@@ -0,0 +1,798 @@
...
+     -- | Options specific to a certain namespace
+    tcOptions            :: Map.Map Namespace [ConfigOption]
+     -- | Options for trace-forwarder    
+  , tcForwarder          :: RemoteAddr
+  , tcForwarderQueueSize :: Int
+}
+```
+
+If the configuration file is in Yaml format, the following entry means, that by default
+all messages with Info or higher Priority or higher are shown:
```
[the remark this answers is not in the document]
jutaro: Okay

trace-dispatcher/doc/trace-dispatcher.md
```diff
@@ -0,0 +1,798 @@
...
+data TraceConfig = TraceConfig {
+     -- | Options specific to a certain namespace
+    tcOptions            :: Map.Map Namespace [ConfigOption]
+     -- | Options for trace-forwarder    
+  , tcForwarder          :: RemoteAddr
+  , tcForwarderQueueSize :: Int
+}
+```
+
+If the configuration file is in Yaml format, the following entry means, that by default
```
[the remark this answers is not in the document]
jutaro: e.g. Json is possible as well

trace-dispatcher/doc/trace-dispatcher.md
```diff
@@ -0,0 +1,798 @@
...
+## Documentation
+
+The self-documentation features of `trace-dispatcher` are provided by a combination of:
+
+* __documentation annotations__, expressed by the `Documented a` type, that carry a list of per- __message__ description (of `DocMsg` type), and
+* a special __dispatcher__ execution mode that emits documentation for all annotated traces, using the __tracer namespace__ to guide the document structure.
+
+The per- __message__ `DocMsg` objects combine:
+
+* __trace prototypes__ -- a stubbed __message__ invocation,
```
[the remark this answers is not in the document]
jutaro: Am I correct that a "trace" is the second argument to "traceWith": Yes
And also the a parameter in Documented a: No
A `Trace a` corresponds to a `Documented a`, and the a fixes the possible messages of this Trace.

trace-dispatcher/doc/trace-dispatcher.md
```diff
@@ -0,0 +1,798 @@
...
+  , dmMetricsMD :: [(Text, Text)]
+  , dmMarkdown  :: Text
+}
+```
+
+### Discuss impact of missing documentation entries
+
+What is the worst case impact of a missing `DocMsg`?
+
+How can we reduce that impact?
```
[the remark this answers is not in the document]
jutaro: That looks interesting, will try this approach later

trace-dispatcher/doc/trace-dispatcher.md
jutaro: Okay

Jimbo4350 (approved):

trace-dispatcher/doc/trace-dispatcher.md
deepfire: - "data plane" refers to the code paths that are proportional to tracing
- "control plane" refers to the code paths that are proportional to (re-)configuration

The idea behind this conceptual separation is to ensure that configuration flexibility has minimal impact on the common data paths.

trace-dispatcher/doc/trace-dispatcher.md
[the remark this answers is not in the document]
deepfire: @DavidEichmann, conceptually:

- _tracer_ is a callback expecting messages of a particular type (i.e. a function object)
- _trace_ is a stream of messages of a particular type, emitted by a particular point in the program (i.e. a conceptual entity -- a potentially infinite sequence of messages)

trace-dispatcher/doc/trace-dispatcher.md
deepfire: The concern seems reasonable.

I wonder if it will make sense to detect that we override a private context to be non-private -- and raise an error.

We don't really need `withPrivacy` to be dynamic -- what does our implementation experience tell us @jutaro ?

What I'm getting at, is that we could restrict the API like this:

- default to `Public` (which we already do)
- export `privately` with original semantics
- rename `publicly` to `unsafePublicly`
- define a new `publicly` to fail if the context is `Private`

I think that will help reducing occasional mistakes, @jutaro ?

trace-dispatcher/doc/trace-dispatcher.md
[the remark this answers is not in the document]
deepfire: @DavidEichmann, that's a useful improvement, definitely.  Jurgen said he'll look into experimenting with  this scheme next week.

However, I'm not sure we should delay the integration for that.  

Would you consider it a blocker?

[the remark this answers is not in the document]
DavidEichmann: Plenty ppl don't like arrow notation and perhaps there is some issue I haven't thought of here, so I'll let you decide if it's worth doing.

> Would you consider it a blocker?

Not at all. Can certainly be left as future work.

trace-dispatcher/doc/trace-dispatcher.md
DavidEichmann: I'm still slightly confused here (though this is likely because I've only read through documentation and not the code). Why is this type named `Trace` instead of `Tracer`? Is this just to avoid a name clash with `Tracer` from `contra-tracer`?

trace-dispatcher/doc/trace-dispatcher.md
DavidEichmann: > define a new publicly to fail if the context is Private

Would you proposed function only allow you to make public things public then (a noop)? In that case it only serves as an assertion function. Renaming to `unsafePublicly` might be sufficient: making things private is always safe but making things public dangerous. Perhaps it'll be more clear as the library get's used.

trace-dispatcher/doc/trace-dispatcher.md
DavidEichmann: @jutaro I meant something like this:

```haskell
-- | Format this trace for human readability
humanFormatter :: (LogFormatting a, MonadIO m)
  => Bool
  -- ^ True if this representation is for the console and should be colored.
  -> Text
  -- ^ The application name which will be prepended to the namespace.
  -> Trace m FormattedMessage
  -> m (Trace m a)
```

That way it's more clear which argument you're talking about. But I'm not going to block you on this :-)

trace-dispatcher/doc/trace-dispatcher.md
DavidEichmann: Let's continue this thread here: https://github.com/input-output-hk/cardano-node/pull/3073#discussion_r704604233

trace-dispatcher/doc/trace-dispatcher.md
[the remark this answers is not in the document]
DavidEichmann: Related thread: https://github.com/input-output-hk/cardano-node/pull/3073#discussion_r709344734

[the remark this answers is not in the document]
DavidEichmann: @deepfire you're definitions make sense to me. The confusing part is that `Trace` is a "tracer" and not a "trace" (am I mistaken?). It even says it in the documentation ("The **Trace** carries the underlying **tracer**"):

```haskell
-- | The Trace carries the underlying tracer Tracer from the contra-tracer package.
--   It adds a 'LoggingContext' and maybe a 'TraceControl' to every message.
newtype Trace m a = Trace
  {unpackTrace :: Tracer m (LoggingContext, Maybe TraceControl, a)}
```

Or perhaps you're representing a "trace" with a "tracer"? :thinking:

trace-dispatcher/doc/trace-dispatcher.md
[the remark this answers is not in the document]
DavidEichmann: Recently someone pointed out that JSON is a subset of yaml :eyes:

trace-dispatcher/doc/trace-dispatcher.md
deepfire: We've decided to just remove `publicly` and `privately` -- as indeed they do not appear to be necessary for the node integration.

trace-dispatcher/doc/trace-dispatcher.md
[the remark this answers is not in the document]
deepfire: Right, that's the confusion between the `Trace` type and "trace" notion.

The `Trace` type is just a specialised `Tracer`, indeed.

trace-dispatcher/doc/trace-dispatcher.md
deepfire: It's because we augment the `Tracer` type to carry context, and call it `Trace`, which is a name choice that was carried over from the previous system.

This `Trace` is not the same thing as the conceptual "trace", which is an abstract entity.

It's simplest to think of `Trace` as "just a minor variation of `Tracer`".

trace-resources/src/Cardano/Logging/Resources/Darwin.hsc
```diff
@@ -0,0 +1,114 @@
...
+getMemoryInfo :: ProcessID -> IO MachTaskBasicInfo
+getMemoryInfo pid =
+  allocaBytes 128 $ \ptr -> do
+    res <- c_get_process_memory_info ptr (fromIntegral pid)
+    if res <= 0
+      then do
+        putStrLn $ "c_get_process_memory_info: failure returned: " ++ (show res)
+        return $ MachTaskBasicInfo 0 0 0 (TIME_VALUE_T 0 0) (TIME_VALUE_T 0 0) 0 0
+      else
+        peek ptr
```
coot, line 86: One could also use `throwIfNeg`, though one needs to account that it checks if the value is `< 0` rather than `<= 0`  (maybe this is not even needed).
```suggestion
    rthrowIfNeg_ (\res -> "c_get_process_memory_info: failure returned: " ++ show (pred res))
                  (succ <$> c_get_process_memory_info ptr (fromIntegral pid))
    peek ptr
 
```

trace-resources/src/Cardano/Logging/Resources/os-support-darwin.c
```diff
@@ -0,0 +1,272 @@
...
+    task_t task = MACH_PORT_NULL;
+    if (task_for_pid(current_task(), pid, &task) != KERN_SUCCESS) {
+        return -2;
+    }
+    struct mach_task_basic_info t_info;
+    mach_msg_type_number_t t_info_count = MACH_TASK_BASIC_INFO_COUNT;
+    if (task_info(task, MACH_TASK_BASIC_INFO, (task_info_t)counters, &t_info_count) != KERN_SUCCESS) {
+        return -1;
+    }
+    return 1;
```
coot, line 46: In the unix it is more common to return `0` on success.  Then `throwIfNeg` from base can be used directly.

trace-resources/trace-resources.cabal
```diff
@@ -0,0 +1,76 @@
...
+version:              0.1.0.0
+
+synopsis:             Package for tracing resources for linux, mac and windows
+author:               Juergen Nicklisch-Franken
+maintainer:           operations@iohk.io
+copyright:            2021 IOHK
+extra-source-files:   CHANGELOG.md
+                      README.md
+                      src/Cardano/Logging/Resources/os-support-darwin.h
+                      src/Cardano/Logging/Resources/os-support-win.h
```
coot, line 12: It's more common to put these files in a top level directory `./include` and add
```
include-dirs: include
```

trace-resources/trace-resources.cabal
```diff
@@ -0,0 +1,76 @@
...
+                       -Wredundant-constraints
+                       -Wmissing-export-lists
+                       -Wno-incomplete-patterns
+
+  if os(linux)
+    exposed-modules:   Cardano.Logging.Resources.Linux
+  if os(windows)
+    exposed-modules:   Cardano.Logging.Resources.Windows
+    c-sources:         src/Cardano/Logging/Resources/os-support-win.c
+    include-dirs:      src/Cardano/Logging/Resources/
```
coot, line 46: It's more common to use `cbits` and `include` directories for `*.c` and header files.

coot (changes requested):

trace-dispatcher/doc/trace-dispatcher.md
```diff
@@ -0,0 +1,803 @@
...
+
+## Dispatcher routing toolkit
+
+During definition of the __trace dispatcher__, it is sometimes useful to have a number of functions to route them.
+
+To send the message of a trace to different tracers depending on some criteria use the following function:
+
+```haskell
+routingTrace :: Monad m => (a -> Trace m a) -> Trace m a
+let resTrace = routingTrace routingf (tracer1 <> tracer2)
```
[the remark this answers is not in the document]
jutaro: No, unfortunately not. The type is wrong. It must be routingTrace :: Monad m => (a -> m (Trace m a)) -> Trace m a -> m (Trace m a).
That is not a nice part of the library, as the typechecker doesn't check this constraint: The second argument must mappend all tracers used in the routing trace function to one tracer.

trace-dispatcher/doc/trace-dispatcher.md
```diff
@@ -0,0 +1,803 @@
...
+}
+```
+
+If the configuration file is in Yaml format, the following entry means, that by default
+all messages with Info or higher Priority are shown:
+
+```yaml
+TraceOptionSeverity:
+  - ns: ''
+    severity: InfoF
```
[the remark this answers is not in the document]
jutaro: I have two different severities. One is the severity that messages carry, and one is the Severity for filtering, which have the additional F. The filter severity has the additional SilenceF constructor.

trace-dispatcher/trace-dispatcher.cabal
```diff
@@ -0,0 +1,151 @@
...
+                      , bytestring
+                      , cborg
+                      , containers
+                      , contra-tracer
+                      , ekg
+                      , ekg-core
+                      , ekg-forward
+                      , hostname
+                      , network
+                      , ouroboros-network-framework
```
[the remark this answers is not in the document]
jutaro: The dependency comes from the integration of trace-forward in Cardano.Logging.Tracer.Forward done from Denis.

trace-dispatcher/src/Control/Tracer/Arrow.hs
```diff
@@ -0,0 +1,90 @@
+{-|
+Module      : Control.TracerA.Arrow
```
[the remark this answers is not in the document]
jutaro: I wrote this in the README:
"We integrated contra-tracer-0.1.0.0 into the source tree, because the
iohk-monitoring framework depends on the non-arrow based contra-tracer framework.
This should become a dependency later".
So after old tracing will be removed it will become a dependeny.

trace-dispatcher/test/Cardano/Logging/Test/Tracer.hs
```diff
@@ -0,0 +1,20 @@
...
+
+import           Data.IORef
+import           Control.Monad.IO.Class
+
+import           Cardano.Logging
+
+testTracer :: MonadIO m
+  => IORef [FormattedMessage]
+  -> m (Trace m FormattedMessage)
+testTracer ioRef = liftIO $ do
```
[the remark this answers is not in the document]
jutaro: Okay

trace-dispatcher/src/Cardano/Logging/Types.hs
```diff
@@ -0,0 +1,346 @@
...
+  -- | To which backend to pass
+  --   Default is [EKGBackend, Forwarder, Stdout HumanFormatColoured]
+  | CoBackend [BackendConfig]
+  -- | Construct a limiter with name (Text) and limiting to the Double,
+  -- which represents frequency in number of messages per second
+  | CoLimiter Text Double
+  deriving (Eq, Ord, Show)
+
+newtype RemoteAddr
+  = LocalSocket FilePath
```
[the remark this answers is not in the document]
jutaro: The type of address has changed over time. I will rename to ForwarderAddr, to be more specific.

trace-dispatcher/src/Cardano/Logging/Types.hs
```diff
@@ -0,0 +1,346 @@
...
+  , ldBackends   :: [(BackendConfig, FormattedMessage)]
+  , ldFiltered   :: [SeverityF]
+  , ldLimiter    :: [(Text, Double)]
+} deriving(Eq, Show)
+
+emptyLogDoc :: Text -> [(Namespace, Text)] -> LogDoc
+emptyLogDoc d m = LogDoc d (Map.fromList m) [] [] [] [] [] [] []
+
+-- | Type for a Fold
+newtype Folding a b = Folding b
```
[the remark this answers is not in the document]
jutaro: It is used to make this function safe and typeable:
-- | Folds the cata function with acc over a.
-- Uses an MVar to store the state
foldTraceM
  :: forall a acc m . (MonadUnliftIO m)
  => (acc -> LoggingContext -> a -> acc)
  -> acc
  -> Trace m (Folding a acc)
  -> m (Trace m a)

trace-dispatcher/src/Cardano/Logging/Tracer/Standard.hs
```diff
@@ -0,0 +1,94 @@
...
+    msg   <- readChan outChan
+    state <- readIORef stateRef
+    case stTarget state of
+        LogFile f -> do
+                        TIO.appendFile f msg
+                        TIO.appendFile f "\n"
+        LogStdout -> do
+                        TIO.putStrLn msg
+                        hFlush stdout
+  modifyIORef stateRef (\ st ->
```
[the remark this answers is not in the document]
jutaro: Sounds scary. But what kind of async exception should happen here?

trace-dispatcher/src/Cardano/Logging/Tracer/Standard.hs
```diff
@@ -0,0 +1,94 @@
...
+    st {stRunning = Just (inChan, outChan, threadId)})
+
+-- | Converts a curried function to a function on a triple.
+uncurry3 :: (a -> b -> c -> d) -> (a, b, c) -> d
+uncurry3 f ~(a,b,c) = f a b c
```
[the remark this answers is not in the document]
jutaro: Okay, changed it.

trace-dispatcher/src/Cardano/Logging/Tracer/Forward.hs
```diff
@@ -0,0 +1,205 @@
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE DeriveGeneric       #-}
+{-# LANGUAGE FlexibleInstances   #-}
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE PackageImports      #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving  #-}
+
+{-# OPTIONS_GHC -Wno-orphans #-}
```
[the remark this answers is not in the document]
jutaro: This is again the trace-forward integration, which should not penetrate the whole of trace-dispatcher.

trace-dispatcher/doc/trace-dispatcher.md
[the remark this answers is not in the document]
deepfire: @jutaro, can we change the naming -- to give the user an F-free name -- and use something suffixed internally?

trace-dispatcher/src/Cardano/Logging/Tracer/Forward.hs
```diff
@@ -0,0 +1,205 @@
...
+         { miniProtocolNum    = MiniProtocolNum num
+         , miniProtocolLimits = MiniProtocolLimits { maximumIngressQueue = maxBound }
+         , miniProtocolRun    = prot
+         }
+      | (prot, num) <- protocols
+      ]
+
+-- | Converts a curried function to a function on a triple.
+uncurry3 :: (a -> b -> c -> d) -> ((a, b, c) -> d)
+uncurry3 f ~(a,b,c) = f a b c
```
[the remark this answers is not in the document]
jutaro: You see everything, do you? I moved it to utils.

trace-dispatcher/src/Cardano/Logging/Tracer/EKG.hs
```diff
@@ -0,0 +1,84 @@
...
+
+ekgTracer :: MonadIO m => Either Metrics.Store Server-> m (Trace m FormattedMessage)
+ekgTracer storeOrServer = liftIO $ do
+    rgsGauges   <- newIORef Map.empty
+    rgsLabels   <- newIORef Map.empty
+    rgsCounters <- newIORef Map.empty
+    pure $ Trace $ T.arrow $ T.emit $
+      output rgsGauges rgsLabels rgsCounters
+  where
+    output rgsGauges rgsLabels rgsCounters
```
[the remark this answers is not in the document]
jutaro: Okay

trace-dispatcher/src/Cardano/Logging/Tracer/Composed.hs
```diff
@@ -0,0 +1,120 @@
...
+  -> (evt -> [Text])
+  -> (evt -> SeverityS)
+  -> (evt -> Privacy)
+  -> IO (Trace IO evt)
+mkCardanoTracer trStdout trForward mbTrEkg name namesFor severityFor privacyFor =
+    mkCardanoTracer' trStdout trForward mbTrEkg name namesFor severityFor
+        privacyFor noHook
+  where
+    noHook :: Trace IO evt -> IO (Trace IO evt)
+    noHook tr = pure tr
```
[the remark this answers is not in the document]
jutaro: Didn't know that. Changed it

trace-dispatcher/src/Cardano/Logging/Trace.hs
```diff
@@ -0,0 +1,261 @@
...
+--- given one
+filterTraceBySeverity :: Monad m =>
+     Maybe SeverityF
+  -> Trace m a
+  -> Trace m a
+filterTraceBySeverity (Just minSeverity) =
+    filterTrace $
+      \case
+        (_lc, Just _, _a)  -> True
+        (lc, _, _e)                ->
```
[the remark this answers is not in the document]
jutaro: Okay

trace-dispatcher/src/Cardano/Logging/Trace.hs
```diff
@@ -0,0 +1,261 @@
...
+  ref <- liftIO (newMVar initial)
+  let trr = mkTracer ref
+  pure $ Trace (T.Tracer trr)
+ where
+    mkTracer ref = T.emit $
+      \case
+        (lc, Nothing, v) -> do
+          x' <- modifyMVar ref $ \x ->
+            let ! accu = cata x lc v
+            in pure $ join (,) accu
```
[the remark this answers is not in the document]
jutaro: Agree, changed

trace-dispatcher/src/Cardano/Logging/FrequencyLimiter.hs
```diff
@@ -0,0 +1,190 @@
...
+limitFrequency
+  :: forall a m . (MonadIO m, MonadUnliftIO m)
+  => Double   -- messages per second
+  -> Text     -- name of this limiter
+  -> Trace m a -- the limited trace
+  -> Trace m LimitingMessage -- the limiters messages
+  -> m (Trace m a) -- the original trace
+limitFrequency thresholdFrequency limiterName vtracer ltracer = do
+    timeNow <- systemTimeToSeconds <$> liftIO getSystemTime
+--    trace ("limitFrequency called " <> unpack limiterName) $ pure ()
```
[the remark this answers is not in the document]
jutaro: Removed it at two places!

trace-dispatcher/src/Cardano/Logging/FrequencyLimiter.hs
```diff
@@ -0,0 +1,190 @@
...
+-- Finally it sends a StopLimiting message on the ltracer and traces all
+-- messages on the vtracer again.
+limitFrequency
+  :: forall a m . (MonadIO m, MonadUnliftIO m)
+  => Double   -- messages per second
+  -> Text     -- name of this limiter
+  -> Trace m a -- the limited trace
+  -> Trace m LimitingMessage -- the limiters messages
+  -> m (Trace m a) -- the original trace
+limitFrequency thresholdFrequency limiterName vtracer ltracer = do
```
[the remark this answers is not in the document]
jutaro: Yet no test for this. Should look into this.

trace-dispatcher/src/Cardano/Logging/Formatter.hs
```diff
@@ -0,0 +1,262 @@
...
+
+-- | Format this trace as metrics
+metricsFormatter
+  :: forall a m . (LogFormatting a, MonadIO m)
+  => Text
+  -> Trace m FormattedMessage
+  -> m (Trace m a)
+metricsFormatter application (Trace tr) = do
+  let trr = mkTracer
+  pure $ Trace (T.arrow trr)
```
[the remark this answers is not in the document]
jutaro: Yes :) changed this

trace-dispatcher/src/Cardano/Logging/Configuration.hs
```diff
@@ -0,0 +1,485 @@
...
+                        liftIO $ writeIORef ref $ Right val
+                        Trace tt <- withConfig (Just val) tr
+                        T.traceWith tt (lc, Just Optimize, m)
+            _      -> let decidingDict =
+                            foldl
+                              (\acc e -> Map.insertWith (+) e (1 :: Int) acc)
+                              Map.empty
+                              (Map.elems cmap)
+                          (mostCommon, _) = maximumBy
+                                              (\(_, n') (_, m') -> compare n' m')
```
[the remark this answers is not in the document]
deepfire: To clarify:

> ``maximumBy (compare `on` snd)``

trace-dispatcher/src/Cardano/Logging/FrequencyLimiter.hs
```diff
@@ -0,0 +1,190 @@
...
+      let spendReward      = min 0.5 (max (-0.5) normaSpendReward)
+      let newBudget        = min 1.0 (max (-1.0) (spendReward + frBudget))
+      -- trace ("elapsedTime " ++ show elapsedTime
+      --        ++ " thresholdPeriod " ++ show thresholdPeriod
+      --        ++ " rawSpendReward " ++ show rawSpendReward
+      --        ++ " normaSpendReward "  ++ show normaSpendReward
+      --        ++ " spendReward "       ++ show spendReward
+      --        ++ " newBudget "       ++ show newBudget $
+      case frActive of
+        Nothing -> -- not active
```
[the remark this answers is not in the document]
jutaro: Okay

trace-dispatcher/src/Cardano/Logging/Formatter.hs
```diff
@@ -0,0 +1,262 @@
...
+ where
+    mkTracer = T.emit $
+      \ case
+        (lc, Nothing, v) ->
+          let metrics = asMetrics v
+          in T.traceWith tr (lc { lcNamespace = application : lcNamespace lc}
+                                , Nothing
+                                , FormattedMetrics metrics)
```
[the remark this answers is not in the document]
jutaro: Done

trace-dispatcher/src/Cardano/Logging/Formatter.hs
```diff
@@ -0,0 +1,262 @@
...
+        (lc, Just ctrl, v) ->
+          let metrics = asMetrics v
+          in T.traceWith tr (lc { lcNamespace = application : lcNamespace lc}
+                                , Just ctrl
+                                , FormattedMetrics metrics)
```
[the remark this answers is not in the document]
jutaro: Done

trace-dispatcher/src/Cardano/Logging/Formatter.hs
```diff
@@ -0,0 +1,262 @@
...
+        (lc, Nothing, v) -> do
+          let fh = forHuman v
+          text <- liftIO $ formatContextHuman withColor hn application lc fh
+          T.traceWith tr (lc { lcNamespace = application : lcNamespace lc}
+                             , Nothing
+                             , FormattedHuman withColor text)
+        (lc, Just ctrl, _v) -> do
+          T.traceWith tr (lc { lcNamespace = application : lcNamespace lc}
+                             , Just ctrl
+                             , FormattedHuman withColor "")
```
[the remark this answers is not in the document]
jutaro: Okay

trace-dispatcher/src/Cardano/Logging/Formatter.hs
```diff
@@ -0,0 +1,262 @@
...
+  pure $ Trace (T.arrow trr)
+ where
+    mkTracer hn = T.emit $
+      \ case
+        (lc, Nothing, v) -> do
+          let fh = forHuman v
+          text <- liftIO $ formatContextHuman withColor hn application lc fh
+          T.traceWith tr (lc { lcNamespace = application : lcNamespace lc}
+                             , Nothing
+                             , FormattedHuman withColor text)
```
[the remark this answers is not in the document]
jutaro: Okay

trace-dispatcher/src/Cardano/Logging/Formatter.hs
```diff
@@ -0,0 +1,262 @@
...
+ where
+    mkTracer hn = T.emit $
+      \case
+        (lc, Nothing, v) -> do
+          let detailLevel = fromMaybe DNormal (lcDetails lc)
+          obj <- liftIO $ formatContextMachine hn application lc (forMachine detailLevel v)
+          T.traceWith tr (lc { lcNamespace = application : lcNamespace lc}
+                             , Nothing
+                             , FormattedMachine (decodeUtf8 (BS.toStrict
+                                    (AE.encodingToLazyByteString obj))))
```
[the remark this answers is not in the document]
jutaro: Okay

trace-dispatcher/src/Cardano/Logging/Formatter.hs
```diff
@@ -0,0 +1,262 @@
...
+        (lc, Just c, _v) -> do
+          T.traceWith tr (lc { lcNamespace = application : lcNamespace lc}
+                             , Just c
+                             , FormattedMachine "")
```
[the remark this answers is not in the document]
jutaro: Okay

trace-dispatcher/src/Cardano/Logging/Configuration.hs
```diff
@@ -0,0 +1,485 @@
...
+            Just v  -> do
+              if v == val
+                then do
+                  Trace tt <- withConfig (Just val) tr
+                  T.traceWith tt (lc, Just (Config c), m)
+                else error $ "Inconsistent trace configuration with context "
+                                  ++ show (lcNamespace lc)
+        Right _val -> error $ "Trace not reset before reconfiguration (1)"
+                            ++ show (lcNamespace lc)
+        Left (_cmap, Just _v) -> error $ "Trace not reset before reconfiguration (2)"
```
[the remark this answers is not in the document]
jutaro: No, these are errors that can never pop up in production

trace-dispatcher/src/Cardano/Logging/Configuration.hs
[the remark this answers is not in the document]
jutaro: Would like to keep it? Not used to use on

trace-resources/src/Cardano/Logging/Resources/Darwin.hsc
jutaro: As I have no Mac it's a big pain to do changes here. I have to wait for the mac and windows builds on github to see if it compiles at all. So someone with a mac could try this?

trace-resources/src/Cardano/Logging/Resources/os-support-darwin.c
jutaro: See above

trace-resources/trace-resources.cabal
jutaro: See above

trace-resources/trace-resources.cabal
jutaro: See above

trace-dispatcher/doc/trace-dispatcher.md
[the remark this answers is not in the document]
jutaro: Not sure how to do it, as user see filter severities in the configuration and message severities in log files. So we can do some hack, but maybe the result is more confusing then helpful.

trace-dispatcher/src/Control/Tracer/Arrow.hs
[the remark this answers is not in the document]
coot: That's ok then.

trace-dispatcher/src/Cardano/Logging/Configuration.hs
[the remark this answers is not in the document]
coot: both are equally readable, so I am fine either way.

trace-resources/src/Cardano/Logging/Resources/Darwin.hsc
coot: I don't have a mac, but this should be the same on linux and macos.  You could try it out first in `Cardano.Logging.Resources.Linux`.

coot: Do you use `nix`? Then there's a command to cross compile it on linux for macos.  @deepfire might know the command.  I'm sometimes cross compiling from linux to windows using `haskell.nix` which is available in our repos.

jutaro: > You could try it out first in `Cardano.Logging.Resources.Linux`.
Linux doesn't have a C part here. 
> Do you use `nix`? Then there's a command to cross compile it on linux for macos. @deepfire might know the command. 
I already spend some time on this without success, and of course asked Serge. Seems these tools are currently not working.
But I will work on making changes to the c code as suggested.

> One could also use `throwIfNeg`, though one needs to account that it checks if the value is `< 0` rather than `<= 0` (maybe this is not even needed).
I don't even know where this function comes from, and don't see the advantage here anyway

trace-resources/src/Cardano/Logging/Resources/os-support-darwin.c
jutaro: Ok

trace-resources/trace-resources.cabal
jutaro: > It's more common to put these files in a top level directory `./include` and add

I searched in repos: cardano-node, cardano-base, ouroboros-network and cardano-ledger-specs, and could not find one example of this practice. So it is surely common in C programming, it doesn't seem common practice in this project. Do I need to do this changes?

trace-resources/trace-resources.cabal
jutaro: Yes, cardano-crypto-praos uses the cbits directory, but no include.

trace-dispatcher/doc/trace-dispatcher.md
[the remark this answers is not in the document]
jutaro: Made the filtering severity a newtype of Maybe Severity, where Nothing denotes Silence.

trace-resources/src/Cardano/Logging/Resources/Darwin.hsc
deepfire: I think we've addressed this with the suggested code change, @coot

trace-dispatcher/doc/trace-dispatcher.md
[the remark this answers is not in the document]
deepfire: Looks a lot nicer now!

trace-resources/trace-resources.cabal
jutaro: Done the changes

trace-resources/trace-resources.cabal
jutaro: Done

trace-resources/src/Cardano/Logging/Resources/Darwin.hsc
jutaro: Done

coot: I still see `res <= 0` above.  The [throwIfNeg_](https://hackage.haskell.org/package/base-4.15.0.0/docs/Foreign-Marshal-Error.html#v:throwIfNeg_) is provided by `base` package.

coot: > I don't even know where this function comes from, and don't see the advantage here anyway

It's a common C practice to use `< 0` for errors and `0` for no-error, keeping this, will be easier for an (independent) reviewer.

jutaro: Changed this for Windows as well. Funny enough the underlying Windows call returns 1 for success and 0 for failure :)

trace-dispatcher/src/Cardano/Logging/Tracer/Standard.hs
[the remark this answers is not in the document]
coot: User hits `Ctrl-c`, lack of memory, or what more common killing a thread with `throwTo` or `cancel`. See [AsyncException](https://hackage.haskell.org/package/base-4.15.0.0/docs/Control-Exception-Base.html#t:AsyncException).

trace-resources/src/Cardano/Logging/Resources/Darwin.hsc
coot: yes I think it's more common for Windows to use `0` for failure, though I am not sure how consistent they are.

trace-dispatcher/src/Cardano/Logging/Tracer/Forward.hs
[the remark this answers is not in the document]
coot: `SeverityS`  is defined [here](https://github.com/input-output-hk/cardano-node/pull/3073/files/7b3d8403ea947ddaff3393c932e7ffd538cb9645#diff-fff977c0d5f3c2c6a8e7c4f961c5ebcabc7e9a89c637433524c1e453f8a59b64R159), so why not to move `Generic` and `Serialisable` instance there?

trace-dispatcher/src/Cardano/Logging/FrequencyLimiter.hs
```diff
@@ -0,0 +1,190 @@
...
+  -> m (Trace m a) -- the original trace
+limitFrequency thresholdFrequency limiterName vtracer ltracer = do
+    timeNow <- systemTimeToSeconds <$> liftIO getSystemTime
+--    trace ("limitFrequency called " <> unpack limiterName) $ pure ()
+    foldMTraceM
+      (checkLimiting (1.0 / thresholdFrequency))
+      (FrequencyRec Nothing timeNow 0.0 0.0 Nothing)
+      (Trace $ T.contramap unfoldTrace (unpackTrace (filterTraceMaybe vtracer)))
+  where
+    checkLimiting :: Double -> FrequencyRec a -> LoggingContext -> a -> m (FrequencyRec a)
```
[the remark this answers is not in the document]
jutaro: Did a revision of this, please check if it clearer now

trace-dispatcher/src/Cardano/Logging/FrequencyLimiter.hs
```diff
@@ -0,0 +1,190 @@
...
+    checkLimiting thresholdPeriod fs@FrequencyRec {..} lc message = do
+      -- trace ("Limiter " <> unpack limiterName <> " receives " <> show (lcNamespace lc))
+      --         $ pure ()
+      timeNow <- liftIO $ systemTimeToSeconds <$> getSystemTime
+      let elapsedTime      = timeNow - frLastTime
+      let rawSpendReward   = elapsedTime - thresholdPeriod
+                                    -- negative if shorter, positive if longer
+      let normaSpendReward = rawSpendReward * thresholdFrequency -- TODO not really normalized
```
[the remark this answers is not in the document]
jutaro: Did a revision of this, please check if it clearer now

trace-dispatcher/src/Cardano/Logging/Tracer/Forward.hs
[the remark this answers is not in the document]
jutaro: Done

trace-dispatcher/src/Cardano/Logging/Tracer/Standard.hs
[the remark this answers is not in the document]
jutaro: @coot What code change would you suggest?

trace-dispatcher/src/Cardano/Logging/Tracer/EKG.hs
```diff
@@ -0,0 +1,94 @@
...
+import           Data.Text (pack, Text)
+import qualified System.Metrics as Metrics
+import qualified System.Metrics.Counter as Counter
+import qualified System.Metrics.Gauge as Gauge
+import qualified System.Metrics.Label as Label
+import           System.Remote.Monitoring (Server, getCounter, getGauge,
+                     getLabel)
+
+
+ekgTracer :: MonadIO m => Either Metrics.Store Server-> m (Trace m FormattedMessage)
```
coot, line 25: @karknu I think you'd like to check this ekg tracer.  I think you'd prefer to have a tracer which does not need to make a namespace lookup but has an access to the gauge directly, which would make it very fast.

trace-dispatcher/src/Cardano/Logging/Tracer/Standard.hs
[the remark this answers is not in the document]
coot: I'd use `withAsync` - the caller will need to make sure that it does not exit from its continuation too early (which would kill the logging thread), and it also has a chance to do something if the logging thread dies.  since you use `forever` the async can have a type `Async Void` which indicates that the spawn thread will never produce a value, it could only throw an exception.

trace-dispatcher/src/Cardano/Logging/Tracer/EKG.hs
jutaro: Just did a benchmark with this solution, and it sends 1000 messages, each containing 5 metrics in less then 2 microseconds in the current non optimized version:)

trace-dispatcher/src/Cardano/Logging/Types.hs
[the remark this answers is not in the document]
coot: ok, it makes sense.  I would mention the function `foldTraceM` in the haddock, so one can follow the link and check its type signature.

trace-dispatcher/src/Cardano/Logging/Tracer/EKG.hs
deepfire: @jutaro, how does it compare to direct submission?

karknu: That sounds like it fast enough.

coot: Was the map was almost empty? But even with `100`'s entries it shouldn't be that much slower.

trace-dispatcher/src/Cardano/Logging/Types.hs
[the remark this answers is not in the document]
jutaro: Okay

trace-dispatcher/src/Cardano/Logging/FrequencyLimiter.hs
```diff
@@ -0,0 +1,190 @@
...
+                              --   arrive in a longer period budget is earned.
+                              --   A value between 1.0 and -1.0. If -1.0 is reached start limiting,
+                              --   and if 1.0 is reached stop limiting.
+  , frActive   :: Maybe (Int, Double)
+                              -- ^ Just is active and carries the number
+                              --   of suppressed messages and the time of last send message
+} deriving (Show)
+
+-- | Limits the frequency of messages to nMsg which is given per minute.
+
```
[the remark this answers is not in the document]
jutaro: Okay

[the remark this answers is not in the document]
jutaro: Okay

trace-dispatcher/src/Cardano/Logging/Tracer/Standard.hs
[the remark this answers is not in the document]
jutaro: I changed to async, but in my version of withAsync, the channel read blocks forever, no idea why? I checked the not working code in as comment. Can you see what I'm doing wrong?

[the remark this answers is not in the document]
coot: Have you pushed it? I don't see the non working version.  If you try to use the channel outside of the continuation it would block, because the thread would be dead.  Could that be the cause?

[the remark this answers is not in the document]
jutaro: @coot should be their: 
    -- this doesn't work for unknown reason
    -- withAsync (writerThread stateRef outChan) (\ as ->
    --   modifyIORef stateRef (\ st ->
    --     st {stRunning = Just (inChan, outChan, as)})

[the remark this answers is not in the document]
deepfire: 1. We have dropped the file writer option
2. We consider `stdout` writes blocking to be an extreme corner case -- since it's only possible with `stdout` redirection to a file system -- and also considering that no reasonable deployment would do this, since it makes log rotation impossible.

deepfire:
The `stdout` thread is now `link`-ed to the main thread, thereby propagation exceptions and shutting down the node, in case that trivial thread raises any exceptions (i.e. `stdout` being closed or redirected `stdout` write hitting a full disc).

deepfire (approved):
This looks absolutely great.

Thank you a huge lot @jutaro !

PR #3086, opened 2021-08-16
stake credential history tool

JaredCorduan:
I made this tool, which has come in handy for debugging several issues. The tool takes a stake credential uses `foldBlocks` to replay the chain and log every event related to this credential. Currently it only works for key credentials.

I think it makes sense to check this tool in.


erikd (changes requested):
Needs to also work on `shelley_qa` .

erikd:
bors merge

PR #3093, opened 2021-08-18
Make script invalid flag an assertion

newhoggy:
This means if `--script-invalid` is specified, the build transaction will fail if the scripts are valid.

### DIFF
--- a/cardano-api/src/Cardano/Api/Fees.hs
+++ b/cardano-api/src/Cardano/Api/Fees.hs
@@ -647,6 +647,9 @@ data TxBodyErrorAutoBalance =
        -- | One or more of the scripts fails to execute correctly.
      | TxBodyScriptExecutionError [(ScriptWitnessIndex, ScriptExecutionError)]
 
+       -- | One or more of the scripts were expected to fail validation, but none did.
+     | TxBodyScriptBadScriptValidity
+
        -- | The balance of the non-ada assets is not zero. The 'Value' here is
        -- that residual non-zero balance. The 'makeTransactionBodyAutoBalance'
        -- function only automatically balances ada, not other assets.
@@ -706,6 +709,9 @@ instance Error TxBodyErrorAutoBalance where
                 ++ " failed with " ++ displayError failure
               | (index, failure) <- failures ]
 
+  displayError TxBodyScriptBadScriptValidity =
+      "One or more of the scripts were expected to fail validation, but none did."
+
   displayError (TxBodyErrorAssetBalanceWrong _value) =
       "The transaction does not correctly balance in its non-ada assets. "
    ++ "The balance between inputs and outputs should sum to zero. "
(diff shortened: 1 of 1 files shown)


newhoggy:
```
$ scripts/plutus/always-fails.sh
Socket path: example/node-bft1/node.sock
Socket path: /Users/jky/wrk/iohk/cardano-node
srwxr-xr-x  1 jky  staff  0 18 Aug 15:25 example/node-bft1/node.sock
Always succeeds Plutus script in use. Any datum and redeemer combination will succeed.
Script at: ./scripts/plutus/scripts/always-succeeds-spending.plutus
{
    "16b0125dc3fb55644bc189aedea012efa4a48382480d879c4593eb3833deeff0#0": {
        "address": "addr_test1vz680r6cw6g5kd492arhgc3mug3z25jyx2cglsjhx36n4dsa7yh3t",
        "value": {
            "lovelace": 1000000000000
        }
    }
}Transaction successfully submitted.
Pausing for 5 seconds...
{
    "e0ae1388a7866ebadfbdd14409fde0c6bdb4a4e25a2755fc9eab89ef232bfacd#0": {
        "address": "addr_test1vz680r6cw6g5kd492arhgc3mug3z25jyx2cglsjhx36n4dsa7yh3t",
        "value": {
            "lovelace": 666666666394
        }
    }
}Plutus txin
e0ae1388a7866ebadfbdd14409fde0c6bdb4a4e25a2755fc9eab89ef232bfacd#1
Collateral
e0ae1388a7866ebadfbdd14409fde0c6bdb4a4e25a2755fc9eab89ef232bfacd#0
Command failed: transaction build  Error: One or more of the scripts were expected to fail validation, but none did.
```

erikd (approved):

cardano-api/src/Cardano/Api/Fees.hs
```diff
@@ -751,19 +757,26 @@ handleExUnitsErrors ::
...
-        byScriptValidity :: (ScriptWitnessIndex, ScriptExecutionError) -> Bool
-        byScriptValidity (_, e) = case scriptValidity of
-          ScriptValid -> True
-          ScriptInvalid -> case e of
-            ScriptErrorEvaluationFailed _ -> False
+      else Left (TxBodyScriptExecutionError failures)
+  where failures :: [(ScriptWitnessIndex, ScriptExecutionError)]
+        failures = Map.toList failuresMap
+handleExUnitsErrors ScriptInvalid failuresMap exUnitsMap
+  | null scriptFailures = Left TxBodyScriptBadScriptValidity
```
Jimbo4350, line 767: We should plumb through the `ScriptWitnessIndex` with the error. Otherwise in the event where there are multiple scripts, the user will not know which script(s) are failing.

Jimbo4350 (approved):
Nice, one minor change.

cardano-api/src/Cardano/Api/Fees.hs
newhoggy: In this case, we're erroring out because there are no script failures, so there aren't any indexes to plumb through.

newhoggy:
bors merge

PR #3109, opened 2021-08-20
Add Nick and Jared to CODEOWNERS

nc6:
I'm now doing a fair amount of work in node, and we should have an approver in the US timezone(s).


disassembler (approved):

newhoggy (approved):

nc6:
bors merge

PR #3116, opened 2021-08-23
Remove ci/hydra-eval

newhoggy:


rvl (approved):
The advantage of removing `ci/hydra-eval` is that Hydra gets a chance to retry evals which failed or partially finished due to OOM, without causing the bors batch to immediately fail.

The disadvantage of removing `ci/hydra-eval` is that real evaluation errors will be hidden, unless you know to look at [the eval errors page](https://hydra.iohk.io/jobset/Cardano/cardano-node-bors-staging#tabs-errors). In the case of a real eval error, nothing will get merged (because ci/hydra-build:required depends on eval), and the bors batch will eventually time out.

Perhaps there is some `release.nix` eval check which could be added to GitHub actions -- @jbgi would know.

jbgi (approved):
:+1:  eval errors for bors/staging are already quite volatile (a bors merge request from another PR will clear it) so a separate check would be more useful anyway.

jbgi:
bors merge

PR #3118, opened 2021-08-24
Suggestion to clarify wording around the appropriate GHC version to use

cloudeigh:
Clarified wording to instruct users to install version 8.10.4 of GHC.

The previous version directed users to install the latest version, which results in compile issues. It's best to instruct them to use the version that these instructions work.

Later in the document it was noted that using the latest version causes compile errors - calling it out this way causes rework.


Jimbo4350:
This is now out of date as we instruct to use `8.10.7`. Thank you for the contribution however!

PR #3121, opened 2021-08-24
Bump cardano-ledger-specs.

nc6:
This commit incorporates the following updates:
- Additional failure information for failed Plutus scripts, sufficient
  to allow local replication of the error.
- A (partial) reversion of a bugfix, which it turns out has occurred on
  the testnet and so must be accounted for.


Jimbo4350 (approved):

nc6:
bors merge

PR #3122, opened 2021-08-25
Update deps 20210825

Jimbo4350:


newhoggy (approved):

nc6 (changes requested):
I would prefer to not merge this until we get the changelog notes in for 1.29.0, otherwise we end up going on a release branch. I should have them ready later today, if you don't mind holding off until tomorrow

PR #3123, opened 2021-08-25
Wire up required signers to build and build-raw cli commands

Jimbo4350:


james-iohk:
Confirmed this works, passes my script validation when checking for tx signatories.
Also confirmed the negative case where I try to submit tx without required signature, throws on submit: `(MissingRequiredSigners (fromList [KeyHash "1ed5f183a76122092ac0f7f23427ca22385d5d6e8166300eb71a8b28"]))`

PR #3125, opened 2021-08-25
Merge queries in run tx build

newhoggy:


cardano-cli/src/Cardano/CLI/Shelley/Run/Transaction.hs
```diff
@@ -423,26 +425,27 @@ runTxBuild (AnyCardanoEra era) (AnyConsensusModeParams cModeParams) networkId mS
...
                             (AnyConsensusMode CardanoMode) (AnyCardanoEra era))
 
-      let utxoQuery = QueryInEra eInMode $ QueryInShelleyBasedEra sbe
-                        (QueryUTxO . QueryUTxOByTxIn $ Set.fromList onlyInputs)
-      let pParamsQuery = QueryInEra eInMode $ QueryInShelleyBasedEra sbe QueryProtocolParameters
-      utxo <- executeQuery era cModeParams localConnInfo utxoQuery
-      pparams <- executeQuery era cModeParams localConnInfo pParamsQuery
-      (eraHistory, systemStart) <- firstExceptT ShelleyTxCmdAcquireFailure
-                                     $ newExceptT $ queryEraHistoryAndSystemStart localNodeConnInfo Nothing
+      (utxo, pparams, eraHistory, systemStart) <-
```
Jimbo4350, line 434: I really don't like this new style with `joinEither` and `joinEitherM`. It obfuscates the code IMO and we have to introduce the aforementioned functions for the sake of only using them here.

PR #3127, opened 2021-08-25
Script context equivalence shell scripts

Jimbo4350:
This PR implements two Plutus scripts that check the equivalence of the Plutus `ScriptContext` with a custom redeemer built from the accompanying transaction. The custom redeemer mimics the `ScriptContext` data type. Depends on #3122 and #3123


cardano-api/gen/Gen/Cardano/Api/Typed.hs
```diff
@@ -170,8 +175,8 @@ genScriptData =
     genInteger :: Gen Integer
     genInteger = Gen.integral
                   (Range.linear
-                    (-fromIntegral (maxBound :: Word64) :: Integer)
-                    ( fromIntegral (maxBound :: Word64) :: Integer))
```
newhoggy, line 174: What's the reason for this change?

cardano-cli/src/Cardano/CLI/Shelley/Output.hs
```diff
@@ -31,12 +32,14 @@ data QueryTipLocalState = QueryTipLocalState
...
   , eraHistory :: EraHistory CardanoMode
   , mSystemStart :: Maybe SystemStart
+  , epochInfo :: EpochInfo (Either TransactionValidityIntervalError)
   }
 
 data QueryTipLocalStateOutput = QueryTipLocalStateOutput
   { mEra :: Maybe AnyCardanoEra
   , mEpoch :: Maybe EpochNo
   , mSyncProgress :: Maybe Text
+  , mSystemStartOut :: Maybe SystemStart
```
newhoggy, line 42: Was there a new requirement to output the system start?

Jimbo4350: Needed here: https://github.com/input-output-hk/cardano-node/pull/3127/files#diff-ce34c52a5affc68370e5db5016ace60dd627d0c9e7b495191bc52bf5ccc7bd65R224

plutus-example/plutus-example/src/Cardano/PlutusExample/ScriptContextChecker.hs
```diff
@@ -0,0 +1,372 @@
...
+  sbe <- getSbe $ cardanoEraStyle cEra
+  SocketPath sockPath <- firstExceptT EnvVarSocketErr readEnvSocketPath
+  case consensusModeOnly cModeParams of
+    CardanoMode -> do
+      let localNodeConnInfo = LocalNodeConnectInfo cModeParams network sockPath
+
+      (_chainTip, eQueryTipLocalState) <- liftIO $ queryQueryTip localNodeConnInfo Nothing
+      qTipLocalState <- firstExceptT AcquireFail $ hoistEither eQueryTipLocalState
+      let eInfo = epochInfo qTipLocalState
+      sStart <- hoistMaybe NoSystemStartTimeError $ mSystemStart qTipLocalState
```
newhoggy, line 286: `queryQueryTip` also queries for era:

```
era <- queryExpr (QueryCurrentEra CardanoModeIsMultiEra)
```

Which is thrown away.  Each `queryExpr` is a roundtrip to the server.  `_chainTip` is also discarded.

Generally speaking it's better to not query for things you don't need to avoid latency and placing load on the node if we don't need to.

It's better to inline `queryQueryTip`, replace the call to `executeLocalStateQueryExprWithChainSync` with `executeLocalStateQueryExpr` and and drop the call to `queryExpr (QueryCurrentEra CardanoModeIsMultiEra)` do avoid querying for things you don't need.

cardano-api/gen/Gen/Cardano/Api/Typed.hs
Jimbo4350: I undid this. There was a test failure that is no longer appearing.

newhoggy (approved):

PR #3130, opened 2021-08-27
Separate connection for chain tip

newhoggy:
Unfortunately, it unwinds a change we want which is to run more than one mini-protocol over the same connection, but the change is small and ought to be a good stop gap until we figure out the cause of the error.

This fixes https://github.com/input-output-hk/cardano-node/issues/3042


erikd (approved):

erikd:
bors merge

PR #3143, opened 2021-08-30
Add `chainSyncClientWithLedgerState` and `chainSyncClientPipelinedWithLedgerState`

DavidEichmann:
These wrap a ChainSyncClient(Pipelined) with logic to track the ledger state while also supporting rollback.

### DIFF
--- a/cardano-api/src/Cardano/Api/LedgerState.hs
+++ b/cardano-api/src/Cardano/Api/LedgerState.hs
@@ -20,6 +20,7 @@ module Cardano.Api.LedgerState
       )
   , initialLedgerState
   , applyBlock
+  , ValidationMode(..)
 
     -- * Traversing the block chain
   , foldBlocks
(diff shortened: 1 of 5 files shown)


cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -388,36 +367,212 @@ foldBlocks nodeConfigFilePath cardanoModeParams socketPath enableValidation stat
...
             ChainTip _ _ bno -> At bno
 
+-- | Wrap a 'ChainSyncClient' with logic that tracks the ledger state.
+chainSyncClientWithLedgerState
+  :: forall m a.
+     Monad m
+  => Env
+  -> LedgerState
+  -- ^ Initial ledger state
+  -> Bool
```
Jimbo4350, line 399: Can we use a custom sum type instead of a `Bool` here? Maybe `ValidationMode` or something like that would be suitable.

Jimbo4350 (approved):

cardano-api/src/Cardano/Api/LedgerState.hs
DavidEichmann: Done!

DavidEichmann:
bors merge

DavidEichmann:
bors merge

PR #3145, opened 2021-08-30
Fix AlonzoGenesisHash

DavidEichmann:
I've found this issue: https://github.com/input-output-hk/cardano-node/issues/2896 which was closed with https://github.com/input-output-hk/cardano-node/commit/933063c0bf01909547b939d64278f45efcd66a83, but that seems to only affect the nix built docker image and solves it by copying configs from a nix script e.g. https://github.com/input-output-hk/iohk-nix/blob/ced24db2fbed1e56563e17fd2d47436c3a4649f3/cardano-lib/mainnet-config.nix. That unfortunately doesn't fix non-docker related runs, which is why I've created this PR. I've personally seen this bug pop up in my work on #3143.


Jimbo4350 (approved):

disassembler (approved):
These configuration files are only for developer use with local testing and not properly maintained. I recommend having your image pull the configs from hydra instead.

That being said. I approve of this PR but it will need merged by developers as these are managed by them.

disassembler (changes requested):
Correction, the alonzo genesis file is wrong too. Just fixing hash isn't enough.

DavidEichmann:
So I've added a script that will build the relevant nix derivation and copy over the mainnet config files. I notice that there are a lot of other config files in the `configuration` directory. Perhapas more files could be handled by this script, though I'm happy with this for now.

disassembler (approved):

Issue #3112, opened 2021-08-21
[FR] - Reject non-key-locked collateral inputs


JaredCorduan:
**Internal/External**
*Internal*

**Area**
*Alonzo*

**Describe the feature you'd like**
It would be helpful if the `transaction build` command on the CLI rejected using collateral inputs which are not key locked. The ledger rejects them.

In order to do this, we will need to lookup the collateral inputs in the UTxO, and inspect the addresses. If the cardano-api doesn't already have a function for determining if an address is key-locked, we can use this one:

https://github.com/input-output-hk/cardano-ledger-specs/blob/91e1c199c31ad3d93e4ac4715b3e7bb8526e97cd/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs#L224-L227

james-iohk:
Proven that #3151 fixes this by throwing error like:
`Command failed: transaction build  Error: Expected key witnessed tx inputs but got script witnessed tx inputs: 
49e462b220ddfe932447c32fd75f47d57b5e4392e37d13b3b8e2b3fe030b4ac5#1`

Issue #2996, opened 2021-07-30
[FR] - Add input utxo validation to transaction build cmd


james-iohk:
**Internal/External**
*Internal*   if an IOHK staff member.

**Area**
*Plutus*    Related to Plutus Scripts (Alonzo).

**Describe the feature you'd like**
The new `transaction build` cmd doesn't validate whether `--tx-in` utxo exists, it just says (where I was using a non-existing utxo and trying to send 5ada and it thought fee was 0.163785ada)
`"The net balance of the transaction is negative: Lovelace (-5163785) lovelace."`
So, it treats the utxo as value 0 instead of explicitly saying what's up. It may not be so obvious with complex transactions, or if the utxo gets spent it'd be good to know.

**Describe alternatives you've considered**
Would be good to error with list of all the `--tx-in` utxos that don't exist.

**Additional context / screenshots**
n/a

newhoggy:
The balance check is done in the following places:

* https://github.com/input-output-hk/cardano-node/blob/aab244a42ab9a1e79f82c34a8e8f11faf1fa8a57/cardano-api/src/Cardano/Api/Fees.hs#L816
* https://github.com/input-output-hk/cardano-node/blob/aab244a42ab9a1e79f82c34a8e8f11faf1fa8a57/cardano-api/src/Cardano/Api/Fees.hs#L820

dorin100:
the correct validation is done for the `--tx-out`

james-iohk:
Proven #3151 fixes this. Additional validation may error like:
```
Command failed: transaction build  Error: The following tx input(s) were not present in the UTxO: fac4b496fc50d0fc15cc9dd60ae1f0e411154311144d020334dfaa0522fef92a#1
fac4b496fc50d0fc15cc9dd60ae1f0e411154311144d020334dfaa0522fef92a#2
```

PR #3151, opened 2021-09-01
Update build cmd to check tx inputs

Jimbo4350:
Solves #2996 and #3112

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Shelley/Run/Transaction.hs
+++ b/cardano-cli/src/Cardano/CLI/Shelley/Run/Transaction.hs
@@ -19,6 +19,7 @@ import           Prelude (String, error)
 import qualified Data.Aeson as Aeson
 import qualified Data.ByteString.Char8 as BS
 import qualified Data.ByteString.Lazy.Char8 as LBS
+import           Data.List (intersect, (\\))
 import qualified Data.Map.Strict as Map
 import qualified Data.Set as Set
 import qualified Data.Text as Text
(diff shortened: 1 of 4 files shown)


newhoggy (approved):

cardano-cli/src/Cardano/CLI/Shelley/Run/Transaction.hs
```diff
@@ -233,6 +236,13 @@ renderShelleyTxCmdError err =
     ShelleyTxCmdByronEraQuery -> "Query not available in Byron era"
     ShelleyTxCmdLocalStateQueryError err' -> renderLocalStateQueryError err'
     ShelleyTxCmdBalanceTxBody err' -> Text.pack $ displayError err'
+    ShelleyTxCmdExpectedKeyLockedTxIn txins ->
+      "Expected key witnessed tx inputs but got script witnessed tx inputs: " <>
+      Text.singleton '\n' <>
+      Text.intercalate (Text.singleton '\n') (map renderTxIn txins)
+    ShelleyTxCmdTxInsDoNotExist txins ->
+      "The following tx input(s) were not present in the UTxO: " <>
```
james-iohk, line 244: Imo it's easier to read if each missing inout is on a new line (not just from the 2nd one), so add new line at end of message.

james-iohk:
I've noticed that if the `tx-in-collateral` utxo is missing then only this is listed, even if the `tx-in` utxos are also missing. Then after fixing collateral input the same error is displayed with the missing `tx-in` utxos.  I kinda get why this would happen, is it possible or even worth making it list them all together?

cardano-cli/src/Cardano/CLI/Shelley/Run/Transaction.hs
Jimbo4350: I thought I fixed this. Good catch I'll re-push and bors.

PR #3157, opened 2021-09-02
tx generator refactoring

MarcFontaine:
Refactor the tx-generator:
* Split the wallet into a `FundSource` and `FundToStore`
* Get UTxO for Tx inputs via `FundSource` and put new Tx outputs into the wallet via `FundToStore`
* Use same code path for regular Txs and Plutus Txs (it were two independent functions before).


deepfire (approved):
LGTM, with the caveat that we should have a slightly better PR description.

PR #3162, opened 2021-09-03
Add an example for using a Plutus script as stake credential

kantp:


cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
```diff
@@ -786,7 +786,7 @@ updateTemplate (SystemStart start)
         cModel = case Alonzo.CostModel <$> defaultCostModelParams of
                    Just (Alonzo.CostModel m) ->
                      if Alonzo.validateCostModelParams m
-                     then Map.singleton Alonzo.PlutusV1 (Alonzo.CostModel m)
```
Jimbo4350, line 789: I need to fix this

Jimbo4350:
We should not merge the commit in `master`  than enables building of staking script addresses

Jimbo4350 (approved):
LGTM! We just need to adapt the scripts to use the local Alonzo cluster. I can take this over if you like.

kantp:
That would be great @Jimbo4350, thanks!

PR #3177, opened 2021-09-07
Update cli to allow datum values in the tx

Jimbo4350:


cardano-api/gen/Gen/Cardano/Api/Typed.hs
```diff
@@ -170,8 +169,8 @@ genScriptData =
     genInteger :: Gen Integer
     genInteger = Gen.integral
                   (Range.linear
-                    (-fromIntegral (maxBound :: Word64) :: Integer)
-                    ( fromIntegral (maxBound :: Word64) :: Integer))
+                    (fromIntegral (minBound :: Word64) :: Integer)
+                    (fromIntegral (maxBound :: Word64) :: Integer))
```
newhoggy, line 173: What's the reason for changing the generated value to be in the range `[-maxBound, maxBound]` to `[0, maxBound]`?

newhoggy (approved):

james-iohk:
Tested on `testnet` by using `--tx-out-datum-embed-value` to include additional datum and confirmed minting policy passes validation when expecting exact size of datum mapping.

cardano-api/gen/Gen/Cardano/Api/Typed.hs
Jimbo4350: https://input-output-rnd.slack.com/archives/C21UF2WVC/p1631459558223700?thread_ts=1631236451.216300&cid=C21UF2WVC

PR #3179, opened 2021-09-08
Use new GetChainBlockNo and GetChainPoint queries in query tip

newhoggy:


newhoggy:
Output against updated `node`.

```
CARDANO_NODE_SOCKET_PATH=example/node-bft1/node.sock  $(./scripts/bin-path.sh cardano-cli) query tip --testnet-magic 42 --cardano-mode
{
    "epoch": 0,
    "hash": "e26ae91f83160b5bcda6a4496bdd8f0aefc392eed6e510d9452446183e0fa894",
    "slot": 42,
    "block": 1,
    "era": "Alonzo",
    "syncProgress": "100.00"
}
```

Fall back when using earlier alonzo era node:

```
CARDANO_NODE_SOCKET_PATH=example/node-bft1/node.sock  $(./scripts/bin-path.sh cardano-cli) query tip --testnet-magic 42 --cardano-mode
Warning: Local header state query unavailable.  Falling back to chain sync query
{
    "epoch": 22,
    "hash": "85d006f24c4458d403391b211757c2659afbd3bcdab90c52568da947a1a4e177",
    "slot": 33828,
    "block": 2365,
    "era": "Alonzo",
    "syncProgress": "100.00"
}
```

dcoutts (dismissed):
See the suggestion below about reusing the existing API type for the tip.

cardano-api/src/Cardano/Api/IPC.hs
```diff
@@ -189,7 +189,8 @@ connectToLocalNode localNodeConnectInfo handlers
 -- protocol handlers parameterized on the negotiated node-to-client protocol
 -- version.
 --
-connectToLocalNodeWithVersion :: LocalNodeConnectInfo mode
+connectToLocalNodeWithVersion :: forall mode.
```
dnadales, line 192: Is the `forall mode` part needed?

cardano-api/src/Cardano/Api/IPC.hs
```diff
@@ -4,6 +4,7 @@
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
```
dnadales, line 7: I cannot see type applications in the diff :thinking:

cardano-api/src/Cardano/Api/Query.hs
```diff
@@ -127,6 +133,35 @@ data QueryInMode mode result where
   QuerySystemStart
     :: QueryInMode mode SystemStart
 
+  QueryHeaderStateTip
+    :: QueryInMode mode (WithOrigin (HeaderStateTip mode))
+
+data HeaderStateTip mode where
+  HeaderStateTip
+    :: ( ConsensusBlockForMode mode ~ blk
+      --  , HeaderHash blk ~ Hash BlockHeader
```
dnadales, line 142: Dangling comment.

cardano-api/src/Cardano/Api/Query.hs
```diff
@@ -496,6 +535,12 @@ fromConsensusQueryResult QuerySystemStart q' r' =
         -> r'
       _ -> fromConsensusQueryResultMismatch
 
+fromConsensusQueryResult QueryHeaderStateTip q' r' =
+    case q' of
```
dnadales, line 539: Maybe it'd be easier to read if the part after the `of` would be aligned in 2 lines.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -210,25 +224,50 @@ runQueryTip (AnyConsensusModeParams cModeParams) network mOutFile = do
...
+          { O.era = era
+          , O.eraHistory = eraHistory
+          , O.mSystemStart = mSystemStart
+          , O.mHeaderStateTip = mHeaderStateTip
+          }
 
       mLocalState <- hushM (first ShelleyQueryCmdAcquireFailure eLocalState) $ \e ->
         liftIO . T.hPutStrLn IO.stderr $ "Warning: Local state unavailable: " <> renderShelleyQueryCmdError e
 
+      chainTip <- case mLocalState >>= O.mHeaderStateTip of
```
dnadales, line 246: If I'm not mistaken `chainTip` and the long code block needed to compute it is only needed to calculate `tipSlotNo`. I'd rather have this code block abstracted away to avoid obscuring the logic of `runQueryTip`.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -210,25 +224,50 @@ runQueryTip (AnyConsensusModeParams cModeParams) network mOutFile = do
...
+          case headerStateTip of
+            Origin -> return Origin
+            At hst -> return $ At O.HeaderStateTipOutput
+              { O.slotNo = headerStateTipToSlotNo hst
+              , O.blockNo = headerStateTipToBlockNo hst
+              , O.headerHash = renderHeaderHash Proxy (headerStateTipToHeaderHash hst)
+              }
+        Nothing -> do
+          liftIO . T.hPutStrLn IO.stderr $
+            "Warning: Local header state query unavailable.  Falling back to chain sync query"
```
dnadales, line 257: Is the extra space intentional?

cardano-api/src/Cardano/Api/IPC.hs
newhoggy: Removed.

cardano-api/src/Cardano/Api/IPC.hs
newhoggy: Removed.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: The chain tip is actually used later on as well.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: An old habit I developed because my high school teacher insisted this is how sentences are supposed to be delimited 😆

cardano-api/src/Cardano/Api/Query.hs
newhoggy: I hesitate to break with the existing formatting.  I'll let my team decide if reformatting is warranted here.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
dnadales: Hehe, yeah, I think that's the way typists break their sentences :) Even Emacs requires two spaces after the period to consider a sentence ended.

cardano-api/src/Cardano/Api/Query.hs
```diff
@@ -127,6 +134,34 @@ data QueryInMode mode result where
...
+  QueryChainBlockNo
+    :: QueryInMode mode (WithOrigin BlockNo)
+
+  QueryChainPoint
+    :: QueryInMode mode (Point mode)
+
+data Point mode where
+  Point
+    :: ( ConsensusBlockForMode mode ~ blk
+       )
```
Jimbo4350, line 146: Why not ` :: (ConsensusBlockForMode mode ~ blk)`? Seems weird to newline for a bracket.

cardano-api/src/Cardano/Api/Query.hs
```diff
@@ -479,7 +512,9 @@ consensusQueryInEraInMode erainmode =
 --
 
 fromConsensusQueryResult :: forall mode block result result'.
-                            ConsensusBlockForMode mode ~ block
+                            ( ConsensusBlockForMode mode ~ block
+                            -- , HeaderHash block ~ Hash BlockHeader
```
Jimbo4350, line 516: Left over comment?

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -53,6 +54,8 @@ import           Shelley.Spec.Ledger.EpochBoundary
 import           Shelley.Spec.Ledger.LedgerState hiding (_delegations)
 import           Shelley.Spec.Ledger.Scripts ()
 import           Text.Printf (printf)
+-- import           Ouroboros.Consensus.Block.Abstract (toRawHash)
```
Jimbo4350, line 57: Left over comments?

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -72,10 +75,21 @@ import qualified Ouroboros.Network.Protocol.LocalStateQuery.Type as LocalStateQu
...
+import           Ouroboros.Consensus.Block (ConvertRawHash (..))
+import           Ouroboros.Network.Block (HeaderHash)
+import           Ouroboros.Consensus.Cardano.Block (CardanoBlock)
+
 {- HLINT ignore "Reduce duplication" -}
 {- HLINT ignore "Use const" -}
 {- HLINT ignore "Use let" -}
 
+-- | Hex encode and render a 'HeaderHash' as text.
+renderHeaderHash :: (ConvertRawHash blk, blk ~ CardanoBlock StandardCrypto) => proxy blk -> HeaderHash blk -> Text
```
Jimbo4350, line 89: I think it makes more sense to wrap `HeaderHash blk` in `cardano-api` and implement a `SerialiseAsRawBytes` instance instead of exposing `ouroboros-network` types in the cli.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -210,25 +241,45 @@ runQueryTip (AnyConsensusModeParams cModeParams) network mOutFile = do
...
-            else return Nothing
-          return O.QueryTipLocalState
-            { O.era = era
-            , O.eraHistory = eraHistory
-            , O.mSystemStart = mSystemStart
-            }
+      eLocalState <- liftIO $ executeLocalStateQueryExpr localNodeConnInfo Nothing $ \ntcVersion -> do
+        era <- queryExpr (QueryCurrentEra CardanoModeIsMultiEra)
+        eraHistory <- queryExpr (QueryEraHistory CardanoModeIsMultiEra)
+        mChainBlockNo <- if ntcVersion >= NodeToClientV_10
```
Jimbo4350, line 247: We really need to come up with a better way of handling which queries are available for a particular version. Is this information available in the consensus library?

cardano-cli/src/Cardano/CLI/Shelley/Output.hs
```diff
@@ -5,32 +5,42 @@ module Cardano.CLI.Shelley.Output
...
 import           Data.Function (($), (.), id)
 import           Data.Maybe
 import           Data.Monoid (mconcat)
 import           Shelley.Spec.Ledger.Scripts ()
+import           Text.Show (Show)
 
 import qualified Data.Aeson as J
 import qualified Data.Aeson.Encoding as JE
 
+data ChainTipInfo = ChainTipInfo
```
Jimbo4350, line 28: `ChainTipInfo` and `ChainTip` are extremely similar. Why can't we combine them or use one? Using `Maybe ChainTipInfo` isn't as informative. Does the `Nothing` case represent when the node's version does not support the query?

Jimbo4350 (changes requested):
Nice. A few comments.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: At the moment, it's document in the code here: https://github.com/input-output-hk/ouroboros-network/blob/c25d36e62549d5bec9876cb57d1c52ee41bdec3a/ouroboros-network/src/Ouroboros/Network/NodeToClient/Version.hs#L45

cardano-api/src/Cardano/Api/Query.hs
newhoggy: Fixed!

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: Fixed!

cardano-cli/src/Cardano/CLI/Shelley/Output.hs
newhoggy: `ChainTip` is defined like this:

```
data ChainTip = ChainTipAtGenesis
              | ChainTip !SlotNo !(Hash BlockHeader) !BlockNo
  deriving (Eq, Show)
```

This is the type returned by the chain sync protocol and is a pre-existing type.

```
getLocalChainTip :: LocalNodeConnectInfo mode -> IO ChainTip
```

In contrast, the local state query returns a `Point` which internally represents the hash as a `HeaderHash blk`.

`ChainTipInfo` is the new type and is defined like this:

```
data ChainTipInfo = ChainTipInfo
  { mBlockNo :: BlockNo
  , mSlotNo :: SlotNo
  , mHeaderHash :: Text
  } deriving Show
```

I introduced this type because the header hash is represented as `Text`.  I now how to convert a `HeaderHash blk` to `Text` and `Hash BlockHeader` to `Text`.

In order for me to use `ChainTip`, I need some way to convert `HeaderHash blk` to `Hash BlockHeader`.

newhoggy: I think I may be able to use `ChainPoint` type.  Let me take a look.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: I'm no longer using this ype and I don' need the function anymore.

cardano-api/src/Cardano/Api/Query.hs
newhoggy: This code is deleted.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -213,21 +220,54 @@ runQueryTip (AnyConsensusModeParams cModeParams) network mOutFile = do
...
+        -- The chain tip is unavailable via local state query because we are connecting with an older
+        -- node to client protocol so we use chain sync instead which necessitates another connection.
+        -- At some point when we can stop supporting the older node to client protocols, this fallback
+        -- can be removed.
+        Nothing -> queryChainTipViaChainSync localNodeConnInfo
+
+      let tipSlotNo :: SlotNo = case chainTip of
             ChainTipAtGenesis -> 0
-            ChainTip slotNo _ _ -> slotNo
+            (ChainTip slotNo _ _) -> slotNo
```
Jimbo4350, line 257: Unnecessary brackets

Jimbo4350 (approved):
LGTM! Can you clean up the commits?

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -241,24 +281,27 @@ runQueryTip (AnyConsensusModeParams cModeParams) network mOutFile = do
...
       case mOutFile of
-        Just (OutputFile fpath) -> liftIO $ LBS.writeFile fpath jsonOutput
-        Nothing                 -> liftIO $ LBS.putStrLn        jsonOutput
+        Just (OutputFile fpath) -> liftIO $ LBS.writeFile fpath $ encodePretty localStateOutput
+        Nothing                 -> liftIO $ LBS.putStrLn        $ encodePretty localStateOutput
 
     mode -> left (ShelleyQueryCmdUnsupportedMode (AnyConsensusMode mode))
 
+
+makeChainTip :: WithOrigin BlockNo -> ChainPoint -> ChainTip
```
Jimbo4350, line 298: This belongs in `cardano-api`

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -213,21 +220,54 @@ runQueryTip (AnyConsensusModeParams cModeParams) network mOutFile = do
...
 
-      (chainTip, eLocalState) <- liftIO $ queryQueryTip localNodeConnInfo Nothing
+      eLocalState <- liftIO $ executeLocalStateQueryExpr localNodeConnInfo Nothing $ \ntcVersion -> do
+        era <- queryExpr (QueryCurrentEra CardanoModeIsMultiEra)
+        eraHistory <- queryExpr (QueryEraHistory CardanoModeIsMultiEra)
+        mChainBlockNo <- if ntcVersion >= NodeToClientV_10
+          then Just <$> queryExpr QueryChainBlockNo
+          else return Nothing
+        mChainPoint <- if ntcVersion >= NodeToClientV_10
+          then Just <$> queryExpr (QueryChainPoint CardanoMode)
```
newhoggy, line 230: I'm curious if there is a way to not pass in `CardanoMode` the constructor argument like this.

cardano-api/src/Cardano/Api/Query.hs
```diff
@@ -496,6 +502,18 @@ fromConsensusQueryResult QuerySystemStart q' r' =
         -> r'
       _ -> fromConsensusQueryResultMismatch
 
+fromConsensusQueryResult QueryChainBlockNo q' r' =
+    case q' of
+      Consensus.GetChainBlockNo
+        -> r'
+      _ -> fromConsensusQueryResultMismatch
+
+fromConsensusQueryResult (QueryChainPoint mode) q' r' =
```
newhoggy, line 511: The reason I need the `mode` is because below, `fromConsensusPointInMode` requires `mode`.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: Moved!

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: Removed!

PR #3181, opened 2021-09-08
Update min utxo calculation cli command

### LINKED ISSUE #3165: [BUG] - TODO alonzo: runTxCalculateMinValue using new protocol params
**External**

**Area**
*Other*   calculate-min-value cli command

**Summary**
When running `cardano-cli transaction calculate-min-value...` command in Alonzo, we get an error.

```
$ cardano-cli transaction calculate-min-value --protocol-params-file /tmp/protocol-parameters-1097911063.json --multi-asset '1 b45fde8ab44e77aaa825e838715c5f2a1b60013c19efa639fa96da96.BlueCheese'

cardano-cli: FatalError {fatalErrorMessage = "TODO alonzo: runTxCalculateMinValue using new protocol params"}
```

**Steps to reproduce**
1. In an alonzo network, run the above command
2. See the error

**Expected behavior**
Expect to correctly calculate costs in Alonzo

- OS Name: Ubuntu
- OS Version: 20.04
- Node version: 1.29.0
- CLI version: 1.29.0

Jimbo4350:
Resolves #3165 

Breaking change: `calculate-min-req-utxo` requires a transaction output (`TxOut era`) instead of a `Value` because to calculate the min required UTxO in the Alonzo era we require a transaction output. It also requires the specification of the era.

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
+++ b/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
@@ -598,18 +598,29 @@ pTransaction =
     , subParser "policyid"
         (Opt.info pTransactionPolicyId $ Opt.progDesc "Calculate the PolicyId from the monetary policy script.")
     , subParser "calculate-min-fee"
-        (Opt.info pTransactionCalculateMinFee $ Opt.progDesc "Calculate the minimum fee for a transaction")
-    , subParser "calculate-min-value"
-        (Opt.info pTransactionCalculateMinValue $ Opt.progDesc "Calculate the minimum value for a transaction")
-
+        (Opt.info pTransactionCalculateMinFee $ Opt.progDesc "Calculate the minimum fee for a transaction.")
+    , subParser "calculate-min-required-utxo"
+        (Opt.info pTransactionCalculateMinReqUTxO $ Opt.progDesc "Calculate the minimum required UTxO for a transaction output.")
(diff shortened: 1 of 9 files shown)


cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -598,18 +598,29 @@ pTransaction =
...
+  -- Backwards compatible parsers
+  calcMinValueInfo :: ParserInfo TransactionCmd
+  calcMinValueInfo =
+    Opt.info pTransactionCalculateMinValue
+      $ Opt.progDesc "Calculate the minimum value for a transaction"
+
+  pCalculateMinRequiredUtxoBackwardCompatible :: Parser TransactionCmd
+  pCalculateMinRequiredUtxoBackwardCompatible =
+    Opt.subparser
+      $ Opt.command "calculate-min-value" calcMinValueInfo <> Opt.internal
```
newhoggy, line 622: Are these deprecated?  Should there be a recommendation in the help to use something else instead?

newhoggy (approved):

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
Jimbo4350: Yeah I should include a deprecation warning.

PR #3186, opened 2021-09-10
trace-forward: flexible buffering mechanism.

deepfire:
This updates the `trace-forward` / `trace-dispatcher` integration:

1. improve the abstraction between the two components
2. we introduce an adaptive buffering mechanism, which:
    - provides sufficient buffering in situations when we don't have a connection to `cardano-tracer` (either initially, or after connection loss)
    - reverts to a smaller buffer (short queue) when the connection is established -- to improve resource efficiency


deepfire:
As soon as we test the full node integration, we can merge.

denisshevchenko (approved):

PR #3198, opened 2021-09-14
Fix rendering of RemoteConnectionId for tracing

cblp:
https://input-output.atlassian.net/browse/CAD-3411

### DIFF
--- a/cardano-node/src/Cardano/Tracing/Peer.hs
+++ b/cardano-node/src/Cardano/Tracing/Peer.hs
@@ -6,6 +6,7 @@
 module Cardano.Tracing.Peer
   ( Peer (..)
   , getCurrentPeers
+  , ppCid
   , ppPeer
   , tracePeers
   ) where
--- a/cardano-node/src/Cardano/Tracing/Peer.hs
+++ b/cardano-node/src/Cardano/Tracing/Peer.hs
@@ -14,13 +14,13 @@ import           Cardano.Prelude hiding (atomically)
 import           Prelude (String)
 
 import qualified Control.Monad.Class.MonadSTM.Strict as STM
-
 import           Data.Aeson (ToJSON (..), Value (..), toJSON, (.=))
 import qualified Data.Map.Strict as Map
 import qualified Data.Set as Set
 import qualified Data.Text as Text
+import           Network.Socket (SockAddr (SockAddrInet, SockAddrInet6))
+import           NoThunks.Class (AllowThunk (..), NoThunks)
 import           Text.Printf (printf)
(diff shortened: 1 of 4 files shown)


cardano-node/src/Cardano/Tracing/Peer.hs
```diff
@@ -57,7 +58,14 @@ ppPeer (Peer cid _af status inflight) =
...
 ppCid :: RemoteConnectionId -> String
-ppCid = takeWhile (/= ':') . show . remoteAddress
+ppCid cid =
+  case r of
+    SockAddrInet{} -> takeWhile (/= ':') s
+    SockAddrInet6{} -> takeWhile (/= ']') s ++ [']']
+    _ -> s
+  where
+    r = remoteAddress cid
+    s = show r
```
karknu, line 68: The port number is part of the connection id. Two TCP connections from the same IP address but different remote ports are treated as different peers by ourboros-network.

karknu (changes requested):

cardano-node/test/cardano-node-test.hs
```diff
@@ -15,14 +15,10 @@ import qualified Test.Cardano.Node.POM
 
 main :: IO ()
 main = defaultMain
+  ( [ Test.Cardano.Node.Json.tests
+    , Test.Cardano.Node.POM.tests
```
Jimbo4350, line 20: This can be merged with the commit that introduces the new property test.

cardano-node/src/Cardano/Tracing/Peer.hs
```diff
@@ -6,6 +6,7 @@
 module Cardano.Tracing.Peer
   ( Peer (..)
   , getCurrentPeers
+  , ppCid
```
Jimbo4350, line 9: This can be merged with the previous commit

Jimbo4350 (commented):
I'll let  @coot have a look before approving

cardano-node/src/Cardano/Tracing/Peer.hs
cblp: oh, the previous implementation was incorrect for IPv4, too?

cardano-node/src/Cardano/Tracing/Peer.hs
cblp: Is adding an export ok?

cardano-node/src/Cardano/Tracing/Peer.hs
karknu: Yes.

cblp:
see #3199

PR #3202, opened 2021-09-14
Upgrade to cabal-3.6.0.0

newhoggy:


newhoggy:
We will upgrade to `cabal-3.6.2.0` instead in https://github.com/input-output-hk/cardano-node/pull/3334

PR #3206, opened 2021-09-16
Revert "Don't assume AssetName is UTF8"

disassembler:
This reverts commit 2144866b074698e5adec894fc24a7c5b79304c35.

Reverting as this was only changed in display, but transaction building still assumes latin1. It makes more sense to do this as one piece when devs have time to work on this to limit churn with community tools.

### DIFF
--- a/cardano-api/src/Cardano/Api/Value.hs
+++ b/cardano-api/src/Cardano/Api/Value.hs
@@ -58,7 +58,7 @@ import           Prelude
 
 import           Data.Aeson hiding (Value)
 import qualified Data.Aeson as Aeson
-import           Data.Aeson.Types (Parser)
+import           Data.Aeson.Types (Parser, toJSONKeyText)
 import           Data.ByteString (ByteString)
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.Char8 as BSC
@@ -160,7 +160,6 @@ scriptPolicyId = PolicyId . hashScript
 newtype AssetName = AssetName ByteString
     deriving stock (Eq, Ord)
     deriving newtype (Show)
-    deriving (ToJSON, FromJSON, ToJSONKey, FromJSONKey) via UsingRawBytesHex AssetName
 
 instance IsString AssetName where
     fromString s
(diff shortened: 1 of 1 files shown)


nc6:
Looks reasonable to me. @sjoerdvisscher does this interfere with the reasons you needed this merged?

disassembler:
This was opened as a result of community feedback. Two complaints:

sjoerdvisscher:
To be clear: I did not anticipate that this change would change the structure, and we don't need that. We only need support for non-UTF8 asset names. 

I have asked @j-mueller to weigh in regarding how much reverting this will be a problem.

j-mueller:
@disassembler : Without this change we won't be able to send certain transactions from the PAB to the wallet backend. So I would very much prefer not to revert it. The spec is pretty clear on this: `AssetName` can be any bytestring.

nc6:
@j-mueller Would the PAB be able to continue developing against the existing version, if we back this out for this release and promise to have it (properly!) working again by the time the PAB is ready? @disassembler I guess this would mean we'd need to push a new release for PAB compatibility (but that wouldn't need a new HF)

Jimbo4350:
I've asked @cblp to work on this, discuss the changes needed with him.

AndrewWestberg:
If this change is made, to make the token names as HEX due to a requirement of the PAB, I would also kindly suggest that cardano-cli support hex names at the same time.

j-mueller:
@nc6 : Yes we will be able to keep working on the wallet backend integration, which is our main task currently. But after the HF, when our change is put back in, we'll also need a new release of the wallet backend, because that's where we send the transactions that use this encoding.

And just to be clear what the consequences are: Any apps that use hashes for `AssetName`s won't be able to send transactions to the network via wallet backend as long as the latin1 encoding is used. This includes everything that uses the thread tokens feature of our state machine library.

cblp:
If I understand correctly, no-assuming-utf-8 works correctly, and reverting it will fail utxo queries. So, I object against assuming UTF8 and 3206. It's safer to show hex strings.

JaredCorduan (approved):
I'm approving so that we can keep things moving, and knowing that @cblp will start on the proper fix tomorrow with high priority.

gitmachtl:
Do we have an idea on how the switch to hex will look like for the normal query utxo output and also for the cli parameters --tx-out and --mint?  

Is the idea to keep the <policyID>.<assetName> structure with the '.' in the middle as separator for all the outputs/inputs, or will it be a combined hexstring with <policyID><assetName> like the metadata-registry subject?

gitmachtl:
or maybe with an additional parameter for the cli query utxo command '--decode' to get out a readable format and defaulting to pure hex output?

gitmachtl:
resolved via https://github.com/input-output-hk/cardano-node/pull/3211

PR #3208, opened 2021-09-16
Add update proposal friendly rendering

cblp:
Including Alonzo-specific updates


disassembler:
Passing a signed tx looks great, but passing a txbody throws a deserialise error:

cardano-cli/test/data/golden/alonzo/genesis/verification.key
```diff
@@ -0,0 +1,5 @@
+{
+    "type": "GenesisVerificationKey_ed25519",
+    "description": "Genesis Verification Key",
```
Jimbo4350, line 3: We should remove these keys and use them directly from `/home/jordan/Repos/Work/cardano-haskell/cardano-node/cardano-cli/test/data/golden/shelley/keys/genesis_keys/verification_key`

Likewise for the signing key. We're duplicating things here.

cardano-cli/test/Test/Golden/TxView.hs
```diff
@@ -173,3 +206,48 @@ golden_view_mary =
       execCardanoCLI
         ["transaction", "view", "--tx-body-file", transactionBodyFile]
     diffVsGoldenFile result "test/data/golden/mary/transaction-view.out"
+
+golden_view_alonzo :: Property
+golden_view_alonzo =
+  propertyOnce $
+  moduleWorkspace "tmp" $ \tempDir -> do
```
Jimbo4350, line 213: Missing indent

cardano-cli/test/data/golden/shelley/genesis/verification.key
```diff
@@ -0,0 +1,5 @@
+{
+    "type": "GenesisVerificationKey_ed25519",
+    "description": "Genesis Verification Key",
```
Jimbo4350, line 3: Same here. These already exist in: `/home/jordan/Repos/Work/cardano-haskell/cardano-node/cardano-cli/test/data/golden/shelley/keys/genesis_keys/`

Jimbo4350 (changes requested):
Looking good, minor changes.

cardano-cli/test/data/golden/alonzo/genesis/verification.key
cblp: I think making tests independent is a good idea. But in this case I see no problem in reusing keys.

cardano-cli/test/Test/Golden/TxView.hs
cblp: Is indent really needed here?

Jimbo4350: Only if you want to follow the style doc :)

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -2645,31 +2645,31 @@ pPoolInfluence :: Parser Rational
...
-      <> Opt.metavar "DOUBLE"
+      <> Opt.metavar "RATIONAL"
       <> Opt.help "Monetary expansion."
       )
 
 pDecentralParam :: Parser Rational
 pDecentralParam =
     Opt.option readRationalUnitInterval
       (  Opt.long "decentralization-parameter"
-      <> Opt.metavar "DOUBLE"
```
Jimbo4350, line 2672: Let's leave it as `DOUBLE` as we can assume some technical knowledge if somebody is using `cardano-cli`.

cardano-cli/test/Test/Golden/TxView.hs
```diff
@@ -53,7 +53,39 @@ golden_view_shelley :: Property
...
+    transactionBodyFile <- noteTempFile tempDir "transaction-body"
+
+    -- Create update proposal
+    void $
+      execCardanoCLI
+        [ "governance", "create-update-proposal"
+        , "--decentralization-parameter", "63/64"
+        , "--epoch", "64"
+        , "--extra-entropy", "c0ffee"
+          --  hash(c0ffee) =
```
Jimbo4350, line 66: So rather than leaving comments with the hashes we should output this in the logs using `note_` from `Hedgehog.Extras.Test`.

Jimbo4350 (changes requested):
Almost there

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
cblp: Maybe DECIMAL?

cblp: The reason is we accept strings like "31/57". I can't call it DOUBLE[-PRECISION FLOATING POINT].

cardano-cli/test/Test/Golden/TxView.hs
cblp: done

cardano-cli/test/Test/Golden/TxView.hs
cblp: done

scripts/reconfigure-hlint.sh
```diff
@@ -29,6 +29,6 @@ gen_rules() {
   echo "# This file is generated from .hlint.template.yaml by scripts/reconfigure-hlint.sh"
   echo ""
   cat .hlint.template.yaml \
-    | sed '/^# BEGIN-GENERATED/,/^# END-GENERATED/{/^# BEGIN-GENERATED/!{/^# END-GENERATED/!d}}' \
```
Jimbo4350, line 32: Does this still work on linux?

Jimbo4350 (approved):
LGTM

scripts/reconfigure-hlint.sh
cblp: Fixed. Works now.

Jimbo4350 (approved):

cblp:
bors merge

PR #3211, opened 2021-09-17
Serialize AssetName to JSON as hex string

cblp:
Fix (second attempt) of incorrect UTF-8 decoding of non-textual (non-UTF-8) asset names.

I also changed tx-view to test JSON repr.

BEFORE:
```js
{
  "51f056b530aba5969bf996d1be51c7aa3581f6e0d7b9e0911a5c0c1d": {
    "TheDrifter13": 1,
    "GOLDtest": 5,
    "��\"3DUfw��������": 42 // UTF-8 fail, actually
  }
}
```

AFTER:
```js
{
  "51f056b530aba5969bf996d1be51c7aa3581f6e0d7b9e0911a5c0c1d": {
    "546865447269667465723133": 1,
    "474f4c4474657374": 5,
    "00112233445566778899AABBCCDDEEFF": 42
  }
}
```

This is how it already looks like in UI (Cardano Explorer):
> asset1...
> Ticker:
> Name:
> Description:
> Policy ID: 51f056b530aba5969bf996d1be51c7aa3581f6e0d7b9e0911a5c0c1d
> Asset Name: 474f4c4474657374

Future Daedalus:
> Name: GoodCoin
> Asset name: 676f6f64636f696e (ASCII: goodcoin)

### DIFF
--- a/cardano-api/src/Cardano/Api/SerialiseUsing.hs
+++ b/cardano-api/src/Cardano/Api/SerialiseUsing.hs
@@ -10,14 +10,14 @@ module Cardano.Api.SerialiseUsing
 
 import           Prelude
 
+import qualified Data.Aeson.Types as Aeson
+import           Data.ByteString (ByteString)
 import qualified Data.ByteString.Base16 as Base16
 import qualified Data.ByteString.Char8 as BSC
 import           Data.String (IsString (..))
 import qualified Data.Text as Text
 import qualified Data.Text.Encoding as Text
-import           Data.Typeable
-
-import qualified Data.Aeson.Types as Aeson
+import           Data.Typeable (Typeable, tyConName, typeRep, typeRepTyCon)
 
 import           Cardano.Api.Error
 import           Cardano.Api.HasTypeProxy
@@ -63,33 +63,39 @@ instance SerialiseAsRawBytes a => Show (UsingRawBytesHex a) where
     show (UsingRawBytesHex x) = show (serialiseToRawBytesHex x)
 
 instance SerialiseAsRawBytes a => IsString (UsingRawBytesHex a) where
-    fromString str =
-      case Base16.decode (BSC.pack str) of
-        Right raw -> case deserialiseFromRawBytes ttoken raw of
-          Just x  -> UsingRawBytesHex x
-          Nothing -> error ("fromString: cannot deserialise " ++ show str)
-        Left msg -> error ("fromString: invalid hex " ++ show str ++ ", " ++ msg)
-      where
-        ttoken :: AsType a
-        ttoken = proxyToAsType Proxy
+    fromString = either error id . deserialiseFromRawBytesBase16 . BSC.pack
 
 instance SerialiseAsRawBytes a => ToJSON (UsingRawBytesHex a) where
     toJSON (UsingRawBytesHex x) = toJSON (serialiseToRawBytesHexText x)
 
 instance (SerialiseAsRawBytes a, Typeable a) => FromJSON (UsingRawBytesHex a) where
-    parseJSON =
-      Aeson.withText tname $ \str ->
-        case Base16.decode (Text.encodeUtf8 str) of
-          Right raw -> case deserialiseFromRawBytes ttoken raw of
-            Just x  -> return (UsingRawBytesHex x)
-            Nothing -> fail ("cannot deserialise " ++ show str)
-          Left msg  -> fail ("invalid hex " ++ show str ++ ", " ++ msg)
-      where
-        ttoken = proxyToAsType (Proxy :: Proxy a)
-        tname  = (tyConName . typeRepTyCon . typeRep) (Proxy :: Proxy a)
+  parseJSON =
+    Aeson.withText tname $
+      either fail pure . deserialiseFromRawBytesBase16 . Text.encodeUtf8
+    where
+      tname  = (tyConName . typeRepTyCon . typeRep) (Proxy :: Proxy a)
 
-instance SerialiseAsRawBytes a => ToJSONKey (UsingRawBytesHex a)
-instance (SerialiseAsRawBytes a, Typeable a) => FromJSONKey (UsingRawBytesHex a)
+instance SerialiseAsRawBytes a => ToJSONKey (UsingRawBytesHex a) where
+  toJSONKey =
+    Aeson.toJSONKeyText $ \(UsingRawBytesHex x) -> serialiseToRawBytesHexText x
+
+instance
+  (SerialiseAsRawBytes a, Typeable a) => FromJSONKey (UsingRawBytesHex a) where
+
+  fromJSONKey =
+    Aeson.FromJSONKeyTextParser $
+    either fail pure . deserialiseFromRawBytesBase16 . Text.encodeUtf8
+
+deserialiseFromRawBytesBase16 ::
+  SerialiseAsRawBytes a => ByteString -> Either String (UsingRawBytesHex a)
+deserialiseFromRawBytesBase16 str =
+  case Base16.decode str of
+    Right raw -> case deserialiseFromRawBytes ttoken raw of
+      Just x  -> Right (UsingRawBytesHex x)
+      Nothing -> Left ("cannot deserialise " ++ show str)
+    Left msg  -> Left ("invalid hex " ++ show str ++ ", " ++ msg)
+  where
+    ttoken = proxyToAsType (Proxy :: Proxy a)
 
 
 -- | For use with @deriving via@, to provide instances for any\/all of 'Show',
--- a/cardano-api/src/Cardano/Api/Utils.hs
+++ b/cardano-api/src/Cardano/Api/Utils.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE LambdaCase #-}
+
 -- | Internal utils for the other Api modules
 --
 module Cardano.Api.Utils
@@ -6,6 +8,7 @@ module Cardano.Api.Utils
   , formatParsecError
   , noInlineMaybeToStrictMaybe
   , runParsecParser
+  , note
   ) where
 
 import           Prelude
@@ -42,3 +45,8 @@ runParsecParser parser input =
   case Parsec.parse (parser <* Parsec.eof) "" (Text.unpack input) of
     Right txin -> pure txin
     Left parseError -> fail $ formatParsecError parseError
+
+note :: MonadFail m => String -> Maybe a -> m a
+note msg = \case
+  Nothing -> fail msg
+  Just a -> pure a
--- a/cardano-api/src/Cardano/Api/Value.hs
+++ b/cardano-api/src/Cardano/Api/Value.hs
@@ -56,9 +56,9 @@ module Cardano.Api.Value
 
 import           Prelude
 
-import           Data.Aeson hiding (Value)
+import           Data.Aeson (FromJSON, FromJSONKey, ToJSON, object, parseJSON, toJSON, withObject)
 import qualified Data.Aeson as Aeson
-import           Data.Aeson.Types (Parser, toJSONKeyText)
+import           Data.Aeson.Types (Parser, ToJSONKey)
 import           Data.ByteString (ByteString)
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.Char8 as BSC
@@ -74,9 +74,9 @@ import qualified Data.Text.Encoding as Text
 import qualified Cardano.Chain.Common as Byron
 
 import qualified Cardano.Ledger.Coin as Shelley
+import           Cardano.Ledger.Crypto (StandardCrypto)
 import qualified Cardano.Ledger.Mary.Value as Mary
 import qualified Cardano.Ledger.ShelleyMA.Rules.Utxo as Shelley
-import           Cardano.Ledger.Crypto (StandardCrypto)
 
 import           Cardano.Api.HasTypeProxy
 import           Cardano.Api.Script
@@ -158,8 +158,10 @@ scriptPolicyId = PolicyId . hashScript
 
 
 newtype AssetName = AssetName ByteString
-    deriving stock (Eq, Ord)
-    deriving newtype (Show)
+  deriving stock (Eq, Ord)
+  deriving newtype (Show)
+  deriving (ToJSON, FromJSON, ToJSONKey, FromJSONKey)
+    via UsingRawBytesHex AssetName
 
 instance IsString AssetName where
     fromString s
@@ -176,18 +178,6 @@ instance SerialiseAsRawBytes AssetName where
     deserialiseFromRawBytes AsAssetName bs
       | BS.length bs <= 32 = Just (AssetName bs)
       | otherwise          = Nothing
-
-instance ToJSON AssetName where
-  toJSON (AssetName an) = Aeson.String $ Text.decodeUtf8 an
-
-instance FromJSON AssetName where
-  parseJSON = withText "AssetName" (return . AssetName . Text.encodeUtf8)
-
-instance ToJSONKey AssetName where
-  toJSONKey = toJSONKeyText (\(AssetName asset) -> Text.decodeUtf8 asset)
-
-instance FromJSONKey AssetName where
-  fromJSONKey = FromJSONKeyText (AssetName . Text.encodeUtf8)
 
 
 data AssetId = AdaAssetId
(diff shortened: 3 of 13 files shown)


gitmachtl:
Just as a note: It should also be than `<policyID><assetName>` combined hexstring for the normal query utxo output too, also for the tx-out and mint parameters for CLI. like the "subject" parameter for the metadata-registry, same format for all.

cblp:
Removed textual AssetName everywhere, hex only

gitmachtl:
> Removed textual AssetName everywhere, hex only

do we really need the "." separator now than for policyID.assetName ?  polID is 56 chars long, rest is the assetName if present.

gitmachtl:
> Removed textual AssetName everywhere, hex only

minted some tokens, looking good

gitmachtl:
I noticed in the JSON output, that the address key is now showing the bech32 address of the requested address. A few versions ago this showed the hex-representation of the address. Does not make a huge difference, but is this address field actively used somewhere? Looks like too me, that this is an additional information nobody really needs?

cardano-api/src/Cardano/Api/Value.hs
```diff
@@ -354,9 +345,12 @@ valueFromNestedRep (ValueNestedRep bundles) =
...
   toJSON (ValueNestedRep bundles) = object $ map toPair bundles
     where
-     toPair :: ValueNestedBundle -> (Text, Aeson.Value)
-     toPair (ValueNestedBundleAda q) = ("lovelace", toJSON q)
-     toPair (ValueNestedBundle pid assets) = (renderPolicyId pid, toJSON assets)
+      toPair :: ValueNestedBundle -> (Text, Aeson.Value)
+      toPair (ValueNestedBundleAda q) = ("lovelace", toJSON q)
+      toPair (ValueNestedBundle pid assets) =
+        ( renderPolicyId pid
+        , toJSON $ Map.mapKeys serialiseToRawBytesHexText assets
```
Jimbo4350, line 352: Let's extract this logic to a  `ToJSON AssetName` instance.

cardano-api/src/Cardano/Api/Value.hs
```diff
@@ -366,11 +360,28 @@ instance FromJSON ValueNestedRep where
...
       parsePid :: (Text, Aeson.Value) -> Parser ValueNestedBundle
       parsePid ("lovelace", q) = ValueNestedBundleAda <$> parseJSON q
-      parsePid (pid, q) =
-        case deserialiseFromRawBytesHex AsScriptHash (Text.encodeUtf8 pid) of
-          Just sHash -> ValueNestedBundle (PolicyId sHash) <$> parseJSON q
-          Nothing -> fail $ "Failure when deserialising PolicyId: "
-                         <> Text.unpack pid
+      parsePid (pid, quantityBundleJson) = do
+        sHash <-
+          note ("Failure when deserialising PolicyId: " <> Text.unpack pid) $
```
Jimbo4350, line 365: "Failed to deserialise PolicyId: " would be a better error message.

cardano-api/src/Cardano/Api/Value.hs
```diff
@@ -366,11 +360,28 @@ instance FromJSON ValueNestedRep where
...
+          deserialiseFromRawBytesHex AsScriptHash $ Text.encodeUtf8 pid
+        quantityBundleRaw <- parseJSON quantityBundleJson
+        quantityBundle <-
+          Map.fromList
+          <$> for (Map.assocs quantityBundleRaw) \(nameText, quantity) -> do
+                name <-
+                  note
+                    ("Failure when deserialising AssetName"
```
Jimbo4350, line 373: Like wise for here. We should extract this to a `FromJSON AssetName` and then implement a JSON roundtrip test.

cardano-api/src/Cardano/Api/Value.hs
```diff
@@ -366,11 +360,28 @@ instance FromJSON ValueNestedRep where
...
+                      <> Text.unpack nameText) $
+                  deserialiseFromRawBytesHex AsAssetName $
+                  Text.encodeUtf8 nameText
+                pure (name, quantity)
+        pure $ ValueNestedBundle (PolicyId sHash) quantityBundle
+
+note :: MonadFail m => String -> Maybe a -> m a
```
Jimbo4350, line 380: `failWith` or something similar would be a better name.

cardano-api/src/Cardano/Api/Value.hs
```diff
@@ -406,6 +417,9 @@ renderPolicyId (PolicyId scriptHash) = serialiseToRawBytesHexText scriptHash
...
 renderAssetId :: AssetId -> Text
 renderAssetId AdaAssetId = "lovelace"
-renderAssetId (AssetId polId (AssetName assetName))
-  | BS.null assetName = renderPolicyId polId
-  | otherwise         = renderPolicyId polId <> "." <> Text.decodeUtf8 assetName
+renderAssetId (AssetId polId (AssetName "")) = renderPolicyId polId
+renderAssetId (AssetId polId assetName) =
+  renderPolicyId polId <> "." <> renderAssetName assetName
+
+renderAssetName :: AssetName -> Text
```
Jimbo4350, line 424: `renderAssetName` is overkill. We just use `serialiseToRawBytesHexText`.

cardano-cli/src/Cardano/CLI/Mary/ValueParser.hs
```diff
@@ -111,10 +113,15 @@ decimal = do
...
-assetName =
-    toAssetName <$> some alphaNum
-  where
-    toAssetName = AssetName . Text.encodeUtf8 . Text.pack
+assetName = do
+  hexText <- some hexDigit
+  note "AssetName deserisalisation failed" $
+    deserialiseFromRawBytesHex AsAssetName $ BSC.pack hexText
+
+note :: MonadFail m => String -> Maybe a -> m a
```
Jimbo4350, line 121: Why have you re-implemented the same function twice? DRY - Do not Repeat Yourself.

Jimbo4350 (changes requested):
Nice, a few comments

cardano-api/src/Cardano/Api/Value.hs
```diff
@@ -1,6 +1,8 @@
+{-# LANGUAGE BlockArguments #-}
```
Jimbo4350, line 1: Let's remove this pragma. I don't think its work it for the sake of not specifying parentheses in your lamba function.

gitmachtl:
@cblp any updates here from your side? would be nice to see this in 1.31.0 😄

cardano-api/src/Cardano/Api/Value.hs
cblp: done

cardano-api/src/Cardano/Api/Value.hs
cblp: done

cardano-api/src/Cardano/Api/Value.hs
cblp: this name is used in Cardano.Prelude (but with a slightly different type) and https://hackage.haskell.org/package/errors-2.3.0/docs/Control-Error-Util.html#v:note. I consider it somewhat traditional.

cardano-cli/src/Cardano/CLI/Mary/ValueParser.hs
cblp: each one has type incompatible with others

cblp: Parser has no MonadFail

cardano-api/src/Cardano/Api/Value.hs
cblp: done

cardano-api/src/Cardano/Api/Value.hs
cblp: done

cardano-api/src/Cardano/Api/Value.hs
cblp: done

gitmachtl:
how does the output of "Friendly output multi-asset with ASCII version where possible" look like? i would leave it up to the software behind the CLI to decide to show it as readable or as hex/binary. makes it more complicated to post-process if the output is not hex all the time.

cblp:
> how does the output of "Friendly output multi-asset with ASCII version where possible" look like?

Here is an example:
```yaml
mint:
  policy 52dc3d43b6d2465e96109ce75ab61abe5e9c1d8a3c9ce6ff8a3af528:
    asset 736b79 (sky): 142
```

> i would leave it up to the software behind the CLI to decide to show it as readable or as hex/binary. makes it more complicated to post-process if the output is not hex all the time.

`cardano-cli transaction view` is an end-user software that uses Cardano API.

API version of cardano-cli is to be built in #3213

gitmachtl:
ah, so the normal `utxo query` is not affected by this, only in the `transaction view` output?

cblp:
@gitmachtl `utxo query` is affected only in part of switching from text to hex.

gitmachtl:
> @gitmachtl `utxo query` is affected only in part of switching from text to hex.

thats perfect, so hex output in the normal `utxo query` output, also in the JSON version of the output. and friendly names in the `transaction view` output.

cblp:
@dcoutts should I fix docs (couttscoins etc.) in this PR?

cardano-cli/src/Cardano/CLI/Mary/ValueParser.hs
cblp: Sorry, Aeson.Parser has MonadFail, but not MonadError, which is used by Cardano.Prelude.note.

And I can't figure out the right place for MonadFail version, so I repeated the code to save ourselves from bad module structure.

Can you suggest a place for this function?

gitmachtl:
ping 😄

cardano-cli/src/Cardano/CLI/Mary/ValueParser.hs
Jimbo4350: I don't follow. The `note` in `cardano-api/src/Cardano/Api/Value.hs` has the type signature `note :: MonadFail m => String -> Maybe a -> m a` which is identical to the `note` in `cardano-cli/src/Cardano/CLI/Mary/ValueParser.hs`.

Why not put `note` in `Cardano.Api.Utils`?

cardano-api/test/Test/Cardano/Api/Typed/Value.hs
```diff
@@ -65,7 +69,21 @@ canonicalise =
...
+  property $ do
+    v <- forAll genAssetName
+    tripping (Map.singleton v ()) encode eitherDecode
+
+
 -- -----------------------------------------------------------------------------
 
 tests :: TestTree
-tests = fromGroup $$discover
+tests = $testGroupGenerator
```
Jimbo4350, line 89: Why the change to `testGroupGenerator`?

Jimbo4350 (changes requested):
A few comments

cardano-api/src/Cardano/Api/SerialiseUsing.hs
```diff
@@ -63,33 +63,40 @@ instance SerialiseAsRawBytes a => Show (UsingRawBytesHex a) where
...
+    where
+      tname  = (tyConName . typeRepTyCon . typeRep) (Proxy :: Proxy a)
+
+instance SerialiseAsRawBytes a => ToJSONKey (UsingRawBytesHex a) where
+  toJSONKey =
+    Aeson.toJSONKeyText $ \(UsingRawBytesHex x) -> serialiseToRawBytesHexText x
+
+instance
+  (SerialiseAsRawBytes a, Typeable a) => FromJSONKey (UsingRawBytesHex a)
+  where
```
Jimbo4350, line 84: Strange indent here.

`(SerialiseAsRawBytes a, Typeable a) => FromJSONKey (UsingRawBytesHex a) where` should be fine

Jimbo4350:
> @dcoutts should I fix docs (couttscoins etc.) in this PR?
Yes, please update the docs with respect to the serialization change.

cardano-api/src/Cardano/Api/SerialiseUsing.hs
```diff
@@ -63,33 +63,40 @@ instance SerialiseAsRawBytes a => Show (UsingRawBytesHex a) where
...
-        case Base16.decode (Text.encodeUtf8 str) of
-          Right raw -> case deserialiseFromRawBytes ttoken raw of
-            Just x  -> return (UsingRawBytesHex x)
-            Nothing -> fail ("cannot deserialise " ++ show str)
-          Left msg  -> fail ("invalid hex " ++ show str ++ ", " ++ msg)
-      where
-        ttoken = proxyToAsType (Proxy :: Proxy a)
-        tname  = (tyConName . typeRepTyCon . typeRep) (Proxy :: Proxy a)
-
-instance SerialiseAsRawBytes a => ToJSONKey (UsingRawBytesHex a)
```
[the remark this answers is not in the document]
cblp: The default ToJSONKey implementation is the Value-based encoding; we need the Text-based here.

cardano-api/test/Test/Cardano/Api/Typed/Value.hs
cblp: There is already grouping ability in `tasty` library, we don't need our home-grown version.

cardano-cli/src/Cardano/CLI/Mary/ValueParser.hs
cblp: `Cardano.Api.Utils` is internal in `cardano-api`, I use `Cardano.CLI.Helpers`.

cardano-api/src/Cardano/Api/Value.hs
```diff
@@ -70,13 +72,14 @@ import           Data.String (IsString (..))
 import           Data.Text (Text)
 import qualified Data.Text as Text
 import qualified Data.Text.Encoding as Text
+-- import           Data.Traversable (for)
```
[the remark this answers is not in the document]
cblp: removed

cardano-api/test/Test/Cardano/Api/Typed/Value.hs
cblp: Indeed, this change is lateral, reverted. It should be done more generally in a separate PR.

cardano-cli/src/Cardano/CLI/Mary/ValueParser.hs
cblp: Code moved to `Cardano.Api.Utils`

cardano-api/src/Cardano/Api/SerialiseUsing.hs
Jimbo4350: The limit is 100 chars as per the style document

cardano-api/src/Cardano/Api/SerialiseUsing.hs
```diff
@@ -63,33 +63,39 @@ instance SerialiseAsRawBytes a => Show (UsingRawBytesHex a) where
...
+    Aeson.toJSONKeyText $ \(UsingRawBytesHex x) -> serialiseToRawBytesHexText x
+
+instance
+  (SerialiseAsRawBytes a, Typeable a) => FromJSONKey (UsingRawBytesHex a) where
+
+  fromJSONKey =
+    Aeson.FromJSONKeyTextParser $
+    either fail pure . deserialiseFromRawBytesBase16 . Text.encodeUtf8
+
+deserialiseFromRawBytesBase16 ::
```
Jimbo4350, line 89: We now have two implementations of essentially the same thing `deserialiseFromRawBytesHex` and  `deserialiseFromRawBytesBase16`. Not to mention we Base16 decode the input `ByteString` twice in order to get an error message in `deserialiseFromRawBytesBase16`. The proper solution here is to update `deserialiseFromRawBytesHex` to return an `Either`. This refactor would be better as a separate commit.

cardano-api/src/Cardano/Api/Utils.hs
```diff
@@ -42,3 +45,8 @@ runParsecParser parser input =
   case Parsec.parse (parser <* Parsec.eof) "" (Text.unpack input) of
     Right txin -> pure txin
     Left parseError -> fail $ formatParsecError parseError
+
+note :: MonadFail m => String -> Maybe a -> m a
```
Jimbo4350, line 49: :+1:

cardano-api/src/Cardano/Api/Value.hs
Jimbo4350: `note` needs to be removed from here

cardano-cli/src/Cardano/CLI/Helpers.hs
```diff
@@ -126,3 +128,8 @@ hushM :: forall e m a. Monad m => Either e a -> (e -> m ()) -> m (Maybe a)
 hushM r f = case r of
   Right a -> return (Just a)
   Left e -> f e >> return Nothing
+
+note :: MonadFail m => String -> Maybe a -> m a
```
Jimbo4350: This needs to be deleted from here.

Jimbo4350 (approved):
Nice :+1:

gitmachtl:
@JaredCorduan  @newhoggy maybe someone of you is having the time to review it also?

cardano-api/src/Cardano/Api/SerialiseUsing.hs
cblp: Here it is #3304

gitmachtl:
do we have a change to get this into the 1.31.0 release?

gitmachtl:
@nc6 do we have a change to get this into the 1.31.0 release?

cblp:
bors merge

gitmachtl:
wohoo! :-)

PR #3223, opened 2021-09-20
tx-generator: tx-generator.service for Plutus

MarcFontaine:
tx-generator: Add Plutus support to JSON config.
* Add payMode.PayToScript and spendMode.SpendScript
* JSON-configurable execution units.
* Use protocolParamPrices in tx-generator.
* Better handling of collaterals in tx-generator.
* Update tx-generator-service.nix and hard-coded test-scripts.
* TODO: configurable ScriptRedemer and ScriptDatum.


nix/nixos/tx-generator-service.nix
```diff
@@ -1,5 +1,7 @@
 pkgs:
 let
+  plutus = true;
+  plutusScriptFile = "../../../bench/script/sum1ToN.plutus";
```
MarcFontaine, line 4: Here the Plutus settings are hardcoded.
Instead proper parameters from the nix-service definition should be used instead.

deepfire (approved):
Looks good, thank you Marc!

PR #3253, opened 2021-09-27
Add missing AsType for current Tx eras

koslambrou:
Motivation: Trying to deserialize a `Tx era` with `deserialiseFromTextEnvelope` requires as parameter of type `AsType`. Currently we only have `AsByronTx` and `AsShelleyTx`. Therefore, this PR extends this to all current eras.


dcoutts (changes requested):
Thanks. What we should do is deprecate `pattern AsByronTx` and `pattern AsShelleyTx`. They should be deprecated in favour of `AsTx AsByronEra` and `AsTx AsShelleyEra`. And all the other eras follow that pattern, using `AsTx` with the "As" type for the era.

koslambrou:
@dcoutts Thanks for the comment! I deprecated instead those patterns.

newhoggy (approved):

dcoutts (approved):
LGTM! Thanks!

koslambrou:
bors merge

PR #3255, opened 2021-09-27
Include script context test in ci

Jimbo4350:
This PR includes the script context equivalence test in CI.

### DIFF
--- a/plutus-example/plutus-example/app/create-script-context.hs
+++ b/plutus-example/plutus-example/app/create-script-context.hs
@@ -2,6 +2,7 @@ import           Prelude
 
 import           Cardano.Api
 
+import           Control.Monad.IO.Class
 import           Control.Monad.Trans.Except (runExceptT)
 import qualified Data.ByteString.Lazy as LB
 import           Data.Foldable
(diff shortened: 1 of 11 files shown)


newhoggy:
If you're introducing a new binary to be called from test you will need to update some `nix` expressions.

To find the locations you need to change search for `CARDANO_CLI` and `packages.cardano-cli` in `*.nix` files.

plutus-example/plutus-example/app/create-script-context.hs
```diff
@@ -41,22 +46,32 @@ parseScriptContextCmd = parseGenerateDummy <|> parseGenerateTxBody
...
+runScriptContextCmd (GenerateDummyScriptContextRedeemer outFp) =
+  LB.writeFile outFp sampleTestScriptContextDataJSON
+runScriptContextCmd (GenerateScriptContextRedeemerTxBody txbodyfile cModeParams nid outFp) = do
       eTxBodyRedeemer <- runExceptT $ txToRedeemer txbodyfile cModeParams nid
       case eTxBodyRedeemer of
-        Left err -> print err
-        Right () -> return ()
+        Left err -> do liftIO . LB.writeFile outFp $ "FAILED: " <> C8.pack (show err)
+                       error $ "Error creating redeemer from: " <> txbodyfile <>
+                               " Error: " <> show err
```
newhoggy, line 72: Why do we write the error to the output file rather than just standard output?

plutus-example/plutus-example/src/Cardano/PlutusExample/ScriptContextChecker.hs
```diff
@@ -360,14 +359,11 @@ txToRedeemer
   :: FilePath
   -> AnyConsensusModeParams
   -> NetworkId
-  -> ExceptT ScriptContextError IO ()
-txToRedeemer txFp anyCmodeParams nid = do
+  -> ExceptT ScriptContextError IO LB.ByteString
+txToRedeemer txFp anyCmodeParams nid  = do
```
newhoggy, line 363: Seems like an extra space got introduced here.

Jimbo4350:
bors merge

PR #3267, opened 2021-10-05
CDA-717 decodeUtf8With to ignore invalid chars

CodiePP:
the problem: invalid chars in native tokens' asset names were making the cli fail. For example, it was not possible to dump the UTxO set anymore.
the solution: byte-encoded characters in these names, if invalid, will be ignoredd when decoded to UTF-8.  Now, the dump of UTxO set is working again.


Jimbo4350:
This has been fixed in `master`

PR #3274, opened 2021-10-06
Including minting script context test in CI

Jimbo4350:


cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -387,6 +397,26 @@ instance ToJSON TxIn where
 instance ToJSONKey TxIn where
   toJSONKey = toJSONKeyText renderTxIn
 
+instance FromJSON TxIn where
+  parseJSON = withText "TxIn" $ \txinStr -> runParsecParser parseTxIn txinStr
+
+parseTxId :: Parsec.Parser TxId
+parseTxId = do
+  str <- Parsec.many1 Parsec.hexDigit Parsec.<?> "transaction id (hexadecimal)"
```
newhoggy, line 405: So it takes two hex digits to make a byte.  I'm curious what happens if we supply an odd number of digits?

cardano-testnet/test/Spec/Plutus/Direct/ScriptContextEqualityMint.hs
```diff
@@ -0,0 +1,268 @@
...
+import           Testnet.Cardano (defaultTestnetOptions, testnet)
+import qualified Testnet.Cardano as TC
+import qualified Testnet.Conf as H
+
+{- HLINT ignore "Redundant <&>" -}
+{- HLINT ignore "Redundant return" -}
+{- HLINT ignore "Use let" -}
+
+hprop_plutus_script_context_mint_equality :: Property
+hprop_plutus_script_context_mint_equality = H.integration . H.runFinallies . H.workspace "chairman" $ \tempAbsBasePath' -> do
```
newhoggy, line 45: I'm thinking rename the shell script name to match the haskell code or the reverse.  Currently one says equality and the other says equivalence.

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -970,6 +1047,47 @@ instance ToJSON (TxOutValue era) where
...
+           pure $ valueFromList [(AdaAssetId, Quantity $ toInteger ll)]
+         Nothing ->
+           fail $ "Expected a Bounded number but got: " <> show sci
+     decodeAssetId wrong = fail $ "Expected a policy id and a JSON object but got: " <> show wrong
+
+     decodeAssets :: Aeson.Value -> Aeson.Parser [(AssetName, Quantity)]
+     decodeAssets (Aeson.Object assetNameHm) =
+       let l = HMS.toList assetNameHm
+       in mapM (\(aName, q) -> (,) <$> parseAssetName aName <*> decodeQuantity q) l
+     decodeAssets wrong = fail $ "Expected aeso object but got: " <> show wrong
```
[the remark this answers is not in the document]
Jimbo4350: Good point, I'll make the change :+1:

cardano-api/src/Cardano/Api/TxBody.hs
Jimbo4350: `parseTxId` fails but:
```
>let testParser = many1 hexDigit <?> "transaction id (hexadecimal)"
>parse testParser "" "a"

<interactive>:25:21: warning: [-Wtype-defaults]
    • Defaulting the following constraint to type ‘[ghc-prim-0.6.1:GHC.Types.Char]’
        Data.String.IsString s0 arising from the literal ‘"a"’
    • In the third argument of ‘parse’, namely ‘"a"’
      In the expression: parse testParser "" "a"
      In an equation for ‘it’: it = parse testParser "" "a"

<interactive>:25:21: warning: [-Wtype-defaults]
    • Defaulting the following constraint to type ‘[ghc-prim-0.6.1:GHC.Types.Char]’
        Data.String.IsString s0 arising from the literal ‘"a"’
    • In the third argument of ‘parse’, namely ‘"a"’
      In the expression: parse testParser "" "a"
      In an equation for ‘it’: it = parse testParser "" "a"
Right "a"

Jimbo4350:
bors merge

PR #3287, opened 2021-10-11
trace-dispatcher: active/passive mode.

denisshevchenko:
Now `trace-dispatcher` can _initiates_ connection with `cardano-tracer` or _accepts_ connection from `cardano-tracer`. So now we have two scenarios:

1. passive node <--- active tracer 
2. active node ---> passive tracer


deepfire (approved):
LGTM!

PR #3300, opened 2021-10-19
EKG counter for tips served

karknu:
Add an EKG counter, cardano.node.metrics.served.block.latest.count, which
can be used to track how usefull the node is to its down stream peers.

This counter can be used as follows `rate(cardano_node_metrics_served_block_latest_count_int[1h])/rate(cardano_node_metrics_blockNum_int[1h])` . Which gives a value close to 0 for unpopular node. And a value around 1 or more for a very useful node.

### DIFF
--- a/cardano-node/src/Cardano/Tracing/Tracers.hs
+++ b/cardano-node/src/Cardano/Tracing/Tracers.hs
@@ -498,9 +498,6 @@ isRollForward :: TraceChainSyncServerEvent blk -> Bool
 isRollForward (TraceChainSyncRollForward _) = True
 isRollForward _ = False
 
-isTraceBlockFetchServerBlockCount :: TraceBlockFetchServerEvent blk -> Bool
-isTraceBlockFetchServerBlockCount (TraceBlockFetchServerSendBlock _) = True
-
 mkConsensusTracers
   :: forall blk peer localPeer.
      ( Show peer
@@ -533,7 +530,9 @@ mkConsensusTracers mbEKGDirect trSel verb tr nodeKern fStats = do
   forgeTracers <- mkForgeTracers
   meta <- mkLOMeta Critical Public
 
-  tBlocksServed <- STM.newTVarIO @Int 0
+  tBlocksServed <- STM.newTVarIO 0
+  tLocalUp <- STM.newTVarIO 0
+  tMaxSlotNo <- STM.newTVarIO $ SlotNo 0
   tSubmissionsCollected <- STM.newTVarIO 0
   tSubmissionsAccepted <- STM.newTVarIO 0
   tSubmissionsRejected <- STM.newTVarIO 0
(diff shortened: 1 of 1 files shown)


cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -621,6 +616,54 @@ mkConsensusTracers mbEKGDirect trSel verb tr nodeKern fStats = do
...
+traceBlockFetchServerMetrics (Just ekgDirect) tBlocksServed tLocalUp tMaxSlotNo tracer = Tracer bsTracer
+
+  where
+    bsTracer :: TraceBlockFetchServerEvent blk -> IO ()
+    bsTracer e@(TraceBlockFetchServerSendBlock p) = do
+      traceWith tracer e
+      let mSlotNo = case pointSlot p of
+                         Origin -> Nothing
+                         At slotNo -> Just slotNo
+
```
coot, line 637: It doesn't seem necessary to transform one maybe type into another one.  A few lines below we pattern match on `mSlotNo`, equally well we can just pattern match on `pointSlot p`.

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -621,6 +616,54 @@ mkConsensusTracers mbEKGDirect trSel verb tr nodeKern fStats = do
...
+          served <- STM.modifyReadTVar' tBlocksServed (+1)
+          maxSlotNo <- STM.readTVar tMaxSlotNo
+          case mSlotNo of
+               Nothing ->
+                   return (served, Nothing)
+               Just slotNo ->
+                   case compare maxSlotNo slotNo of
+                        LT -> do
+                            STM.writeTVar tMaxSlotNo slotNo
+                            lu <- STM.modifyReadTVar' tLocalUp (+1)
```
coot, line 648: What is `modifyReadTVar'`?

Is it equialvent to:
```
STM.modifyTVar' tLocalUP (+1) >> STM.readTVar tLocalUP
```

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -621,6 +616,54 @@ mkConsensusTracers mbEKGDirect trSel verb tr nodeKern fStats = do
...
+                            return (served, Just lu)
+                        GT -> do
+                            return (served, Nothing)
+                        EQ -> do
+                            lu <- STM.modifyReadTVar' tLocalUp (+1)
+                            return (served, Just lu)
+      sendEKGDirectInt ekgDirect
```
coot, line 655: Please add an empty line above.

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -621,6 +616,54 @@ mkConsensusTracers mbEKGDirect trSel verb tr nodeKern fStats = do
...
+  -> STM.TVar Int64
+  -> STM.TVar Int64
+  -> STM.TVar SlotNo
+  -> Tracer IO (TraceBlockFetchServerEvent blk)
+  -> Tracer IO (TraceBlockFetchServerEvent blk)
+traceBlockFetchServerMetrics Nothing _ _ _ tracer = tracer
+traceBlockFetchServerMetrics (Just ekgDirect) tBlocksServed tLocalUp tMaxSlotNo tracer = Tracer bsTracer
+
+  where
+    bsTracer :: TraceBlockFetchServerEvent blk -> IO ()
```
coot, line 631: This adds the local upstreaminess metric which tries to answer the question: how many times we serve a block which slot is either equal or ahead of last served block.

I would like to see a description how this related to the upstreaminess metric that is used to choose peers.

cardano-node/src/Cardano/Tracing/Tracers.hs
karknu: https://github.com/input-output-hk/cardano-node/blob/aa7d41199227029cf80b1d5672c5401d1c73f57f/cardano-node/src/Cardano/Node/STM.hs#L24-L30

cardano-node/src/Cardano/Tracing/Tracers.hs
karknu: I don't think we should mention that at all. This is a useful metric on its own, regardless of if that is how down stream peers may judge the node in the future.

dorin100:
more useful details for this new metric:
```
A value of 0 means that no one is fetching blocks from your relay.
A value of 1 means that for every new block forged someone is fetching it.
Above 2 means that for every block forged two nodes are fetching that block from you
```

PR #3303, opened 2021-10-20
CAD-3511: datapoint-forward library.

denisshevchenko:
The third forwarding library for the new tracing infrastructure (https://github.com/input-output-hk/cardano-node/pull/2847). It allows forwarding arbitrary data that provides `ToJSON` instance. The `trace-dispatcher` will use this library to forward different structured data to external acceptor apps (for example, `cardano-tracer`).

**For reviewers**: please note that this library is based on the existing `trace-forward` library, but with one important difference: there is no "infinite stream" of data when we ask for it every `N` seconds. Instead, acceptor app asks for each `DataPoint` explicitly, when it needs this data.


datapoint-forward/test/Test/DataPoint/Forward/Types.hs
```diff
@@ -0,0 +1,25 @@
...
+module Test.DataPoint.Forward.Types
+  ( NodeInfo (..)
+  , BlockchainStatus (..)
+  ) where
+
+import qualified Data.Aeson as A
+import           Data.Text (Text)
+import           GHC.Generics
+
+data NodeInfo = NodeInfo
```
deepfire, line 13: @denisshevchenko, can we rename it so it doesn't cause "grep conflicts" with the real `NodeInfo`?

denisshevchenko: Yes, I'll do it.

denisshevchenko:
Please note that after it will be merged, `NodeInfo` will be removed from `trace-forward` library (https://github.com/input-output-hk/cardano-node/blob/master/trace-forward/src/Trace/Forward/Protocol/Type.hs#L125-L132), because it will be forwarded as one of `DataPoint`s.

Jimbo4350 (approved):

datapoint-forward/src/DataPoint/Forward/Utils.hs
```diff
@@ -0,0 +1,131 @@
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NamedFieldPuns #-}
+
+module DataPoint.Forward.Utils
```
coot, line 5: I would rename it as `DataPoint.Forward.Store`

datapoint-forward/src/DataPoint/Forward/Utils.hs
```diff
@@ -0,0 +1,131 @@
...
+import           Data.Aeson
+import qualified Data.HashMap.Strict as HM
+import           System.Time.Extra (sleep)
+
+import           DataPoint.Forward.Configuration
+import           DataPoint.Forward.Protocol.Forwarder
+import           DataPoint.Forward.Protocol.Type
+
+-- | Type wrapper for some value of type 'v'. The only reason we need this
+--   wrapper is an ability to store different values in the same 'DataPointStore'.
```
coot, line 30: Note that this has consequences: when you read if from the store you will only be able to access it as unstructured `JSON` unlike Haskell algebraic data types.  It's probably fine for your application, but it might be worth mentioning the trade-off.

datapoint-forward/src/DataPoint/Forward/Utils.hs
```diff
@@ -0,0 +1,131 @@
...
+data DataPoint where
+  DataPoint :: ToJSON v => v -> DataPoint
+
+type DataPointStore = TVar (HM.HashMap DataPointName DataPoint)
```
coot, line 34: A `Map` is preferred, especially if this will run on the `cardano-node` side.

datapoint-forward/test/Test/DataPoint/Forward/Protocol/Tests.hs
```diff
@@ -0,0 +1,27 @@
...
+import           Network.TypedProtocol.Codec
+
+import           DataPoint.Forward.Protocol.Codec
+import           DataPoint.Forward.Protocol.Type
+
+import           Test.DataPoint.Forward.Protocol.Codec ()
+
+tests :: TestTree
+tests = testGroup "DataPoint.Forward.Protocol"
+  [ testProperty "codec" prop_codec_DataPointForward
```
coot, line 19: You should expand the test in this module. Please check [keep-alive](https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-network/protocol-tests/Ouroboros/Network/Protocol/KeepAlive/Test.hs) test suite as a template.

datapoint-forward/datapoint-forward.cabal
```diff
@@ -0,0 +1,91 @@
+cabal-version:       2.4
+name:                datapoint-forward
+version:             0.1.0
+synopsis:            See README for more info
```
coot, line 4: Please provide one sentence info about the library.

datapoint-forward/datapoint-forward.cabal
```diff
@@ -0,0 +1,91 @@
...
+description:         See README for more info
```
coot, line 5: There's no `README` included, so I'd suggest to actually provide a description.

coot (changes requested):
Could you provide a high level description how does it fit into node and the `trace-forwarder` library.

I provided various feedback, the most important is to not use `UnversionedProtocol`, but provided it's own versioning scheme which would distniguish it from all other protocols (`trace-forwarder`, `node-to-client` or even `node-to-node`).

datapoint-forward/src/DataPoint/Forward/Protocol/Codec.hs
```diff
@@ -0,0 +1,83 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE PolyKinds #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module DataPoint.Forward.Protocol.Codec
+  ( codecDataPointForward
```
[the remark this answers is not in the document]
denisshevchenko: Yes, it's for Unix sockets only.

denisshevchenko:
Closed because of https://github.com/input-output-hk/cardano-node/pull/3342

PR #3305, opened 2021-10-21
Add EKG counter for number of forks seen

karknu:
Add cardano_node_metrics_forks_int for counting the number of forks seen
by the node.

### DIFF
--- a/cardano-node/src/Cardano/Tracing/OrphanInstances/Consensus.hs
+++ b/cardano-node/src/Cardano/Tracing/OrphanInstances/Consensus.hs
@@ -686,6 +686,8 @@ instance ( ConvertRawHash blk
                [ "kind" .= String "TraceAddBlockEvent.SwitchedToAFork"
                , "newtip" .= renderPointForVerbosity verb (AF.headPoint new)
                , "chainLengthDelta" .= new `chainLengthΔ` old
+               -- Check that the SwitchedToAFork event was triggered by a proper fork.
+               , "realFork" .= not (AF.withinFragmentBounds (AF.headPoint old) new)
                ]
             ++ [ "headers" .= toJSON (toObject verb `map` addedHdrsNewChain old new)
                | verb == MaximalVerbosity ]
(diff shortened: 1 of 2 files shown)


cardano-node/src/Cardano/Tracing/OrphanInstances/Consensus.hs
```diff
@@ -686,6 +686,7 @@ instance ( ConvertRawHash blk
                [ "kind" .= String "TraceAddBlockEvent.SwitchedToAFork"
                , "newtip" .= renderPointForVerbosity verb (AF.headPoint new)
                , "chainLengthDelta" .= new `chainLengthΔ` old
+               , "realFork" .= not (AF.withinFragmentBounds (AF.headPoint old) new)
```
nfrisby, line 689: Maybe rename this to `"sanityCheck"`? Or add a comment explaining that we're just double-checking that the `SwitchedToAFork` event really was incurred by a proper fork.

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -417,28 +420,29 @@ traceChainMetrics
...
-traceChainMetrics Nothing _ _ _ = nullTracer
-traceChainMetrics (Just _ekgDirect) _blockConfig _fStats tr = do
+traceChainMetrics Nothing _ _ _ _ = nullTracer
+traceChainMetrics (Just _ekgDirect) tForks _blockConfig _fStats tr = do
   Tracer $ \ev ->
     fromMaybe (pure ()) $ doTrace <$> chainTipInformation ev
   where
     chainTipInformation :: ChainDB.TraceEvent blk -> Maybe ChainInformation
     chainTipInformation = \case
       ChainDB.TraceAddBlockEvent ev -> case ev of
```
nfrisby, line 435: If we generally trust that `SwitchedToAFork` is only emitted when it should be, then the changes to this case expression could be simpler; we just need to pass a `True` to `chainInformation` in one branch and a `False` in the other.

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -1214,21 +1221,29 @@ data ChainInformation = ChainInformation
...
   => ChainDB.NewTipInfo blk
+  -> Maybe (AF.AnchoredFragment (Header blk))
   -> AF.AnchoredFragment (Header blk)
   -> Int64
   -> ChainInformation
-chainInformation newTipInfo frag blocksUncoupledDelta = ChainInformation
-    { slots = unSlotNo $ fromWithOrigin 0 (AF.headSlot frag)
-    , blocks = unBlockNo $ fromWithOrigin (BlockNo 1) (AF.headBlockNo frag)
-    , density = fragmentChainDensity frag
+chainInformation newTipInfo oldFrag_m newFrag blocksUncoupledDelta = ChainInformation
```
nfrisby, line 1235: See my related comment in `traceChainMetrics`: we could get by with just a new `Bool` argument instead.

PR #3316, opened 2021-10-25
Remove mainnet mirror

newhoggy:


newhoggy:
Already merged as part of https://github.com/input-output-hk/cardano-node/pull/3334

PR #3325, opened 2021-10-30
Upgrade to cabal 3.6.2.0

newhoggy:


.github/workflows/haskell.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       id: setup-haskell
       with:
         ghc-version: ${{ matrix.ghc }}
-        cabal-version: 3.4.0.0
```
Jimbo4350, line 45: what about `.github/workflows/github-page.yml`?

PR #3334, opened 2021-11-02
Temporarily disable problem test on Windows

newhoggy:


cardano-testnet/test/Spec/Plutus/Script/TxInLockingPlutus.hs
```diff
@@ -47,28 +48,29 @@ hprop_plutus = H.integration . H.runFinallies . H.workspace "chairman" $ \tempAb
...
-          , ("UTXO_VKEY", tempAbsPath </> "shelley/utxo-keys/utxo1.vkey")
-          , ("UTXO_SKEY", tempAbsPath </> "shelley/utxo-keys/utxo1.skey")
-          , ("CARDANO_NODE_SOCKET_PATH", IO.sprocketArgumentName (head bftSprockets))
-          , ("TESTNET_MAGIC", show @Int testnetMagic)
-          , ("PATH", path)
-          , ("RESULT_FILE", resultFile)
-          ]
-        , H.execConfigCwd = Last $ Just tempBaseAbsPath
-        }
+  unless (OS.isWin32) $ do
```
Jimbo4350, line 51: I think it would be more obvious to use `ignoreOnWindows` in `cardano-testnet/test/Main.hs` instead rather than having `OS.isWin32` buried here.

newhoggy: Fixed.  Thanks!

.github/workflows/haskell.yml
```diff
@@ -137,9 +137,6 @@ jobs:
     - name: Build
       run: retry 2 cabal build cardano-node cardano-cli cardano-node-chairman cardano-submit-api plutus-example
 
-    - name: Git clone
```
Jimbo4350, line 140: This should result in a significant speed up right?

Jimbo4350 (approved):

cardano-testnet/test/Main.hs
```diff
@@ -25,7 +25,8 @@ tests = do
   pure $ T.testGroup "test/Spec.hs"
     [ T.testGroup "Spec"
       [ H.testProperty "Spec.Plutus.Direct.TxInLockingPlutus" Spec.Plutus.Direct.TxInLockingPlutus.hprop_plutus
-      , H.testProperty "Spec.Plutus.Script.TxInLockingPlutus" Spec.Plutus.Script.TxInLockingPlutus.hprop_plutus
+      -- This hangs on Windows for an unknown reason
```
cblp, line 28: please open an issue and put its number here

.github/workflows/haskell.yml
newhoggy: Yes, but also CI currently fails because the repository is many gigabytes large and the CI VM runs out of disk space.

newhoggy:
bors merge

PR #3337, opened 2021-11-02
CAD-3628: remove NodeInfo from trace-forward.

denisshevchenko:
Since the new `datapoint-forward` library (https://github.com/input-output-hk/cardano-node/pull/3303) will be used to forward structured data from the node, `NodeInfo` type should be removed from `trace-forward` library.


deepfire (approved):
Oh yes, that gives me goosebumps -- thank you @denisshevchenko !

This was a really not-so-nice corner case we had to handle..

:heart: :+1:

nc6 (approved):

PR #3345, opened 2021-11-04
Update dependencies

erikd:


cardano-api/src/Cardano/Api/Fees.hs
```diff
@@ -495,8 +496,8 @@ evaluateTransactionExecutionUnits _eraInMode systemstart history pparams utxo tx
              (toLedgerEpochInfo history)
              systemstart
              (toAlonzoCostModels (protocolParamCostModels pparams))
-        of Left  err   -> Left err
-           Right exmap -> Right (fromLedgerScriptExUnitsMap exmap)
+        of Left  err   -> error $ show err
+           Right exmap -> Right (fromLedgerScriptExUnitsMap $ fromRight mempty exmap)
```
erikd, line 500: This still needs to be fixed.

cardano-api/src/Cardano/Api/LedgerEvent.hs
```diff
@@ -153,23 +155,33 @@ pattern LEMirTransfer rp tp rtt ttr <-
...
+                    (convertSumRewardsMapSimple -> tp)
                     (fromShelleyDeltaLovelace -> rtt)
                     (fromShelleyDeltaLovelace -> ttr)
                   )
               )
           )
       )
 
-convertSumRewardsMap ::
+convertRewardEventMapNew ::
```
nc6, line 167: Can we drop the `New` here? The names don't clash

cardano-api/src/Cardano/Api/LedgerEvent.hs
```diff
@@ -153,23 +155,33 @@ pattern LEMirTransfer rp tp rtt ttr <-
...
+
+convertSumRewardsMapSimple ::
   Map
     ( Cardano.Ledger.Credential.StakeCredential
         Cardano.Ledger.Crypto.StandardCrypto
     )
     Cardano.Ledger.Coin.Coin ->
   Map StakeCredential Lovelace
-convertSumRewardsMap =
+convertSumRewardsMapSimple =
```
nc6, line 184: Likewise with `Simple` here

nc6 (changes requested):
Minor changes requested, otherwise looks fine

cardano-node/src/Cardano/Tracing/OrphanInstances/Shelley.hs
```diff
@@ -395,6 +391,9 @@ instance ( ShelleyBasedEra era
   toObject _verb InvalidMetadata =
     mkObject [ "kind" .= String "InvalidMetadata"
              ]
+  toObject _verb (ExtraneousScriptWitnessesUTXOW _) =
+    mkObject [ "kind" .= String "ExtraneousScriptWitnessesUTXOW"
```
nc6, line 395: This drops the extra information provided - we should also print the set of extra scripts.

cardano-api/src/Cardano/Api/Fees.hs
```diff
@@ -488,15 +489,14 @@ evaluateTransactionExecutionUnits _eraInMode systemstart history pparams utxo tx
                          (Map ScriptWitnessIndex
                               (Either ScriptExecutionError ExecutionUnits))
     evalAlonzo era tx =
-      case Alonzo.evaluateTransactionExecutionUnits
+      fromLedgerScriptExUnitsMap . fromRight mempty <$>
```
JaredCorduan, line 492: We should at the very least leave a `TODO` comment saying that this `fromRight mempty` is hiding a basic validation error in `Alonzo.evaluateTransactionExecutionUnits`. This will be very confusing to anyone hitting the `Left` case here!

cardano-api/src/Cardano/Api/LedgerEvent.hs
erikd: We need that `Simple` to differentiate it from `convertRewardEventMap` which is different.

cardano-api/src/Cardano/Api/LedgerEvent.hs
erikd: There are two functions named `convertSumRewardsMap*`. How would you differentiate them?

cardano-api/src/Cardano/Api/LedgerEvent.hs
nc6: They have different names - `convertRewardEventMap` and `convertSumRewardsMap`

disassembler:
bors try

disassembler:
bors try

PR #3347, opened 2021-11-05
Expose textEnvelopeToJSON

hasufell:
Otherwise we're forced to use writeFileTextEnvelope
or duplicate internal functions, without a clean option
to e.g. print to stdout.


newhoggy:
Thanks for the PR.  The function is exported in this PR: https://github.com/input-output-hk/cardano-node/pull/3354

newhoggy:
Thanks for the PR. Closing now as it is implemented in the other PR.

PR #3349, opened 2021-11-08
Disable idle GC in the default nix build

weebl2000:
I guess this is what's causing cardano-node constantly high CPU usage when running in Daedalus.

My suggestion would be to also add "-c" - always compact oldest generation instead of copying collection. Since most heap data is very long-lived, this can cause considerably less memory usage.

### DIFF
--- a/nix/nixos/cardano-node-service.nix
+++ b/nix/nixos/cardano-node-service.nix
@@ -404,7 +404,7 @@ in {
 
       rtsArgs = mkOption {
         type = types.listOf types.str;
-        default = [ "-N2" "-A16m" "-qg" "-qb" "--disable-delayed-os-memory-return" ];
+        default = [ "-N2" "-I0" "-A16m" "-qg" "-qb" "--disable-delayed-os-memory-return" ];
         apply = args: if (args != [] || cfg.profilingArgs != []) then
           ["+RTS"] ++ cfg.profilingArgs ++ args ++ ["-RTS"]
           else [];


weebl2000:
@jbgi Not sure who to tag, but I've seen you've worked on this file. :)

jbgi:
I think this rts option is already set at compile time:
https://github.com/input-output-hk/cardano-node/blob/2cbe363874d0261bc62f52185cf23ed492cf4859/cardano-node/cardano-node.cabal#L166

weebl2000:
> I think this rts option is already set at compile time:
> 
> https://github.com/input-output-hk/cardano-node/blob/2cbe363874d0261bc62f52185cf23ed492cf4859/cardano-node/cardano-node.cabal#L166
> I think this rts option is already set at compile time:
> 
> https://github.com/input-output-hk/cardano-node/blob/2cbe363874d0261bc62f52185cf23ed492cf4859/cardano-node/cardano-node.cabal#L166


Hmm. I thought it was maybe overriding the default, but apparently not. I was trying to debug running the Daedalus / cardano-node combination but no luck so far with the nix build. But it seems cardano-node is always using 100% of one core when running from Daedalus. I'll try to investigate further.

weebl2000:
> I think this rts option is already set at compile time:
> 
> https://github.com/input-output-hk/cardano-node/blob/2cbe363874d0261bc62f52185cf23ed492cf4859/cardano-node/cardano-node.cabal#L166

One more question. I'm not very familiair with nixos or native binaries and their linking to libraries and stuff. But do you think that running a `strip` and `patchelf --shrink-rpath` might remove the default GHC RTS options assigned at compile time?

weebl2000:
I'm getting closer, it's not here. :P

weebl2000:
@jbgi It seems the NixOS parameters are used when running from NixOS. I think -I0 would be a good addition still.

image

jbgi (approved):

jbgi:
bors retry

weebl2000:
@jbgi Do I have to make any changes to fix the build?

PR #3352, opened 2021-11-09
bench:  generalised result processing

deepfire:
This introduces a basic generalised result processor that:

- accepts a bunch of directories with JSOH representing the run result matrix (profiles x iterations)
- classifies per-run data into inputs and outputs
- performs a statistical tally on the run outputs
- outputs one of the implemented table formats (CSV, github markdown)

The format specification is reasonably generic, and currently supports just the
`cardano-memory-benchmark` output -- see bench/membenches_v1.jq

### DIFF
--- a/bench/process/membenches_v1.jq
+++ b/bench/process/membenches_v1.jq
@@ -1,0 +1,39 @@
+def format_specs:
+
+{ config_specs:
+    [{ key: "config",  header: "Config name"
+     , path: ["config_name"]}
+    ,{ key: "flags",   header: "RTS Flags"
+     , path: ["data", "flags"]
+     }
+    ,{ key: "failed",  header: "Failed"
+     , path: ["data", "failed"]
+     # , post: not
+     }]
+, result_specs:
+    [{ key: "RssAvg",  header: "Avg RSS, MB"
+     , path: ["RSS",  "avg"], round: true
+     }
+    ,{ key: "RssMax",  header: "Max RSS, MB"
+     , path: ["RSS",  "max"], round: true
+     }
+    ,{ key: "HeapAvg", header: "Avg heap, MB"
+     , path: ["Heap", "avg"], round: true
+     }
+    ,{ key: "HeapMax", header: "Max heap, MB"
+     , path: ["Heap", "max"], round: true
+     }
+    ,{ key: "WallSec", header: "Wall, s"
+     , path: ["totaltime"], round: true
+     }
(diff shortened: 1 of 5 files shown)


bench/process/membenches_v1.jq
```diff
@@ -0,0 +1,39 @@
...
+    ,{ key: "RssMax",  header: "Max RSS, MB"
+     , path: ["RSS",  "max"], round: true
+     }
+    ,{ key: "HeapAvg", header: "Avg heap, MB"
+     , path: ["Heap", "avg"], round: true
+     }
+    ,{ key: "HeapMax", header: "Max heap, MB"
+     , path: ["Heap", "max"], round: true
+     }
+    ,{ key: "WallSec", header: "Wall, s"
```
jutaro, line 26: Why not sec instead of s?

bench/process/process.jq
```diff
@@ -0,0 +1,54 @@
...
+## Aggregation
+##
+def mean:     if length == 0
+                then 0
+                else reduce .[] as $n (0; . + $n) / length end;
+def pow2:     . * .;
+def variance: . | mean as $mean | map_values(. - $mean | pow2) | mean;
+def stddev:   . | (variance ?) | sqrt;
+
+def samples_to_variable (n):
```
jutaro, line 11: May at comments to the more elaborate functions? Specially for the ones that don't know jq to well, like me ;)

bench/process/process.sh
```diff
@@ -0,0 +1,121 @@
...
+        render )     op_render;;
+
+        report )     op_collect "$@" | op_process | op_render;;
+
+        call )       eval "$@";;
+        * ) echo "ERROR:  operation must be one of:  collect process render report" >&2; exit 1;; esac
+}
+
+hardcoded_branch='membench'
+hardcoded_commit='a7ee17d1af44b571c6e476916bd24ed65db97e15'
```
jutaro, line 43: Why is it necessary to put it here? Seams to me that it breaks the abstract nature of the other part of the script.

jutaro (approved):
Looks good to me. Some high level comments in the bash and jq scripts would be extremly helpfull for understanding it.

bench/process/process.sh
deepfire: Oh, yes, I've gone out of my way to make it painfully obvious so I have a chance to do it later.

Perhaps later has finally occured! : -)

MarcFontaine (approved):
LGTM But Jürgen had some comments.

PR #3358, opened 2021-11-10
Better error message for query utxo 2

newhoggy:
This PR adds better messages for `query utxo`.  It replaces an earlier PR https://github.com/input-output-hk/cardano-node/pull/2957.

This PR improves on the earlier version by making it so that the monadic expressions no longer need to know what version each query requires.  If the node that is connected to does not support the query being made, then the query expression throws `QueryErrorUnsupportedVersion` which can be handled in a convenient top-level location.

### DIFF
--- a/cardano-api/src/Cardano/Api/IPC.hs
+++ b/cardano-api/src/Cardano/Api/IPC.hs
@@ -67,7 +67,10 @@ module Cardano.Api.IPC (
     ConsensusMode(..),
     consensusModeOnly,
 
-    NodeToClientVersion(..)
+    NodeToClientVersion(..),
+
+    QueryError(..),
+    MinNodeToClientVersion
   ) where
 
 import           Prelude
@@ -159,6 +162,14 @@ data LocalNodeConnectInfo mode =
        localNodeNetworkId       :: NetworkId,
        localNodeSocketPath      :: FilePath
      }
+
+type MinNodeToClientVersion = NodeToClientVersion
+
+data QueryError a
+  = QueryErrorAcquireFailure !Net.Query.AcquireFailure
+  | QueryErrorUnsupportedVersion !MinNodeToClientVersion !NodeToClientVersion
+  | QueryErrorOf a
+  deriving (Eq, Show)
 
 localConsensusMode :: LocalNodeConnectInfo mode -> ConsensusMode mode
 localConsensusMode LocalNodeConnectInfo {localConsensusModeParams} =
@@ -475,35 +486,32 @@ convLocalStateQueryClient mode =
 -- | Establish a connection to a node and execute a single query using the
 -- local state query protocol.
 --
-queryNodeLocalState :: forall mode result.
+queryNodeLocalState :: forall e mode result.
                        LocalNodeConnectInfo mode
                     -> Maybe ChainPoint
                     -> QueryInMode mode result
-                    -> IO (Either Net.Query.AcquireFailure result)
+                    -> IO (Either (QueryError e) result)
 queryNodeLocalState connctInfo mpoint query = do
     resultVar <- newEmptyTMVarIO
     connectToLocalNode
       connctInfo
-      LocalNodeClientProtocols {
-        localChainSyncClient    = NoLocalChainSyncClient,
-        localStateQueryClient   = Just (singleQuery mpoint resultVar),
-        localTxSubmissionClient = Nothing
+      LocalNodeClientProtocols
+      { localChainSyncClient    = NoLocalChainSyncClient
+      , localStateQueryClient   = Just (singleQuery mpoint resultVar)
+      , localTxSubmissionClient = Nothing
       }
     atomically (takeTMVar resultVar)
   where
     singleQuery
       :: Maybe ChainPoint
-      -> TMVar (Either Net.Query.AcquireFailure result)
+      -> TMVar (Either (QueryError e) result)
       -> Net.Query.LocalStateQueryClient (BlockInMode mode) ChainPoint
                                          (QueryInMode mode) IO ()
     singleQuery mPointVar' resultVar' =
-      LocalStateQueryClient $ do
-      pure $
-        Net.Query.SendMsgAcquire mPointVar' $
-        Net.Query.ClientStAcquiring
-          { Net.Query.recvMsgAcquired =
-              pure $ Net.Query.SendMsgQuery query $
-                Net.Query.ClientStQuerying
+      LocalStateQueryClient . pure $ do
+        Net.Query.SendMsgAcquire mPointVar' $ Net.Query.ClientStAcquiring
+          { Net.Query.recvMsgAcquired = do
+              pure $ Net.Query.SendMsgQuery query $ Net.Query.ClientStQuerying
                   { Net.Query.recvMsgResult = \result -> do
                     atomically $ putTMVar resultVar' (Right result)
 
@@ -511,7 +519,7 @@ queryNodeLocalState connctInfo mpoint query = do
                       pure $ Net.Query.SendMsgDone ()
                   }
           , Net.Query.recvMsgFailure = \failure -> do
-              atomically $ putTMVar resultVar' (Left failure)
+              atomically $ putTMVar resultVar' (Left (QueryErrorAcquireFailure failure))
               pure $ Net.Query.SendMsgDone ()
           }
 
@@ -523,10 +531,10 @@ submitTxToNodeLocal connctInfo tx = do
     resultVar <- newEmptyTMVarIO
     connectToLocalNode
       connctInfo
-      LocalNodeClientProtocols {
-        localChainSyncClient    = NoLocalChainSyncClient,
-        localTxSubmissionClient = Just (localTxSubmissionClientSingle resultVar),
-        localStateQueryClient   = Nothing
+      LocalNodeClientProtocols
+      { localChainSyncClient    = NoLocalChainSyncClient
+      , localTxSubmissionClient = Just (localTxSubmissionClientSingle resultVar)
+      , localStateQueryClient   = Nothing
       }
     atomically (takeTMVar resultVar)
   where
@@ -552,10 +560,10 @@ getLocalChainTip localNodeConInfo = do
     connectToLocalNode
       localNodeConInfo
       LocalNodeClientProtocols
-        { localChainSyncClient = LocalChainSyncClient $ chainSyncGetCurrentTip resultVar
-        , localTxSubmissionClient = Nothing
-        , localStateQueryClient = Nothing
-        }
+      { localChainSyncClient = LocalChainSyncClient $ chainSyncGetCurrentTip resultVar
+      , localTxSubmissionClient = Nothing
+      , localStateQueryClient = Nothing
+      }
     atomically $ takeTMVar resultVar
 
 chainSyncGetCurrentTip
--- a/cardano-api/src/Cardano/Api/IPC.hs
+++ b/cardano-api/src/Cardano/Api/IPC.hs
@@ -75,6 +75,7 @@ module Cardano.Api.IPC (
 
 import           Prelude
 
+import           Data.Bifunctor (first)
 import           Data.Void (Void)
 
 import qualified Data.ByteString.Lazy as LBS
@@ -165,10 +166,9 @@ data LocalNodeConnectInfo mode =
 
 type MinNodeToClientVersion = NodeToClientVersion
 
-data QueryError a
+data QueryError
   = QueryErrorAcquireFailure !Net.Query.AcquireFailure
   | QueryErrorUnsupportedVersion !MinNodeToClientVersion !NodeToClientVersion
-  | QueryErrorOf a
   deriving (Eq, Show)
 
 localConsensusMode :: LocalNodeConnectInfo mode -> ConsensusMode mode
@@ -489,9 +489,10 @@ convLocalStateQueryClient mode =
 queryNodeLocalState :: forall e mode result.
                        LocalNodeConnectInfo mode
                     -> Maybe ChainPoint
+                    -> (QueryError -> e)
                     -> QueryInMode mode result
-                    -> IO (Either (QueryError e) result)
-queryNodeLocalState connctInfo mpoint query = do
+                    -> IO (Either e result)
+queryNodeLocalState connctInfo mpoint mapQueryError query = do
     resultVar <- newEmptyTMVarIO
     connectToLocalNode
       connctInfo
@@ -500,11 +501,11 @@ queryNodeLocalState connctInfo mpoint query = do
       , localStateQueryClient   = Just (singleQuery mpoint resultVar)
       , localTxSubmissionClient = Nothing
       }
-    atomically (takeTMVar resultVar)
+    atomically (first mapQueryError <$> takeTMVar resultVar)
   where
     singleQuery
       :: Maybe ChainPoint
-      -> TMVar (Either (QueryError e) result)
+      -> TMVar (Either QueryError result)
       -> Net.Query.LocalStateQueryClient (BlockInMode mode) ChainPoint
                                          (QueryInMode mode) IO ()
     singleQuery mPointVar' resultVar' =
--- a/cardano-api/src/Cardano/Api/IPC/Monad.hs
+++ b/cardano-api/src/Cardano/Api/IPC/Monad.hs
@@ -59,7 +59,7 @@ executeLocalStateQueryExpr
   :: forall e mode a .
      LocalNodeConnectInfo mode
   -> Maybe ChainPoint
-  -> ExceptT (QueryError e) (LocalStateQueryExpr (BlockInMode mode) ChainPoint (QueryInMode mode) () IO) a
+  -> LocalStateQueryExpr (BlockInMode mode) ChainPoint (QueryInMode mode) () IO (Either (QueryError e) a)
   -> IO (Either (QueryError e) a)
 executeLocalStateQueryExpr connectInfo mpoint f = do
   tmvResultLocalState <- newEmptyTMVarIO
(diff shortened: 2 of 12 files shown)


cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -42,15 +50,15 @@ import qualified Ouroboros.Network.Protocol.LocalStateQuery.Type as Net.Query
 -- In order to make pipelining still possible we can explore the use of Selective Functors
 -- which would allow us to straddle both worlds.
 newtype LocalStateQueryExpr block point query r m a = LocalStateQueryExpr
-  { runLocalStateQueryExpr :: ContT (Net.Query.ClientStAcquired block point query m r) m a
+  { runLocalStateQueryExpr :: ReaderT NodeToClientVersion (ContT (Net.Query.ClientStAcquired block point query m r) m) a
```
newhoggy, line 53: The `ReaderT` allows the monad expression to know the node to client version of the node server.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -220,18 +236,12 @@ runQueryTip (AnyConsensusModeParams cModeParams) network mOutFile = do
...
         eraHistory <- queryExpr (QueryEraHistory CardanoModeIsMultiEra)
-        mChainBlockNo <- if ntcVersion >= NodeToClientV_10
-          then Just <$> queryExpr QueryChainBlockNo
-          else return Nothing
-        mChainPoint <- if ntcVersion >= NodeToClientV_10
-          then Just <$> queryExpr (QueryChainPoint CardanoMode)
-          else return Nothing
-        mSystemStart <- if ntcVersion >= NodeToClientV_9
-          then Just <$> queryExpr QuerySystemStart
-          else return Nothing
```
newhoggy, line 234: This is simplified a lot.  Not only is it shorter, but we no longer need to know what node version a given query requires.  Instead if the query is not supported by the node then `Nothing` is returned.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -846,10 +851,12 @@ determineEra cModeParams localNodeConnInfo =
     ByronMode -> return $ AnyCardanoEra ByronEra
     ShelleyMode -> return $ AnyCardanoEra ShelleyEra
     CardanoMode -> do
-      eraQ <- liftIO . queryNodeLocalState localNodeConnInfo Nothing
-                     $ QueryCurrentEra CardanoModeIsMultiEra
```
newhoggy, line 850: We no longer need `queryNodeLocalState` because they can all by replaced by `executeLocalStateQueryExpr`/`queryExpr`.

plutus-example/plutus-example/src/Cardano/PlutusExample/ScriptContextChecker.hs
```diff
@@ -327,20 +333,15 @@ readCustomRedeemerFromTx fp (AnyConsensusModeParams cModeParams) network = do
                    (ConsensusModeMismatch (AnyConsensusMode CardanoMode) (AnyCardanoEra cEra))
                    $ toEraInMode cEra CardanoMode
 
-      eResult <-
-        liftIO $ executeLocalStateQueryExpr localNodeConnInfo Nothing
-          $ \ntcVersion -> do
-              (EraHistory _ interpreter) <- queryExpr $ QueryEraHistory CardanoModeIsMultiEra
```
newhoggy, line 333: We no longer need to worry about `ntcVersion` lambda when executing queries because of the use of `ReaderT`.

plutus-example/plutus-example/src/Cardano/PlutusExample/ScriptContextChecker.hs
```diff
@@ -327,20 +333,15 @@ readCustomRedeemerFromTx fp (AnyConsensusModeParams cModeParams) network = do
                    (ConsensusModeMismatch (AnyConsensusMode CardanoMode) (AnyCardanoEra cEra))
                    $ toEraInMode cEra CardanoMode
 
-      eResult <-
-        liftIO $ executeLocalStateQueryExpr localNodeConnInfo Nothing
-          $ \ntcVersion -> do
```
newhoggy, line 332: We no longer need to worry about `ntcVersion` lambda when executing queries because of the use of `ReaderT`.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -298,6 +303,12 @@ runQueryTip (AnyConsensusModeParams cModeParams) network mOutFile = do
 -- via the local state query protocol.
 --
 
+minNtcVersionForQueryUTxOFilter :: QueryUTxOFilter -> NodeToClientVersion
```
Jimbo4350: Should probably add a comment like "See `data NodeToClientVersion` in `ouroboros-network` for more details". There also appears to be more version related information by `data QueryVersion`. We should aim to have all queries in the cli accounted for here so when we have a new query the type checker complains we have a missing pattern match.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -347,7 +359,7 @@ runQueryPoolParams (AnyConsensusModeParams cModeParams) network poolid = do
     & hoistMaybe (ShelleyQueryCmdEraConsensusModeMismatch (AnyConsensusMode cMode) anyE)
 
   let qInMode = QueryInEra eInMode . QueryInShelleyBasedEra sbe $ QueryDebugLedgerState
-  result <- executeQuery era cModeParams localNodeConnInfo qInMode
+  result <- executeQuery era cModeParams NodeToClientV_1 localNodeConnInfo qInMode
```
Jimbo4350: So this is the type of situation we don't want. If we can implement a function that takes `qInMode` and returns a minimum required node version, that would be ideal.

cardano-api/src/Cardano/Api/IPC.hs
```diff
@@ -166,9 +166,10 @@ data LocalNodeConnectInfo mode =
 
 type MinNodeToClientVersion = NodeToClientVersion
 
-data QueryError
+data QueryError a
   = QueryErrorAcquireFailure !Net.Query.AcquireFailure
   | QueryErrorUnsupportedVersion !MinNodeToClientVersion !NodeToClientVersion
+  | QueryErrorOf a
```
Jimbo4350, line 171: I'm not sure I see the benefit of parameterizing `QueryError`. Was the `QueryErrorOf` constructor defined to indicate that a particular error is a query error?

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -141,6 +144,11 @@ runQueryCmd cmd =
     QueryUTxO' consensusModeParams qFilter networkId mOutFile ->
       runQueryUTxO consensusModeParams qFilter networkId mOutFile
 
+queryErrorToShelleyQueryCmdError :: QueryError ShelleyQueryCmdError -> ShelleyQueryCmdError
```
Jimbo4350, line 150: Why not add an additional constructor to `ShelleyQueryCmdError`? E.g `ShelleyQueryRelatedCmdError QueryError`

cardano-cli/src/Cardano/CLI/Shelley/Run/Transaction.hs
```diff
@@ -378,6 +378,11 @@ runTxBuildRaw (AnyCardanoEra era)
     firstExceptT ShelleyTxCmdWriteFileError . newExceptT $
       writeFileTextEnvelope fpath Nothing txBody
 
+queryErrorToShelleyTxCmdError :: QueryError ShelleyTxCmdError -> ShelleyTxCmdError
```
Jimbo4350, line 381: Likewise here, why not add a constructor to `ShelleyTxCmdError`? E.g `ShelleyTxQueryRelatedCmdError QueryError`?

Jimbo4350 (changes requested):
Right direction, a couple of comments :+1:

cardano-api/src/Cardano/Api/Query.hs
```diff
@@ -140,6 +142,14 @@ data QueryInMode mode result where
     :: ConsensusMode mode
     -> QueryInMode mode ChainPoint
 
+instance NtcVersionOf (QueryInMode mode result) where
```
Jimbo4350, line 145: This is exactly what I had in mind except as a function rather than a type class :smile:.

cardano-api/src/Cardano/Api/IPC/NtcVersionOf.hs
```diff
@@ -0,0 +1,10 @@
+module Cardano.Api.IPC.NtcVersionOf
+  ( NtcVersionOf (..)
+  ) where
+
+import           Ouroboros.Network.NodeToClient.Version (NodeToClientVersion (..))
+
+-- | The query 'a' is a versioned query, which means it requires the Node to support a minimum
+-- Node-to-Client version.
```
Jimbo4350, line 8: We probably don't need a type class for this. We can implement this as a function in the `Query` module.

cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -74,36 +84,56 @@ setupLocalStateQueryExpr ::
...
-      Net.Query.ClientStQuerying
-      { Net.Query.recvMsgResult = f
-      }
+-- | Get the node server's Node-to-Client version.
+getNtcVersion :: ExceptT (QueryError e) (LocalStateQueryExpr block point (QueryInMode mode) r IO) NodeToClientVersion
+getNtcVersion =  lift $ LocalStateQueryExpr $ ReaderT pure
+
+-- | Lift a query value into a monadic query expression.
+-- Use 'queryExpr' in a do block to construct monadic local state queries.
+queryExpr :: QueryInMode mode a -> ExceptT (QueryError e) (LocalStateQueryExpr block point (QueryInMode mode) r IO) a
```
Jimbo4350, line 111: I disagree with lifting this into `ExceptT` because then this is now imposed on users. Also we don't seem to gain much by doing this outside of things looking a little nicer in the cli.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: I'm not sure what happened, but `minNtcVersionForQueryUTxOFilter` doesn't exist anymore.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: Same as before.

cardano-api/src/Cardano/Api/IPC/NtcVersionOf.hs
newhoggy: If we implement this as a function, it won't be one function, but a function per type.  Something like `ntcVersionOfQueryTypeName` where `QueryTypeName` is the name of the type.  Is that okay?

cardano-api/src/Cardano/Api/IPC/Monad.hs
newhoggy: So if I understand your comment, it's okay to use `ExceptT` in CLI, but not API?

newhoggy: This commit removes `ExceptT` from type signatures of exported functions in `Cardano.Api.IPC.Monad`. https://github.com/input-output-hk/cardano-node/pull/3358/commits/c51800580d7c924a52c4af06c984fb47b7f5a45d

cardano-api/src/Cardano/Api/IPC/NtcVersionOf.hs
newhoggy: Converting from function to typeclass also gives users the opportunity to not use our function but substitute their own, which most likely will be wrong too.

cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -74,38 +85,59 @@ setupLocalStateQueryExpr ::
...
+    else throwE $ QueryErrorUnsupportedVersion minNtcVersion ntcVersion
+
+-- | Lift a query value into a monadic query expression returning Maybe of a result.
+-- This is the same as 'queryExpr' except if the query is not supported by the server, will return Nothing instead
+-- of throwing an error.
+maybeQueryExpr :: QueryInMode mode a -> LocalStateQueryExpr block point (QueryInMode mode) r IO (Either QueryError (Maybe a))
+maybeQueryExpr q = runExceptT $ catchE (Just <$> ExceptT (queryExpr q)) $ \e ->
+  case e of
+    QueryErrorUnsupportedVersion _ _ -> return Nothing
+    _ -> throwE e
```
Jimbo4350, line 133: So generally we use: https://hackage.haskell.org/package/transformers-except-0.1.1/docs/Control-Monad-Trans-Except-Extra.html

I think `throwE` is misleading vs `left` as we're not throwing an exception here. What are your thoughts?

cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -74,38 +85,59 @@ setupLocalStateQueryExpr ::
...
+  let minNtcVersion = ntcVersionOf q
+  ntcVersion <- ExceptT getNtcVersion
+  if ntcVersion >= minNtcVersion
+    then lift
+      $ LocalStateQueryExpr $ ReaderT $ \_ -> ContT $ \f -> pure $
+        Net.Query.SendMsgQuery q $
+          Net.Query.ClientStQuerying
+          { Net.Query.recvMsgResult = f
+          }
+    else throwE $ QueryErrorUnsupportedVersion minNtcVersion ntcVersion
```
Jimbo4350, line 124: So generally we use: https://hackage.haskell.org/package/transformers-except-0.1.1/docs/Control-Monad-Trans-Except-Extra.html

I think `throwE` is misleading vs `left` as we're not throwing an exception here. What are your thoughts?

cardano-api/src/Cardano/Api/IPC/NtcVersionOf.hs
Jimbo4350: > but a function per type

I've had a second look and I think you're right :+1:

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -136,6 +141,10 @@ runQueryCmd cmd =
     QueryUTxO' consensusModeParams qFilter networkId mOutFile ->
       runQueryUTxO consensusModeParams qFilter networkId mOutFile
 
+mapQueryError :: QueryError -> ShelleyQueryCmdError
```
Jimbo4350, line 144: These should be additional data constructors in `ShelleyQueryCmdError`. Generally we use `firstExceptT ...` to wrap the underlying error types.

cardano-cli/src/Cardano/CLI/Shelley/Run/Transaction.hs
```diff
@@ -113,6 +113,8 @@ data ShelleyTxCmdError
   | ShelleyTxCmdTxInsDoNotExist ![TxIn]
   | ShelleyTxCmdMinimumUTxOErr !MinimumUTxOError
   | ShelleyTxCmdPParamsErr !ProtocolParametersError
+  | ShelleyTxCmdUnsupportedVersion !MinNodeToClientVersion !NodeToClientVersion
```
Jimbo4350, line 116: So this should be `ShelleyTxCmdQueryError QueryError`

plutus-example/plutus-example/src/Cardano/PlutusExample/ScriptContextChecker.hs
```diff
@@ -243,6 +243,7 @@ data ScriptContextError = NoScriptsInByronEra
                         | QueryError ShelleyQueryCmdError
                         | ConsensusModeMismatch AnyConsensusMode AnyCardanoEra
                         | EraMismatch !Consensus.EraMismatch
+                        | ScriptContextUnsupportedVersion !MinNodeToClientVersion !MinNodeToClientVersion
```
Jimbo4350, line 246: And likewise here `ScriptContextUnsupportedVersion QueryError`

Jimbo4350 (changes requested):
Couple more comments but right direction

cardano-api/src/Cardano/Api/IPC.hs
```diff
@@ -475,43 +486,41 @@ convLocalStateQueryClient mode =
 -- | Establish a connection to a node and execute a single query using the
 -- local state query protocol.
 --
-queryNodeLocalState :: forall mode result.
+queryNodeLocalState :: forall e mode result.
```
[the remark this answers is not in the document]
newhoggy: The reason is I want to be able to return errors other than those described by the `QueryError` constructor in monad query expressions.

For example:

* https://github.com/input-output-hk/cardano-node/blob/0242e81cf452889e09e862e8ae21ae877174f7f5/cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs#L723
* https://github.com/input-output-hk/cardano-node/blob/0242e81cf452889e09e862e8ae21ae877174f7f5/cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs#L725
* https://github.com/input-output-hk/cardano-node/blob/0242e81cf452889e09e862e8ae21ae877174f7f5/plutus-example/plutus-example/src/Cardano/PlutusExample/ScriptContextChecker.hs#L338
* https://github.com/input-output-hk/cardano-node/blob/0242e81cf452889e09e862e8ae21ae877174f7f5/cardano-cli/src/Cardano/CLI/Shelley/Run/Transaction.hs#L484
* https://github.com/input-output-hk/cardano-node/blob/0242e81cf452889e09e862e8ae21ae877174f7f5/cardano-cli/src/Cardano/CLI/Shelley/Run/Transaction.hs#L468
* https://github.com/input-output-hk/cardano-node/blob/0242e81cf452889e09e862e8ae21ae877174f7f5/cardano-cli/src/Cardano/CLI/Shelley/Run/Transaction.hs#L460
* https://github.com/input-output-hk/cardano-node/blob/0242e81cf452889e09e862e8ae21ae877174f7f5/cardano-cli/src/Cardano/CLI/Shelley/Run/Transaction.hs#L475

And many more.

The idea is that any user code can construct a complex monadic expression and be able to throw their own errors that are not `QueryError`s.

cardano-api/src/Cardano/Api/IPC/Monad.hs
newhoggy: The main reason I `throwE` here is because I `catchE`, as they are duals of each other.

cardano-cli/src/Cardano/CLI/Shelley/Run/Transaction.hs
newhoggy: I  merged `ShelleyQueryCmdAcquireFailure` and `ShelleyQueryCmdUnsupportedVersion` constructors into `ShelleyQueryCmdQueryError`, I think this was what you meant?

newhoggy:
Re-implemented in https://github.com/input-output-hk/cardano-node/pull/4788

PR #3367, opened 2021-11-15
Update dependencies

erikd:


cardano-api/src/Cardano/Api/Orphans.hs
```diff
@@ -266,6 +269,13 @@ instance Crypto.Crypto crypto => ToJSON (Shelley.SnapShot crypto) where
 instance Crypto.Crypto crypto => ToJSON (Shelley.Stake crypto) where
   toJSON (Shelley.Stake s) = toJSON s
 
+instance (VMap.Vector vp a, VMap.Vector vb k, ToJSON k, ToJSON a, ToJSONKey k)
+        => ToJSON (VMap vb vp k a) where
+  toJSON vm = toJSON (VMap.toMap vm)
+
+instance ToJSON (Shelley.CompactForm Shelley.Coin) where
+  toJSON cc = toJSON (Ledger.fromCompact cc)
```
nc6, line 277: Odd that this is needed where it wasn't before, this type is hardly new...

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -548,14 +550,14 @@ writeStakeSnapshot (StakePoolKeyHash hk) qState =
...
-    (Stake s) = poolStake hash (_delegations ss) (_stake ss)
+    Coin pStake = fold (Map.map Ledger.fromCompact $ VMap.toMap s)
+    Stake s = poolStake hash (_delegations ss) (_stake ss)
 
 -- | Sum the active stake from a snapshot
 getAllStake :: SnapShot crypto -> Integer
 getAllStake (SnapShot stake _ _) = activeStake
   where
-    Coin activeStake = fold . unStake $ stake
+    Coin activeStake = fold . Map.map Ledger.fromCompact . VMap.toMap $ unStake stake
```
nc6, line 560: I suggest we merge https://github.com/input-output-hk/cardano-ledger/pull/2550/files and make use of `sumAllStake` there

nc6 (changes requested):

cardano-api/src/Cardano/Api/Orphans.hs
nc6: I guess we're using the compact form in more places, so this is ok

nc6 (approved):

nc6:
bors merge

erikd:
bors merge

PR #3378, opened 2021-11-19
cardano-node:  factor tracing-independent mini-API as Cardano.Node.Queries

deepfire:
The current tracing system makes a number of projections of the system state, which are not really tracing-system specific.

We consolidate those as modules under `Cardano.Node.Queries`, so this can be reused by old and new tracing systems:

  - Cardano.Node.Queries.ConvertTxId
  - Cardano.Node.Queries.KES
  - Cardano.Node.Queries.Ledger


newhoggy (approved):
LGTM

cardano-node/src/Cardano/Node/Queries/KES.hs
```diff
@@ -0,0 +1,123 @@
...
+{-# LANGUAGE DisambiguateRecordFields #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+module Cardano.Node.Queries.KES
+  (
```
Jimbo4350, line 12: Extra newline here

Jimbo4350 (changes requested):
Rather than creating three separate modules, lets put all of this into one module.

cardano-node/cardano-node.cabal
```diff
@@ -73,16 +73,15 @@ library
                        Cardano.Node.Protocol.Shelley
                        Cardano.Node.Protocol.Types
                        Cardano.Node.Parsers
+                       Cardano.Node.Queries
```
Jimbo4350, line 76: Thanks :+1:

Jimbo4350 (approved):
LGTM!

PR #3380, opened 2021-11-22
startup logging

coot:
- Added StartupTracer
- Use StartupTracer to log start time and db validation startup
- Added a todo

### DIFF
--- a/cardano-node/src/Cardano/Tracing/Startup.hs
+++ b/cardano-node/src/Cardano/Tracing/Startup.hs
@@ -13,6 +13,8 @@ import           Data.Map (Map)
 import qualified Data.Map as Map
 import           Data.Text (Text)
 import qualified Data.Text as Text
+import           Data.Time.Clock (UTCTime)
+import           Data.Time.Clock.POSIX (POSIXTime, utcTimeToPOSIXSeconds)
 
 import           Cardano.Slotting.Slot (SlotNo (..))
 import           Cardano.Node.Configuration.TopologyP2P (UseLedger (..))
@@ -55,11 +57,14 @@ data StartupTrace blk =
 
   -- | Log peer-to-peer diffusion mode
   | StartupP2PInfo DiffusionMode
+    
+  | StartupTime UTCTime
 
   | StartupNetworkMagic NetworkMagic
 
   | StartupSocketConfigError SocketConfigError
 
+  | StartupDBValidation
 
   -- | Log that the network configuration is being updated.
   --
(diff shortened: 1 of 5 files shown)


deepfire (approved):
Looks good to me, thank you @coot !

coot:
@deepfire I added proper json instances, and better pretty printing, I added proper severity levels, and  I checked that it actually works.

cardano-node/src/Cardano/Tracing/Startup.hs
```diff
@@ -0,0 +1,244 @@
...
+
+  | NetworkConfigUpdateError Text
+
+  | NetworkConfig [(Int, Map RelayAccessPoint PeerAdvertise)]
+                  [RelayAccessPoint]
+                  UseLedgerAfter
+
+  | P2PWarning
+
+  | P2PWarningDevelopementNetworkProtocols
```
karknu, line 72: We should drop the P2P from this message. Currently `DevelopementNetworkProtocols` means p2p protocols, but that will not be the case in the future.

karknu (approved):
LGTM.

cardano-node/src/Cardano/Tracing/Startup.hs
coot: This is a specific warning for p2p: it requires `TestEnableDevelopmentNetworkProtocols`, so when it is not set it will show a message that one needs to enable it.  But I take your point that we should have another warning that says that development network protocols are in use.

coot: Done, it's either:
```
[waldorf:cardano.node.nodeconfig:Warning:5] [2021-11-23 11:50:47.86 UTC] unsupported and unverified version of `cardano-node` with peer-to-peer networking capabilities
[waldorf:cardano.node.nodeconfig:Warning:5] [2021-11-23 11:50:47.86 UTC] using development network protocols: [NodeToNodeV_8] []
```
or
```
[waldorf:cardano.node.nodeconfig:Warning:5] [2021-11-23 11:53:36.04 UTC] unsupported and unverified version of `cardano-node` with peer-to-peer networking capabilities
[waldorf:cardano.node.nodeconfig:Warning:5] [2021-11-23 11:53:36.04 UTC] peer-to-peer requires TestEnableDevelopmentNetworkProtocols to be set to True
```
or nothing, or just the development network protocols message.
```

**Review by jutaro — APPROVED** [review]
LGTM

File: cardano-node/src/Cardano/Tracing/Startup.hs [hunk 2]
```diff
@@ -0,0 +1,273 @@
...
+       ++ intercalate "\n"
+          (map (\(v, bv) -> show v ++ "\t" ++ show bv)
+        . Map.assocs
+        $ supportedNodeToNodeVersions)
+    , "node-to-client versions:\n"
+       ++ intercalate "\n"
+          (map (\(v, bv) -> show v ++ "\t" ++ show bv)
+        . Map.assocs
+        $ supportedNodeToCleintVersions)
+    ]
```
newhoggy, line 196: Seems like a lot of string appends.  Do we need to worry about performance?

cardano-node/src/Cardano/Tracing/Startup.hs
```diff
@@ -0,0 +1,273 @@
...
+    , "node-to-node versions:\n"
+       ++ intercalate "\n"
+          (map (\(v, bv) -> show v ++ "\t" ++ show bv)
+        . Map.assocs
+        $ supportedNodeToNodeVersions)
+    , "node-to-client versions:\n"
+       ++ intercalate "\n"
+          (map (\(v, bv) -> show v ++ "\t" ++ show bv)
+        . Map.assocs
+        $ supportedNodeToCleintVersions)
```
newhoggy, line 195: There a typo in this name `supportedNodeToCleintVersions`.  Cleint -> Client

newhoggy:
I'm assuming both TODOs are intentional.

newhoggy (approved):
Left some comments.  Otherwise LGTM.

cardano-node/src/Cardano/Tracing/Startup.hs
coot: No, not at all; this runs once at startup.

cardano-node/src/Cardano/Tracing/Startup.hs
coot: Fixed.

Issue #3338, opened 2021-11-03
[FR] - transaction build cmd required-signer to accept keyhash


james-iohk:
**Internal/External**
*Internal*   if an IOHK staff member.

**Area**
*Plutus*    Related to Plutus Scripts (Alonzo).

**Describe the feature you'd like**
Currently, the `transaction build` cmd's `--required-signer` option takes signing key file. It would be good to also allow keyhash to be used so that keys are not always exposed to the cli, e.g. for multisig.

Note: This option exists on tags 1.30.0 and 1.31.0 (not 1.30.1)

waalge:
Is there a reason to why this wouldn't only except the pkh rather than the skey?
(ie why would I ever want to supply the skey?)

robertwb:
I just ran into this as well; it seems for transaction building one should be able to assert what expected signatures exist, not require the signing keys (which may be performed by a separate entity).

BlakeBrown:
+1 for this enhancement

We're currently running a dApp (www.jpg.store) where our customers use Nami wallet to sign transactions. It's impossible to get someone's signing key to construct an identical transaction in the CLI. Even if it was possible, requiring a signing key to be passed around is a huge security risk.

jfischoff:
I have a branch that implements this implements this here: https://github.com/Canonical-LLC/cardano-node/pull/1

PR #3382, opened 2021-11-22
Accept keyhashes for required signers

Jimbo4350:
Required signers can now be specified via the `--required-signer-hash` cli option
Resolves: #3338

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
+++ b/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
@@ -1430,20 +1430,28 @@ pColdSigningKeyFile =
       )
     )
 
-pRequiredSigner :: Parser WitnessSigningData
+pRequiredSigner :: Parser RequiredSigner
 pRequiredSigner =
-    KeyWitnessSigningData
-      <$>
-        ( SigningKeyFile <$>
-            Opt.strOption
-              (  Opt.long "required-signer"
-              <> Opt.metavar "FILE"
-              <> Opt.help "Input filepath of the signing key (zero or more) whose \
-                          \signature is required."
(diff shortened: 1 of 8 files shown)


mkoura:
QA tested this with the following plutus script with both normal and extended keys, and with key files and key hashes. Works as expected :+1: 
```haskell
mkPolicy :: PubKeyHash -> ScriptContext -> Bool
mkPolicy pkh ctx = traceIfFalse "not signed by redeemer pubkeyhash" checkWitness
  where
    info :: TxInfo
    info  = scriptContextTxInfo ctx

    checkWitness :: Bool
    checkWitness  = txSignedBy info pkh
```

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -1430,20 +1430,36 @@ pColdSigningKeyFile =
...
+      (  Opt.long "required-signer-hash"
+      <> Opt.metavar "HASH"
+      <> Opt.help "Hash of the verification key (zero or more) whose \
+                  \signature is required."
+      )
+  sExtendedPayKeyHash :: Parser (Hash PaymentExtendedKey)
+  sExtendedPayKeyHash =
+    Opt.option (readerFromParsecParser $ parseHash (AsHash AsPaymentExtendedKey))
+      (  Opt.long "required-signer-hash"
+      <> Opt.hidden
```
newhoggy, line 1460: What is this option for?

Jimbo4350: Ah I actually don't need this

newhoggy (approved):

Jimbo4350:
bors merge

PR #3385, opened 2021-11-23
[cardano-api] Add a FromJSONKey instance for TxIn

ch1bo:
Was trying to build, but `cabal` fails (like on `master`) with:
```
λ cabal build cardano-api
/nix/store/qbv8c3ilaawqm36cbxflvv6gny00nzjj-cabal-install-exe-cabal-3.6.0.0/bin/cabal --project-file=/home/ch1bo/code/iog/cardano-haskell/cardano-node/.nix-shell-cabal.project build cardano-api
Warning: No remote package servers have been specified. Usually you would have
one specified in the config file.
Resolving dependencies...
cabal: Could not resolve dependencies:
[__0] trying: cardano-api-1.31.0 (user goal)
[__1] unknown package: compact-map (dependency of cardano-api)
[__1] fail (backjumping, conflict set: cardano-api, compact-map)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: cardano-api, compact-map
```

### DIFF
--- a/cardano-api/src/Cardano/Api/TxBody.hs
+++ b/cardano-api/src/Cardano/Api/TxBody.hs
@@ -399,7 +399,10 @@ instance ToJSONKey TxIn where
   toJSONKey = toJSONKeyText renderTxIn
 
 instance FromJSON TxIn where
-  parseJSON = withText "TxIn" $ \txinStr -> runParsecParser parseTxIn txinStr
+  parseJSON = withText "TxIn" $ runParsecParser parseTxIn
+
+instance FromJSONKey TxIn where
+  fromJSONKey = Aeson.FromJSONKeyTextParser $ runParsecParser parseTxIn
 
 parseTxId :: Parsec.Parser TxId
 parseTxId = do


Jimbo4350:
There appears to be a build failure but I've also asked Philipp to add to you the cardano-sl team so that CI gets triggered properly.

cardano-api/src/Cardano/Api/Utils.hs
```diff
@@ -40,7 +39,7 @@ formatParsecError err =
     "expecting" "unexpected" "end of input"
     $ Parsec.errorMessages err
 
-runParsecParser :: Parsec.Parser a -> Text -> Aeson.Parser a
+runParsecParser :: MonadFail m => Parsec.Parser a -> Text -> m a
```
Jimbo4350, line 42: Is there a specific reason why you have generalized this?

Jimbo4350 (changes requested):
One comment

cardano-api/src/Cardano/Api/Utils.hs
ch1bo: The alternative would be to have `txInFromText :: Text -> Aeson.Parser TxIn` which then would only be possible to be used in an Aeson context, or not factor out `txInFromText`. The fact that nothing Aeson.Parser specific is used in `runParsecParser` and Aeson is only imported in Cardano.Api.Utils for this function kind of sealed the deal for me, but kan change it to one of the other alternatives if you want.

Jimbo4350: I think its better to have the types concrete until we need to generalize them. So yeah, until there is a need for `txInFromText` in a different context lets leave it as `txInFromText :: Text -> Aeson.Parser TxIn`

ch1bo: Done

ch1bo:
All done as commented.. bumping this, although not critical to us right now.

Jimbo4350 (approved):

ch1bo:
Closed in favor of https://github.com/input-output-hk/cardano-node/pull/3438

PR #3386, opened 2021-11-24
Restrict updating CDF counters after restart.

karknu:
After restart the block delayes doesn't reflect the node's network
condition. This change waits until the node sees a block with a delay
less than 20s. Then we collect 45 samples (about 15 minutes) until
we start to provide values for the CDF counters.


gufmar:
looks good on a mainnet relay here

2021-11-25_15h45_53

it is useful for alarm trigger settings to not have "wrong" low values (delayed blocks), while the node is catching up after start.

before it was like this

image

the floating average from low (restarting) values then distorted the value for a longer period of time.

Jimbo4350 (approved):
LGTM, just fix the indent from 4 spaces to 2 spaces please.

PR #3390, opened 2021-11-25
Better error message for query utxo plucky poc

newhoggy:


newhoggy:
Re-implemented in https://github.com/input-output-hk/cardano-node/pull/4788

PR #3391, opened 2021-11-25
Tx-generator full auto mode Plutus benchmarks

MarcFontaine:


bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
```diff
@@ -632,6 +643,37 @@ createChangeGeneric submitMode createCoins addressMsg value count = do
   chunkList _ [] = []
   chunkList n xs = as : chunkList n bs where (as,bs) = splitAt n xs
 
+spendAutoScript :: SubmitMode -> FilePath -> ThreadName -> NumberOfTxs -> TPSRate -> ActionM ()
```
deepfire, line 646: It'd be nice to have a short verbal description of how this works.

bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
```diff
@@ -632,6 +643,37 @@ createChangeGeneric submitMode createCoins addressMsg value count = do
...
+spendAutoScript submitMode scriptFile threadName txCount tps = do
+  protocolParameters <- queryProtocolParameters
```
deepfire, line 648: By the way, the initial version of `cardano-node` started with this pattern of passing information through the underlying monad.

In Byron Rewrite we dropped this in favor of explicit parameter passing -- because it makes the data dependencies easier to understand.

bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
MarcFontaine: It does a binary search in the range of [0..10000] for the redeemer parameter of
the `loop.plutus` script ( lets call it n)  such that `loop 1 to n` is just within the resource limits. i.e. `loop 1 to (n+1)` is out of limits

bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
```diff
@@ -632,6 +643,37 @@ createChangeGeneric submitMode createCoins addressMsg value count = do
...
+  redeemer <- case startSearch costFn 0 100000 of
+    Left err -> throwE $ ApiError $ "cannot find fitting redeemer :" ++ err
+    Right x -> return $ ScriptDataNumber $ x + 1000000
+  runPlutusBenchmark submitMode scriptFile PreExecuteScript (ScriptDataNumber 0) redeemer threadName txCount tps
+  where
+    startSearch f a b = do
+       l <- f a
+       h <- f b
+       if not l || h
+          then Left $ "binary search: bad inital bounds : " ++ show (a,b)
```
deepfire, line 668: I think we might benefit from a more clear error message here -- since there are two distinct cases:

1. the sought value is above the upper boundary
2. the sought value is below the lower boundary

bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
deepfire: @MarcFontaine, sorry, I've meant in a code comment, since here the information will be lost.

bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
MarcFontaine: This is not really the underlying monad though.
This is a real/full query for the protocol parameters of the locally connected node.

bench/script/loop.plutus
```diff
@@ -0,0 +1,5 @@
+{
+    "type": "PlutusScriptV1",
+    "description": "",
+    "cborHex": "588d588b01000033322233223222322533530083322333573466e2000800403002d40092080897a13263530063357389211572656465656d6572206973203c2031303030303030000074984c01540084dd680099000991800800911991299a9803999ab9a3370e002904044bd0048040803099180180099b8100148008c00800800448004800448800848800480041"
```
deepfire, line 4: I still wish we could move away from pre-compiled Plutus scripts.

Not in the least because the compiler does improve.

bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
```diff
@@ -632,6 +643,37 @@ createChangeGeneric submitMode createCoins addressMsg value count = do
...
+  script <- liftIO $ readScript scriptFile
+  let
+    costFn :: Integer -> Either String Bool
+    costFn n = case preExecuteScript protocolParameters script (ScriptDataNumber 0) (ScriptDataNumber $ n + 1000000) of
+      Left err -> Left err
+      Right use -> Right $ (executionSteps use <= executionSteps budget) && (executionMemory use <= executionMemory budget)
+
+  redeemer <- case startSearch costFn 0 100000 of
+    Left err -> throwE $ ApiError $ "cannot find fitting redeemer :" ++ err
+    Right x -> return $ ScriptDataNumber $ x + 1000000
```
deepfire, line 661: We're hardcoding a number of things here and there.

It'd be nice to have named top-level variables, instead.

bench/script/loop.plutus
MarcFontaine: This is strictly only used by me for testing the stuff locally.
I often run the test-scripts.json directly and by-pass the tx-generator-service.
The tx-generator-service the latest build result.

bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
MarcFontaine: Yes..And 1_000_000 is really hard coded !
The 1_000_000 must match the hard coded value in the Plutus script.
The 100_000 is just an arbitrary large number.

bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
MarcFontaine: Right the sought value is below the lower boundary,
or the thought value is above the upper boundary....

bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
MarcFontaine: Yes there are hardly any comments in this code anywhere.
I think I have only added warnings and TODO comments to the really buggy parts of the code.

bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
deepfire: Oh, that explains -- so at the start of a particular benchmark the parameters are always up to date.

That's actually a useful feature, thank you for explaining Marc!

bench/script/loop.plutus
deepfire: OIC, thanks for explaining Marc!

bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
deepfire: The way to do this is to introduce named "packages", which come from outside:
```
{
   scriptFile: "foo.json",
   scriptConstantA: 1000000

deepfire: ..and then you suddenly can reformulate script processing around well-defined ADTs..

MarcFontaine: I think we should leave it as it is for now.
The proposed indirection just moves the magic number from inside the Haskell code to
some JSON file and causes a 20 times code-blow-up.
This number is strictly non-serviceable for now.
It must match the number that is hard coded in the plutus-example/loop.hs
This may be changed later (if ever needed).
(It is also possible to handle this just in loop.hs.)

deepfire (approved):

PR #3393, opened 2021-11-25
Apply new topology syntax change in nixos service + new p2p defaults

jbgi:
addrs -> accessPoints
 addr -> address


nix/nixos/cardano-node-service.nix
```diff
@@ -20,6 +20,10 @@ let
         TargetNumberOfKnownPeers = cfg.targetNumberOfKnownPeers;
         TargetNumberOfEstablishedPeers = cfg.targetNumberOfEstablishedPeers;
         TargetNumberOfActivePeers = cfg.targetNumberOfActivePeers;
+        TestEnableDevelopmentNetworkProtocols = true;
+        MaxConcurrencyBulkSync = 2;
+        TraceBlockFetchClient = true;
+        TraceChainSyncClient = true;
```
karknu, line 26: I think you should add a comment that these options are only for p2p testnet. If it turns out to be rather costly they should be disabled in the future.

disassembler (approved):

PR #3399, opened 2021-11-29
Tune the default RTS params

weebl2000:
- **-C0** - context-switch as often as possible, I think this will fix prop delays that are currently experienced. Default is wait 20ms before trying a switch.
- -**A32m** - 32m allocation region per core, the node really generates a lot of data
- **-n1m** - 1m nursery size blocks, make sure allocation areas are filled better before GC
- **-AL512m** - 512MB limit for large object allocation before forcing a GC. Default is the value of **-A**.

Usually a lot of large object allocation happens during snapshot creation and it will hit a lot of times since the mainnet ledger snapshot is around ~1GB. The regular Gen0 collects will make sure large blocks get promoted to Gen 1 anyway.

All in all I think these are safe & useful modifications to the default behavior of cardano-node and will increase performance of the entire network a lot.

Yes, actual useful time spent on computing stuff will decline due to -C0, but really. Most nodes are idling pretty much all of the time except for when starting up / making snapshots. So this small dent on throughput really will not impact real world performance, it will just make the cardano-node handle threads that want to do stuff faster.

I played around with RTS parameters **a lot** and the propagation delays to my nodes dropped significantly. Not entirely due to these parameters, I also connected to more nodes in the network. But that was not the only factor.

image

### DIFF
--- a/cardano-node/cardano-node.cabal
+++ b/cardano-node/cardano-node.cabal
@@ -162,9 +162,9 @@ executable cardano-node
                         -rtsopts
 
   if arch(arm)
-    ghc-options:        "-with-rtsopts=-T -I0 -A16m -N1 --disable-delayed-os-memory-return"
+    ghc-options:        "-with-rtsopts=-T -I0 -C0 -A16m -n1m -N1 --disable-delayed-os-memory-return"
   else
-    ghc-options:        "-with-rtsopts=-T -I0 -A16m -N2 --disable-delayed-os-memory-return"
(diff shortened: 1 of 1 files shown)


weebl2000:
@coot Sorry, shameless tag. I'm not certain how I should bring PRs under attention with appropriate parties.

deepfire:
@weebl2000. thanks, we'll have a look at it!

newhoggy:
This PR is very old with no-one taking it forward.  Closing.

Please re-create if there is still intent to make this change.

PR #3400, opened 2021-11-29
scripts: simple transaction example in alonzo era

Divesh-Otwani:
This MR improves the docs of `/scripts` and adds a script for making a simple transaction in the Alonzo era.

This should also help with https://github.com/input-output-hk/cardano-node/issues/721 by providing an easy example to check the rendering of txids in the node log files.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -3,7 +3,7 @@
 The `scripts` directory consists of the following directories:
 - [benchmarking](#benchmarking)
 - [buildkite](#buildkite)
-- [byron-shelley-allegra-mary](#byron-shelley-allegra-mary)
+- [byron-to-shelley](#byron-to-shelley)
 - [lite](#lite)
 - [shelley-from-scratch](#shelley-from-scratch)
 - [byron-to-alonzo](#byron-to-alonzo)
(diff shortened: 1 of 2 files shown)


scripts/README.md
```diff
@@ -15,28 +15,44 @@ Contains all the scripts relevant to benchmarking `cardano-node`. See the benchm
 #### buildkite
 Contains scripts relevant to IOHK's CI.
 
-#### byron-shelley-allegra-mary
-Contains a script that sets up a cluster beginning in the Byron era and can transition to the Shelley era. You can also start a cluster in the Shelley, Allegra or Mary era by supplying an argument to `mk-files.sh`.
+### Clusters
+
+Most clusters sub-folders contain a `mkfiles.sh` script which creates a top level
```
Jimbo4350, line 20: The language here is too indefinite i.e "most", "possibly", "somtimes". I would just start it with "Generally" and then add more details to the specific sections where this doesn't apply.

scripts/byron-to-alonzo/simple-tx.sh
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env bash
+
+set -e
+set -x
+
+# This script creates, signs, and submits a simple transaction in the Alonzo era.
+# We simply generate a new user and send him some money from a pre-build utxo account.
```
Jimbo4350, line 7: We can leave out the "from a pre-build utxo account.".

scripts/byron-to-alonzo/simple-tx.sh
```diff
@@ -0,0 +1,79 @@
...
+# To use: start an alonzo cluster and then run as ./scripts/byron-to-alonzo/simple-tx.sh
+
+ROOT=example
+TESTNET=42
+WORK=simple-tx-files
+CLI=cardano-cli
+pushd ${ROOT}
+mkdir ${WORK}
+sleep 15 # Wait for things to initialize and the socket to be ready
```
Jimbo4350, line 16: Is the intention here that you want to run the mkfiles.sh and this script at the same time? Rather than manually waiting for the cluster to start?

Jimbo4350 (changes requested):
Looking good! Just a few comments.

scripts/byron-to-alonzo/simple-tx.sh
Divesh-Otwani: :+1:

scripts/README.md
Divesh-Otwani: Yes, makes sense :+1:

scripts/byron-to-alonzo/simple-tx.sh
Divesh-Otwani: Yes, I'll omit that and just put a note at the top to run the script after the cluster has started working and you can query the tip from `node-bft1`, say.

olgahryniuk (approved):

newhoggy:
This can't merge yet because `scripts/byron-to-alonzo/simple-tx.sh` fails due to lack of funds:

```
WARNING: '--protocol-params-file' for 'transaction build' is deprecated
Command failed: transaction build  Error: The transaction does not balance in its use of ada. The net balance of the transaction is negative: Lovelace (-10165545) lovelace. The usual solution is to provide more inputs, or inputs with more ada.
```

This shows there are no funds:

```bash
for x in example/utxo-keys/utxo1.vkey; do
  y="${x%.vkey}.addr"
  echo "== $x $y =="

  CARDANO_NODE_SOCKET_PATH=example/main.sock $(./scripts/bin-path.sh cardano-cli) address build \
    --payment-verification-key-file "$x" \
    --testnet-magic 42 \
    --out-file "$y" && \
  CARDANO_NODE_SOCKET_PATH=example/main.sock $(./scripts/bin-path.sh cardano-cli) query utxo \
    --testnet-magic 42 \
    --address "$(cat "$y")"
done
```

Output:

```
                           TxHash                                 TxIx        Amount
--------------------------------------------------------------------------------------
daf7a0bf1fb6424dca44888c26023339ec05a4e00ffc61dc07ed21be4fcbcce8     0         + TxOutDatumNone
```

newhoggy:
This PR is very old and there hasn't been any progress in a while.

PR #3405, opened 2021-11-30
Print asset name as string literal 

Jimbo4350:
Fixes: https://input-output.atlassian.net/browse/CAD-3747


disassembler (approved):

Jimbo4350:
The issue reported in https://input-output.atlassian.net/browse/CAD-3747 was using an earlier version of the cli and therefore this PR is unnecessary.

PR #3407, opened 2021-11-30
Fix haddock for ScriptDataSupportedInEra

ch1bo:
Stumbled over this when working with the cardano-api.


Jimbo4350:
@ch1bo  Why did you delete the comments?

ch1bo:
@Jimbo4350 because the haddock was seemingly copy pasted and unrelated to script data. Please correct me when I'm wrong, but I believe whether script data is a thing in an era has nothing to do with how value is handled in an era.

PR #3411, opened 2021-12-02
tx-generator: Use generic JSON instances

MarcFontaine:
This PR adds new generic JSON instances for the tx-generator script language.
The old manual instances are still available in the CLI with the `legacy-json` command,
and the tx-generator-service still uses the legacy format.


deepfire (approved):
LGTM, thanks Marc!

PR #3417, opened 2021-12-04
Fixing install doc multiple typos

gideont:
Removed some extra $ during copy and paste commands.
Added mkdir to the commands for copy and paste.


newhoggy:
Thanks for the PR.  This commit will be included in https://github.com/input-output-hk/cardano-node/pull/3502

PR #3419, opened 2021-12-06
Add a normal dev shell in shell.nix

nc6:
This adds a normal nix-shell, which can be used by developers wanting to work on cardano-node. The current nix-shell _itself_ tries to build cardano-node, which is rather annoying, since it prevents using cabal/hls to iterate on node changes, especially when bumping dependencies.

Honestly this should be the default shell and the other shell should be renamed (e.g. "workbench" or similar), but in the short-term we at least make this possible.

One can enter this shell using `nix-shell -A dev`

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -38,6 +38,8 @@ let
 
     devopsShell = shell.devops;
 
+    devShell = shell.dev;
+
     nixosTests = recRecurseIntoAttrs nixosTests;
 
     # so that eval time gc roots are cached (nix-tools stuff)
--- a/flake.nix
+++ b/flake.nix
@@ -62,7 +62,7 @@
 
         inherit (pkgs.commonLib) eachEnv environments;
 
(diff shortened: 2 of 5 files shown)


MarcFontaine:
I find that very useful and would strongly approve a PR in that spirit.

jbgi:
I agree that default shell should not build any cardano-node package. 
Now should default cabal build in dev shell use nix provided dependencies or should cabal build them locally (not using hydra cache) ?

nc6:
Honestly I'm happy with either... if they could do so without having to mangle the cabal.project each time, even better :-D

jbgi:
> if they could do so without having to mangle the cabal.project each time

yeah to use nix caching for dependencies we still need to remove `source-repository-package` entries in a cabal wrapper script. 
I'll add a version of your shell that does that under `nix-shell -A devNixCache` and let people chose.

jbgi (approved):

PR #3420, opened 2021-12-06
Implement shelleyPayAddrToPlutusPubKHash 

Jimbo4350:
- `shelleyPayAddrToPlutusPubKHash` allows easy conversion of a Shelley address to a Plutus `PubKeyHash`

### DIFF
--- a/cardano-api/src/Cardano/Api/Address.hs
+++ b/cardano-api/src/Cardano/Api/Address.hs
@@ -47,6 +47,9 @@ module Cardano.Api.Address (
     StakeKey,
     StakeExtendedKey,
 
+    -- * Conversion functions
+    shelleyPayAddrToPlutusPubKHash,
+
     -- * Internal conversion functions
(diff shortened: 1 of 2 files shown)


JaredCorduan (approved):
LGTM, thanks @Jimbo4350 !

cardano-api/src/Cardano/Api/Address.hs
```diff
@@ -528,6 +536,11 @@ isKeyAddress (AddressInEra (ShelleyAddressInEra _) (ShelleyAddress _ pCred _)) =
     PaymentCredentialByKey _ -> True
     PaymentCredentialByScript _ -> False
 
+shellyAddrToPubKeyHash :: Address ShelleyAddr -> Maybe Plutus.PubKeyHash
```
JaredCorduan, line 539: should we put `payment` somehow in the name? to distinguish from the staking public key...

PR #3427, opened 2021-12-08
tx-generator: add literate keys and manual funds.

MarcFontaine:
This PR adds support for literate keys in benchmarking scripts
and manual definition of funds.


deepfire (approved):
LGTM, thanks @MarcFontaine !

PR #3430, opened 2021-12-09
Building docker for Cardano

szandala:
Training to build cardano in docker


disassembler:
My concern with this is no one will maintain it, and Dockerfile's are not known for their stability over time...

The recommended way to build a docker image is using nix:

PR #3434, opened 2021-12-13
Tracing improvements

coot:
- Verbosity of startup tracer
- connection manager counters trace

### DIFF
--- a/cardano-node/src/Cardano/Tracing/Startup.hs
+++ b/cardano-node/src/Cardano/Tracing/Startup.hs
@@ -23,7 +23,7 @@ import           Cardano.Tracing.OrphanInstances.Network ()
 
 import           Cardano.BM.Tracing (HasPrivacyAnnotation (..),
                    HasSeverityAnnotation (..), Severity (..), ToObject (..),
-                   Transformable (..))
+                   TracingVerbosity (..), Transformable (..))
 import           Cardano.BM.Data.Tracer (HasTextFormatter (..), mkObject,
                    trStructuredText)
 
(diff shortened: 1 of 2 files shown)


jutaro (commented):
Why did you remove the kind field "ConnectionManagerCounters"?

cardano-node/src/Cardano/Tracing/Startup.hs
```diff
@@ -116,20 +116,37 @@ instance ( Show (BlockNodeToNodeVersion blk)
...
+                                supportedNodeToNodeVersions
+                                supportedNodeToClientVersions)
+      = case verb of
+          MaximalVerbosity ->
+            mkObject
+              [ "nodeAddresses" .= Aeson.toJSON (map ppN2NSocketInfo addresses)
+              , "localSocket" .= case localSocket of
+                    Nothing -> Aeson.Null
+                    Just a  -> String (Text.pack . ppN2CSocketInfo $ a)
+              , "nodeToNodeVersions" .=
```
Jimbo4350, line 130: Some repetition here. Why not:
```
       [ "nodeAddresses" .= Aeson.toJSON (map ppN2NSocketInfo addresses)
       , "localSocket" .= case localSocket of
                    Nothing -> Aeson.Null
                    Just a  -> String (Text.pack . ppN2CSocketInfo $ a)
       ]  ++ case verb of 
                  MaximalVerbosity -> [ ... ] 
                  _ -> [ ... ]
```

Jimbo4350 (changes requested):

coot:
> Why did you remove the kind field "ConnectionManagerCounters"?

Because it's a duplicate, the `ConnectionManagerTracer` already adds it: [ref](https://github.com/input-output-hk/cardano-node/blob/cf592b7daa5bc535d13814a0a6a56298001a66bf/cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs#L1845).

cardano-node/src/Cardano/Tracing/Startup.hs
coot: Because it produces a different object with different keys and the duplication is minimal.  But if you want, I can change this.

coot: fixed

Jimbo4350 (approved):

coot:
bors merge

karknu (approved):
LGTM

coot:
bors merge

PR #3443, opened 2021-12-16
Add testnet config files

ArturWieczorek:


ArturWieczorek:
@Jimbo4350 can we add configs for `testnet` ? There are config files for `mainnet` and `shelley-qa` but `testnet` is missing.

Jimbo4350 (approved):

PR #3444, opened 2021-12-16
tx-generator: bug fix: breakup large STM transactions.

MarcFontaine:
This fixed a bug in the tx-generator that affects proper shutdown of
the Node2Node submission clients.
The tx-generator uses an internal TX queue of size 32
and the code for shutting the submission clients down
had tried to write n times Nothing to that queue (n== number of nodes) in one atomic transaction.
TODO: TX-queue is not really needed in the new design and is to be removed altogether.


deepfire (approved):
Thank you @MarcFontaine!

PR #3446, opened 2021-12-16
ouroboros-network:  bump to a fixed version

deepfire:
This bumps consensus to a version that fixes the benchmarking-oriented special case in ledger state setup.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -252,6 +252,7 @@ source-repository-package
     ouroboros-network
     ouroboros-network-framework
     ouroboros-network-testing
+    strict-stm
     typed-protocols
     typed-protocols-cborg
     typed-protocols-examples
--- a/cabal.project
+++ b/cabal.project
@@ -1,4 +1,4 @@
-index-state: 2021-12-15T00:00:00Z
+index-state: 2021-12-07T00:00:00Z
 
 packages:
     cardano-api
--- a/cabal.project
+++ b/cabal.project
@@ -238,7 +238,7 @@ source-repository-package
   type: git
   location: https://github.com/input-output-hk/ouroboros-network
   tag: 7d36feaa3b929d642d11ec9851b0a7fc0260f509
-  --sha256: 1i28c2z40804swaz49qy7k89y1kh4l3dwcfx3aip1m1cfsp2w8mn
+  --sha256: 1vaws0i6vp5mmx9bwpmv8wfk9yivnipknpgyz518w331id0c8x53
   subdir:
     io-sim
     io-classes
--- a/cabal.project
+++ b/cabal.project
@@ -237,8 +237,8 @@ source-repository-package
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/ouroboros-network
-  tag: 67b7c9e7a684e97ad5a5ed7fe85e2b900e8bba11
-  --sha256: 1cjwqpwfdl61pg847pc5h2sm2k1rac7i1f1rv89snmnbzy8s42hj
+  tag: e97580a83c1bcd83a165f02afc06aab81ede4101
+  --sha256: 1i28c2z40804swaz49qy7k89y1kh4l3dwcfx3aip1m1cfsp2w8mn
   subdir:
     io-sim
     io-classes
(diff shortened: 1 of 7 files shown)


cabal.project
```diff
@@ -1,4 +1,4 @@
-index-state: 2021-12-07T00:00:00Z
+index-state: 2021-12-15T00:00:00Z
```
nc6, line 1: I don't see why this bump is necessary. It's already much newer than that used in consensus.

cardano-node/src/Cardano/Tracing/OrphanInstances/Consensus.hs
```diff
@@ -423,26 +419,31 @@ instance ( ConvertRawHash blk
...
           "Chain selection run for block previously from future: " <> renderRealPointAsPhrase pt
       ChainDB.TraceLedgerReplayEvent ev -> case ev of
         LedgerDB.ReplayFromGenesis _replayTo ->
           "Replaying ledger from genesis"
-        LedgerDB.ReplayFromSnapshot snap tip' _replayTo ->
+        LedgerDB.ReplayFromSnapshot snap tip' (LedgerDB.ReplayStart replayFrom) (LedgerDB.ReplayGoal replayTo) ->
           "Replaying ledger from snapshot " <> showT snap <> " at " <>
-            renderRealPointAsPhrase tip'
-        LedgerDB.ReplayedBlock pt _ledgerEvents replayTo ->
+            renderRealPointAsPhrase tip' <> ".  Pushing"
```
nc6, line 436:
```suggestion
            renderRealPointAsPhrase tip' <> ".  Replaying"
```

nc6 (approved):
Looks reasonable to me!

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -245,7 +243,8 @@ instance ElidingTracer (WithSeverity (ChainDB.TraceEvent blk)) where
       return (Just ev, count)
   conteliding _tverb _tr ev@(WithSeverity _ (ChainDB.TraceGCEvent _)) (_old, count) =
       return (Just ev, count)
-  conteliding _tverb tr ev@(WithSeverity _ (ChainDB.TraceLedgerReplayEvent (LedgerDB.ReplayedBlock pt [] replayTo))) (_old, count) = do
+                                                                                                                              -- TODO should this be used?
+  conteliding _tverb tr ev@(WithSeverity _ (ChainDB.TraceLedgerReplayEvent (LedgerDB.ReplayedBlock pt [] (LedgerDB.ReplayStart _replayFrom) (LedgerDB.ReplayGoal replayTo)))) (_old, count) = do
```
newhoggy, line 247: There is an unused `_replayFrom` here.  Does it need to be used?

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
```diff
@@ -490,6 +490,7 @@ instance HasSeverityAnnotation (InboundGovernorTrace addr) where
       InboundGovernor.TrRemoteState {}             -> Debug
       InboundGovernor.TrUnexpectedlyFalseAssertion {}
                                                    -> Error
+      InboundGovernor.TrDemotedToWarmRemote {}     -> Info
```
newhoggy, line 493: Is this a good log level choice for this event?

cardano-node/src/Cardano/Tracing/OrphanInstances/Consensus.hs
dnadales: @EncodePanda @Jasagredo do you agree with this suggestion? :point_up:

cardano-node/src/Cardano/Tracing/Tracers.hs
dnadales: @Jasagredo @EncodePanda

cabal.project
newhoggy: I reverted this change.

cardano-node/src/Cardano/Tracing/OrphanInstances/Consensus.hs
EncodePanda: well, it does not fired during replaying but chain selection and pushing block to the ledger, we have separate events for replaying

PR #3447, opened 2021-12-16
add incremental stake field to UTxOState json

JaredCorduan:
One field was missing from the JSON output of the ledger state

### DIFF
--- a/cardano-api/src/Cardano/Api/Orphans.hs
+++ b/cardano-api/src/Cardano/Api/Orphans.hs
@@ -118,6 +118,11 @@ instance ( Consensus.ShelleyBasedEra era
                          , "delegationState" .= Shelley._delegationState lState
                          ]
 
+instance Crypto.Crypto crypto => ToJSON (ShelleyLedger.IncrementalStake crypto) where
+  toJSON iStake = object [ "credentials" .= Map.toList (ShelleyLedger.credMap iStake)
+                         , "pointers" .= Map.toList (ShelleyLedger.ptrMap iStake)
+                         ]
+
 instance ( Consensus.ShelleyBasedEra era
          , ToJSON (Core.TxOut era)
          , ToJSON (Core.PParamsDelta era)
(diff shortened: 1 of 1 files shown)


cardano-api/src/Cardano/Api/Orphans.hs
```diff
@@ -118,6 +118,11 @@ instance ( Consensus.ShelleyBasedEra era
                          , "delegationState" .= Shelley._delegationState lState
                          ]
 
+instance Crypto.Crypto crypto => ToJSON (ShelleyLedger.IncrementalStake crypto) where
+  toJSON iStake = object [ "stake" .= Map.toList (ShelleyLedger.credMap iStake)
```
lehins, line 122: Nested `stake.stake` is a little confusing, maybe something like `stake.credentials` instead?

```suggestion
  toJSON iStake = object [ "credentails" .= Map.toList (ShelleyLedger.credMap iStake)
```

newhoggy: I like this suggestion.

newhoggy:
I've found that if you rebase on `origin/master` the Haddock will succeed.

JaredCorduan:
I see this property test failure in the CI logs:
```
          714 ┃   H.note_ "Check that the withdrawal from the Plutus staking address was successful"
          715 ┃   lovelaceAtPlutusAddr === pr + 5000000 + 5000000 + 5000000 + 5000000 + Lovelace minrequtxo
        
          This failure can be reproduced by running:
          > recheck (Size 0) (Seed 13016150246274623908 15963492280864531823) Spec.Plutus.Direct.CertifyingAndWithdrawingPlutus
        
      Use '--hedgehog-replay "Size 0 Seed 13016150246274623908 15963492280864531823"' to reproduce.
```

I don't think my change could have caused this failure. Should I make a github issue for this?

Jimbo4350 (approved):

PR #3466, opened 2021-12-26
Expect hexadecimal asset name - ValueParser Message

gitmachtl:
This is a left over, changed to "hexadecimal asset name" as expected. Was "alphanumeric asset name" before.


cblp (approved):

cblp:
bors merge

PR #3478, opened 2022-01-07
Guard against overflows in Shelley TxIns

sevanspowell:
- makeByronTransactionBody guards against overflows in the transaction indices,
but makeShelleyTransactionBody does not.
- I lost some time debugging this here: https://github.com/input-output-hk/cardano-wallet/pull/3071
- Add appropriate guards to makeShelleyTransactionBody.

Now, I have no idea if the `maxShelleyTxInIx` value here is correct, so help would be appreciated!


newhoggy:
To be merged in https://github.com/input-output-hk/cardano-node/pull/4956 to trigger CI

PR #3485, opened 2022-01-09
Revert "Use cardano-config project from github."

tmcgilchrist:
This reverts commit 441115e3fe023144f633dd3e70c657738eca6024.

Further it introduces a new package `cardano-git-rev` to better capture what this code does.
Namely providing a mechanism for both `cabal` and `nix` to provide a git sha in the `--version` flag for `cardano-node`.


cardano-config/README.md
```diff
@@ -1,3 +1,18 @@
...
+
+`cardano-node` support building via `nix` and `cabal`
+
+When building with `nix` the git executable and git metadata isn't available so the
+git revision is embedded as a series of 40 zeros during the build. After the nix build
+is finished the executable is patched with the correct git sha. See [set-git-rev.hs][set-git-rev.hs]
+
+For `cabal` the Template Haskell function `gitRev` executes in the context of the current git checkout,
+requiring the `git` executable and git metadata is available at compile time. This fails on cross-compiling
+ARM executables, for cross-compiling use nix.
```
angerman, line 13: I don't follow.

angerman: We know this doesn't work with `nix`, though I still think we should be able to improve this with nix, and yes cross compilation is exclusive through nix, so this can't work,. It's in the `cabal` section though, so I'm fairly confused.

tmcgilchrist: Perhaps my wording isn't descriptive enough here and I can drop the details about cross-compiling.
The short version is `cabal` builds require `git` and a git checkout while building to get an accurate git sha. 

It took me a while to discover how both nix and cabal, get the right git sha patched into the final executable, I wanted to capture that knowledge somewhere. 

https://github.com/input-output-hk/cardano-node/issues/3440

cardano-config/LICENSE
```diff
@@ -0,0 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2021 Input Output (Hong Kong) Ltd.
```
Jimbo4350, line 190: Should be 2022

cardano-config/NOTICE
```diff
@@ -0,0 +1,14 @@
+Copyright 2021 Input Output (Hong Kong) Ltd.
```
Jimbo4350, line 1: 2022

cardano-config/src/Cardano/Config/Git/RevFromGit.hs
```diff
@@ -0,0 +1,25 @@
...
+    where
+        runGitRevParse :: IO String
+        runGitRevParse = handleJust missingGit (const $ pure "") $ do
+            (exitCode, output, _) <-
+                readProcessWithExitCode "git" ["rev-parse", "--verify", "HEAD"] ""
+            pure $ case exitCode of
+                ExitSuccess -> output
+                _           -> ""
+
+        missingGit e = if isDoesNotExistErrorType (ioeGetErrorType e) then Just () else Nothing
```
angerman, line 25: Why are we reverting this and not incorporate the fixes from `cardano-config`? This code here is clearly broken with some `process` libraries. If `git` is not found, this does not necessarily result in a `isDoesNotExistErrorType` but some other exception that is thrown. It's fairly annoying that the process lirbary got changed to support spawn, and now has a different behaviour. But this code is sadly broken as it is.

angerman:
```
            # src/Cardano/Config/Git/Rev.hs:33:35: error:
            #     • Exception when trying to run compile-time code:
            #         git: readCreateProcessWithExitCode: posix_spawn_file_actions_adddup2(child_end): invalid argument (Invalid argument)
            #       Code: gitRevFromGit
            #     • In the untyped splice: $(gitRevFromGit)
            #    |
            # 33 |         fromGit = T.strip (T.pack $(gitRevFromGit))
            #    |
            ```

Comment (angerman) [reply cardano-config/src/Cardano/Config/Git/RevFromGit.hs:25]: (And similar exceptions...)

File: cardano-git-rev/src/Cardano/Config/Git/RevFromGit.hs [hunk 5]
```diff
@@ -0,0 +1,32 @@
+module Cardano.Config.Git.RevFromGit (
```
erikd, line 1: Maybe this module name should change?

cardano-git-rev/cardano-git-rev.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version: 3.0
 
-name:                   cardano-config
+name:                   cardano-git-rev
```
erikd, line 3: I think I would just have called it `cardano-gitrev` .

cardano-config/src/Cardano/Config/Git/RevFromGit.hs
tmcgilchrist: Copied the code fix across from https://github.com/input-output-hk/cardano-config/pull/4

cardano-git-rev/cardano-git-rev.cabal
Jimbo4350: @tmcgilchrist Does it have to be a separate package or can it be a module?

tmcgilchrist: It's easier to keep this as a small independent package since it's referred to in a number of places. Finding the root of all those to place it was used and putting the code there ended up being the worse option.

cardano-git-rev/src/Cardano/Config/Git/RevFromGit.hs
newhoggy: I agree.  The module name should change.

tmcgilchrist: Changed the module name.

newhoggy (approved):

cardano-git-rev/src/Cardano/Git/RevFromGit.hs
```diff
@@ -0,0 +1,32 @@
+module Cardano.Git.RevFromGit (
+      gitRevFromGit
```
Jimbo4350, line 2: So we format the export list as follows:
```
module Cardano.Git.RevFromGit 
  ( gitRevFromGit
  ) where
```
You can refer to `Export Lists` in the `STYLE.md`

cardano-git-rev/src/Cardano/Git/RevFromGit.hs
```diff
@@ -0,0 +1,32 @@
...
+import           Prelude (String)
+
+import qualified Language.Haskell.TH as TH
+import qualified System.IO as IO
+import           System.IO.Error (isDoesNotExistError)
+import           System.Process (readProcessWithExitCode)
+
+-- | Git revision found by running git rev-parse. If git could not be
+-- executed, then this will be an empty string.
+gitRevFromGit :: TH.Q TH.Exp
```
Jimbo4350, line 15: We also indent with 2 spaces.

cardano-node/src/Cardano/Node/Configuration/Logging.hs
```diff
@@ -1,9 +1,9 @@
-{-# LANGUAGE CPP                 #-}
-{-# LANGUAGE FlexibleContexts    #-}
-{-# LANGUAGE GADTs               #-}
-{-# LANGUAGE OverloadedStrings   #-}
-{-# LANGUAGE PackageImports      #-}
-{-# LANGUAGE Rank2Types          #-}
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE FlexibleContexts #-}
```
Jimbo4350, line 2: :+1:

Jimbo4350 (approved):
LGTM :+1:. Approving with the caveat that the small style changes will be made before merging.

cardano-git-rev/src/Cardano/Git/RevFromGit.hs
tmcgilchrist: Cheers, I didn't know about the style guide. 👍🏻

denisshevchenko (approved):

PR #3493, opened 2022-01-11
tx-generator: use new JSON format in tx-generator-service

MarcFontaine:
Use the auto-derived aeson instances for the tx-generator configs.
This format is easier to maintain in Haskell but not more difficult to read in nix.


MarcFontaine:
subsumed by https://github.com/input-output-hk/cardano-node/pull/3638

PR #3494, opened 2022-01-11
Update leadership-schedule command to calculate following epoch's schedule

### LINKED ISSUE #3546: Upgrade leadership schedule command to predict following epoch's leadership schedule
Currently the leadership schedule command can predict the current epoch's leadership schedule. We should also implement the ability to predict the following epoch's leadership schedule so that SPOs can know when they are expected to mint a block.

Jimbo4350:
`cardano-cli query leadership-schedule ...` now accepts the flag `--next` that allows the leadership schedule calculation for the following epoch. 

Resolves #3546

### DIFF
--- a/cardano-api/src/Cardano/Api/LedgerState.hs
+++ b/cardano-api/src/Cardano/Api/LedgerState.hs
@@ -45,12 +45,14 @@ module Cardano.Api.LedgerState
   , LeadershipError(..)
   , constructGlobals
   , currentEpochEligibleLeadershipSlots
+  , nextEpochEligibleLeadershipSlots
   )
   where
 
 import           Prelude
 
 import           Control.Exception
+import           Control.Monad (when)
 import           Control.Monad.Trans.Class
 import           Control.Monad.Trans.Except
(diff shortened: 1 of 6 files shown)


cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -1256,41 +1278,200 @@ instance Error LeadershipError where
...
+
+
+  relativeStake <- maybe (Left $ LeaderErrStakePoolHasNoStake poolid)
+                         (Right . ShelleyAPI.individualPoolStake) $ Map.lookup poolHash markSnapshotPoolDistr
+
+
+  let Tick.TicknState epochNonce _ = TPraos.csTickn chainDepState
+
+
+  return $ isLeadingSlots sbe rOfInterest epochNonce pParams vrfSkey relativeStake f
```
JaredCorduan, line 1364:
```suggestion
  return $ isLeadingSlots sbe rOfInterest candidateNonce pParams vrfSkey relativeStake f
```

JaredCorduan (approved):
LGTM!

PR #3496, opened 2022-01-13
trace-forward: new API for acceptor's part.

denisshevchenko:
This PR changes API of the acceptor's part only, it doesn't touch the forwarder's part (used by `trace-dispatcher`):

1. `IO`-action returning `DataPointAsker` is provided. It allows avoiding `unsafePerformIO` in the acceptor application.
2. Peer error handlers were added. It allows the acceptor application to know if the connection with the node was dropped.

Additionally:

1. `ekg-forward` dependency is updated.
2. `DataPoint` protocol is fixed: now there is a check if the reply with `DataPoint`s is empty.

**Please note that `cardano-tracer` service, the next part of the new tracing infrastructure, depends on this PR. After it will be merged, the PR for `cardano-tracer` will be opened.**

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -279,8 +279,8 @@ source-repository-package
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/ekg-forward
-  tag: 2adc8b698443bb10154304b24f6c1d6913bb65b9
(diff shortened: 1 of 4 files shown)


deepfire (approved):
@denisshevchenko, `s/Asker/Requestor`, otherwise it's perfect!

Jimbo4350 (approved):

PR #3497, opened 2022-01-13
Update ouroboros-network & tracing changes

coot:
Make the following traces on by default:
* TraceConnectionManager
* TraceInboundGovernor
* TracePeerSelection
* TracePeerSelectionActions

All of them are not chatty if network conditions are stable.

* Fix peer selection tracer.
* Add transition tracers 

input-output-hk/ouroboros-network#3571

### DIFF
--- a/cardano-node/src/Cardano/Node/Tracing/Tracers.hs
+++ b/cardano-node/src/Cardano/Node/Tracing/Tracers.hs
@@ -51,7 +51,7 @@ import qualified Ouroboros.Consensus.Node.Tracers as Consensus
 import qualified Ouroboros.Consensus.Storage.ChainDB as ChainDB
 import qualified Ouroboros.Consensus.Storage.LedgerDB.OnDisk as LedgerDB
 
-import qualified Ouroboros.Network.BlockFetch.ClientState as BlockFetch
+import           Network.Mux.Trace (TraceLabelPeer (..))
 import           Ouroboros.Network.ConnectionId (ConnectionId)
 import qualified Ouroboros.Network.Diffusion as Diffusion
 import qualified Ouroboros.Network.Diffusion.NonP2P as NonP2P
@@ -68,7 +68,7 @@ mkDispatchTracers
 
   , LogFormatting (LedgerEvent blk)
   , LogFormatting
-    (BlockFetch.TraceLabelPeer
+    (TraceLabelPeer
       (ConnectionId RemoteAddress) (TraceChainSyncClientEvent blk))
   )
   => NodeKernelData blk
@@ -198,7 +198,7 @@ mkConsensusTracers :: forall blk.
   ( Consensus.RunNode blk
   , LogFormatting (ChainDB.InvalidBlockReason blk)
   , TraceConstraints blk
-  , LogFormatting (BlockFetch.TraceLabelPeer
+  , LogFormatting (TraceLabelPeer
                     (ConnectionId RemoteAddress) (TraceChainSyncClientEvent blk))
   )
   => Trace IO FormattedMessage
--- a/cardano-node/src/Cardano/Node/Tracing/Tracers.hs
+++ b/cardano-node/src/Cardano/Node/Tracing/Tracers.hs
@@ -593,6 +593,13 @@ mkDiffusionTracersExtra trBase trForward mbTrEKG _trDataPoint trConfig EnabledP2
       severityConnectionManager
       allPublic
     configureTracers trConfig docConnectionManager [connectionManagerTr]
+    connectionManagerTransitionsTr  <-  mkCardanoTracer
+      trBase trForward mbTrEKG
+      "ConnectionManagerTransition"
+      (namesForConnectionManagerTransition @RemoteAddress)
+      severityConnectionManagerTransition
+      allPublic
+    configureTracers trConfig docConnectionManagerTransition [connectionManagerTransitionsTr]
     serverTr  <-  mkCardanoTracer
       trBase trForward mbTrEKG
       "Server"
@@ -605,6 +612,13 @@ mkDiffusionTracersExtra trBase trForward mbTrEKG _trDataPoint trConfig EnabledP2
       "InboundGovernor"
       namesForInboundGovernor
       severityInboundGovernor
+      allPublic
+    configureTracers trConfig docInboundGovernorRemote [inboundGovernorTr]
+    inboundGovernorTransitionsTr  <-  mkCardanoTracer
+      trBase trForward mbTrEKG
+      "InboundGovernorTransition"
+      namesForInboundGovernorTransition
+      severityInboundGovernorTransition
       allPublic
     configureTracers trConfig docInboundGovernorRemote [inboundGovernorTr]
     localConnectionManagerTr  <-  mkCardanoTracer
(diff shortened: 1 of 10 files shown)


cardano-node/src/Cardano/Node/Tracing/Tracers/NodeToClient.hs
```diff
@@ -213,6 +222,152 @@ docTChainSync = Documented [
...
+    namesForTTxMonitor''' LTM.MsgHasTx {}         = ["HasTx"]
+    namesForTTxMonitor''' LTM.MsgReplyHasTx {}    = ["ReplyHasTx"]
+    namesForTTxMonitor''' LTM.MsgGetSizes {}      = ["GetSizes"]
+    namesForTTxMonitor''' LTM.MsgReplyGetSizes {} = ["ReplyGetSizes"]
+    namesForTTxMonitor''' LTM.MsgRelease {}       = ["Release"]
+    namesForTTxMonitor''' LTM.MsgDone {}          = ["Done"]
+
+instance LogFormatting (AnyMessageAndAgency (LTM.LocalTxMonitor txid tx slotNo)) where
+  forMachine _dtal (AnyMessageAndAgency stok LTM.MsgAcquire {}) =
+    mkObject [ "kind" .= String "MsgAcuire"
```
newhoggy, line 263: Typo: `MsgAcuire`.

Also below.

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
```diff
@@ -809,6 +837,52 @@ instance (forall result. Show (query result))
              , "agency" .= String (pack $ show stok)
              ]
 
+instance ToObject (AnyMessageAndAgency (LocalTxMonitor txid tx slotno)) where
+  toObject _verb (AnyMessageAndAgency stok LocalTxMonitor.MsgAcquire {}) =
+    mkObject [ "kind" .= String "MsgAcuire"
```
newhoggy, line 842: Typo `MsgAcuire`.

Also below.

karknu (approved):
LGTM.
I'd like a newer ouroboros-network version once https://github.com/input-output-hk/ouroboros-network/pull/3535 is merged.

cardano-node/src/Cardano/Node/Tracing/Tracers.hs
```diff
@@ -363,6 +363,13 @@ mkNodeToClientTracers trBase trForward mbTrEKG _trDataPoint trConfig = do
                 severityTChainSync
                 allPublic
     configureTracers trConfig docTChainSync [chainSyncTr]
+    txMonitorTr  <-  mkCardanoTracer
+                trBase trForward mbTrEKG
```
Jimbo4350, line 367: Can you fix the indent here? It should be two spaces:
```
mkCardanoTracer
 trBase trForward mbTrEKG
 "TxMonitorClient"
  ...
```

cardano-node/src/Cardano/Node/Tracing/Tracers/NodeToClient.hs
```diff
@@ -213,6 +222,152 @@ docTChainSync = Documented [
...
+  where
+    namesForTTxMonitor' (TraceSendMsg msg) = "Send"    : namesForTTxMonitor'' msg
+    namesForTTxMonitor' (TraceRecvMsg msg) = "Receive" : namesForTTxMonitor'' msg
+
+    namesForTTxMonitor'' (AnyMessageAndAgency _agency msg) = namesForTTxMonitor''' msg
+
+    namesForTTxMonitor''' :: Message (LTM.LocalTxMonitor (GenTxId blk) (GenTx blk) SlotNo)
+                                     from to
+                          -> [Text]
+    namesForTTxMonitor''' LTM.MsgAcquire {}       = ["Acquire"]
```
Jimbo4350, line 249: Please don't align code

cardano-node/src/Cardano/Tracing/Config.hs
```diff
@@ -179,7 +179,7 @@ traceConfigParser v ctor =
       chainSyncProtocol :: OnOff TraceChainSyncProtocol
       chainSyncProtocol = OnOff False
       connectionManager :: OnOff TraceConnectionManager
-      connectionManager = OnOff False
+      connectionManager = OnOff True
```
Jimbo4350, line 186: So I think the correct place to add defaults would be in `defaultPartialNodeConfiguration` due to: https://github.com/input-output-hk/cardano-node/blob/e5d32b3558f436d3178296d19b23578524681839/cardano-node/src/Cardano/Node/Run.hs#L115

Because then the user can override the tracers if they want to via the config yaml file.

I don't think you can actually turn off the tracing if you change the values here.

Jimbo4350 (changes requested):
A couple minor formatting changes and 1 question concerning the setting of default tracers. Looks good otherwise :+1:

cardano-node/src/Cardano/Node/Tracing/Tracers.hs
coot: This is consistent with the surrounding.  I'll change it consistently in the code around.

cardano-node/src/Cardano/Node/Tracing/Tracers/NodeToClient.hs
coot: That's done in the whole module it will look inconsistent if we do that.

cardano-node/src/Cardano/Node/Tracing/Tracers.hs
coot: I pushed a commit which fixes it.

cardano-node/src/Cardano/Tracing/Config.hs
coot: oh, there are some other tracers with the same problem, you might want to look into it at some point.

coot: I'd like to leave it out as is; there seems to be work to be done to make it right.  One cannot use `Last` for entire tracing config (`pncTracingConfig :: Last TraceOptions`), since when one is passing config it will almost always overwrite all values.  Instead, `pncTracingConfig` should have type `TraceOptions` and each of its field should be wrapped in `Last`, and a monoid instance for `TraceOptions` should be derived generically.  Doing this is beyond this pr.

cardano-node/src/Cardano/Node/Tracing/Tracers/NodeToClient.hs
Jimbo4350: If we leave the aligned code it will further encourage others to do so. I'm trying to prevent additional alignment of code like this. It's ok if it looks inconsistent as in the future I will have to undo alignment in the library.

cardano-node/src/Cardano/Tracing/Config.hs
Jimbo4350: Yes this is beyond the scope of the PR.

cardano-node/src/Cardano/Node/Tracing/Tracers/NodeToClient.hs
coot: Though is this a problem, usually when submitting a PR which needs to re-align some of the code, one can filter out space only changes.  Most tools allow that, including github UI.  Or maybe, there's something else to it?

Jimbo4350: The problem is future changes to one line of the aligned code can potentially introduce semantically null changes in all of the surrounding code that needs to be re-aligned.

coot: Indeed, one can filter  out space only changes by [ticking a box](https://github.blog/2018-05-01-ignore-white-space-in-code-review/) in github diff tool.  The same in various editors.

coot: but if you still want unaligned code it's up to you

Jimbo4350: I'd like to keep the code formatting in `cardano-node` as consistent as possible and as such we have a  `STYLE.md`: 
> Indentation
    Do not vertically align code. Mainly because the need to change identifiers and re-aligning the other lines causes semantically null changes in the diff.

coot: Fixed

cardano-node/src/Cardano/Tracing/Config.hs
coot: I created an issue for this https://github.com/input-output-hk/cardano-node/issues/3534

newhoggy (approved):

Jimbo4350 (approved):

coot:
bors merge

coot:
bors merge

deepfire (approved):

coot:
bors merge

PR #3498, opened 2022-01-13
cardano-tracer: logging/monitoring service for node

denisshevchenko:
This PR introduces `cardano-tracer`, the last part of the new tracing infrastructure.

Previously, the node handled all the logging by itself. Moreover, it provided monitoring tools as well: two web-servers, for Prometheus and for EKG monitoring page. `cardano-tracer` is a result of _moving_ all the logging/monitoring-related stuff from the node to a separate service. As a result, the node became smaller, faster, and simpler.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -10,6 +10,7 @@ packages:
     cardano-node-chairman
     cardano-submit-api
     cardano-testnet
+    cardano-tracer
     bench/cardano-topology
     bench/locli
     bench/tx-generator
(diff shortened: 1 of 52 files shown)


cabal.project
```diff
@@ -15,6 +15,7 @@ packages:
     trace-dispatcher
     trace-resources
     trace-forward
+    cardano-tracer
```
deepfire, line 18: Sorry for the nitpicking, but I imagine people prefer to keep this list sorted.

deepfire (approved):
Thank you @denisshevchenko

cabal.project
denisshevchenko: Fixed

Jimbo4350 (approved):

PR #3505, opened 2022-01-18
Output as ledger compliant CDDL flag

### LINKED ISSUE #3370: [BUG] - Cardano-cli creates transactions with witness-set as an array, not a map - this is invalid.
**Summary**
Cardano-cli generates transactions which do not meet the ledger specification. Furthermore, if you're trying to sign these transactions in a browser wallet such as Nami - it will not parse them, because serialization-lib (correctly) detects the transactions produced by Cardano-cli as invalid. This is a major stumbling block for dApps looking to integrate a Cardano-node backend with any browser wallet which uses serialization-lib - as things stand, that's just not possible - due to such a simple error:

In the case of an empty witness set, the field should be an empty map {}, not an empty array [].

**Steps to reproduce**
Generate a transaction with Cardano-cli and observe the witness-set field - in the case of an empty witness set, you will find that it is an empty array, when the ledger spec indicated that it should be a map. 

**Expected behavior**
Witness set field should be an empty map: {}

Latest everything

Example cbor: 
86a600828258207a9ae75e073a713791cc3045c2c2f62b925c4d33ae670da445ec80f20c459d7f0082582093ce7d39fc764e8ec9026a50f1f41f0d18649e18378718e91c45060fead0c703010d80018382583900af035fb9a547276bbb86a1936882524463eccb82465fcdf44489013713a39f099c533eb97b848aea0ae05850b4b40ec1ec05541a3f84c9b11a0483d91a83581d70eb18d99e0c392b18a436d595b5223e8fff559c4643118c1d403dfaab821a001e8480a1581c79e78a324e27842aa2ccaa21a710198b1db2c2a835b70d663b68a9dca1534d79546573746e6574547231303030303237310158201db43f14402b617c93f860381ec852799320d35c1be86386dffa53d792d846d582583900af035fb9a547276bbb86a1936882524463eccb82465fcdf44489013713a39f099c533eb97b848aea0ae05850b4b40ec1ec05541a3f84c9b1821a01312d00a2581c6b8d07d69639e9413dd637a1a815a7323c69c86abbafb66dbfdb1aa7a14002581c79e78a324e27842aa2ccaa21a710198b1db2c2a835b70d663b68a9dca2534d79546573746e65745472313030303032363701534d79546573746e65745472313030303032363801021a0002d8a50e8007582053d1932595c943555dd467aa2d3a36efdc865caf35ad77c04c270edcfd2dc39c9fff8080f5d90103a100a119013863626172

**Additional context**
We're trying to complete our smart validator integration at Artifct.app - our architecture involved having cardano-cli generate transactions which were then sent to Nami wallet for signing in the browser. Nami uses serialization-lib to parse the transactions, but it fails on every transaction we generate from Cardano-cli due to this invalid witness set field. I have discussed this with an Emurgo developer on their issues tracker:
https://github.com/Emurgo/cardano-serialization-lib/issues/259

Jimbo4350:
We implement an alternative `cardano-ledger` CDDL compliant serialization for the cli. The plan is to eventually deprecate the cli's intermediate format and to fully embrace the ledger's CDDL spec. 

There is now an optional `--cddl-format` flag in the `build` and `build-raw` commands that serialize the txbody as a **key** unwitnessed transaction. Any script witnesses specified at this stage are included in the witness set.

Resolves #3370

### DIFF
--- a/cardano-api/src/Cardano/Api/SerialiseLedgerCddl.hs
+++ b/cardano-api/src/Cardano/Api/SerialiseLedgerCddl.hs
@@ -18,6 +18,12 @@ module Cardano.Api.SerialiseLedgerCddl
 
   , writeTxFileTextEnvelopeCddl
   , writeTxWitnessFileTextEnvelopeCddl
+
+  -- Exported for testing
+  , serialiseTxLedgerCddl
+  , deserialiseTxLedgerCddl
+  , serialiseWitnessLedgerCddl
+  , deserialiseWitnessLedgerCddl
   )
   where
 
@@ -94,7 +100,7 @@ data TextEnvelopeCddlError
       Text   -- ^ Actual types
   | TextEnvelopeCddlErrUnknownType Text
   | TextEnvelopeCddlErrByronKeyWitnessUnsupported
-  deriving Show
+  deriving (Show, Eq)
 
 instance Error TextEnvelopeCddlError where
   displayError (TextEnvelopeCddlErrCBORDecodingError decoderError) =
@@ -190,11 +196,11 @@ deserialiseWitnessLedgerCddl
 deserialiseWitnessLedgerCddl era TextEnvelopeCddl{teCddlRawCBOR,teCddlDescription} =
   --TODO: Parse these into types
   case teCddlDescription of
-    "Key BootstrapWitness Shelley" -> do
+    "Key BootstrapWitness ShelleyEra" -> do
       w <- first TextEnvelopeCddlErrCBORDecodingError
              $ CBOR.decodeAnnotator "Shelley Witness" fromCBOR (LBS.fromStrict teCddlRawCBOR)
       Right $ ShelleyBootstrapWitness era w
-    "Key Witness Shelley" -> do
+    "Key Witness ShelleyEra" -> do
       w <- first TextEnvelopeCddlErrCBORDecodingError
              $ CBOR.decodeAnnotator"Shelley Witness" fromCBOR (LBS.fromStrict teCddlRawCBOR)
       Right $ ShelleyKeyWitness era w
--- a/cardano-api/src/Cardano/Api/SerialiseLedgerCddl.hs
+++ b/cardano-api/src/Cardano/Api/SerialiseLedgerCddl.hs
@@ -112,8 +112,7 @@ instance Error TextEnvelopeCddlError where
   displayError (TextEnvelopeCddlTypeError expTypes actType) =
     "TextEnvelopeCddl type error: "
        <> " Expected one of: "
-       <> List.intercalate ", "
-            [Text.unpack expType | expType <- expTypes]
+       <> List.intercalate ", " (map Text.unpack expTypes)
        <> " Actual: " <> Text.unpack actType
   displayError (TextEnvelopeCddlErrUnknownType unknownType) =
     "Unknown TextEnvelopeCddl type:" <> Text.unpack unknownType
@@ -194,7 +193,9 @@ deserialiseWitnessLedgerCddl
   -> TextEnvelopeCddl
   -> Either TextEnvelopeCddlError (KeyWitness era)
 deserialiseWitnessLedgerCddl era TextEnvelopeCddl{teCddlRawCBOR,teCddlDescription} =
-  --TODO: Parse these into types
+  --TODO: Parse these into types because this will increase code readability and
+  -- will make it easier to keep track of the different Cddl descriptions via
+  -- a single sum data type.
   case teCddlDescription of
     "Key BootstrapWitness ShelleyEra" -> do
       w <- first TextEnvelopeCddlErrCBORDecodingError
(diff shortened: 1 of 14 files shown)


cardano-api/src/Cardano/Api/SerialiseLedgerCddl.hs
```diff
@@ -0,0 +1,318 @@
...
+    "TextEnvelopeCDDL CBOR decoding error: " <> show decoderError
+  displayError (TextEnvelopeCddlAesonDecodeError fp aesonErr) =
+    "Could not JSON decode TextEnvelopeCddl file at: " <> fp <> " Error: " <> aesonErr
+  displayError TextEnvelopeCddlUnknownKeyWitness =
+    "Unknown key witness specified"
+  displayError (TextEnvelopeCddlTypeError expTypes actType) =
+    "TextEnvelopeCddl type error: "
+       <> " Expected one of: "
+       <> List.intercalate ", "
+            [Text.unpack expType | expType <- expTypes]
```
newhoggy, line 116: Can this be replaced with `fmap Text.unpack expTypes`?

cardano-api/src/Cardano/Api/SerialiseLedgerCddl.hs
```diff
@@ -0,0 +1,318 @@
...
+  witEra ShelleyBasedEraAllegra = "TxWitness AllegraEra"
+  witEra ShelleyBasedEraMary = "TxWitness MaryEra"
+  witEra ShelleyBasedEraAlonzo = "TxWitness AlonzoEra"
+
+deserialiseWitnessLedgerCddl
+  :: ShelleyBasedEra era
+  -> TextEnvelopeCddl
+  -> Either TextEnvelopeCddlError (KeyWitness era)
+deserialiseWitnessLedgerCddl era TextEnvelopeCddl{teCddlRawCBOR,teCddlDescription} =
+  --TODO: Parse these into types
```
newhoggy, line 197: Can you elaborate on the TODO?

cardano-api/src/Cardano/Api/SerialiseLedgerCddl.hs
```diff
@@ -0,0 +1,318 @@
...
+
+textEnvelopeCddlJSONConfig :: Config
+textEnvelopeCddlJSONConfig =
+  defConfig { confCompare = textEnvelopeCddlJSONKeyOrder }
+
+textEnvelopeCddlJSONKeyOrder :: Text -> Text -> Ordering
+textEnvelopeCddlJSONKeyOrder = keyOrder ["type", "description", "cborHex"]
+
+-- | This GADT allows us to deserialise a tx or key witness without
+--having to provide the era.
```
newhoggy, line 239: Add space before `having`.

cardano-api/src/Cardano/Api/SerialiseLedgerCddl.hs
```diff
@@ -0,0 +1,318 @@
...
+
+   expectedTypes :: [Text]
+   expectedTypes = [ typ | FromCDDLTx typ _f <- types ]
+
+   matching :: FromSomeTypeCDDL TextEnvelopeCddl b -> Bool
+   matching (FromCDDLTx ttoken _f) = actualType == ttoken
+   matching (FromCDDLWitness ttoken _f)  = actualType == ttoken
+
+-- TODO: This is atrocious but the plan is to parse the 'teCddlType'
+-- field in the future.
```
newhoggy, line 282: Is it sufficient to just write `TODO Parse the 'teCddlType' field in the future`, so is the atrocious bit pointing to something worth noting explicity?

cardano-api/src/Cardano/Api/Tx.hs
```diff
@@ -348,7 +349,9 @@ pattern AsShelleyWitness :: AsType (KeyWitness ShelleyEra)
 pattern AsShelleyWitness = AsKeyWitness AsShelleyEra
 {-# COMPLETE AsShelleyWitness #-}
 
-
+-- We implement a custom serialization instance that differs from
+-- cardano-ledger because we want to be able to tell the difference between
+-- on disk witnesses for the cli's 'assemble' command.
```
newhoggy, line 354: Suggest: `s/We implement a custom serialization instance that differs/This custom instance differs`

cardano-cli/src/Cardano/CLI/Types.hs
```diff
@@ -73,6 +74,15 @@ data OutputFormat
   | OutputFormatBech32
   deriving (Eq, Show)
 
+-- | Specify whether to serialise  a value according to the ledger's CDDL spec
```
newhoggy, line 77: Remove double space before `a`

cardano-cli/src/Cardano/CLI/Types.hs
```diff
@@ -73,6 +74,15 @@ data OutputFormat
...
+-- or the cli' intermediate format. Note the intermediate format is defined
```
newhoggy, line 78: `s/cli'/cli's`

newhoggy (approved):

cardano-api/src/Cardano/Api/SerialiseLedgerCddl.hs
Jimbo4350: Yep, this was a copy paste.

cardano-api/src/Cardano/Api/SerialiseLedgerCddl.hs
Jimbo4350: I updated the comment though :+1:

Jimbo4350:
bors cancel

PR #3525, opened 2022-01-24
cardano-cli: Add friendly printing of script data for tx outputs (Alonzo)

cblp:


cardano-api/src/Cardano/Api/ScriptData.hs
```diff
@@ -383,11 +383,14 @@ scriptDataToJsonNoSchema = conv
...
 
     -- Script data allows any value as a key, not just string as JSON does.
-    -- For simple types we just convert them to string dirctly.
+    -- For simple types we just convert them to string directly.
     -- For structured keys we render them as JSON and use that as the string.
     convKey :: ScriptData -> Text
     convKey (ScriptDataNumber n) = Text.pack (show n)
-    convKey (ScriptDataBytes bs) = bytesPrefix
+    convKey (ScriptDataBytes bs)
+      | BSC.all (\c -> Char.isAscii c && Char.isPrint c) bs
```
Jimbo4350, line 391: Why this change?

Jimbo4350 (changes requested):

cardano-api/src/Cardano/Api/ScriptData.hs
cblp: A map key is often a program identifier, and it's nicer to view it decoded than in hex.

cblp: without this,
```json
{"random name": "Vincent Ford"}
```
looks
```json
{"0x72616e646f6d206e616d65": "Vincent Ford"}
```

cblp: I followed the comment "For simple types we just convert them to string directly."

Jimbo4350: You missed: 
> -- * JSON hex strings with \"0x\" prefix represented as CBOR byte strings
We can't render `ScriptDataBytes` as ASCII because when parsing back to `ScriptData` the prefix `0x` followed by hex indicates that we want to encode this on chain as a CBOR bytestring.

Jimbo4350: Have a look at `doc/reference/tx-metadata.md`

cblp: What about strings that don't start with `0x`?

Is there another tool to convert a JSON-like datum to corresponding JSON?

cblp: So, if a txout datum input is
```js
{"random name": "Vincent Ford"}
```
then inside the tx we have
```hs
txOutDatum = ScriptDataMap [(ScriptDataBytes "random name",ScriptDataBytes "Vincent Ford")]
```

Both key and value are ScriptDataBytes, so they they both must be displayed with no schema as 0x-base-16 strings?

cblp: So, since this contradicts tx-metadata.md, I removed such behaviour from `scriptDataToJsonNoSchema`.

cardano-api/src/Cardano/Api/ScriptData.hs
```diff
@@ -362,11 +361,6 @@ scriptDataToJsonNoSchema = conv
     conv :: ScriptData -> Aeson.Value
     conv (ScriptDataNumber n) = Aeson.Number (fromInteger n)
     conv (ScriptDataBytes bs)
-      | Right s <- Text.decodeUtf8' bs
```
Jimbo4350, line 365: Why? This is plutus script data not tx metadata.

cardano-cli/src/Cardano/CLI/Run/Friendly.hs
```diff
@@ -149,29 +153,52 @@ friendlyTxOut (TxOut addr amount mdatum) =
...
-   renderDatum TxOutDatumNone = Aeson.Null
-   renderDatum (TxOutDatumHash _ h) =
-     Aeson.String $ serialiseToRawBytesHexText h
-   renderDatum (TxOutDatum _ sData) =
-     scriptDataToJson ScriptDataJsonDetailedSchema sData
+
+    -- | 'ScriptData' does not support text, only byte strings.
+    -- 'ScriptData.scriptDataToJson' formats byte strings as Base16 blobs.
+    -- So here we try to display text data where byte strings look like text.
+    conv :: ScriptData -> Aeson.Value
```
Jimbo4350, line 184: There is a lot of repetition of code here from `scriptDataToJson`. I suggest using `scriptDataToJson`, pattern matching on the `Value`s of interest (`Object` in your case) and dealing with the Map keys as you see fit.

Jimbo4350 (changes requested):

cardano-cli/src/Cardano/CLI/Run/Friendly.hs
cblp: That will be an extra 0x-base-16 encoding followed by extra 0x-base-16 decoding. Is it better?

cardano-api/src/Cardano/Api/ScriptData.hs
cblp: You sent me tx-metadata docs before. I supposed, tx-metadata rules apply here. Don't they?

cardano-cli/src/Cardano/CLI/Run/Friendly.hs
Jimbo4350: I don't understand what you mean. Can you clarify?

Jimbo4350: I don't understand what you mean. Can you clarify?

Jimbo4350: So this is what I have in mind:
```
conv :: ScriptData -> ScriptDataJsonSchema -> Aeson.Value
conv sd schema = 
  case scriptDataToJson schema sd of 
    Aeson.Array a -> ...
    ...
```

cblp:
@Jimbo4350 I've removed changes in ScriptData.hs, which don't add to the purpose of the PR.

PR #3538, opened 2022-01-26
Retry testnet tests

newhoggy:
This will retry any testnet that failed the first time.  The hope is this is a quick fix to reduce chairman test flakiness before we are able to have proper investigation into why the chairman tests fail in the first place.


erikd (approved):

Jimbo4350 (approved):
LGTM! Just squash the commits :smile:

newhoggy:
Looks like this is already merged as part of another PR.  Closing.

Issue #3545, opened 2022-01-27
Incremental transaction signing


Jimbo4350:
Implement the ability to incrementally sign a transaction. I.e the ability to sign a transaction, send that signed transaction to another party who will also add his/her signature.

dorin100:
from an end-user perspective, it would be best to be able to see:
- what keys (or at least how many) need to sign the transaction
- what keys already signed the transaction

PR #3549, opened 2022-01-28
Update transaction sign command to allow for incremental signing

Jimbo4350:
You can now specify an already signed transaction via `--tx-file` in order to add more signatures to it:  

`cardano-cli transaction sign (--tx-body-file FILE | --tx-file FILE) ...`

Resolves: #3545

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Types.hs
+++ b/cardano-cli/src/Cardano/CLI/Types.hs
@@ -22,7 +22,6 @@ module Cardano.CLI.Types
   , ScriptDatumOrFile (..)
   , TransferDirection(..)
   , TxBodyFile (..)
-  , TxBodyOrTxFile (..)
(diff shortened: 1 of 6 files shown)


newhoggy (approved):

cardano-cli/src/Cardano/CLI/Types.hs
```diff
@@ -241,3 +247,15 @@ data EpochLeadershipSchedule
...
 
+newtype TxBodyFile
+  = TxBodyFile FilePath
+  deriving Show
+
+newtype TxFile
+  = TxFile FilePath
+  deriving Show
+
+data TxBodyOrTxFile
```
cblp, line 258: InputTxFile already exists

cblp (changes requested):

cblp (approved):

PR #3550, opened 2022-01-28
flake.nix:  fix membench Hydra invocation

deepfire:
This should fix the flake's definition of the membench job.


deepfire:
Thanks to @cleverca22 for figuring this out.

PR #3571, opened 2022-02-07
Set -O2 flag by default

Jimbo4350:
`cardano-node` was being built with `-O1` by default but we should take advantage of further optimisation with `-O2`.

See: https://downloads.haskell.org/~ghc/9.0.1/docs/html/users_guide/using-optimisation.html#options-optimise


Jimbo4350:
Closed, decided to not bother using -O2 unless there are clear benefits to doing so.

PR #3575, opened 2022-02-08
DWARF debug info: enable

deepfire:
This is to enable DWARF debug info in the closure of `cardano-node`, so as to enable near-overhead-free profiling.


deepfire:
Turns out, @duog says we still need a GHC that had `--enable-dwarf` passed to its `configure`.

We'll be chatting with @angerman

PR #3561, opened 2022-02-02
Added TrInboundGovernorError Trace

bolt12:

Issue #3568, opened 2022-02-04
[BUG] - cardano-cli fails with `MuxError` when p2p is enabled


mkoura:
*Internal*

**Area**
*Other*   p2p

**Summary**
On local network where p2p is enabled, all cardano-cli commands fail:
```
$ cardano-cli query tip --testnet-magic 42
MuxError (MuxIOException writev: resource vanished (Broken pipe)) "(sendAll errored)"

cardano-cli: MuxIOException writev: resource vanished (Broken pipe) "(sendAll errored)"
```

`cardano-node` processes are running, there are no errors in log files. The issue is present in latest master, first it was observed in nightly job that ran with rev 40cf9b428f4bd6156cc8d2aef01ac9475f5a62c5
The issue is not present when p2p is disabled and legacy topology files are used.

**Steps to reproduce**
1. start local cardano network with p2p enabled
2. run any `cardano-cli` command

**System info (please complete the following information):**
- OS Name: Linux
- Node version: 40cf9b428f4bd6156cc8d2aef01ac9475f5a62c5
- CLI version: 40cf9b428f4bd6156cc8d2aef01ac9475f5a62c5

cardano-apexpool:
This bug is still present in 1.34.0, actually even worse. Every query fails. 

        $ cardano-cli query tip --testnet-magic 1097911063 
        MuxError (MuxIOException writev: resource vanished (Broken pipe)) "(sendAll errored)"
        
        cardano-cli: MuxIOException writev: resource vanished (Broken pipe) "(sendAll errored)"

I was trying to check if #3659 was resolved in 1.34.0, but it is not, probably because of this one.

PR #3587, opened 2022-02-09
Bump ledger and network dependencies

nc6:
The changes here come from a few areas:

- Refactoring of the ledger code to move the protocol concerns into their own package.
- Some additional errors for Plutus scripts (not yet enabled).
- Additional ledger events.
- Remove ToJSON instances which have now migrated to the ledger codebase.
- Change `TxIx` to be a `Word16`, which is its new bound in Shelley+ eras.

This also incorporates @bolt12's changes from #3561.


cabal.project
```diff
@@ -252,8 +252,8 @@ source-repository-package
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/ouroboros-network
-  tag: 4fac197b6f0d2ff60dc3486c593b68dc00969fbf
-  --sha256: 1b43vbdsr9m3ry1kgag2p2ixpv54gw7a4vvmndxl6knqg8qbsb8b
+  tag: 4ae7d33231611111780e0ab36b6d648a6527f995
```
coot, line 255: I would like it be set past input-output-hk/ouroboros-network@173d9395c.  This will include the fix for #3568.

cardano-node/src/Cardano/Node/Tracing/Tracers/P2P.hs
```diff
@@ -1422,6 +1432,10 @@ docInboundGovernor peerAddr = Documented
   --     (InboundGovernor.TrUnexpectedlyFalseAssertion protoIGAssertionLocation)
   --     []
   --     ""
+  -- ,  DocMsg
+  --     (InboundGovernor.TrInboundGovernorError protoSomeException)
+  --     []
+  --     ""
```
coot, line 1438: I don't think this should be commented out, the same for the above `TrUnexpectedlyFalseAssertion`.  Could you uncomment both?

cardano-node/src/Cardano/Tracing/OrphanInstances/Shelley.hs
```diff
@@ -987,6 +996,12 @@ instance ToJSON (Alonzo.CollectError StandardCrypto) where
           , "error" .= String "NoCostModel"
           , "language" .= toJSON lang
           ]
+      -- TODO What do we want to show here?
```
JaredCorduan, line 999: I think that `"kind" .= String "PlutusTranslationError"` would make sense.

The `_` has type `TranslationError`, which is an enum with four constructors. The `error` could mention which one it is:
* byron input in the presence of a plutus script
* byron output in the presence of a plutus script
* unnexpected/logic error regarding translating inputs
* unnexpected/logic error regarding translating redeemers

JaredCorduan (approved):
from the ledger side, looks great to me! thanks @nc6 ! (I'll approve, but obviously @coot 's comments need addressing)

deepfire (approved):

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -426,7 +428,7 @@ renderTxIn (TxIn txId (TxIx ix)) =
   serialiseToRawBytesHexText txId <> "#" <> Text.pack (show ix)
 
 
-newtype TxIx = TxIx Word
+newtype TxIx = TxIx Word16
```
Jimbo4350, line 431: This is also used in Byron transactions so if we we should do the appropriate check similarly to below: `guard (fromIntegral txix <= maxByronTxInIx) ?! TxBodyInIxOverflow` in `makeShelleyTransactionBody`.

cardano-node/src/Cardano/Tracing/OrphanInstances/Shelley.hs
```diff
@@ -94,6 +94,9 @@ import           Cardano.Ledger.Shelley.Rules.Tick
 import           Cardano.Ledger.Shelley.Rules.Upec
 import           Cardano.Ledger.Shelley.Rules.Utxo
 import           Cardano.Ledger.Shelley.Rules.Utxow
+import Cardano.Protocol.TPraos.OCert (KESPeriod (KESPeriod))
```
Jimbo4350, line 97: Alignment  and ordering need to be fixed

Jimbo4350 (changes requested):
LGTM other than a couple minor changes

cabal.project
nc6: Now bumped to 48e498cfea694b89f99fd92ae7a86b9fcd5f29f0

cardano-api/src/Cardano/Api/TxBody.hs
nc6: OK. That means that `toShelleyTxIn` can overflow, which feels bad, but I'll change for consistency.

cardano-client-demo/StakeCredentialHistory.hs
```diff
@@ -34,6 +34,8 @@ import           Options.Applicative (Parser, (<|>), (<**>))
 import qualified Options.Applicative as Opt
 import           Ouroboros.Consensus.Shelley.Eras (StandardCrypto)
 import qualified Ouroboros.Consensus.Shelley.Ledger as Shelley
+import qualified Cardano.Ledger.UnifiedMap as UM
```
Jimbo4350, line 37: Import ordering

cabal.project
```diff
@@ -7,14 +7,11 @@ packages:
     cardano-git-rev
     cardano-node
     cardano-node-capi
-    cardano-node-chairman
```
Jimbo4350, line 10: Why have these been removed?

Jimbo4350 (approved):
LGTM!

coot (approved):
LGTM

JaredCorduan (changes requested):
Bad news, a node with these changes does not sync mainnet. I assume this means we do not want to merge this PR until we find the bug.

```
[2022-02-14 20:17:11.26 UTC] Invalid block 1c5d618069f0663f4dea9b6a710e35672af64f59f4f11a50a6be7c09500e2a2e at slot 42762467: ExtValidationErrorLedger (HardForkLedgerErrorFromEra S (S (S (S (Z (WrapLedgerErr {unwrapLedgerErr = BBodyError (BlockTransitionError [ShelleyInAlonzoPredFail (LedgersFailure (LedgerFailure (UtxowFailure (WrappedShelleyEraFailure (UtxoFailure (UtxosFailure (ValidationTagMismatch (IsValid True) (FailedUnexpectedly [PlutusFailure "\nThe 3 arg plutus script (PlutusScript PlutusV1 ScriptHash \"0695a020afc3f4b3299bb9eedee0f21422ce9ad8c82c2baa48ec9721\") fails.\nCekError An error has occurred:  User error:\nThe budget was overspent...
```

EDIT: I am able to sync past this slot if I change this PR to _not_ update Plutus. I think something has change with the cost model calculation.

cardano-node/src/Cardano/Node/Tracing/Tracers/P2P.hs
bolt12: I will, sorry about the delay

JaredCorduan (approved):
:+1:

nc6:
With the latest update, I can now sync past that point:
```
[lorien:cardano.node.ChainDB:Info:5] [2022-02-16 15:48:16.03 UTC] Replayed block: slot 43523999 out of 50390030. Progress: 86.37%
[lorien:cardano.node.ChainDB:Info:5] [2022-02-16 15:48:24.70 UTC] Replayed block: slot 43545598 out of 50390030. Progress: 86.42%
```

nc6:
bors merge

PR #3591, opened 2022-02-10
Trace NodeState: demo

denisshevchenko:
This is the first step toward `NodeState` tracing. The `NodeState` represents the current state of the node. So any external application (for example, `cardano-wallet` or `cardano-tracer`) will be able to ask this state (as a `DataPoint`) and use it.

Currently, `NodeState` is almost useless, it only demonstrates the changing the state of a node in real-time.

This PR introduces `demo-acceptor` as well: a tiny application allowing to ask `DataPoint` from the node.


deepfire (approved):
bors r+

cardano-node/src/Cardano/Node/StateRep.hs
```diff
@@ -0,0 +1,30 @@
+{-# LANGUAGE DeriveGeneric #-}
```
Jimbo4350, line 1: Can we move this into the `Tracing` folder?

Jimbo4350 (commented):
Can we also squash the first two commits?

cardano-node/src/Cardano/Node/StateRep.hs
denisshevchenko: @Jimbo4350 Done.

denisshevchenko:
@Jimbo4350 Done, now there are only 2 commits.

nc6 (approved):

nc6:
bors merge

PR #3608, opened 2022-02-16
Add IsString (Hash BlockHeader)

andreabedini:
Hi there :wave:

This PR adds the instance `IsString (Hash BlockHeader)` following the same pattern of many other keys in `Cardano.Api`. The use case is simple, being able to refer to a specific point on the chain. E.g.

```haskell
recentPoint :: ChainPoint
recentPoint = ChainPoint (SlotNo 53427524) "5e2bde4e504a9888a4f218dafc79a7619083f97d48684fcdba9dc78190df8f99"
```

I am also changing the Show instance, as it seems to be the common pattern.


Jimbo4350 (approved):

andreabedini:
Replaced by #3619

PR #3610, opened 2022-02-16
Export toLedgerEpochInfo

v0d1ch:
It is a bit hard to construct `EpochInfo` value if you are not sure where exactly to look for it ([related issue in cardano-ledger]( https://github.com/input-output-hk/cardano-ledger/issues/2655#issuecomment-1039531922))
This PR purpose is to expose `toLedgerEpochInfo` so api users have more pleasant experience.


Jimbo4350:
Thank you for your contribution but we have already exported this via `Cardano.Api`

PR #3611, opened 2022-02-16
flake.lock:  bump membench

deepfire:
This updates the membench for the https://hydra.iohk.io/jobset/Cardano/cardano-node-flake-temp


deepfire:
I'll only wait for CI to pass, and then will merge manually, since this has no impact on anyhing but the referenced Hydra jobset.

PR #3623, opened 2022-02-18
Remove op cert check when using bulk credentials

Jimbo4350:


newhoggy:
Please document reason for change in the PR.

erikd (approved):

PR #3634, opened 2022-02-23
cardano-node-3633: init forwarding if needed.

denisshevchenko:
Currently, we always initialize forwarding using `initForwarding` (in `Cardano.Node.Tracing.API.initTraceDispatcher` function). But since we have `Forwarder` backend in `Cardano.Logging.Types.BackendConfig`, we should initialize forwarding **only** if `Forwarder` backend is presented in the node's configuration.


jutaro (approved):
Looks good to me

PR #3638, opened 2022-02-23
tx-generator: Use multiple named wallets

MarcFontaine:
This PR allows the tx-generator to use multiple named wallets (instead of one global wallet).
This makes it easier to control which UTxO is used when, inside tx-generator scripts.


deepfire (approved):
bors r+

PR #3640, opened 2022-02-23
cardano-tracer-3637: correct JSON logs.

denisshevchenko:
@jutaro reported:

Currently the data field in JSON formats are strings, which are escaped:

```
{"host":"deusXmachina", "data":"{\"kind\":\"OpenEvent.OpenedImmutableDB\"}","thread":"5"}
```

But data should not be a string, but a JSON object:

```
{"host":"deusXmachina", "data":{"kind":"OpenEvent.OpenedImmutableDB"}, "thread":"5"}
```

Additionally, now the order of JSON-pairs is the same as in `trace-dispatcher`.


jutaro (approved):
The code looks good for me.

PR #3641, opened 2022-02-23
RTS workaround converting SIGTERM into SIGINT

### LINKED ISSUE #1697: [BUG] - Node should cleanly exit on SIGTERM
**Internal/Exernal**

**Summary**
The node only cleanly exits on SIGINT (ctrl-c) and not SIGTERM. This makes the default systemd signal handler stop the node in a way it takes longer to restart. We work around this by changing the signal it gets in the systemd service, but this should probably be fixed in the node.

**Steps to reproduce**

**Expected behavior**

- OS: [e.g. Ubuntu]
- Version [e.g. 20.04]
- Node version

- If applicable, add screenshots, config files and/or logs to help explain the problem.

**Additional context**

DavidEichmann:
This is important for graceful shutdown, cleaning up resources in
`bracket` statements.

Fixes #1697

### DIFF
--- a/cardano-node/src/Cardano/Node/Run.hs
+++ b/cardano-node/src/Cardano/Node/Run.hs
@@ -38,6 +38,7 @@ import           System.Directory (canonicalizePath, createDirectoryIfMissing, m
 import           System.Environment (lookupEnv)
 
 #ifdef UNIX
+import           GHC.Weak (deRefWeak)
 import           System.Posix.Files
 import qualified System.Posix.Signals as Signals
 import           System.Posix.Types (FileMode)
@@ -105,6 +106,8 @@ runNode
   :: PartialNodeConfiguration
   -> IO ()
 runNode cmdPc = do
+    installSigTermHandler
+
     -- TODO: Remove sodiumInit: https://github.com/input-output-hk/cardano-base/issues/175
     Crypto.sodiumInit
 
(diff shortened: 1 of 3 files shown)


DavidEichmann:
I should actually throw to the main thread rather than raise SIGINT. That's what the RTS does. I should also upstream this to GHC (see TopHandler).

DavidEichmann:
I'm working on a GHC patch that'll make this the default behavior of the RTS https://gitlab.haskell.org/DavidEichmann/ghc/-/tree/davide/sigterm

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -105,6 +106,23 @@ runNode
...
+#ifdef UNIX
+    -- Similar implementation to the RTS's handling of SIGINT (see GHC's
+    -- TopHandler.hs)
+    runThreadIdWk <- mkWeakThreadId =<< myThreadId
+    void $ Signals.installHandler
+      Signals.sigTERM
+      (Signals.CatchOnce $ do
+        runThreadIdMay <- deRefWeak runThreadIdWk
+        case runThreadIdMay of
+          Nothing -> return ()
```
Jimbo4350, line 120: Why not a `putStrLn "thread id no longer exists"` or something to that effect?

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -105,6 +106,23 @@ runNode
...
+          Just runThreadId -> killThread runThreadId
```
Jimbo4350, line 121: Should we also put some output here signalling that the thread was successfully killed?

Jimbo4350 (commented):
Looks good, a couple comments

cardano-node/src/Cardano/Node/Run.hs
DavidEichmann: This code was lifted from GHC. We use a weak pointer so that the main thread can be garbage collected. If `doRefWeak` returns `Nothing`, then the main thread must have already exited. I can only imagine that might be possible if the main thread had shut down by some other means and a SIGTERM happened just after that. In that case I'd expect the process to exit gracefully any way, so simply doing nothing here (`return ()`) sounds correct. Is that convincing or do you still think we should output some message?

DavidEichmann:
I've responded to comments by @newhoggy in https://github.com/input-output-hk/cardano-node/pull/3670

cardano-node/src/Cardano/Node/Run.hs
Jimbo4350: ^^ Fair points. I think it's fine as is.

Jimbo4350 (approved):

PR #3648, opened 2022-02-24
Update simple-scripts.md

raymonstah:
fix transaction validity intervals


Jimbo4350:
Subsumed by  #4567. @raymonstah what email should I use for you in my commit message?

PR #3654, opened 2022-02-25
cardano-tracer-3652: fix JSON data.

denisshevchenko:
Now `cardano-tracer` produces correct JSON: `data` contains a `Value`, not a string.


deepfire (approved):
Thank you @denisshevchenko

PR #3661, opened 2022-02-28
Update ledger, plutus and consensus (for benchmarking)

lehins:


erikd (changes requested):
This update does not include ledger commit 525844be05adae151e82069dcd0000f3301ca0d0 which is needed by `db-sync`.

erikd:
Probably should fix this as well:
```
src/Cardano/Analysis/ChainFilter.hs:56:32: error:
    • Overlapping instances for NFData EpochNo
        arising from the 'deriving' clause of a data type declaration
      Matching instances:
        instance NFData EpochNo -- Defined in ‘Cardano.Slotting.Slot’
        instance NFData EpochNo
          -- Defined at src/Cardano/Analysis/ChainFilter.hs:44:1
    • When deriving the instance for (NFData SlotCond)
   |
56 |   deriving (FromJSON, Generic, NFData, Show, ToJSON)
```

erikd (changes requested):
This needs to be rebased against `master` which now contains two fixes needed by `db-sync`.

lehins:
@erikd Sorry for the confusion, I should have made a comment earlier. This draft PR is mostly for analyzing performance. There will be a final update of ledger in node once we get performance in line.

PR #3668, opened 2022-03-01
Fix tracer config mismatches

### LINKED ISSUE #3663: [BUG] - TraceMempool (and perhaps others?) config setting is ignored
**Internal/External**
*External* 

**Area**
*Other*  cardano-node tracing config

**Summary**
It seems TraceMempool configuration is ignored in cardano-node 1.34.0

**Steps to reproduce**
1. Upgrade to 1.34.0
2. Set TraceMepool to false in config

**Expected behavior**
I do not see mempool trace logging

Ubuntu 20.04 x64, cardano-node 1.34.0 built from source

**Additional context**
I see a lot of changes regarding tracing. Should we configure this differently? Anyway, this seems to be a regression for 1.34.0 vs 1.33.0 at least.

jbgi:
Tracer mismatches, among which was mempool which has known performance impacts.

closes #3663


weebl2000:
This will hopefully fix #3663

Jimbo4350 (approved):

PR #3669, opened 2022-03-01
Fix mismatched tracers

Jimbo4350:
Some of the tracers were mismatched in the configuration machinery. This is now no longer possible as the code won't compile if this happens.


coot (approved):
LGTM

JaredCorduan (approved):

PR #3670, opened 2022-03-02
Add --shutdown-on-slot-synced test and ensure ExitSuccess

DavidEichmann:
Fixes https://github.com/input-output-hk/cardano-node/issues/3646

This does 2 things:

1. Adds a test that starts a node with `--shutdown-on-slot-synced x` and checks that (A) the exit code is `0` and that (B) the chain tip slot is close to `x`.
2. Rethrows the `ExceptionInLinkedThread _ ExitSuccess` exception as an an unwrapped `ExitSuccess` exception. This means the process's exit code will be `0` rather than `1` (this is asserted by the new test).

Without the changes of this PR, property 1.A fails (the exit code is `1`) but 1.B succeeds.

### DIFF
--- a/cardano-node/src/Cardano/Node/Handlers/TopLevel.hs
+++ b/cardano-node/src/Cardano/Node/Handlers/TopLevel.hs
@@ -48,6 +48,7 @@ import           Prelude
 
 import           Control.Exception
 
+import           Control.Monad.Class.MonadAsync (ExceptionInLinkedThread (..))
 import           System.Environment
 import           System.Exit
 import           System.IO
@@ -74,9 +75,16 @@ toplevelExceptionHandler prog = do
       ]
   where
     -- Let async exceptions rise to the top for the default GHC top-handler.
-    -- This includes things like CTRL-C.
+    -- This includes things like CTRL-C. If a linked thread throws ExitSuccess,
+    -- then we rethrow ExitSuccess. This happens for example when using the
+    -- `--shutdown-on-slot-synced` option.
     rethrowAsyncExceptions :: SomeAsyncException -> IO a
-    rethrowAsyncExceptions = throwIO
+    rethrowAsyncExceptions full@(SomeAsyncException e) = do
+      case fromException (toException e) of
+        Just (ExceptionInLinkedThread _ eInner)
+          | Just ExitSuccess <- fromException eInner
+          -> throwIO ExitSuccess
+        _ -> throwIO full
 
     -- We don't want to print ExitCode, and it should be handled by the default
     -- top handler because that sets the actual OS process exit code.
--- a/cardano-testnet/test/Spec/ShutdownOnSlotSynced.hs
+++ b/cardano-testnet/test/Spec/ShutdownOnSlotSynced.hs
@@ -1,0 +1,75 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Spec.ShutdownOnSlotSynced
+  ( hprop_shutdownOnSlotSynced
+  ) where
+
+import           Control.Monad
+import           Data.Bool ((&&))
+import           Data.Either (Either (Right), isRight)
+import           Data.Function
+import           Data.Int
+import           Data.List (find, isInfixOf, lines, reverse, words, (++))
+import           Data.Maybe
+import           Data.Ord
+import           GHC.IO.Exception (ExitCode (ExitSuccess))
+import           GHC.Num
+import           GHC.Stack (callStack)
+import           Hedgehog (Property, assert, (===))
+import           Prelude (fromIntegral, round)
+import           Text.Read (readMaybe)
+import           Text.Show (Show (..))
+
+import qualified Hedgehog.Extras.Test.Base as H
+import qualified Hedgehog.Extras.Test.File as H
+import qualified Hedgehog.Extras.Test.Process as H
+import qualified Test.Base as H
+import           Testnet.Cardano (TestnetNode (..), TestnetNodeOptions (TestnetNodeOptions),
+                   TestnetOptions (..), TestnetRuntime (..), defaultTestnetNodeOptions,
+                   defaultTestnetOptions, testnet)
+import qualified Testnet.Cardano as TC
+import qualified Testnet.Conf as H
+
+hprop_shutdownOnSlotSynced :: Property
+hprop_shutdownOnSlotSynced = H.integration . H.runFinallies . H.workspace "chairman" $ \tempAbsBasePath' -> do
+  -- Start a local test net
+  conf <- H.noteShowM $ H.mkConf tempAbsBasePath' Nothing
+  let maxSlot = 1500
+      slotLen = 0.01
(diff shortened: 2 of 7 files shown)


cardano-node/src/Cardano/Node/Handlers/TopLevel.hs
```diff
@@ -74,9 +75,16 @@ toplevelExceptionHandler prog = do
...
+    -- then we rethrow ExitSuccess. This happens for example when using the
+    -- `--shutdown-on-slot-synced` option.
     rethrowAsyncExceptions :: SomeAsyncException -> IO a
-    rethrowAsyncExceptions = throwIO
+    rethrowAsyncExceptions full@(SomeAsyncException e) = do
+      case fromException (toException e) of
+        Just (ExceptionInLinkedThread _ eInner)
+          | Just ExitSuccess <- fromException eInner
+          -> throwIO ExitSuccess
+        _ -> throwIO full
```
newhoggy, line 87: How does propagating `ExitSuccess` this way change the observed behaviour of the program?

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -105,6 +106,23 @@ runNode
   :: PartialNodeConfiguration
   -> IO ()
 runNode cmdPc = do
+    -- Workaround to ensure that the main thread throws an async exception on
+    -- receiving a SIGTERM signal.
+#ifdef UNIX
```
newhoggy, line 111: Does this only work on POSIX?

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -105,6 +106,23 @@ runNode
...
+    void $ Signals.installHandler
+      Signals.sigTERM
+      (Signals.CatchOnce $ do
+        runThreadIdMay <- deRefWeak runThreadIdWk
+        case runThreadIdMay of
+          Nothing -> return ()
+          Just runThreadId -> killThread runThreadId
+      )
+      Nothing
+#endif
```
newhoggy, line 124: This function is fairly large already.  Can this be moved to a separate function with a meaningful name and called from here?

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -105,6 +106,23 @@ runNode
   :: PartialNodeConfiguration
   -> IO ()
 runNode cmdPc = do
+    -- Workaround to ensure that the main thread throws an async exception on
+    -- receiving a SIGTERM signal.
+#ifdef UNIX
+    -- Similar implementation to the RTS's handling of SIGINT (see GHC's
+    -- TopHandler.hs)
```
newhoggy, line 113: I think it's worth providing a permalink to `TopHandler.hs` so it's easy to find that resource.

newhoggy:
Can you document how the test fails without the changes so it's easier to understand what this PR does to the behaviour of `cardano-node`?

cardano-node/src/Cardano/Node/Handlers/TopLevel.hs
DavidEichmann: I've added info in the PR description. See point 2.

cardano-node/src/Cardano/Node/Run.hs
DavidEichmann: Yes, so I've limited scope of the PR to posix, aiming to leave the windows behavior unchanged. I'm honestly not sure what the counterpart is on windows or if one even exits.

cardano-node/src/Cardano/Node/Run.hs
DavidEichmann: Will do in https://github.com/input-output-hk/cardano-node/pull/3641

cardano-node/src/Cardano/Node/Run.hs
DavidEichmann: Will do in https://github.com/input-output-hk/cardano-node/pull/3641

DavidEichmann:
I've removed one commit that is already in another PR: https://github.com/input-output-hk/cardano-node/pull/3641

cardano-node/src/Cardano/Node/Run.hs
deepfire: The primary user of this is system-level benchmarking -- and we don't really run that on Windows, since that platform is wallet-only, @newhoggy.

So we're perfectly find with `#ifdef UNIX`

deepfire (approved):
Thanks a lot @DavidEichmann!

newhoggy (approved):

DavidEichmann:
I've increased the timeout of the test while checking for termination every second in hopes that the test will succeed on CI.

DavidEichmann:
I've added a new commit "[Allow tests to specify custom node CLI options for local testnet](https://github.com/input-output-hk/cardano-node/pull/3670/commits/bd463a4e4b43f931c7ed36a4a2967f9081fdbd67)" and used this to greatly simplify the test added in this PR.

cardano-testnet/test/Spec/ShutdownOnSlotSynced.hs
```diff
@@ -0,0 +1,67 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE RecordWildCards #-}
```
Jimbo4350, line 4: Can we use `NamedFieldPuns` instead? I find `RecordWildCards` too opaque.

cardano-testnet/test/Spec/ShutdownOnSlotSynced.hs
```diff
@@ -0,0 +1,67 @@
...
+  mExitCodeRunning <- H.waitSecondsForProcess (80 :: Int) nodeProcessHandle
+
+  -- Check results
+  when (isRight mExitCodeRunning) $ do
+    H.cat nodeStdout
+    H.cat nodeStderr
+  mExitCodeRunning === Right ExitSuccess
+  log <- H.readFile nodeStdout
+  slotTip <- H.noteShow
+              $ read @Integer
```
Jimbo4350, line 59: Let's also catch this potential exception here and fail properly. `*** Exception: Prelude.read: no parse` may potentially be a nightmare to debug.

Jimbo4350 (changes requested):
Nice :+1:. A couple of minor comments.

cardano-testnet/test/Spec/ShutdownOnSlotSynced.hs
DavidEichmann: I'll just get rid of it and use projection functions

cardano-testnet/test/Spec/ShutdownOnSlotSynced.hs
DavidEichmann: I'm not sure what the "proper" way of handling this is, but I'll use `fail` and explicit error messages.

DavidEichmann:
I had to increase the running time of the test as it seems that some times the test would finish without any log messages re. chain tip.

Jimbo4350 (approved):

PR #3671, opened 2022-03-02
trace-dispatcher: double backslashes for forwarding.

denisshevchenko:
Double backslashes for proper substrings inside JSON-strings.


jutaro (approved):
Approved

Issue #3627, opened 2022-02-21
Always display the metrics dict in the output of kes-period-info command


saratomaz:
**Internal**

**Area**
*cli command*

**Describe the feature you'd like**
Always get the metrics dict as a output of the command, now is only retrieving the metrics on an "all valid" scenario.

**Additional context / screenshots**
_"all valid" scenario example with the metrics_
```
✓ The counters match what is in the node's protocol state
✓ The counters in the operational certificate and operational certificate issue counter file are the same
✓ Operational certificate's kes period is within the correct KES period interval
{
    "qKesNodeStateOperationalCertificateNumber": 6,
    "qKesCurrentKesPeriod": 404,
    "qKesOnDiskOperationalCertificateNumber": 6,
    "qKesRemainingSlotsInKesPeriod": 3760228,
    "qKesMaxKESEvolutions": 62,
    "qKesKesKeyExpiry": "2022-03-20T21:44:51Z",
    "qKesEndKesInterval": 434,
    "qKesStartKesInterval": 372,
    "qKesSlotsPerKesPeriod": 129600
}
```

PR #3683, opened 2022-03-04
cardano-cli query kes-period-info: Always display metrics (#3627)

cblp:


Jimbo4350 (approved):
LGTM. Can you ask QA to run the command to be sure they are satisfied before merging? And also link the relevant issue in the description?

newhoggy (approved):

saratomaz:
The metrics are returned, but the values are not correct, already exist a PR that fix that: https://github.com/input-output-hk/cardano-node/pull/3697.
Ideally both of the PRs should be accepted at the same time, like that we can truly test everything related to these metrics.

Jimbo4350:
> The metrics are returned, but the values are not correct, already exist a PR that fix that: #3697. Ideally both of the PRs should be accepted at the same time, like that we can truly test everything related to these metrics.

We will still merge this as I won't get to implementing tests for #3697 until Wednesday

PR #3686, opened 2022-03-07
Rename output-format to key-output-format

newhoggy:


cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -1622,14 +1622,14 @@ pOperationalCertificateFile =
     <> Opt.completer (Opt.bashCompleter "file")
     )
 
-pOutputFormat :: Parser OutputFormat
-pOutputFormat =
-  Opt.option readOutputFormat
-    (  Opt.long "output-format"
```
newhoggy, line 1628: It's a bit unfortunate that `--output-format` is already taken and I can't use it for the choosing between the JSON and text for query commands without introducing inconsistency.

PR #3692, opened 2022-03-08
Make LedgerStateEvents a type alias

andreabedini:
IMHO There's no real value in LedgerStateEvents.

- In the exported API it only appears in the type of applyBlock (in positive position)
- It is not exported, preventing users to make use of applyBlock

This replaces the data type with a type alias (to keep the type
signature brief) and changes the signature of applyBlock to not mention
the type.


nc6 (approved):

nc6:
bors merge

PR #3700, opened 2022-03-11
Update dependencies and pins.

nc6:
- The `cardano-base`, `cardano-ledger`, `plutus` and `ouroboros-network`
  dependencies have been updated to recent versions.
- Since the downstream libraries have been updated to more recent
  index-state and hackage revisions, we synchronise this here.
- As a consequence of this update, we switch to a newer version of
  `aeson`, necessitating updates in `ekg-json` and `hedgehog-extras`.
- We switch to ghc8107, to be in sync with other packages.

### DIFF
--- a/cardano-api/src/Cardano/Api/TxMetadata.hs
+++ b/cardano-api/src/Cardano/Api/TxMetadata.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE ViewPatterns #-}
 
 -- | Metadata embedded in transactions
 --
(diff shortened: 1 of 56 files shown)


cardano-api/src/Cardano/Api/TxMetadata.hs
```diff
@@ -76,6 +76,8 @@ import           Cardano.Api.Eras
 import           Cardano.Api.Error
 import           Cardano.Api.HasTypeProxy
 import           Cardano.Api.SerialiseCBOR
+import qualified Data.Aeson.KeyMap as KeyMap
```
Jimbo4350, line 79: These imports should be with `Data.Aeson` above

cardano-api/src/Cardano/Api/Value.hs
```diff
@@ -84,6 +85,7 @@ import           Cardano.Api.SerialiseCBOR
 import           Cardano.Api.SerialiseRaw
 import           Cardano.Api.SerialiseUsing
 import           Cardano.Api.Utils (note)
+import qualified Data.Aeson.KeyMap as KeyMap
```
Jimbo4350, line 88: This import should be with `Data.Aeson` above

cardano-cli/src/Cardano/CLI/Run/Friendly.hs
```diff
@@ -30,6 +30,7 @@ import           Cardano.Ledger.Crypto (Crypto)
 import qualified Cardano.Ledger.Shelley.API as Shelley
 
 import           Cardano.CLI.Helpers (textShow)
+import qualified Data.Aeson.Key as Aeson
```
Jimbo4350, line 33: This import should be  with `Data.Aeson` above

cardano-cli/src/Cardano/CLI/Shelley/Output.hs
```diff
@@ -26,6 +26,7 @@ import           Data.Word
 import           Cardano.CLI.Shelley.Orphans ()
 import           Cardano.Ledger.Shelley.Scripts ()
 import           Cardano.Slotting.Time (SystemStart (..))
+import qualified Data.Aeson.Key as Aeson
```
Jimbo4350, line 29: This import should be with `Data.Aeson` above

cardano-node/cardano-node.cabal
```diff
@@ -186,7 +186,6 @@ library
                       , transformers
                       , transformers-except
                       , typed-protocols
-                      , unordered-containers
```
Jimbo4350, line 189: :tada:

Jimbo4350 (changes requested):

Jimbo4350 (approved):

deepfire (approved):

JaredCorduan (approved):
@nc6 thank you for this huge effort!

coot (approved):
LGTM

nc6:
bors merge

PR #3704, opened 2022-03-14
Newhoggy/enable shutdown test on windows 2

newhoggy:


cardano-node/src/Cardano/Node/Handlers/TopLevel.hs
```diff
@@ -64,6 +65,7 @@ import           System.IO
 --
 toplevelExceptionHandler :: IO a -> IO a
 toplevelExceptionHandler prog = do
+    outPutStrLn "Starting"
```
newhoggy, line 68: Only this gets logged when the test fails.  This tells me no exceptions were intercepted.

Issue #3698, opened 2022-03-10
Expose local tx monitoring protocol in cardano-api


Jimbo4350:
The LocalTxMonitor protocol has been added to consensus: https://github.com/input-output-hk/ouroboros-network/pull/3404

We need to integrate this protocol into the cardano-api and then the cardano-cli. Another ticket will be written up for the cardano-cli integration with testing details.

PR #3706, opened 2022-03-14
Expose the tx mempool monitoring mini protocol in cardano-api

Jimbo4350:
Resolves: #3698

### DIFF
--- a/cardano-api/src/Cardano/Api/IPC.hs
+++ b/cardano-api/src/Cardano/Api/IPC.hs
@@ -79,7 +79,8 @@ import           Data.Void (Void)
 import qualified Data.ByteString.Lazy as LBS
 import qualified Data.Map.Strict as Map
 
-import           Control.Concurrent.STM
+import           Control.Concurrent.STM (TMVar, atomically, newEmptyTMVarIO, putTMVar, takeTMVar,
+                   tryPutTMVar)
 import           Control.Monad (void)
 import           Control.Tracer (nullTracer)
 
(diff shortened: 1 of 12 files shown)


cardano-api/src/Cardano/Api/IPC.hs
```diff
@@ -550,11 +624,98 @@ submitTxToNodeLocal connctInfo tx = do
...
+
+data LocalTxMonitoringQuery mode
+  -- | Query if a particular tx exists in the mempool. Note that, the absence
+  -- of a transaction does not imply anything about how the transaction was
+  -- processed: it may have been dropped, or inserted in a block.
+  = LocalTxMonitoringQueryTx (TxIdInMode mode)
+  -- | The mempool is modeled as an ordered list of transactions and thus, can
+  -- be traversed linearly. 'LocalTxMonitoringSendNextTx' requests the next transaction from the
+  -- current list. This must be a transaction that was not previously sent to
+  -- the client for this particular snapshot.
```
newhoggy, line 650: Is there state associated connection that tracks where the current transaction in the list is?

cardano-api/src/Cardano/Api/InMode.hs
```diff
@@ -6,12 +6,17 @@
 -- | Transactions in the context of a consensus mode, and other types used in
 -- the transaction submission protocol.
 --
-module Cardano.Api.TxInMode (
+module Cardano.Api.InMode (
```
newhoggy, line 9: What's the motivation for the module rename?

Maybe the comment for the module needs adjustment too?

cardano-api/src/Cardano/Api/IPC.hs
Jimbo4350: There are only 3 available things you can get:
```
data ServerStBusy (kind :: StBusyKind) txid tx slot m a where
  SendMsgReplyNextTx
    :: Maybe tx
    -> ServerStAcquired txid tx slot m a
    -> ServerStBusy NextTx txid tx slot m a

  SendMsgReplyHasTx
    :: Bool
    -> ServerStAcquired txid tx slot m a
    -> ServerStBusy HasTx txid tx slot m a

  SendMsgReplyGetSizes
    :: MempoolSizeAndCapacity
    -> ServerStAcquired txid tx slot m a
    -> ServerStBusy GetSizes txid tx slot m 
    ```

newhoggy (approved):

PR #3710, opened 2022-03-16
tx-generator: refactor TPS limited tx-feeder

MarcFontaine:
For a while the tx-generator has been generating transaction just in time,
and the `TxSendQueue` had been used with a list of dummy transactions.
This PR removes the `TxSendQueue` altogether but without changing the observable behaviour.


MarcFontaine:
These change have been included in https://github.com/input-output-hk/cardano-node/pull/3789

PR #3714, opened 2022-03-21
cardano-tracer-3712: try Windows tests.

denisshevchenko:
Currently, cardano-tracer still sometimes has problems with Windows tests. It should be investigated deeply and fixed. Particularly, Cardano.Tracer.Test.Logs.Tests is mostly disabled for Windows.

Now, tests are restored.


deepfire (approved):
bors r+

@denisshevchenko, thank you!

PR #3718, opened 2022-03-23
kes-period-info property test

Jimbo4350:
We test the validity of the operational certificate based on: 
1. If the output of `cardano-cli query kes-period-info` indicates the operational certificate is valid
2. If the node logs indicate that the node has minted a block

I.e we can infer that if 1 is true and the node is minting blocks then `kes-period-info`'s positive case (i.e indicates the certificate is valid) correctly checks an operational certificate.


cardano-cli/src/Cardano/CLI/Shelley/Output.hs
```diff
@@ -24,52 +24,120 @@ import           Data.Time.Clock (UTCTime)
...
+    nodeStateOperationalCertificateNumber <- o .: "qKesNodeStateOperationalCertificateNumber"
+    maxKESEvolutions <- o .: "qKesMaxKESEvolutions"
+    slotsPerKesPeriod <- o .: "qKesSlotsPerKesPeriod"
+    kesKeyExpiry <- o .: "qKesKesKeyExpiry"
+    let opCertIntervalInfo = createOpCertIntervalInfo
+                               currentKesPeriod
+                               startKesInterval
+                               endKesInterval
+                               remainingSlotsInKesPeriod
+    return $  QueryKesPeriodInfoOutput
```
newhoggy, line 107: Double space after `$`.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -405,155 +406,149 @@ runQueryKesPeriodInfo (AnyConsensusModeParams cModeParams) network nodeOpCertFil
...
+      liftIO $ renderOpCertNodeAndOnDiskCounterInformation nodeOpCertFile counterInformation
+
+      let qKesInfoOutput = createQueryKesPeriodInfoOutput opCertIntervalInformation counterInformation eInfo gParams
+          kesPeriodInfoJSON = encodePretty qKesInfoOutput
+      case mOutFile of
+        Just (OutputFile oFp) -> do
+          liftIO $ LBS.putStrLn kesPeriodInfoJSON
+          handleIOExceptT (ShelleyQueryCmdWriteFileError . FileIOError oFp)
+            $ LBS.writeFile oFp kesPeriodInfoJSON
+        Nothing -> liftIO $ LBS.putStrLn kesPeriodInfoJSON
```
newhoggy, line 443: To prevent duplication, suggest:

```
      liftIO $ LBS.putStrLn kesPeriodInfoJSON
      forM_ mOutFile $ \Just (OutputFile oFp) -> do
          handleIOExceptT (ShelleyQueryCmdWriteFileError . FileIOError oFp)
```

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -405,155 +406,149 @@ runQueryKesPeriodInfo (AnyConsensusModeParams cModeParams) network nodeOpCertFil
...
+   opCertStartingKesPeriod :: OperationalCertificate -> OpCertStartingKesPeriod
+   opCertStartingKesPeriod = OpCertStartingKesPeriod . fromIntegral . getKesPeriod
+
+   opCertEndKesPeriod :: GenesisParameters -> OperationalCertificate -> OpCertEndingKesPeriod
+   opCertEndKesPeriod gParams oCert =
+     let OpCertStartingKesPeriod start = opCertStartingKesPeriod oCert
+         maxKesEvo = fromIntegral $ protocolParamMaxKESEvolutions gParams
+     in OpCertEndingKesPeriod $ start + maxKesEvo
+
+   -- See OCERT rule
```
newhoggy, line 464: Is there a link or pointer to where this rule is?

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -405,155 +406,149 @@ runQueryKesPeriodInfo (AnyConsensusModeParams cModeParams) network nodeOpCertFil
...
+       in O.createOpCertIntervalInfo c s e (Just slotsTillExp)
+
+   opCertNodeAndOnDiskCounters
+     :: OpCertOnDiskCounter
+     -> Maybe OpCertNodeStateCounter
+     -> OpCertNodeAndOnDiskCounterInformation
+   opCertNodeAndOnDiskCounters o@(OpCertOnDiskCounter odc) (Just n@(OpCertNodeStateCounter nsc))
+     | odc < nsc = OpCertOnDiskCounterBehindNodeState o n
+     | otherwise = OpCertOnDiskCounterMoreThanOrEqualToNodeState o n
+   opCertNodeAndOnDiskCounters o Nothing = OpCertNoBlocksMintedYet o
```
newhoggy, line 486: What does the node state counter and on disk counter have to do with whether or blocks are minted yet?

cardano-cli/src/Cardano/CLI/Types.hs
```diff
@@ -71,11 +82,86 @@ newtype CddlTx = CddlTx {unCddlTx :: InAnyCardanoEra Tx}
...
 
+data OpCertNodeAndOnDiskCounterInformation
+  = OpCertOnDiskCounterMoreThanOrEqualToNodeState
+      OpCertOnDiskCounter
+      OpCertNodeStateCounter
+  | OpCertOnDiskCounterBehindNodeState
+      OpCertOnDiskCounter
+      OpCertNodeStateCounter
+  | OpCertNoBlocksMintedYet
+      OpCertOnDiskCounter
```
newhoggy, line 106: Some documentation on these constructors would be handy.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -405,155 +406,149 @@ runQueryKesPeriodInfo (AnyConsensusModeParams cModeParams) network nodeOpCertFil
...
+     -> OpCertNodeAndOnDiskCounterInformation
+     -> EpochInfo (Either Text)
+     -> GenesisParameters
+     -> O.QueryKesPeriodInfoOutput
+   createQueryKesPeriodInfoOutput oCertIntervalInfo oCertCounterInfo eInfo gParams  =
+     let (_s, e, _c, mStillExp) = case oCertIntervalInfo of
+                                  OpCertWithinInterval st end curr sTillExp -> (st, end, curr, Just sTillExp)
+                                  OpCertStartingKesPeriodIsInTheFuture st end curr -> (st, end, curr, Nothing)
+                                  OpCertExpired st end curr -> (st, end, curr, Nothing)
+                                  OpCertSomeOtherError st end curr -> (st, end, curr, Nothing)
```
newhoggy, line 527: Why does the case return `st` and `curr` only to ignore them?

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -567,86 +562,28 @@ runQueryKesPeriodInfo (AnyConsensusModeParams cModeParams) network nodeOpCertFil
...
+
+renderOpCertIntervalInformation :: FilePath -> OpCertIntervalInformation -> String
+renderOpCertIntervalInformation _ (OpCertWithinInterval _start _end _current _stillExp) =
+  "✓ Operational certificate's kes period is within the correct KES period interval"
+renderOpCertIntervalInformation opCertFile
+  (OpCertStartingKesPeriodIsInTheFuture (OpCertStartingKesPeriod start) (OpCertEndingKesPeriod end) (CurrentKesPeriod current)) =
+   "✗ Node operational certificate at: " <> opCertFile <> " has an incorrectly specified starting KES period. " <> "\n" <>
+   "Current KES period: " <> show current <> "\n" <>
+   "Operational certificate's starting KES period: " <> show start <> "\n" <>
+   "Operational certificate's expiry KES period: " <> show end
```
newhoggy, line 576: Perhaps we should indent this lines so the output looks like this to aid readability?

```
✗ Node operational certificate at ...
   Current KES period: ...
   Operational certificate's starting KES period: ...
   Operational certificate's expiry KES period: ...
```

Similarly below.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -405,155 +406,149 @@ runQueryKesPeriodInfo (AnyConsensusModeParams cModeParams) network nodeOpCertFil
...
+
+   renderOpCertNodeAndOnDiskCounterInformation :: FilePath -> OpCertNodeAndOnDiskCounterInformation -> IO ()
+   renderOpCertNodeAndOnDiskCounterInformation opCertFile opCertCounterInfo =
+     case opCertCounterInfo of
+      OpCertOnDiskCounterMoreThanOrEqualToNodeState _ _ ->
+        putStrLn @String "✓ The operational certificate counter agrees with the node protocol state counter"
+      OpCertOnDiskCounterBehindNodeState onDiskC nodeStateC ->
+        putStrLn $ "✗ The protocol state counter is greater than the counter in the operational certificate at: " <> opCertFile <> "\n" <>
+                   "On disk operational certificate counter: " <> show (unOpCertOnDiskCounter onDiskC) <> "\n" <>
+                   "Protocol state counter: " <> show (unOpCertNodeStateCounter nodeStateC)
```
newhoggy, line 510: Perhaps we should indent this lines so the output looks like this to aid readability?

```
✗ The protocol state counter is greater than the counter in the operational certificate at: ...
    On disk operational certificate counter: ...
    Protocol state counter: ...
```

Similarly below.

cardano-testnet/test/Properties/Cli/KesPeriodInfo.hs
```diff
@@ -0,0 +1,50 @@
+module Properties.Cli.KesPeriodInfo where
```
newhoggy, line 1: Where is this module referenced?

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
Jimbo4350: Have a look at `currentIssueNo`: https://github.com/input-output-hk/cardano-ledger/blob/ac51494e151af0ad99b937a787458ce71db0aaea/libs/cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/OCert.hs#L67
We currently only check the counter map for the existence of the corresponding stake pool's operational certificate counter. If it's not in the counter map it means the stake pool hasn't minted a block yet.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
Jimbo4350: This was prior to a refactor I did. I will adjust the case to return a tuple.

newhoggy (approved):

PR #3724, opened 2022-03-28
Fix typo in Transaction.hs

miguelaeh:
This PR fixes a minor typo into `cardano-cli/src/Cardano/CLI/Shelley/Run/Transaction.hs`.


Jimbo4350:
Subsumed by #4567

PR #3728, opened 2022-03-30
fix - some statements are not consistent with the examples shown

kindofdev:
- I think the conditions for evaluating **RequireTimeBefore** and **RequireTimeAfter** in the **Time locking** section (**172**, **180**) are not correct. Notice that these statements are not consistent with the examples shown. 

- Descriptions at **333** and **368** are not correct.

- The scripts about **time locking** (**602**, **637**) are not correct based on the example descriptions.


Jimbo4350:
Subsumed by #4567 @kindofdev what email should I use for you in my commit message?

kindofdev:
> Subsumed by #4567 @kindofdev what email should I use for you in my commit message?

jose.velasco.home@gmail.com

PR #3744, opened 2022-04-03
stakeKey registration stake.skey witness not needed

gitmachtl:
Since the start of Shelley we sign a stakeKey registration also with the stake.skey. Turnes out, this is wrong and not needed at all. 😱 

According to https://hydra.iohk.io/build/6752481/download/1/delegation_design_spec.pdf
Chapter 3.3: "We do not require a witness to register a stake address (besides, of course, any witnesses needed for the transaction that is used to post the certificate)."

So only the certificate is needed and the witness for the payment of course.

### DIFF
--- a/doc/stake-pool-operations/register_key.md
+++ b/doc/stake-pool-operations/register_key.md
@@ -29,7 +29,7 @@ For the transaction draft, --tx.out, --invalid-hereafter and --fee can be set to
     --tx-body-file tx.draft \
     --tx-in-count 1 \
     --tx-out-count 1 \
-    --witness-count 2 \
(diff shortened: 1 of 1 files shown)


JaredCorduan (approved):
Great catch @gitmachtl. Note that the Shelley ledger spec also confirms this, see the `certWitsNeededTx` function in "Figure 18: Functions used in witness rule".

olgahryniuk (approved):

newhoggy:
To be merged in https://github.com/input-output-hk/cardano-node/pull/4975

newhoggy:
Thanks for the contribution!

PR #3773, opened 2022-04-06
Update cardano-api's TxOut with inline datum

Jimbo4350:
Updates the TxOut data type with the ability to specify an inline datum and propagate the changes throughout the api and cli.

### DIFF
--- a/cardano-api/src/Cardano/Api/TxBody.hs
+++ b/cardano-api/src/Cardano/Api/TxBody.hs
@@ -51,7 +51,7 @@ module Cardano.Api.TxBody (
     CtxTx, CtxUTxO,
     TxOut(..),
     TxOutValue(..),
-    TxOutDatum(TxOutDatumNone, TxOutDatumHash, TxOutDatum, TxOutInlineDatum),
+    TxOutDatum(TxOutDatumNone, TxOutDatumHash, TxOutDatumInTx, TxOutDatumInline),
     toCtxUTxOTxOut,
     lovelaceToTxOutValue,
     prettyRenderTxOut,
@@ -508,8 +508,8 @@ toCtxUTxOTxOut (TxOut addr val d) =
   let dat = case d of
               TxOutDatumNone -> TxOutDatumNone
               TxOutDatumHash s h -> TxOutDatumHash s h
-              TxOutDatum' s h _ -> TxOutDatumHash s h
-              TxOutInlineDatum s sd -> TxOutInlineDatum s sd
+              TxOutDatumInTx' s h _ -> TxOutDatumHash s h
+              TxOutDatumInline s sd -> TxOutDatumInline s sd
   in TxOut addr val dat
 
 instance IsCardanoEra era => ToJSON (TxOut ctx era) where
@@ -522,13 +522,13 @@ instance IsCardanoEra era => ToJSON (TxOut ctx era) where
            , "value"     .= val
            , "datumhash" .= h
            ]
-  toJSON (TxOut addr val (TxOutDatum' _ h d)) =
+  toJSON (TxOut addr val (TxOutDatumInTx' _ h d)) =
     object [ "address"   .= addr
            , "value"     .= val
            , "datumhash" .= h
            , "datum"     .= scriptDataToJson ScriptDataJsonDetailedSchema d
            ]
-  toJSON (TxOut addr val (TxOutInlineDatum _ d)) =
+  toJSON (TxOut addr val (TxOutDatumInline _ d)) =
     object [ "address"         .= addr
            , "value"           .= val
            , "inlineDatumhash" .= hashScriptData d
@@ -556,7 +556,7 @@ instance (IsShelleyBasedEra era, IsCardanoEra era)
                     fail $ "Error parsing TxOut JSON: " <> displayError err
                   Right sData -> do
                     dHash <- runParsecParser (parseHash $ AsHash AsScriptData) dHashTxt
-                    pure . TxOut addr val $ TxOutDatum' supported dHash sData
+                    pure . TxOut addr val $ TxOutDatumInTx' supported dHash sData
               (mDatVal, wrongDatumHashFormat) ->
                 fail $ "Error parsing TxOut's datum hash/data: " <>
                        "\nData value:" <> show mDatVal <>
@@ -674,7 +674,7 @@ toAlonzoTxOutDataHash :: TxOutDatum CtxUTxO era
                       -> StrictMaybe (Alonzo.DataHash StandardCrypto)
 toAlonzoTxOutDataHash  TxOutDatumNone                        = SNothing
 toAlonzoTxOutDataHash (TxOutDatumHash _ (ScriptDataHash dh)) = SJust dh
-toAlonzoTxOutDataHash (TxOutInlineDatum _ sd) =
+toAlonzoTxOutDataHash (TxOutDatumInline _ sd) =
   let ScriptDataHash alonzoDataHash = hashScriptData sd
   in SJust alonzoDataHash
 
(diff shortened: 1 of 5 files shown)


cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -670,30 +669,30 @@ fromShelleyTxOut era ledgerTxOut =
...
+toAlonzoTxOutDataHash
+  :: TxOutDatum CtxUTxO AlonzoEra
+  -> StrictMaybe (Alonzo.DataHash StandardCrypto)
 toAlonzoTxOutDataHash  TxOutDatumNone                        = SNothing
 toAlonzoTxOutDataHash (TxOutDatumHash _ (ScriptDataHash dh)) = SJust dh
-toAlonzoTxOutDataHash (TxOutInlineDatum _ sd) =
-  let ScriptDataHash alonzoDataHash = hashScriptData sd
-  in SJust alonzoDataHash
+toAlonzoTxOutDataHash (TxOutDatumInline inlineDatumSupp _sd) =
+  case inlineDatumSupp :: InlineDatumSupportedInEra AlonzoEra of {}
```
newhoggy, line 680: Can you explain this change?

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -670,30 +669,30 @@ fromShelleyTxOut era ledgerTxOut =
...
-  :: Ledger.Crypto (ShelleyBasedEra era) ~ StandardCrypto
-  => TxOutDatum CtxUTxO era -> Babbage.Datum (ShelleyBasedEra era)
-_toBabbageTxOutDatum  TxOutDatumNone                        = Babbage.NoDatum
-_toBabbageTxOutDatum (TxOutDatumHash _ (ScriptDataHash dh)) = Babbage.DatumHash dh
-_toBabbageTxOutDatum (TxOutInlineDatum _ sd) = scriptDataToBinaryData sd
--- TODO: Rename TxOutDatum' to TxOutAddDatumToDatumMap or something similar
--- as we need to distinguish between this feature (which is an Alonzo feature)
--- and the inline datum feature (which is a Babbage feature).
--- toBabbageTxOutDatum (TxOutDatum' _ (ScriptDataHash dh) _) = Babbage.DatumHash dh
+fromAlonzoTxOutDataHash s (SJust dh)   = TxOutDatumHash s (ScriptDataHash dh)
```
newhoggy, line 686: What's the reason for renaming `era` to `s`?

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -670,30 +669,30 @@ fromShelleyTxOut era ledgerTxOut =
...
+
+-- TODO: If ledger creates an open type family for datums
+-- we can consolidate this function with the Alonzo version
+toBabbageTxOutDatum
+  :: Ledger.Crypto (ShelleyLedgerEra era) ~ StandardCrypto
+  => TxOutDatum CtxUTxO era -> Babbage.Datum (ShelleyLedgerEra era)
+toBabbageTxOutDatum  TxOutDatumNone = Babbage.NoDatum
+toBabbageTxOutDatum (TxOutDatumHash _ (ScriptDataHash dh)) = Babbage.DatumHash dh
+toBabbageTxOutDatum (TxOutDatumInline _ sd) = scriptDataToBinaryData sd
```
newhoggy, line 695: The code looks the same, but the TODO is gone.  Is the TODO not needed anymore?

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -2882,14 +2884,27 @@ toShelleyTxOutAny _ (TxOut addr (TxOutValue MultiAssetInAlonzoEra value) txoutda
...
+toAlonzoTxOutDataHash' (TxOutDatumInline inlineDatumSupp _sd) =
+  case inlineDatumSupp :: InlineDatumSupportedInEra AlonzoEra of {}
+
+-- TODO: Consolidate with alonzo function and rename
+toBabbageTxOutDatum'
+  :: Ledger.Crypto (ShelleyLedgerEra era) ~ StandardCrypto
+  => TxOutDatum ctx era -> Babbage.Datum (ShelleyLedgerEra era)
+toBabbageTxOutDatum'  TxOutDatumNone = Babbage.NoDatum
+toBabbageTxOutDatum' (TxOutDatumHash _ (ScriptDataHash dh)) = Babbage.DatumHash dh
+toBabbageTxOutDatum' (TxOutDatumInTx'    _ (ScriptDataHash dh) _) = Babbage.DatumHash dh
```
newhoggy, line 2906: Too much indentation before first `_`?

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -2882,14 +2884,27 @@ toShelleyTxOutAny _ (TxOut addr (TxOutValue MultiAssetInAlonzoEra value) txoutda
...
-toAlonzoTxOutDataHash' :: TxOutDatum ctx era
-                      -> StrictMaybe (Alonzo.DataHash StandardCrypto)
+
+toAlonzoTxOutDataHash' :: TxOutDatum ctx AlonzoEra
+                       -> StrictMaybe (Alonzo.DataHash StandardCrypto)
 toAlonzoTxOutDataHash'  TxOutDatumNone                          = SNothing
 toAlonzoTxOutDataHash' (TxOutDatumHash _ (ScriptDataHash dh))   = SJust dh
-toAlonzoTxOutDataHash' (TxOutDatum'    _ (ScriptDataHash dh) _) = SJust dh
-toAlonzoTxOutDataHash' _ = error "TODO: Babbage"
+toAlonzoTxOutDataHash' (TxOutDatumInTx'    _ (ScriptDataHash dh) _) = SJust dh
```
newhoggy, line 2896: Too much indentation before first _?

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -2020,7 +2022,7 @@ fromAlonzoTxOut multiAssetInEra scriptDataInEra txdatums
     fromAlonzoTxOutDatum _          SNothing = TxOutDatumNone
     fromAlonzoTxOutDatum supported (SJust dh)
       | Just d <- Map.lookup dh txdatums
-                  = TxOutDatum'    supported (ScriptDataHash dh)
+                  = TxOutDatumInTx'    supported (ScriptDataHash dh)
```
newhoggy, line 2025: Too much indentation before first `supported`?

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -1213,29 +1212,32 @@ data TxOutDatum ctx era where
...
 
-pattern TxOutDatum :: ScriptDataSupportedInEra era
-                   -> ScriptData
-                   -> TxOutDatum CtxTx era
-pattern TxOutDatum s d  <- TxOutDatum' s _ d
+pattern TxOutDatumInTx
+  :: ScriptDataSupportedInEra era
+  -> ScriptData
+  -> TxOutDatum CtxTx era
+pattern TxOutDatumInTx s d  <- TxOutDatumInTx' s _ d
```
newhoggy, line 1235: Since the line is changing we can take the opportunity to remove the extra space before `<-`.

cardano-api/src/Cardano/Api/TxBody.hs
Jimbo4350: Sure so basically in the AlonzoEra we can only have datum hashes or nothing. Whereas in Babbage era we can have inline datums, datum hashes or nothing. Because there is no open type family for datums, I need to define two separate functions that return the correct type based on the era. I've asked the ledger team to define an open type family for datums and they will probably do this after the hardfork. When they do this, we will be able to have a single function that instantiates the correct type based on the `ShelleyBasedEra era` we feed it.

Jimbo4350: `case inlineDatumSupp :: InlineDatumSupportedInEra AlonzoEra of {}` tells the compiler it's not possible to pattern match `TxOutDatumInline` because only `InlineDatumSupportedInEra BabbageEra` exists.

cardano-api/src/Cardano/Api/TxBody.hs
Jimbo4350: This is a mistake :)

cardano-api/src/Cardano/Api/TxBody.hs
Jimbo4350: It's not needed anymore as `TxOutDatum'` was renamed to `TxOutDatumInTx'`

cardano-api/src/Cardano/Api/TxBody.hs
newhoggy: Oh I see.  Thanks!

newhoggy (approved):

newhoggy:
bors merge

PR #3779, opened 2022-04-08
Update cardano-api txout with reference scripts

Jimbo4350:
Update cardano-api's txout data type with reference scripts


MarcFontaine (approved):
LGTM regarding the tx-generator

newhoggy (approved):

PR #3790, opened 2022-04-12
  DO NOT MERGE: Bump ouroboros-network version to the latest version of…

dnadales:
… feature/utxo-hd

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -192,7 +192,7 @@ source-repository-package
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-ledger
-  tag: 1db68a3ec0a2dcb5751004beb22b906162474f23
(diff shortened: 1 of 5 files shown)


nfrisby:
Looks good to me, but I'll let Javier Approve it.

Also: I removed all reviewers other than Javier, since this PR targets the branch that only you and he care about.

cabal.project
```diff
@@ -28,8 +28,9 @@ package cardano-cli
 package cardano-git-rev
   ghc-options: -Werror
 
-package cardano-node
-  ghc-options: -Werror
+-- FIXME: this needs to be enabled again!
```
Jasagredo, line 31: I would mention that this is because there are some incomplete patterns related to `Orphans/Network`

Jasagredo (approved):
This looks awesome. In fact we might not need to do much work in the state queries in case they work.

Thanks for doing this PR :pray:

PR #3796, opened 2022-04-13
Docs: update Haskell installation method and mention libsecp256k1

Jasagredo:
The installation method through `Cabal` didn't mention the need for `libsecp256k1`. And also nowadays the recommended way to install the Haskell environment is through GHCup.

### DIFF
--- a/doc/getting-started/install.md
+++ b/doc/getting-started/install.md
@@ -46,56 +46,11 @@ sudo apt-get install automake build-essential pkg-config libffi-dev libgmp-dev l
 
 If you are using a different flavor of Linux, you will need to use the correct package manager for your platform instead of `yum` or `apt-get`, and the names of the packages you need to install might differ.  On MacOSX, use the Homebrew (`brew`) installer.
 
-#### Downloading, unpacking, installing, and updating Cabal:
+#### Installing the Haskell environment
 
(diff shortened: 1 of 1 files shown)


coot (approved):
LGTM

Jasagredo:
> Can we please inform the user of the next steps to get the node up and running?
> What is the next page with instructions to fetch config files and actually run the node?

I'm unsure what the next page would be, even if it is documented somewhere. Perhaps we can merge this and point to that page in a follow-up PR once we know which one it is (or if it would need to be created)?

doc/getting-started/install.md
```diff
@@ -132,13 +92,33 @@ make
...
+```bash
+mkdir -p ~/src
+cd ~/src
+```
+
+Download and install `libsecp256k1`:
+
+```bash
+git clone https://github.com/bitcoin-core/secp256k1
+cd secp256k1
```
rdlrt, line 115: Are we asking users to follow main branch instead of specific commit used in github workflows of this repo?

Jasagredo: Indeed this is a good point, and as I don't have the knowledge to say that master would be OK, I changed it to use the same commit as the github workflow. Thanks.

coot:
@Jasagredo can this PR be merged?

PR #3797, opened 2022-04-13
Dependency bump 2022-04-19

Jimbo4350:


MarcFontaine (approved):
LGTM

nc6 (approved):

cardano-api/gen/Gen/Cardano/Api.hs
```diff
@@ -83,19 +87,23 @@ genExUnits = do
...
-genCostModel :: Range Int -> Gen Text -> Gen Integer -> Gen Alonzo.CostModel
-genCostModel r gt gi = do
-  map' <- Gen.map r ((,) <$> gt <*> gi)
-  return $ Alonzo.CostModel map'
+genCostModels :: Gen Alonzo.CostModels
+genCostModels = do
+  CostModel cModel <- genCostModel
+  lang <- genLanguage
+  case Alonzo.mkCostModel lang cModel of
+    Left err -> panic . Text.pack $ "genCostModels: " <> err
```
newhoggy, line 95: Are we sure this case never happens?

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -2244,12 +2244,13 @@ fromLedgerTxTotalCollateral era txbody =
   case totalAndReturnCollateralSupportedInEra $ shelleyBasedToCardanoEra era of
     Nothing -> TxTotalCollateralNone
     Just supp ->
-      let totColl = obtainTotalCollateralHasFieldConstraint supp $ getField @"totalCollateral" txbody
-      in TxTotalCollateral supp $ fromShelleyLovelace totColl
+      case obtainTotalCollateralHasFieldConstraint supp $ getField @"totalCollateral" txbody of
+        SNothing ->TxTotalCollateralNone
```
newhoggy, line 2248: Need space after `->`

JaredCorduan (approved):
LGTM!

newhoggy (approved):

cardano-api/gen/Gen/Cardano/Api.hs
Jimbo4350: We're not, it depends on `mkEvaluationContext` but: https://input-output-rnd.slack.com/archives/C21UF2WVC/p1649876509637199

It currently returns nothing so for now we just panic. When the plutus team gives more debugging info we can improve.

PR #3803, opened 2022-04-19
Babbage functionality integration in cardano api Part 1

Jimbo4350:


cardano-api/src/Cardano/Api/Fees.hs
```diff
@@ -1039,12 +1080,17 @@ calculateMinimumUTxO era txout@(TxOut _ v _ _) pparams' =
...
-        Just (Lovelace costPerWord) -> do
-          Right . lovelaceToValue
-            $ Lovelace (Alonzo.utxoEntrySize (toShelleyTxOutAny era txout) * costPerWord)
-        Nothing -> Left PParamsUTxOCostPerWordMissing
-    ShelleyBasedEraBabbage -> error "TODO: Babbage"
+      let lTxOut = toShelleyTxOutAny era txout
+          babPParams = toAlonzoPParams pparams'
+          minUTxO = Shelley.evaluateMinLovelaceOutput babPParams lTxOut
+          val = lovelaceToValue $ fromShelleyLovelace minUTxO
+      in Right val
```
newhoggy, line 1087: I notice that this changed and that the new version doesn't fail.  What's the story here?

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -528,56 +531,55 @@ txOutToJsonValue era (TxOut addr val dat refScript) =
...
+        , "referenceScript" .= refScriptJsonVal refScript
         ]
  where
-   datHashJsonVal :: TxOutDatum ctx era -> Maybe Aeson.Pair
+   datHashJsonVal :: TxOutDatum ctx era -> Aeson.Pair
    datHashJsonVal d =
      case d of
-       TxOutDatumNone -> Nothing
+       TxOutDatumNone ->
+         "datumhash" .= Aeson.Null
```
newhoggy, line 554: Behavioural change where absence maps to `field = null` rather than absence of field.  Is this intended?

newhoggy (approved):

cardano-api/src/Cardano/Api/Fees.hs
Jimbo4350: The ledger provides an instance called `CLI` with a function `evaluateMinLovelaceOutput`. I think after I implemented the logic manually here they created that class that duplicates the logic.

cardano-api/src/Cardano/Api/TxBody.hs
Jimbo4350: Yeah the JSON instances were a pain to get round tripping. I opted for `Null` instead to make it simpler. Even then this needs to be revisited and refactored because its too complicated.

Jimbo4350:
bors merge

PR #3810, opened 2022-04-21
Vasil cardano-cli update

Jimbo4350:
Add the following options to the `build` and `build-raw` command:
- `--tx-out-inline-datum` - specify an inline datum in a tx output
- `--tx-in-reference` - specify a reference tx input
- `--reference-script-file` - specify a reference script
- `--tx-out-return-collateral` - specify a return collateral tx out
- `--tx-total-collateral` - specify the total collateral that will be collected in the event of a Plutus script failure


newhoggy (approved):

Jimbo4350:
bors merge

PR #3812, opened 2022-04-21
Benchmarking: forgestress

deepfire:
1. Log analysis has been rewritten as a composition of orthogonal chunks -- micro-ops, with clear inputs and outputs
2. The workbench log output has been clarified
3. Log analysis filters are now specified by profiles
4. Profiles now provide run-time expectations to the workbench -- informed by shutdown slot and tx workload
5. The workbench now enforces the above run-time expectations on the `fixed` scenario profiles (time-bounded)
6. Log analysis can now be requested for runs
7. Runs can be iterated (only makes sense for `fixed` scenario profiles)
8. Genesis handling has been properly moved
9. The profile structure has been cleaned up a lot
10. More intermediate computed values are now specified as part of profile output
11. Log analysis now emits separate components that are easier to use for reports


mgmeier (approved):

deepfire:
This has passed CI on the previous run, so merging directly.

PR #3829, opened 2022-04-30
Use append flag on tee command to prevent it from overwriting the pre…

merkeld:
My nix build failed because the second tee command overwrote the enabling of the flage features. Adding the flag "-a" prevents this.


olgahryniuk (approved):

newhoggy:
It appears that the fixes have already been independently applied by https://github.com/input-output-hk/cardano-node/pull/4613

Thanks for the contribution!

PR #3847, opened 2022-05-09
Generate a cost model basing on the language version.

ghost:
Fixes the `genCostModels` function by generating cost models basing on the language version instead of using `defaultCostModelParams` which is default only for plutusV2 (with 170 params) while plutusV1 has 166 params.


koslambrou:
@Jimbo4350

PR #3852, opened 2022-05-10
cardano-tracer: init RTView

denisshevchenko:
This is pre-MVP for RTView.


deepfire (approved):
Excellent, thank you Denis!

deepfire:
bors r_

PR #3861, opened 2022-05-14
Updated docker-compose.yml with latest versions & added CARDANO_NODE_SOCKET_PATH

os11k:
Updated docker-compose.yml with latest versions(1.34.1) for cardano-node & cardano-submit-api, additionally added variable CARDANO_NODE_SOCKET_PATH, so we can run now:

`docker exec -ti cardano-node-iog_cardano-node_1 /usr/local/bin/cardano-cli query tip --testnet-magic 1097911063`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3.5"
 
 services:
   cardano-node:
-    image: inputoutput/cardano-node:${CARDANO_NODE_VERSION:-1.34.1}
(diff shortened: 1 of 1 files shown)


os11k:
Hi! Any updates? I think it is quite good idea to have CARDANO_NODE_SOCKET_PATH env set out of the box, so we can query tip, just after docker-compose up, without setting any additional variables...

os11k:
Added one more change for mounting /data/db as folder on host, so we can copy existing DB to that folder and no need to wait to sync via internet, might be quite useful.

PR #3862, opened 2022-05-15
Update index state to 2022-07-01T00:00:00Z

newhoggy:


newhoggy:
Thanks @hamishmack for the fixes!

cardano-api/test/Test/Cardano/Api/Json.hs
```diff
@@ -3,6 +3,8 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE TemplateHaskell #-}
 
+{-# OPTIONS_GHC -Wno-deprecations #-} -- TODO Fix deprecations
```
Jimbo4350: What deprecations?

cardano-api/test/Test/Cardano/Api/Typed/CBOR.hs
```diff
@@ -191,4 +186,37 @@ roundtrip_Tx_Cddl (AnyCardanoEra era) =
 -- -----------------------------------------------------------------------------
 
 tests :: TestTree
-tests = $testGroupGenerator
+tests = testGroup "Test.Cardano.Api.Typed.CBOR"
+  [ testPropertyNamed "roundtrip witness byron CBOR"                         "roundtrip witness byron CBOR"                         prop_roundtrip_witness_byron_CBOR
```
Jimbo4350, line 190: Can we just avoid all together using `testPropertyNamed`? It seems like a lot of additional trouble for not much benefit. Hedgehog will display the property name by default.

Jimbo4350 (changes requested):

cardano-api/test/Test/Cardano/Api/Json.hs
newhoggy: Not sure why this appeared in your review, but it's not in the PR.

bench/locli/src/Cardano/Analysis/Chain.hs
```diff
@@ -1,6 +1,9 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving#-}
-{-# OPTIONS_GHC -Wno-deprecations -Wno-orphans #-}
+
+{-# OPTIONS_GHC -Wno-orphans #-}
```
Jimbo4350, line 3: :+1:

cardano-api/cardano-api.cabal
```diff
@@ -217,7 +217,6 @@ test-suite cardano-api-test
                       , tasty
                       , tasty-hedgehog
                       , tasty-quickcheck
-                      , tasty-th
```
Jimbo4350, line 220: :+1: Great. The less deps we use the better.

Jimbo4350 (approved):
LGTM! Thanks for keeping an eye on this.

cleverca22 (approved):

MarcFontaine (approved):
Lgtm

denisshevchenko (approved):

PR #3873, opened 2022-05-18
Porting ekg-direct metrics to new-tracing

jutaro:
Direct EKG is a way to record measurments without using the tracing system. We ported it over to new tracing, partly with using standard cardano-tracer, and partly with sending it directly to the EKGBackend


trace-dispatcher/src/Cardano/Logging/DocuGenerator.hs
```diff
@@ -354,21 +353,21 @@ documentMarkdown (Documented documented) tracers = do
...
+    --       l   -> fromText "> Severities: "
+    --               <> mconcat (intersperse (singleton ',')
+    --                     (map (asCode . fromString . show) l))
+    --   <>
+    --     case nub ldPrivacy of
+    --       []  -> fromText "\nPrivacy:   " <> asCode (fromString (show Public))
+    --       [p] -> fromText "\nPrivacy:   " <> asCode (fromString (show p))
+    --       l   -> fromText "\nPrivacies: "
+    --               <> mconcat (intersperse (singleton ',')
+    --                     (map (asCode . fromString . show) l))
```
deepfire, line 370: @jutaro, do we need this commented section?

deepfire (approved):
@jutaro, looks good -- just one comment/question!

trace-dispatcher/src/Cardano/Logging/DocuGenerator.hs
jutaro: Removed, hope I can recover it one day

jutaro:
bors merge

PR #3876, opened 2022-05-19
trace-dispatcher: seperate routing of metrics

jutaro:
Metrics should be send in any case, and not be filtered by configuration.


deepfire (approved):
Thank you @jutaro!

jutaro:
bors merge

PR #3884, opened 2022-05-23
github-pages:  try switching to actions-gh-pages, for a compacted history

deepfire:


deepfire:
Let's test this out -- we can back this out easily if it breaks.

PR #3896, opened 2022-05-24
Install a dummy SIGHUP handler for non p2p mode

### LINKED ISSUE #3872: [BUG] - SIGHUP causes full re-validation
**Internal/External**
*External*

**Area**
*Other*

**Summary**
Sending a SIGHUP signal to the cardano-node process as per [understanding-your-configuration-files-and-how-to-use-them](https://github.com/input-output-hk/cardano-node/blob/master/doc/getting-started/understanding-config-files.md#understanding-your-configuration-files-and-how-to-use-them):
> You can tell the node that the topology configuration file changed by sending a SIGHUP signal to the cardano-node process, e.g. pkill -HUP cardano-node. After receiving the signal, cardano-node will re-read the file and restart all dns resolution. Please note that this only applies to the topology configuration file!

results in a full re-validation.

**Steps to reproduce**
go to a terminal on a machine running a node instance.
issue command `pkill -HUP cardano-node`

**Expected behavior**
The node should just continue synchronising with its new list of peers after re-reading the topology config file and not perform a re-validation.

- OS Name: Linux Mint
- OS Version 20.3
- Node version 1.34.1 - linux-x86_64 - ghc-8.10
- CLI version  1.34.1 - linux-x86_64 - ghc-8.10

coot:
When sending SIGHUP handler, p2p will reload the configuration, while
non-p2p will exit.  Instead of that we install a dummy handler, which
logs that topology reloading for non-p2p is not supported, which
prevents haskell runtime to use the default signal handler (which forces
the process to exit).

Fixes #3872.

### DIFF
--- a/cardano-node/src/Cardano/Node/Run.hs
+++ b/cardano-node/src/Cardano/Node/Run.hs
@@ -423,6 +423,14 @@ handleSimpleNode runP p2pMode tracers nc onKernel = do
               , srnMaybeMempoolCapacityOverride = ncMaybeMempoolCapacityOverride nc
               }
       DisabledP2PMode -> do
+#ifdef UNIX
+        _ <- Signals.installHandler
+              Signals.sigHUP
+              (Signals.Catch $ do
+                traceWith (startupTracer tracers) NetworkConfigUpdateUnsupported
+              )
+              Nothing
+#endif
         eitherTopology <- TopologyNonP2P.readTopologyFile nc
         nt <- either (\err -> panic $ "Cardano.Node.Run.handleSimpleNodeNonP2P.readTopologyFile: " <> err) pure eitherTopology
         let (ipProducerAddrs, dnsProducerAddrs) = producerAddressesNonP2P nt
--- a/cardano-node/src/Cardano/Node/Startup.hs
+++ b/cardano-node/src/Cardano/Node/Startup.hs
@@ -79,6 +79,10 @@ data StartupTrace blk =
   --
   | NetworkConfigUpdate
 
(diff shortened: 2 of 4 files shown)


cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -423,6 +423,15 @@ handleSimpleNode runP p2pMode tracers nc onKernel = do
               , srnMaybeMempoolCapacityOverride = ncMaybeMempoolCapacityOverride nc
               }
       DisabledP2PMode -> do
+#ifdef UNIX
+        _ <- Signals.installHandler
+              Signals.sigHUP
+              (Signals.Catch $ do
+                traceWith (startupTracer tracers)
+                        $ NetworkConfigUpdateError "not supported in non-p2p mode"
```
dnadales, line 431: Don't we need to clarify what is not supported? (configuration reloading in this case I guess :) )

coot: Yes, and actually we should use a different massage.  This is not an error, just a warning.

coot: I changed it to `NetworkConfigUpdateUnsupported` which expands to:
```
Network topology reconfiguration is not supported in non-p2p mode
```

denisshevchenko (approved):

coot:
bors merge

newhoggy (approved):

coot:
bors merge

coot:
bors merge

PR #3911, opened 2022-05-25
Version the plutus scripts

newhoggy:
Move all plutus scripts to under the `v1` directory.

New Babbage scripts will use scripts under the `v2` directory.

If it's possible for a script to run in both Alonzo and Babbage, they will be able to use the corresponding script that is built with the correct version of plutus language by choosing the directory.

`example-txin-locking-plutus-script.sh` modified to run against both Alonzo and Babbage.

Scripts were generated by https://github.com/input-output-hk/plutus-apps/pull/482


Jimbo4350 (approved):
Good stuff :+1:

PR #3924, opened 2022-05-26
Update deps for Ledger, Consensus, et al

nfrisby:
This PR does two things.

- It updates many dependencies.

- The necessary integration is just trivial refactoring. Beyond that, this PR also updates the log representation of some tracer events; typo fixes and a new shape for ChainSync server events.

```
           cardano-ledger f31c29add - Merge pull request #2819 from input-output-hk/andre/babbage (6 days ago) <Alexey Kuleshevich>
iohk-monitoring-framework 066f700   - Merge #627 (5 days ago) <iohk-bors[bot]>
                   io-sim f4183f2   - Merge pull request #2 from input-output-hk/coot/check-stylish-script (12 days ago) <Marcin Szamotulski>
        ouroboros-network c254d02b9 - Merge #3766 (5 hours ago) <iohk-bors[bot]>
                   plutus d24a7540e - Backport: PLT-250: mkEvaluationContext now lives in MonadError CostModelApplyError (#4641) (9 days ago) <Nikolaos Bezirgiannis>
          typed-protocols 181601b   - Merge pull request #4 from input-output-hk/coot/check-stylish (12 days ago) <Marcin Szamotulski>
```

### DIFF
--- a/cardano-api/gen/Gen/Cardano/Api.hs
+++ b/cardano-api/gen/Gen/Cardano/Api.hs
@@ -26,6 +26,7 @@ import qualified Cardano.Ledger.BaseTypes as Ledger
 import qualified Cardano.Ledger.Coin as Ledger
 import           Cardano.Ledger.Shelley.Metadata (Metadata (..), Metadatum (..))
 
+
 import           Hedgehog (Gen, Range)
 import qualified Hedgehog.Gen as Gen
 import qualified Hedgehog.Internal.Range as Range
--- a/cardano-api/gen/Gen/Cardano/Api.hs
+++ b/cardano-api/gen/Gen/Cardano/Api.hs
@@ -92,7 +92,7 @@ genCostModels = do
   CostModel cModel <- genCostModel
   lang <- genLanguage
   case Alonzo.mkCostModel lang cModel of
-    Left err -> panic . Text.pack $ "genCostModels: " <> err
+    Left err -> panic . Text.pack $ "genCostModels: " <> show err
     Right alonzoCostModel ->
       Alonzo.CostModels . conv <$> Gen.list (Range.linear 1 3) (return alonzoCostModel)
  where
--- a/cardano-api/src/Cardano/Api/ProtocolParameters.hs
+++ b/cardano-api/src/Cardano/Api/ProtocolParameters.hs
@@ -752,12 +752,12 @@ newtype CostModel = CostModel (Map Text Integer)
 validateCostModel :: PlutusScriptVersion lang
                   -> CostModel
                   -> Either InvalidCostModel ()
-validateCostModel PlutusScriptV1 (CostModel m)
-  | Alonzo.isCostModelParamsWellFormed m = Right ()
-  | otherwise                        = Left (InvalidCostModel (CostModel m))
-validateCostModel PlutusScriptV2 (CostModel m)
-  | Alonzo.isCostModelParamsWellFormed m = Right ()
-  | otherwise                        = Left (InvalidCostModel (CostModel m))
+validateCostModel PlutusScriptV1 (CostModel m) =
+    bimap (\_err -> InvalidCostModel (CostModel m)) id
+  $ Alonzo.assertWellFormedCostModelParams m
+validateCostModel PlutusScriptV2 (CostModel m) =
+    bimap (\_err -> InvalidCostModel (CostModel m)) id
+  $ Alonzo.assertWellFormedCostModelParams m
 
 -- TODO alonzo: it'd be nice if the library told us what was wrong
 newtype InvalidCostModel = InvalidCostModel CostModel
@@ -797,7 +797,7 @@ fromAlonzoScriptLanguage Alonzo.PlutusV1 = AnyPlutusScriptVersion PlutusScriptV1
 fromAlonzoScriptLanguage Alonzo.PlutusV2 = AnyPlutusScriptVersion PlutusScriptV2
 
 toAlonzoCostModel :: CostModel -> Alonzo.Language -> Either String Alonzo.CostModel
-toAlonzoCostModel (CostModel m) l = Alonzo.mkCostModel l m
+toAlonzoCostModel (CostModel m) l = bimap show id $ Alonzo.mkCostModel l m
 
 fromAlonzoCostModel :: Alonzo.CostModel -> CostModel
 fromAlonzoCostModel m = CostModel $ Alonzo.getCostModelParams m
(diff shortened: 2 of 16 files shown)


nfrisby:
This is Draft because it points at the commit of https://github.com/input-output-hk/ouroboros-network/pull/3774 instead of the merged result of it.

https://github.com/input-output-hk/ouroboros-network/pull/3774 has not yet merged, because it similarly points at https://github.com/input-output-hk/iohk-monitoring-framework/pull/627.

https://github.com/input-output-hk/iohk-monitoring-framework/pull/627 has not yet merged, because `bors merge` has time out four or fives times now :'(

lehins (approved):
LGTM

cardano-api/src/Cardano/Api/ProtocolParameters.hs
```diff
@@ -752,12 +752,12 @@ newtype CostModel = CostModel (Map Text Integer)
...
                   -> CostModel
                   -> Either InvalidCostModel ()
-validateCostModel PlutusScriptV1 (CostModel m)
-  | Alonzo.isCostModelParamsWellFormed m = Right ()
-  | otherwise                        = Left (InvalidCostModel (CostModel m))
-validateCostModel PlutusScriptV2 (CostModel m)
-  | Alonzo.isCostModelParamsWellFormed m = Right ()
-  | otherwise                        = Left (InvalidCostModel (CostModel m))
+validateCostModel PlutusScriptV1 (CostModel m) =
+    bimap (\_err -> InvalidCostModel (CostModel m)) id
```
michaelpj, line 756: do you maybe want to propagate the error?

cardano-api/gen/Gen/Cardano/Api.hs
```diff
@@ -92,7 +92,7 @@ genCostModels = do
   CostModel cModel <- genCostModel
   lang <- genLanguage
   case Alonzo.mkCostModel lang cModel of
-    Left err -> panic . Text.pack $ "genCostModels: " <> err
+    Left err -> panic . Text.pack $ "genCostModels: " <> show err
```
michaelpj, line 95: FYI these errors have a `Pretty` instance if you want to use that.

michaelpj (approved):

cardano-api/src/Cardano/Api/ProtocolParameters.hs
nfrisby: This is a fair question. My goal was this PR was to unblock the dependency bumps.

cc @Jimbo4350 re propagating this error now?

nfrisby:
I have now marked this Ready for Review and will notify @Jimbo4350 and @kosyrevSerge

amesgen (approved):

cardano-api/gen/Gen/Cardano/Api.hs
Jimbo4350: Yep that sounds good

cardano-api/src/Cardano/Api/ProtocolParameters.hs
Jimbo4350: Yeah we may as well do it now. I can take over.

cardano-node/src/Cardano/Tracing/OrphanInstances/Shelley.hs
```diff
@@ -1026,6 +1026,7 @@ instance ToJSON (Alonzo.CollectError StandardCrypto) where
               Alonzo.InlineDatumsNotSupported -> String "Inline datums not supported"
               Alonzo.ReferenceScriptsNotSupported -> String "Reference scripts not supported"
               Alonzo.ReferenceInputsNotSupported -> String "Reference inputs not supported"
+              Alonzo.TimeTranslationPastHorizon -> String "Time translation requested past the horizon"
```
Jimbo4350, line 1029: Where can I find more information about the "horizon"? I think we can improve this error message.

Jimbo4350 (commented):
LGTM. I can take over and make some adjustments.

cardano-node/src/Cardano/Tracing/OrphanInstances/Shelley.hs
nfrisby: 1) I'm unaware of user facing documentation for this. 😬 

2) There's a whole chapter of the Consensus Report about it "Part IV Hard Fork Combinator", "Chapter 17 Time", where "beyond the horizon" means "past the Forecast range". But it's very abstract/fundamental/not necessarily in "Cardano terminology".

3) High-level summary (@JaredCorduan please review this):

* the horizon is either the end of the current epoch or the end of the next epoch
    * it's the end of the current epoch if the ledger state does not determine whether the current epoch is the end of the current era
    * it's the end of the next epoch if the ledger state does determine whether the current epoch is the end of the current era
    * (ie the horizon is the end of the current epoch if and only if we don't know whether it's the last epoch of this era)
    
* when does the ledger state does determine whether the current epoch is the end of the current era?

    * the transition happens at the end of the current epoch if and only if all the necessary proposal/voting transactions are in a block on the chain too deep to be rolled back _and_ there's a stability window's worth of slots between that block and the end of the current epoch.
    * The current epoch is determined as the last of its era when the above is already true in the ledger state.
    * The current epoch is determined as _not_ the last of its era when the above _cannot_ be true, eg it's not true now and we're already within a stability window of the end of the epoch.
    * Otherwise, it's not determined: ie it's not true now but it still might become true.

JaredCorduan: That all sounds correct to me, though I might re-phrase it a bit. I would say that the ledger is ignorant about transitions to future eras, and that the consensus layer uses the ledger state to determine if an era is ending.

In particular, votes to change the protocol parameters in the next epoch _must_ occur within two stability windows of the end of the epoch (the vote states what epoch it is intended for, you can also place a pending vote for the following epoch). After one stability window from the end of the epoch (so one stability window _after_ voting has ended), the consensus layer looks to see if the major protocol version (which is a protocol parameter). is going to change, and hence usher in a new era (provided it is not an intra-era hard fork).

Jimbo4350: I will come back to this another time, thanks for the input

nfrisby:
bors retry

amesgen:
FTR: I asked for code owner approvals on [Slack](https://input-output-rnd.slack.com/archives/CR599HMFX/p1654248700006379)

JaredCorduan (approved):
thank you @nfrisby !

Jimbo4350:
bors merge

deepfire (approved):

nfrisby:
bors retry

PR #3933, opened 2022-05-28
Optional CLI-parameter for forward socket

### LINKED ISSUE #3533: tracing:  add a node CLI flag for the tracing socket path
The wallet would benefit from a non-config-involving way of setting node's tracing socket path.

denisshevchenko:
Closes #3533


deepfire (approved):
Thank you @denisshevchenko !

scripts/lite/configuration/shelley-3.yaml
```diff
@@ -286,6 +286,4 @@ TraceOptionLimiter:
     limiterName: ResourceLimiter
     limiterFrequency: 1.0
 TraceOptionForwarder:
-    address:
```
Jimbo4350, line 289: So the socket configuration is decided with the `cardano-node` executable but the mode is decided with the yaml file?

cardano-node/src/Cardano/Node/Configuration/POM.hs
```diff
@@ -168,6 +171,7 @@ data PartialNodeConfiguration
        , pncLoggingSwitch  :: !(Last Bool)
        , pncLogMetrics     :: !(Last Bool)
        , pncTraceConfig    :: !(Last PartialTraceOptions)
```
Jimbo4350, line 173: The question here is, should be include the socket path and mode configuration within `TraceOptions`/`PartialTraceOptions`?

scripts/lite/configuration/shelley-3.yaml
deepfire: @Jimbo4350, thanks for catching this -- this was actually intended to both be decided by CLI -- fixed.

cardano-node/src/Cardano/Node/Configuration/POM.hs
deepfire: @Jimbo4350, 

- if you mean the `TraceConfig` type, which is structurally carrying the config for new tracing -- it could be, but then it would have to live in the `trace-dispatcher` library, since that `TraceOptions` type is defined there.  I did not think it was a good option, frankly.
- if you indeed mean the `TraceOptions` type, then, well -- it structurally is just a tracing system selector data type.  So not sure socket paths belong in it..

deepfire: It's a difficult question -- but I'm open to discuss!

Jimbo4350: Alright I think let's just rename it to `ncTraceForwardSocket` for now and that should be good.

cardano-node/src/Cardano/Node/Parsers.hs
```diff
@@ -46,6 +46,7 @@ nodeRunParser = do
   topFp <- lastOption parseTopologyFile
   dbFp <- lastOption parseDbPath
   socketFp <- lastOption $ parseSocketPath "Path to a cardano-node socket"
+  fwdSocketFp <- lastOption $ parseForwardSocketPath "Path to a forward socket"
```
Jimbo4350: Can we elaborate a little more on `"Path to a forward socket"`? Maybe `"Path to a trace-dispatcher forwarding socket"`

scripts/lite/configuration/shelley-3.yaml
Jimbo4350: So shouldn't the `TraceOptionForwarder:` be completely removed? But actually the actual current pattern is:
```
case makeNodeConfiguration $ defaultPartialNodeConfiguration <> configYamlPc <> cmdPc of
            Left err -> panic $ "Error in creating the NodeConfiguration: " <> Text.pack err
            Right nc' -> return nc'
```
The cli takes precedence but it should be possible to specify via the config yaml file.

Jimbo4350 (changes requested):
A couple of comments

scripts/lite/configuration/shelley-3.yaml
deepfire: @Jimbo4350, problem is `TraceOptionForwarder` has other fields -- which are normally not changed in the config files, which is why you don't see them:

https://github.com/input-output-hk/cardano-node/blob/issue-3533-node-CLI-socket/trace-dispatcher/src/Cardano/Logging/Types.hs#L325-L331

cardano-node/src/Cardano/Node/Configuration/POM.hs
deepfire: @Jimbo4350, done.

cardano-node/src/Cardano/Node/Parsers.hs
deepfire: That's an outdated bit of code -- it's now documented as:
- `"Accept incoming cardano-tracer connection at local socket"`, and
- `"Connect to cardano-tracer listening on a local socket"`

deepfire:
@Jimbo4350, all replied to!

cardano-node/src/Cardano/Node/Parsers.hs
denisshevchenko: @deepfire Strictly speaking, this socket is for forwarding to _any_ external Haskell program, and `cardano-tracer` is just one example of such a program. So, this socket doesn't know anything about `cardano-tracer`. :-)

cardano-node/src/Cardano/Node/Configuration/POM.hs
Jimbo4350: I'm happy to hear other suggestions `ncForwardSocket` doesn't mean anything to me. `ncTraceForwardSocket` tells me that this socket is forwarding traces. What do you have in mind?

Jimbo4350: I just saw this: https://github.com/input-output-hk/cardano-node/pull/3933/commits/3f64c4c5597bf8cf767b84652d912b70eedad631#r888707952

`ncForwardingSocket` ?

scripts/lite/configuration/shelley-3.yaml
Jimbo4350: Alright fair enough, let's leave it as you prefer it

cardano-node/src/Cardano/Node/Configuration/POM.hs
deepfire: @Jimbo4350, on a second thought that's fine, actually -- my initial intent was to reflect on `cardano-tracer` somehow, but as @denisshevchenko added, that's overly specific.

I was initially slightly dissatisfied by `TraceForward` being too reflective of the legacy `trace-forwarder` backend, but indeed -- we can simply now read that as just "socket for forwarding traces", which makes sense.

Jimbo4350 (approved):

PR #3941, opened 2022-05-30
tx-generator: zero config legacy tracing

MarcFontaine:
This is a fallback until the tx-generator has been ported to the new tracing infrastructure.
The fallback kicks in, if the tx-generator is started with a node configuration that uses the tracing infrastructure.


deepfire (approved):
Thank you @MarcFontaine !

PR #3945, opened 2022-05-31
Fix kes period info command

newhoggy:


newhoggy:
Example output:

```
$ CARDANO_NODE_SOCKET_PATH=example/node-spo1/node.sock cardano-cli query kes-period-info --testnet-magic 42 --op-cert-file example/node-spo1/opcert.cert
✓ Operational certificate's KES period is within the correct KES period interval
✓ The operational certificate counter agrees with the node protocol state counter
{
    "qKesCurrentKesPeriod": 0,
    "qKesEndKesInterval": 60,
    "qKesKesKeyExpiry": null,
    "qKesMaxKESEvolutions": 60,
    "qKesNodeStateOperationalCertificateNumber": 0,
    "qKesOnDiskOperationalCertificateNumber": 0,
    "qKesRemainingSlotsInKesPeriod": 7775071,
    "qKesSlotsPerKesPeriod": 129600,
    "qKesStartKesInterval": 0
}
```

Jimbo4350 (changes requested):

cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -148,6 +149,7 @@ import qualified Ouroboros.Consensus.Mempool.TxLimits as TxLimits
 import qualified Ouroboros.Consensus.Node.ProtocolInfo as Consensus
 import qualified Ouroboros.Consensus.Protocol.Abstract as Consensus
 -- import qualified Ouroboros.Consensus.Protocol.Praos as Consensus
```
Jimbo4350, line 151: Can you remove this comment too?

Jimbo4350 (approved):
LGTM! We should ask @disassembler to test the command for us to double check its working. We should create an issue somewhere that after the hardfork we will only be able to bump the op cert counter by 1. We will need to update the command to do this check in the Babbage era onwards.

gitmachtl:
> Example output `babbage`:
> 
> ```
> $ CARDANO_NODE_SOCKET_PATH=example/node-spo1/node.sock cardano-cli query kes-period-info --testnet-magic 42 --op-cert-file example/node-spo1/opcert.cert
> ✓ Operational certificate's KES period is within the correct KES period interval
> ✓ The operational certificate counter agrees with the node protocol state counter
> {
>     "qKesCurrentKesPeriod": 0,
>     "qKesEndKesInterval": 60,
>     "qKesKesKeyExpiry": null,
>     "qKesMaxKESEvolutions": 60,
>     "qKesNodeStateOperationalCertificateNumber": 0,
>     "qKesOnDiskOperationalCertificateNumber": 0,
>     "qKesRemainingSlotsInKesPeriod": 7775071,
>     "qKesSlotsPerKesPeriod": 129600,
>     "qKesStartKesInterval": 0
> }
> ```

is the plaintext output at the top written to /dev/stderr? otherwise it is a strang mixed format output of plaintext with json, thats not good.

gitmachtl:
question, will the `qKesNodeStateOperationalCertificateNumber` be set again to `null ` if there was no block produced in the current KesPeriod? or will this value show the latest opcert number used on chain in the past?

gitmachtl:
this should not greenlit with two checkmarks right? because the new opcert counter 3 is too high when the last one on chain is only 1. but the current output looks like that:

```
✓ Operational certificate's KES period is within the correct KES period interval
✓ The operational certificate counter agrees with the node protocol state counter
{
    "qKesCurrentKesPeriod": 16,
    "qKesEndKesInterval": 136,
    "qKesKesKeyExpiry": null,
    "qKesMaxKESEvolutions": 120,
    "qKesNodeStateOperationalCertificateNumber": 1,
    "qKesOnDiskOperationalCertificateNumber": 3,
    "qKesRemainingSlotsInKesPeriod": 10342913,
    "qKesSlotsPerKesPeriod": 86400,
    "qKesStartKesInterval": 16
}
```

PR #3953, opened 2022-05-31
Reference script integration

Jimbo4350:
- Implements a shell script that uses a spending plutus v2 reference script. Currently only `build-raw` is working with reference scripts.
- Updates the api and cli to enable the usage of reference scripts
- Adds documentation outlining how to use reference scripts

### DIFF
--- a/cardano-api/src/Cardano/Api/Script.hs
+++ b/cardano-api/src/Cardano/Api/Script.hs
@@ -66,6 +66,7 @@ module Cardano.Api.Script (
 
     -- * The Plutus script language
     PlutusScript(..),
+    PlutusScriptOrReferenceInput(..),
     examplePlutusScriptAlwaysSucceeds,
     examplePlutusScriptAlwaysFails,
 
@@ -159,6 +160,7 @@ import           Cardano.Api.SerialiseJSON
 import           Cardano.Api.SerialiseRaw
 import           Cardano.Api.SerialiseTextEnvelope
 import           Cardano.Api.SerialiseUsing
+import           Cardano.Api.TxIn
 import           Cardano.Api.Utils (failEitherWith)
 
 {- HLINT ignore "Use section" -}
@@ -686,7 +688,6 @@ toScriptInEra era (ScriptInAnyLang lang s) = do
 eraOfScriptInEra :: ScriptInEra era -> ShelleyBasedEra era
 eraOfScriptInEra (ScriptInEra langInEra _) = eraOfScriptLanguageInEra langInEra
 
-
 -- ----------------------------------------------------------------------------
 -- Scripts used in a transaction (in an era) to witness authorised use
 --
@@ -720,6 +721,14 @@ data WitCtx witctx where
      WitCtxMint  :: WitCtx WitCtxMint
      WitCtxStake :: WitCtx WitCtxStake
 
+-- | Scripts can now exist in the UTxO at a transaction output. We can
+-- reference these scripts via specification of a reference transaction input
+-- in order to witness spending inputs, withdrawals, certificates
+-- or to mint tokens. This datatype encapsulates this concept.
+data PlutusScriptOrReferenceInput lang
+  = PScript (PlutusScript lang)
+  | PReferenceScript TxIn
+  deriving (Eq, Show)
 
 -- | A /use/ of a script within a transaction body to witness that something is
 -- being used in an authorised manner. That can be
@@ -743,7 +752,7 @@ data ScriptWitness witctx era where
 
      PlutusScriptWitness :: ScriptLanguageInEra  lang era
                          -> PlutusScriptVersion  lang
-                         -> PlutusScript         lang
+                         -> PlutusScriptOrReferenceInput lang
                          -> ScriptDatum witctx
                          -> ScriptRedeemer
                          -> ExecutionUnits
@@ -785,14 +794,18 @@ data ScriptDatum witctx where
 deriving instance Eq   (ScriptDatum witctx)
 deriving instance Show (ScriptDatum witctx)
 
-
-scriptWitnessScript :: ScriptWitness witctx era -> ScriptInEra era
+-- We cannot always extract a script from a script witness due to reference scripts.
+-- Reference scripts exist in the UTxO, so without access to the UTxO we cannot
+-- retrieve the script.
+scriptWitnessScript :: ScriptWitness witctx era -> Maybe (ScriptInEra era)
 scriptWitnessScript (SimpleScriptWitness langInEra version script) =
-    ScriptInEra langInEra (SimpleScript version script)
+    Just $ ScriptInEra langInEra (SimpleScript version script)
 
-scriptWitnessScript (PlutusScriptWitness langInEra version script _ _ _) =
-    ScriptInEra langInEra (PlutusScript version script)
+scriptWitnessScript (PlutusScriptWitness langInEra version (PScript script) _ _ _) =
+    Just $ ScriptInEra langInEra (PlutusScript version script)
 
+scriptWitnessScript (PlutusScriptWitness _ _ (PReferenceScript _) _ _ _) =
+    Nothing
 
 -- ----------------------------------------------------------------------------
 -- The kind of witness to use, key (signature) or script
@@ -1136,7 +1149,6 @@ toShelleyScript (ScriptInEra langInEra (PlutusScript PlutusScriptV2
                                          (PlutusScriptSerialised script))) =
     case langInEra of
       PlutusScriptV2InBabbage -> Alonzo.PlutusScript Alonzo.PlutusV2 script
-
 
 fromShelleyBasedScript  :: ShelleyBasedEra era
                         -> Ledger.Script (ShelleyLedgerEra era)
--- a/cardano-api/src/Cardano/Api/Script.hs
+++ b/cardano-api/src/Cardano/Api/Script.hs
@@ -966,7 +966,7 @@ fromShelleyScriptHash = ScriptHash
 
 
 -- ----------------------------------------------------------------------------
--- The simple native script language
+-- The simple script language
 --
 
 data SimpleScript lang where
--- a/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
+++ b/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
@@ -229,10 +229,58 @@ pScriptFor name (Just deprecated) help =
         <> Opt.internal
         )
 
+pPlutusReferenceScriptWitnessFiles
+  :: WitCtx witctx
+  -> BalanceTxExecUnits -- ^ Use the @execution-units@ flag.
+  -> Parser (ScriptWitnessFiles witctx)
+pPlutusReferenceScriptWitnessFiles witctx autoBalanceExecUnits =
+   toReferenceScriptWitnessFiles
+     <$> ( (,,,,) <$> pReferenceTxIn
+                  <*> pAnyScriptLang
+                  <*> pScriptDatumOrFile "reference-tx-in" witctx
+                  <*> pScriptRedeemerOrFile "reference-tx-in"
+                  <*> (case autoBalanceExecUnits of
+                        AutoBalance -> pure (ExecutionUnits 0 0)
+                        ManualBalance -> pExecutionUnits "reference-tx-in")
+         )
+ where
+  pReferenceTxIn :: Parser TxIn
+  pReferenceTxIn =
+    Opt.option (readerFromParsecParser parseTxIn)
+      (  Opt.long "tx-in-reference"
+      <> Opt.metavar "TX-IN"
+      <> Opt.help "TxId#TxIx - Specify a reference input. The reference input may or may not have\
+                  \ a plutus reference script attached."
+      )
+  pAnyScriptLang :: Parser AnyScriptLanguage
+  pAnyScriptLang =
+    Opt.flag' (AnyScriptLanguage $ SimpleScriptLanguage SimpleScriptV1)
+      (  Opt.long "simple-script-v1"
+      <> Opt.help "Specify a simple script v1 reference script."
+      ) <|>
+    Opt.flag' (AnyScriptLanguage $ SimpleScriptLanguage SimpleScriptV2)
+      (  Opt.long "simple-script-v2"
+      <> Opt.help "Specify a simple script v2 reference script."
+      ) <|>
+    Opt.flag' (AnyScriptLanguage $ PlutusScriptLanguage PlutusScriptV1)
+      (  Opt.long "plutus-script-v1"
+      <> Opt.help "Specify a plutus script v1 reference script."
+      ) <|>
+    Opt.flag' (AnyScriptLanguage $ PlutusScriptLanguage PlutusScriptV2)
+      (  Opt.long "plutus-script-v2"
+      <> Opt.help "Specify a plutus script v2 reference script."
+      )
+
+
+  toReferenceScriptWitnessFiles
+    :: (TxIn, AnyScriptLanguage, ScriptDatumOrFile witctx, ScriptRedeemerOrFile, ExecutionUnits)
+    -> ScriptWitnessFiles witctx
+  toReferenceScriptWitnessFiles (txin, sLang, d,r, e) = PlutusReferenceScriptWitnessFiles txin sLang d r e
+
 pScriptWitnessFiles :: forall witctx.
                        WitCtx witctx
                     -> BalanceTxExecUnits -- ^ Use the @execution-units@ flag.
-                    -> String
+                    -> String -- ^ Script flag prefix
                     -> Maybe String
                     -> String
                     -> Parser (ScriptWitnessFiles witctx)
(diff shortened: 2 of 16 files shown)


doc/reference/plutus/reference-script-example.md
```diff
@@ -0,0 +1,146 @@
...
+  --tx-out "$plutusscriptaddr+$lovelace" \
+  --tx-out-datum-hash "$scriptdatumhash" \
+  --tx-out "$dummyaddress+$lovelaceattxindiv3" \
+  --reference-script-file "$plutusscriptinuse" \
+  --protocol-params-file "$WORK/pparams.json" \
+  --out-file "$WORK/create-datum-output.body"
+```
+
+There are a couple things happening in the following `build` command.
+Firstly we are sending ADA to the plutus script address along with a datum. This is reflected in the following lines:
```
james-iohk, line 75: "datum hash"

doc/reference/plutus/reference-script-example.md
```diff
@@ -0,0 +1,146 @@
...
+```
+
+
+#### Spending ADA at the script address
+
+Now that there is ADA at our script address, we must construct the appropriate transaction in order to spend it.
+Because we are using the `build` command we only have to concern ourselves with the following:
+
+`$plutusutxotxin` - This is the tx input that sits at the Plutus script address (NB: It has a datum hash).
+`tx-in-reference` -This specifies the reference input you are using to witness a transaction input.
```
james-iohk, line 114: add space after hyphen

doc/reference/plutus/reference-script-example.md
```diff
@@ -0,0 +1,146 @@
+# Plutus Scripts
+
+## What is a reference script?
+
+A reference script is a script that exists at a particular transaction output. It can be used to witness, for example, a UTxO at the corresponding script address of said reference script. Why is this useful? It means we no longer have to include the script in the transaction, greatly reducing its size
```
james-iohk, line 5: end paragraph with `.`

doc/reference/plutus/reference-script-example.md
```diff
@@ -0,0 +1,146 @@
...
+
+
+#### Spending ADA at the script address
+
+Now that there is ADA at our script address, we must construct the appropriate transaction in order to spend it.
+Because we are using the `build` command we only have to concern ourselves with the following:
+
+`$plutusutxotxin` - This is the tx input that sits at the Plutus script address (NB: It has a datum hash).
+`tx-in-reference` -This specifies the reference input you are using to witness a transaction input.
+`plutus-script-v2`- This specifies the version of the reference script at the reference input. There are no restrictions on the type of script we can have at a reference input.
```
james-iohk, line 115: Right, but it's always invalid to execute a referenced V1 script. We should make this limitation clear.

james-iohk: When using `--plutus-script-v1` I always see `Command failed: transaction build  Error: Error translating the transaction context: ReferenceInputsNotSupported`, so should it be an option?

doc/reference/plutus/reference-script-example.md
```diff
@@ -0,0 +1,146 @@
...
+./scripts/babbage/mkfiles.sh
+```
+
+To start your babbage cluster you need to run the `example/run/all.sh` shell script.
+The remainder of this guide briefly talks about the [shell script example](scripts/plutus/example-reference-script-usage.sh) that automatically does everything for you.
+
+#### Creating a reference script at a transaction output and
+#### sending ADA to the script address (with a datum)
+
+In order to use a reference script we must first create said reference script at a particular address. Note you cannot create a reference script at the same correspnding script address, this will result in a failure when trying to spend the transaction input at the script address.
```
james-iohk, line 56: > Note you cannot create a reference script at the same correspnding script address, this will result in a failure when trying to spend the transaction input at the script address.
Not sure this is accurate. I have successfully created an output with inline script at its own script address and then referred to it to spend a different utxo at that same script address.

doc/reference/plutus/reference-script-example.md
olgahryniuk, line 5: Suggest changing to:
'This is useful because the script does not have to be included in the transaction anymore, which significantly reduces the transaction size.

doc/reference/plutus/reference-script-example.md
```diff
@@ -0,0 +1,146 @@
+# Plutus Scripts
+
+## What is a reference script?
+
+A reference script is a script that exists at a particular transaction output. It can be used to witness, for example, a UTxO at the corresponding script address of said reference script. Why is this useful? It means we no longer have to include the script in the transaction, greatly reducing its size
+
+### An example of using a Plutus V2 reference script
+
+Below is an example that shows how to use a Plutus spending script. We will walk though the [shell script example of how to use a reference script to spend a tx input](scripts/plutus/example-reference-script-usage.sh).This is a step-by-step process involving:
```
olgahryniuk, line 9: Suggest changing to:
'Here we discuss a [shell script example of how to use a reference script to spend a tx input](scripts/plutus/example-reference-script-usage.sh).'

doc/reference/plutus/reference-script-example.md
```diff
@@ -0,0 +1,146 @@
...
+
++ the creation of the `Required Redeemer` Plutus txin script
++ the creation of the `Required Redeemer` Plutus script at a transaction output (creation of the reference script)
++ sending ADA and a datum to the Plutus script address
```
olgahryniuk, line 13: In our style guide, we always use 'ada' instead of 'ADA'. Just a suggestion to change this, but up to you @Jimbo4350

doc/reference/plutus/reference-script-example.md
```diff
@@ -0,0 +1,146 @@
...
+
+Run the following command:
+
+```bash
+cabal install cardano-cli
+cabal install cardano-node
+./scripts/babbage/mkfiles.sh
+```
+
+To start your babbage cluster you need to run the `example/run/all.sh` shell script.
```
olgahryniuk, line 50: 'To start your Babbage cluster, you need......'

doc/reference/plutus/reference-script-example.md
```diff
@@ -0,0 +1,146 @@
...
+The remainder of this guide briefly talks about the [shell script example](scripts/plutus/example-reference-script-usage.sh) that automatically does everything for you.
```
olgahryniuk, line 51: 'Further, we briefly discuss the [shell script example](scripts/plutus/example-reference-script-usage.sh), which automates your following actions'.
@Jimbo4350 Please see if this makes sense^. If not, I'd suggest changing 'automatically does everything for you' into a clearer statement. Like what exactly does it automate?

doc/reference/plutus/reference-script-example.md
olgahryniuk, line 56: 'In order to use a reference script, we must first create this script at a particular address. Note that you cannot create a reference script at the same corresponding script address as this will result in a failure when trying to spend the transaction input at the script address.'

doc/reference/plutus/reference-script-example.md
```diff
@@ -0,0 +1,146 @@
...
+  --tx-out "$utxoaddr+$lovelace" \
+  --tx-out "$plutusscriptaddr+$lovelace" \
+  --tx-out-datum-hash "$scriptdatumhash" \
+  --tx-out "$dummyaddress+$lovelaceattxindiv3" \
+  --reference-script-file "$plutusscriptinuse" \
+  --protocol-params-file "$WORK/pparams.json" \
+  --out-file "$WORK/create-datum-output.body"
+```
+
+There are a couple things happening in the following `build` command.
```
olgahryniuk, line 74: How about:
'The following should be noted about this `build` command:

doc/reference/plutus/reference-script-example.md
olgahryniuk, line 75: Firstly, we are sending ADA to the Plutus script address along with a datum. This is reflected in the following lines:

doc/reference/plutus/reference-script-example.md
```diff
@@ -0,0 +1,146 @@
...
+```bash
+...
+--tx-out "$plutusscriptaddr+$lovelace" \
+--tx-out-datum-hash "$scriptdatumhash" \
+...
+```
+
+We have seen this before in the [plutus-spending-script-example.md](doc/reference/plutus/plutus-spending-script-example.md).
+
+Secondly we are creating a reference script at an address, which cannot be the same as the reference script address, of our choosing in the following lines:
```
olgahryniuk, line 86: Secondly, we are creating a reference script at an address, which cannot be the same as the reference script address of our choice:

doc/reference/plutus/reference-script-example.md
olgahryniuk: I'd also suggest bullet-pointing these items, and starting an explanation with a regular-case letter.

doc/reference/plutus/reference-script-example.md
```diff
@@ -0,0 +1,146 @@
...
+  --protocol-params-file "$WORK/pparams.json"
+
+cardano-cli transaction sign \
+  --tx-body-file $WORK/test-alonzo-ref-script.body \
+  --testnet-magic "$TESTNET_MAGIC" \
+  --signing-key-file "${UTXO_SKEY}" \
+  --out-file $WORK/alonzo-ref-script.tx
+```
+
+If there is ADA at `$dummyaddress2` then the Plutus script was successfully executed. Conversely, if the Plutus script failed, the collateral input would have been consumed.
```
olgahryniuk, line 143: Add a comma before 'then'

MarcFontaine (approved):
LGTM for the tx-generator

doc/reference/plutus/reference-script-example.md
Jimbo4350: I've also pinged @JaredCorduan. It was my understanding that this was not possible.

doc/reference/plutus/reference-script-example.md
Jimbo4350: Right so there is some confusion here. It was my understanding that we can use V1 scripts as reference scripts to witness the spending of a particular tx input. Is this the case @JaredCorduan ?

Jimbo4350: If it is, we shouldn't get `ReferenceInputsNotSupported` when using a V1 reference script to witness a tx input

JaredCorduan: > 



> Right so there is some confusion here. It was my understanding that we can use V1 scripts as reference scripts to witness the spending of a particular tx input. Is this the case @JaredCorduan ?

it is not the case, the context translation will fail.

doc/reference/plutus/reference-script-example.md
JaredCorduan: I'll need to break this down a bit before I'm confident I understand. I don't see how the prose and @james-iohk 's comments are in conflict.

JaredCorduan: we had a video chat about this, what @james-iohk observed is expected, abeit not a normal workflow we'd expect.

Jimbo4350: thanks for catching this :+1:

james-iohk:
Consider renaming `--reference-script-file` to follow convention of other tx-out flags. For example, it could be changed to `--tx-out-reference-script-file`

doc/reference/plutus/reference-script-example.md
```diff
@@ -0,0 +1,148 @@
...
+```bash
+...
+--tx-out "$plutusscriptaddr+$lovelace" \
+--tx-out-datum-hash "$scriptdatumhash" \
+...
+```
+
+We have seen this before in the [plutus-spending-script-example.md](doc/reference/plutus/plutus-spending-script-example.md).
+
+Secondly, we are creating a reference script at a tx output, which cannot be the same as the reference script address of our choice:
```
james-iohk, line 88: > which cannot be the same as the reference script address of our choice`

It can. We could say "there is no restriction on which address we choose." or "a reference script can be guarded by any type of address and there's no restriction on it being referenced."

doc/reference/plutus/reference-script-example.md
```diff
@@ -0,0 +1,148 @@
...
+
+
+#### Spending ada at the script address
+
+Now that there is ada at our script address, we must construct the appropriate transaction in order to spend it.
+Because we are using the `build` command, we should only note the following:
+
+`$plutusutxotxin` - This is the tx input that sits at the Plutus script address (NB: It has a datum hash).
+`tx-in-reference` - This specifies the reference input you are using to witness a transaction input.
+`plutus-script-v2`- This specifies the version of the reference script at the reference input. There are no restrictions on the type of script we can have at a reference input.
```
james-iohk, line 117: > There are no restrictions on the type of script we can have at a reference input.

Did we want to remove this altogether, or explain that V1 scripts can be referenced but not executed?

james-iohk:
If we're going to leave `--plutus-script-v1` flag in then perhaps add to its help text `Specify a plutus script v1 reference script. Note: Ledger rules prevent referenced V1 scripts from being executed`

james-iohk:
```
--simple-script-v1       Specify a simple script v1 reference script.
--simple-script-v2       Specify a simple script v2 reference script.
```
@Jimbo4350 Do these refer to the Shelley(v1) and Allegra(v2) native scripts?  If so it may be worth clarifying that in the help text because it wasn't obvious to me.

doc/reference/plutus/reference-script-example.md
Jimbo4350: Yeah let's remove that. I will add a separate cli parser for "stand alone" reference inputs that don't witness anything

doc/reference/plutus/reference-script-example.md
Jimbo4350: I'll just remove it :+1:

james-iohk (approved):

cardano-api/src/Cardano/Api/Script.hs
```diff
@@ -953,7 +966,7 @@ fromShelleyScriptHash = ScriptHash
 
 
 -- ----------------------------------------------------------------------------
--- The simple native script language
+-- The simple simple script language
```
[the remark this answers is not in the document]
Jimbo4350: It's _super_ simple ("simple simple") . Thanks for catching this.

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -229,10 +229,56 @@ pScriptFor name (Just deprecated) help =
...
+      <> Opt.metavar "TX-IN"
+      <> Opt.help "TxId#TxIx - Specify a reference input. The reference input may or may not have\
+                  \ a plutus reference script attached."
+      )
+  pAnyScriptLang :: Parser AnyScriptLanguage
+  pAnyScriptLang =
+    Opt.flag' (AnyScriptLanguage $ SimpleScriptLanguage SimpleScriptV1)
+      (  Opt.long "simple-script-v1"
+      <> Opt.help "Specify a simple script v1 reference script. \
+                  \See documentation at doc/reference/simple-scripts.md"
```
james-iohk, line 260: This document doesn't refer to v1 and v2. How can we make that distinction clear?

james-iohk: `Specify a simple script v1 (Shelley) reference script`
`Specify a simple script v2 (Allegra) reference script.`
?

cardano-cli/src/Cardano/CLI/Shelley/Run/Transaction.hs
```diff
@@ -509,6 +517,8 @@ runTxBuild (AnyCardanoEra era) (AnyConsensusModeParams cModeParams) networkId mS
         newExceptT . fmap (join . first ShelleyTxCmdAcquireFailure) $
           executeLocalStateQueryExpr localNodeConnInfo Nothing $ \_ntcVersion -> runExceptT $ do
             unless (null txinsc) $ do
+              -- TODO: Question, why do we not need the collateralUtxo to be included in
+              -- the utxo?
```
JaredCorduan, line 521: I'm confused by this as well, the ledger rules do require that the collateral utxo are legitimate.

doc/reference/plutus/reference-script-example.md
```diff
@@ -0,0 +1,148 @@
...
+Below is an example that shows how to use a Plutus spending script. Here we discuss a [shell script example of how to use a reference script to spend a tx input](scripts/plutus/example-reference-script-usage.sh). This is a step-by-step process involving:
+
++ the creation of the `Required Redeemer` Plutus txin script
++ the creation of the `Required Redeemer` Plutus script at a transaction output (creation of the reference script)
++ sending ada and a datum to the Plutus script address
++ spending ada at the Plutus script address using the Plutus reference script
+
+In this example we will use the [Required Redeemer](scripts/plutus/scripts/v2/required-redeemer.plutus) Plutus spending script. In order to execute a reference Plutus spending script, we require the following:
+
+- Collateral tx input(s) - these are provided and are forfeited in the event the Plutus script fails to execute.
```
JaredCorduan, line 18: Should we ease folks mind by telling them about the fact that a local node will not forward on a failing tx to the relay node unless it is explicitly told to? As you know, I'm sure, the collateral exists to prevent abusive behavior, and under normal circumstances will not be taken.

scripts/babbage/mkfiles.sh
```diff
@@ -151,7 +151,7 @@ $SED -i "${ROOT}/genesis/shelley/genesis.json" \
     -e 's/"minFeeB": 0/"minFeeB": 155381/' \
     -e 's/"minUTxOValue": 0/"minUTxOValue": 1000000/' \
     -e 's/"decentralisationParam": 1.0/"decentralisationParam": 0.7/' \
-    -e 's/"major": 0/"major": 2/' \
+    -e 's/"major": 0/"major": 7/' \
```
JaredCorduan, line 154: :tada:

JaredCorduan (approved):
I made a few minor comments, but overall this is awesome! This was a huge undertaking, thanks @Jimbo4350 !!

cardano-cli/src/Cardano/CLI/Shelley/Run/Transaction.hs
Jimbo4350: Yep I'm not sure why we don't have to include them in the `utxo`. But I will diagnose that some other time.

PR #3965, opened 2022-06-01
Newhoggy/fix leadership schedule command 2

newhoggy:


cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -3134,10 +3135,11 @@ makeShelleyTransactionBody era@ShelleyBasedEraAlonzo
              BuildTxWith Nothing        -> SNothing
              BuildTxWith (Just pparams) ->
                Alonzo.hashScriptIntegrity
-                 (toLedgerPParams ShelleyBasedEraAlonzo pparams)
+                --  (toLedgerPParams ShelleyBasedEraAlonzo pparams)
                  languages
```
newhoggy, line 3139: What do I need to do here?

```
src/Cardano/Api/TxBody.hs:3139:18: error:
    • Couldn't match type ‘Alonzo.Language’ with ‘Babbage.LangDepView’
      Expected type: Set Babbage.LangDepView
        Actual type: Set Alonzo.Language
    • In the first argument of ‘Alonzo.hashScriptIntegrity’, namely
        ‘languages’
      In the expression:
        Alonzo.hashScriptIntegrity languages redeemers datums
      In a case alternative:
          BuildTxWith (Just pparams)
            -> Alonzo.hashScriptIntegrity languages redeemers datums
     |
3139 |                  languages
     |                  ^^^^^^^^^
```

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -3134,10 +3135,11 @@ makeShelleyTransactionBody era@ShelleyBasedEraAlonzo
              BuildTxWith Nothing        -> SNothing
              BuildTxWith (Just pparams) ->
                Alonzo.hashScriptIntegrity
-                 (toLedgerPParams ShelleyBasedEraAlonzo pparams)
```
newhoggy, line 3137: Do I not need to use `pparams` anymore?

PR #3983, opened 2022-06-02
Fix windows builds in GitHub actions

newhoggy:


.github/workflows/cabal.project.local.Windows
```diff
@@ -1,8 +1,7 @@
...
   flags: -external-libsodium-vrf
 
-write-ghc-environment-files: always
-
-ignore-project: False
+package HsOpenSSL
+  flags: +use-pkg-config
 
-reorder-goals: False
+write-ghc-environment-files: always
```
Jimbo4350, line 7: I thought we never want to do this?

Jimbo4350 (approved):

.github/workflows/cabal.project.local.Windows
newhoggy: Removed.

PR #3993, opened 2022-06-03
PLT-106: Add encoder and decoder for `LedgerState`

zliu41:
One use case is for `foldBlocks` to checkpoint the ledger state to avoid having to start from Genisys. cc @JaredCorduan

### DIFF
--- a/cardano-api/src/Cardano/Api/LedgerState.hs
+++ b/cardano-api/src/Cardano/Api/LedgerState.hs
@@ -24,6 +24,8 @@ module Cardano.Api.LedgerState
       , LedgerStateMary
       , LedgerStateAlonzo
       )
+  , encodeLedgerState
+  , decodeLedgerState
   , initialLedgerState
   , applyBlock
   , ValidationMode(..)
@@ -52,6 +54,7 @@ module Cardano.Api.LedgerState
   )
   where
 
+import qualified Cardano.Binary as CBOR
 import           Control.Exception
 import           Control.Monad (when)
 import           Control.Monad.Trans.Class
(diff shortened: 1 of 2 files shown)


JaredCorduan:
It would be absolutely wonderful to have `foldBlocks` be able to checkpoint the ledger state! I'm not certain that such encodings don't already exist somewhere else in this repoo, but @Jimbo4350 probably knows.

zliu41:
Updated!

JaredCorduan (approved):
LGTM!

cardano-api/src/Cardano/Api/LedgerState/Serialisation.hs
```diff
@@ -0,0 +1,46 @@
+{-# LANGUAGE ExplicitNamespaces #-}
```
Jimbo4350, line 1: Does this really deserve its own module?

Jimbo4350 (changes requested):

cardano-api/src/Cardano/Api/LedgerState/Serialisation.hs
zliu41: I can move it to `Cardano.Api.LedgerState`. In the original version the serializer was much more complex, but it has since been simplified, so there's less reason for a separate module.

zliu41: @Jimbo4350 Done.

zliu41:
bors merge

zliu41:
bors try

JaredCorduan:
@zliu41 I think @Jimbo4350 's lock is still on this PR...

zliu41:
> @zliu41 I think @Jimbo4350 's lock is still on this PR...

Yep! I'm working on merging the other PR first, and will come back to this one.

cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -52,6 +54,8 @@ module Cardano.Api.LedgerState
   )
   where
 
+import qualified Codec.CBOR.Decoding as CBOR
```
Jimbo4350, line 57: We should use: `import qualified Cardano.Binary as CBOR`

Jimbo4350 (approved):

PR #4000, opened 2022-06-06
Further explain the `libsodium` installation

Jasagredo:
As discussed with @iquerejeta, using the `-external-libsodium-vrf` flag should not be a problem.

Also `cabal configure` would erase the `cabal.project.local` file as found out by @angerman in [a different issue so](https://github.com/input-output-hk/ouroboros-network/pull/3765) I reverted the instructions to appending with `echo`.


iquerejeta (approved):
LGTM 👍 

The sentence `and adds some others new functions` sounds a bit strange to me though. Maybe `and adds some other new functions`? Otherwise, description is clear.

PR #4004, opened 2022-06-06
Remove David and Nick from CODEOWNERS

Jimbo4350:


JaredCorduan (approved):
I looks like there are some others problems with the CODEOWNERS file (I like this new feature that GitHub tells you this!). But these are unrelated to your changes.

PR #4011, opened 2022-06-07
CAD-3915 define the `BackingStoreSelector` flag

Jasagredo:
Defines new flags:
```
  --in-memory-ledger-db-backend
                           Use the InMemory ledger DB backend. Incompatible with
                           `--lmdb-ledger-db-backend`.
  --lmdb-ledger-db-backend MAPSIZE
                           Use the LMDB ledger DB backend. The mapsize argument
                           must be a number followed by a unit, for example 10Gi
                           for 10 Gibibytes. Incompatible with
                           `--in-memory-ledger-db-backend`.
```
and new options for the JSON config:
```
LedgerDBBackend: (LMDB|InMemory)
LMDBMapSize: <N><Unit> -- for example 12Gi
```

Needs [this PR in ouroboros-network](https://github.com/input-output-hk/ouroboros-network/pull/3795)

### DIFF
--- a/cardano-node/src/Cardano/Node/Configuration/POM.hs
+++ b/cardano-node/src/Cardano/Node/Configuration/POM.hs
@@ -30,6 +30,7 @@ import           Data.Aeson
 import qualified Data.Aeson.Types as Aeson
 import qualified Data.Text as Text
 import           Data.Time.Clock (DiffTime)
+import           Data.Prefix.Units
 import           Data.Yaml (decodeFileThrow)
 import           Generic.Data (gmappend)
 import           Generic.Data.Orphans ()
@@ -38,6 +39,7 @@ import           System.FilePath (takeDirectory, (</>))
 
 import qualified Cardano.Chain.Update as Byron
 import           Cardano.Crypto (RequiresNetworkMagic (..))
+import           Cardano.Node.Configuration.LedgerDB
 import           Cardano.Node.Configuration.Socket (SocketConfig (..))
 import           Cardano.Node.Handlers.Shutdown
 import           Cardano.Node.Protocol.Types (Protocol (..))
@@ -116,6 +118,9 @@ data NodeConfiguration
 
        , ncMaybeMempoolCapacityOverride :: !(Maybe MempoolCapacityBytesOverride)
 
+         -- UTxO-HD configuration
+       , ncLedgerDBBackend :: !BackingStoreSelectorFlag
+
          -- | Protocol idleness timeout, see
          -- 'Ouroboros.Network.Diffusion.daProtocolIdleTimeout'.
          --
@@ -171,6 +176,9 @@ data PartialNodeConfiguration
 
          -- Configuration for testing purposes
        , pncMaybeMempoolCapacityOverride :: !(Last MempoolCapacityBytesOverride)
+
+         -- UTxO-HD configuration
+       , pncLedgerDBBackend :: !(Last BackingStoreSelectorFlag)
 
          -- Network timeouts
        , pncProtocolIdleTimeout   :: !(Last DiffTime)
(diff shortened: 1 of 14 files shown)


cardano-node/src/Cardano/Node/Configuration/POM.hs
```diff
@@ -48,6 +48,7 @@ import           Ouroboros.Consensus.Mempool.API (MempoolCapacityBytes (..),
 import qualified Ouroboros.Consensus.Node as Consensus (NetworkP2PMode (..))
 import           Ouroboros.Consensus.Storage.LedgerDB.DiskPolicy (SnapshotInterval (..))
 import           Ouroboros.Network.NodeToNode (AcceptedConnectionsLimit (..), DiffusionMode (..))
+import Ouroboros.Consensus.Storage.LedgerDB.OnDisk (BackingStoreSelectorFlag(..))
```
jorisdral, line 51: Alignment

cardano-node/src/Cardano/Node/Configuration/POM.hs
```diff
@@ -116,6 +117,7 @@ data NodeConfiguration
 
        , ncMaybeMempoolCapacityOverride :: !(Maybe MempoolCapacityBytesOverride)
 
+       , ncLedgerDBBackend :: !BackingStoreSelectorFlag
```
jorisdral, line 120:
```suggestion
       , ncLedgerDBBackend :: !BackingStoreSelectorFlag
       
```
And maybe a comment to describe this section of the configuration? Something like "UTxO HD configs"

cardano-node/src/Cardano/Node/Configuration/POM.hs
```diff
@@ -172,6 +174,7 @@ data PartialNodeConfiguration
          -- Configuration for testing purposes
        , pncMaybeMempoolCapacityOverride :: !(Last MempoolCapacityBytesOverride)
 
+       , pncLedgerDBBackend :: !(Last BackingStoreSelectorFlag)
```
jorisdral, line 177: Ditto previous comment

cardano-node/src/Cardano/Node/Configuration/POM.hs
```diff
@@ -306,6 +312,13 @@ instance FromJSON PartialNodeConfiguration where
                     "Invalid value for 'MempoolCapacityBytesOverride'.  \
                     \Expecting byte count or NoOverride.  Value was: " <> show invalid
               Nothing -> return Nothing
+
+      parseLedgerDBBackend v = do
+        maybeString :: Maybe String <- v .:? "LedgerDBBackend"
+        case maybeString of
+           Just "MEM" -> return $ Just InMemory
+           _          -> return $ Just LMDB
```
jorisdral, line 320: Maybe an ill-formed flag should fail instead of default to `LMDB`

cardano-node/src/Cardano/Node/Configuration/POM.hs
```diff
@@ -428,6 +441,7 @@ defaultPartialNodeConfiguration =
     , pncLogMetrics = mempty
     , pncTraceConfig = mempty
     , pncMaybeMempoolCapacityOverride = mempty
+    , pncLedgerDBBackend = Last . Just $ LMDB
```
jorisdral, line 444: Is `LMDB` the default?

cardano-node/src/Cardano/Node/Configuration/POM.hs
Jasagredo: Fixed, thanks!

cardano-node/src/Cardano/Node/Configuration/POM.hs
Jasagredo: Fixed, thanks!

cardano-node/src/Cardano/Node/Configuration/POM.hs
Jasagredo: Fixed, thanks!

cardano-node/src/Cardano/Node/Configuration/POM.hs
Jasagredo: This is now the behavior. I think now it is more ergonomic

cardano-node/src/Cardano/Node/Configuration/POM.hs
Jasagredo: I think so (for normal users it should be LMDB), we can always revert it back

PR #4018, opened 2022-06-08
Set local encoding to UTF-8 in cardano-cli

newhoggy:


disassembler (approved):
tested and works without setting LANG

JaredCorduan (approved):
Thanks @newhoggy !

dcoutts (approved):

PR #4021, opened 2022-06-09
Reduce memory usage of create staked command

newhoggy:
The problem arises we build a very large in-memory JSON value and then write it out as `genesis.json`.  The amount of memory used by the in-memory JSON value can be so large as to use up all the memory.

This PR reduces the severity of the memory usage in two ways.

1. It introduces a `ListMap` type, which is almost like `Map`, but has a `[(k, v)]` as its internal representation.  This avoids the serialisation cost of constructing a map only to convert it back into a list.

2. Uses Lazy IO so that generated stuffed utxos are created on demand rather than upfront an all in memory.

3. Introduces a new `LazyToJson` type class which doesn't have the memory retention problems of `aeson` library.

4. For evaluation of fields to `WHNF` to allow parent object to be GCed which allows large fields that have already be serialised to be collected as well.

5. Writes delegations to a single `delegations.jsonl` file which is a newline delimited JSON file.  This file is streamed multiple times so that generation of the `genesis.json` file does not retain memory unnecessarily.

This PR also changes the command to no longer generate payment keys and stake keys.  If we want to have the ability optionally output these files, there is additional work to do.

Addresses https://github.com/input-output-hk/cardano-node/issues/3938


cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis/ListMap.hs
```diff
@@ -0,0 +1,47 @@
...
+
+import qualified Data.Aeson as J
+import qualified Data.Aeson.Encoding as E
+import qualified Data.Aeson.KeyMap as KM
+import qualified Data.List as L
+import qualified Data.Vector as V
+
+newtype ListMap k v = ListMap
+  { unListMap :: [(k, v)]
+  } deriving (Eq, Show)
```
newhoggy, line 21: `ListMap` is just a list of pairs.

newhoggy:
Example run:

```
$ time dist-newstyle/build/aarch64-osx/ghc-8.10.7/cardano-cli-1.33.0/x/cardano-cli/build/cardano-cli/cardano-cli genesis create-staked --genesis-dir example --supply 10000000000000 --gen-utxo-keys 1 --gen-genesis-keys 0 --supply-delegated 2000000000000000 --gen-pools 2 --gen-stake-delegs 1300000 --testnet-magic 42 --num-stuffed-utxo 8000000
generated genesis with: 0 genesis keys, 1 non-delegating UTxO keys, 2 stake pools, 1300000 delegating UTxO keys, 1300000 delegation map entries,
 genesis create-staked --genesis-dir example --supply 10000000000000  1  0     441.74s user 321.01s system 99% cpu 12:47.91 total
```

newhoggy:
Slight improvement and uses about 1GB of memory:

```
$ time dist-newstyle/build/aarch64-osx/ghc-8.10.7/cardano-cli-1.33.0/x/cardano-cli/build/cardano-cli/cardano-cli genesis create-staked --genesis-dir example --supply 10000000000000 --gen-utxo-keys 1 --gen-genesis-keys 0 --supply-delegated 2000000000000000 --gen-pools 2 --gen-stake-delegs 1300000 --testnet-magic 42 --num-stuffed-utxo 8000000
generated genesis with: 0 genesis keys, 1 non-delegating UTxO keys, 2 stake pools, 1300000 delegating UTxO keys, 1300000 delegation map entries,
 genesis create-staked --genesis-dir example --supply 10000000000000  1  0     133.75s user 290.89s system 98% cpu 7:12.31 total
```

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
```diff
@@ -101,6 +103,14 @@ import           Data.Fixed (Fixed(MkFixed))
 import qualified Data.Yaml as Yaml
 import           Text.JSON.Canonical (parseCanonicalJSON, renderCanonicalJSON)
 
+import           Cardano.CLI.Shelley.Run.Genesis.ListMap (ListMap(..))
+import           Cardano.CLI.Shelley.Run.Genesis.LazyToJson (LazyToJson(lazyToJson))
+import           Cardano.CLI.Shelley.Run.Genesis.Types (toOutputTemplate)
+
+import qualified Cardano.CLI.Shelley.Run.Genesis.Types as OT
+import qualified Data.ByteString.Builder as B
```
Jimbo4350, line 111: `Data.ByteString.Builder` should be with the other `Data.*` modules above

Jimbo4350 (changes requested):

Jimbo4350:
I need to make a second pass

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
```diff
@@ -683,38 +693,43 @@ runGenesisCreateStaked (GenesisDir rootdir)
...
-         , let delegIx = delegIxLocal + delegsPerPool * (poolIx - 1)] $
-      uncurry (computeDelegation network stdeldir)
+  let (delegsPerPool, delegsRemaining) = divMod genNumStDelegs genNumPools
+      delegsForPool poolIx = if delegsRemaining /= 0 && poolIx == genNumPools
+        then delegsPerPool
+        else delegsPerPool + delegsRemaining
+      distribution = [pool | (pool, poolIx) <- zip pools [1 ..], _ <- [1 .. delegsForPool poolIx]]
+
+  -- Distribute M delegates across N pools:
+  delegations <- liftIO $ Lazy.forM distribution $ computeDelegation network
```
Jimbo4350, line 703: I would say lets just implement out own version of `forM` and `replicateM` in the cli and remove the `hw-lazy` dependency

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
```diff
@@ -683,38 +693,43 @@ runGenesisCreateStaked (GenesisDir rootdir)
...
+
+  -- NOTE The following code which reads from the same file 'delegations.jsonl' multiple times
+  -- looks like duplication, but it is not.  The file is read lazily, and it is important that
+  -- they be read multiple times because the code is streaming and reading the file multiple
+  -- times ensures that any data structures that are created as a result of the read is not
+  -- retained in memory.
+
+  let numDelegations = length delegations
```
Jimbo4350, line 711: This is suspicious, calling `length` could destroy the laziness. We could get `numDelegations` in this list comprehension: `distribution = [pool | (pool, poolIx) <- zip pools [1 ..], _ <- [1 .. delegsForPool poolIx]]` if we multiply the number of pools by the number of delegations per pool

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
```diff
@@ -924,31 +933,22 @@ writeBulkPoolCredentials dir bulkIx poolIxs = do
...
-    initialUtxoAddr <- case paySVK of
-      APaymentVerificationKey payVK ->
-        firstExceptT ShelleyGenesisCmdAddressCmdError
-        $ buildShelleyAddress payVK (Just . StakeVerifierKey . VerificationKeyFilePath . VerificationKeyFile $ stakeVKF) nw
-      _ -> left $ ShelleyGenesisCmdUnexpectedAddressVerificationKey payVKF "APaymentVerificationKey" paySVK
-
-    StakeVerificationKey stakeVK <- firstExceptT ShelleyGenesisCmdTextEnvReadFileError
-      . newExceptT
-      $ readFileTextEnvelope (AsVerificationKey AsStakeKey) stakeVKF
+computeDelegation :: ()
```
Jimbo4350, line 936: So in this function we are generating two keys but all we are interested in are their hashes. We should create a new primitive in the cli that creates _non-cryptographically secure_  key hashes for testing purposes like this. 

We can use `System.Random` for the entropy and `hashFromBytes` (in `cardano-base`) to generate dummy key hashes. This should speed things up a lot as `generateSigningKey` is a more expensive function.

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis/ListMap.hs
```diff
@@ -0,0 +1,56 @@
...
+import Text.Show (Show(..))
+
+import qualified Data.Aeson as J
+import qualified Data.Aeson.Encoding as E
+import qualified Data.Aeson.KeyMap as KM
+import qualified Data.ByteString.Builder as B
+import qualified Data.List as L
+import qualified Data.Vector as V
+
+newtype ListMap k v = ListMap
```
Jimbo4350, line 27: We should be able to generically derive the JSON instances for `ListMap` and it should do the right thing. @newhoggy can you try this?

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis/Types.hs
```diff
@@ -0,0 +1,185 @@
...
+  , sgActiveSlotsCoeff :: !PositiveUnitInterval
+  , sgSecurityParam :: !Word64
+  , sgEpochLength :: !EpochSize
+  , sgSlotsPerKESPeriod :: !Word64
+  , sgMaxKESEvolutions :: !Word64
+  , sgSlotLength :: !NominalDiffTime
+  , sgUpdateQuorum :: !Word64
+  , sgMaxLovelaceSupply :: !Word64
+  , sgProtocolParams :: !(PParams era)
+  , sgGenDelegs :: !(ListMap (KeyHash 'Genesis (Crypto era)) (GenDelegPair (Crypto era)))
```
Jimbo4350, line 60: The real solution would be to modify the `ShelleyGenesis` datatype in the ledger i.e genDelegs and initialFunds. Changing them from `Map`s to `[(*,*)]`. This should not be a problem, can you talk to @JaredCorduan about doing this? We can then avoid defining the `ListMap` type and this much (if not all) of this module.

Jimbo4350 (changes requested):
Latest review comments courtesy of @dcoutts .

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
newhoggy: Yes, calling `length` destroys laziness, but I've been in discussion with @kosyrevSerge and he is happy to pay the memory cost in this case.

Note, `length` is not the only thing that "destroys laziness".  Using `delegations` more than once does the same thing.  Previously, `delegations` was written to a temporary file and the temporarily file was read multiple times to preserve laziness, but @kosyrevSerge didn't want to pay the serialisation cost of that.

newhoggy: The comment above is out of date however and I've deleted it.

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis/Types.hs
newhoggy: Are you proposing to change the format of the JSON file to use `[["key1", "value1"], ["key2", "value2]]` instead of `{"field1": "value1", "field2", "value2"}`?

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis/ListMap.hs
newhoggy: Note, generically deriving JSON would do one of two things:

`deriving newtype` would change the format of the JSON file to use `[["key1", "value1"], ["key2", "value2"]]` instead of `{"field1": "value1", "field2": "value2"}`.

`DeriveGeneric` would additionally add `{ "unListMap": ... }` wrapper.

cardano-cli/src/Cardano/CLI/Shelley/Run/Address.hs
```diff
@@ -57,37 +58,54 @@ renderShelleyAddressCmdError err =
...
+generateAndWriteKeyFiles :: ()
+  => Key keyrole
+  => AsType keyrole
+  -> VerificationKeyFile
+  -> SigningKeyFile
+  -> ExceptT ShelleyAddressCmdError IO ()
+generateAndWriteKeyFiles asType vkf skf = do
+  uncurry (writePaymentKeyFiles vkf skf) =<< generatePaymentKeys asType
+
+generatePaymentKeys :: ()
```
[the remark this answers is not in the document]
newhoggy: This would create a cyclic dependency because `ShelleyAddressCmdError` is defined in `Cardano.CLI.Shelley.Run.Address` which already imports `Cardano.CLI.Shelley.Key`

The only function I can move is `generatePaymentKeys` because the `ExceptT ShelleyAddressCmdError` wasn't actually necessary for this function.

This has been moved.

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
newhoggy: Moved

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
```diff
@@ -1069,14 +1069,87 @@ updateTemplate (SystemStart start)
     unLovelace :: Integral a => Lovelace -> a
     unLovelace (Lovelace coin) = fromIntegral coin
 
+updateOutputTemplate
```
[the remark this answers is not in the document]
newhoggy: Renamed.

cardano-cli/src/Cardano/CLI/Shelley/Run/Address.hs
```diff
@@ -57,37 +58,54 @@ renderShelleyAddressCmdError err =
...
-    case kt of
-      AddressKeyShelley         -> generateAndWriteKeyFiles AsPaymentKey
-      AddressKeyShelleyExtended -> generateAndWriteKeyFiles AsPaymentExtendedKey
-      AddressKeyByron           -> generateAndWriteKeyFiles AsByronKey
+runAddressKeyGenToFile kt vkf skf = case kt of
+  AddressKeyShelley         -> generateAndWriteKeyFiles AsPaymentKey          vkf skf
+  AddressKeyShelleyExtended -> generateAndWriteKeyFiles AsPaymentExtendedKey  vkf skf
+  AddressKeyByron           -> generateAndWriteKeyFiles AsByronKey            vkf skf
+
+generateAndWriteKeyFiles :: ()
```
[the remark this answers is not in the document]
newhoggy: Removed

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
newhoggy: `cardano-node` doesn't seem the right place to put them given how generic they are.  Can  I put them into `cardano-base`?

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
```diff
@@ -563,7 +573,7 @@ runGenesisCreateCardano (GenesisDir rootdir)
           hashShelleyGenesis genesis = Crypto.hashToTextAsHex gh
             where
               content :: ByteString
-              content = LBS.toStrict $ encodePretty genesis
+              content = LBS.toStrict $ encode genesis
```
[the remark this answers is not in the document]
newhoggy: Reverted.

cardano-cli/src/Cardano/CLI/Shelley/Run/StakeAddress.hs
```diff
@@ -53,20 +53,23 @@ runStakeAddressCmd (StakeCredentialDeRegistrationCert stakeVerifier outputFp) =
...
-      . newExceptT
-      $ writeFileTextEnvelope skFp (Just skeyDesc) skey
-    firstExceptT ShelleyStakeAddressCmdWriteFileError
-      . newExceptT
-      $ writeFileTextEnvelope vkFp (Just vkeyDesc) vkey
-  where
-    skeyDesc, vkeyDesc :: TextEnvelopeDescr
-    skeyDesc = "Stake Signing Key"
-    vkeyDesc = "Stake Verification Key"
+runStakeAddressKeyGenToFile :: ()
```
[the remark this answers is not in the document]
newhoggy: Reverted.

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
```diff
@@ -101,6 +103,14 @@ import           Data.Fixed (Fixed(MkFixed))
 import qualified Data.Yaml as Yaml
 import           Text.JSON.Canonical (parseCanonicalJSON, renderCanonicalJSON)
 
+import           Cardano.CLI.Shelley.Run.Genesis.ListMap (ListMap(..))
+import           Cardano.CLI.Shelley.Run.Genesis.LazyToJson (LazyToJson(lazyToJson))
+import           Cardano.CLI.Shelley.Run.Genesis.Types (toOutputTemplate)
+
+import qualified Cardano.CLI.Shelley.Run.Genesis.Types as OT
```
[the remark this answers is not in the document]
newhoggy: Moved.

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis/Types.hs
```diff
@@ -0,0 +1,181 @@
...
+    }
+  }
+
+instance
+  ( Ledger.Crypto (Crypto era)
+  ) => ToJSON (OutputShelleyGenesis era) where
+  toJSON sg =
+    -- Forced evaluation of each field allows the parent object to no longer be
+    -- referenced, which helps prevent retention of any fields that have already
+    -- been serialised.
```
newhoggy, line 92: Bangs!

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
newhoggy: A new `generateInsecureSigningKey` function has been introduced.

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis/ListMap.hs
newhoggy: Looks like we're settling on changing the ledger code, so this code will disappear once the ledger has been updated.

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis/Types.hs
lehins: My guess would be that `sgGenDelegs` does not need to be a lazy `ListMap`, since that field always contains a very small `Map`. The other two fields below however can be safely switched to `ListMap`, since ledger doesn't even use them.

newhoggy: Using whatever is defined in `cardano-ledger` now.

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis/Types.hs
newhoggy: This code has been moved to `cardano-ledger`.

deepfire (approved):
Thank you a lot for this work @newhoggy !

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
```diff
@@ -1069,14 +1047,89 @@ updateTemplate (SystemStart start)
...
+    -- Non-delegated initial UTxO spec:
+    -> Maybe Lovelace
+    -> Int
+    -> [AddressInEra ShelleyEra]
+    -- Genesis staking: pools/delegation map & delegated initial UTxO spec:
+    -> [(Ledger.KeyHash 'Ledger.StakePool StandardCrypto, Ledger.PoolParams StandardCrypto)]
+    -> [(Ledger.KeyHash 'Ledger.Staking StandardCrypto, Ledger.KeyHash 'Ledger.StakePool StandardCrypto)]
+    -> Lovelace
+    -> Int
+    -> [AddressInEra ShelleyEra] --
```
lehins, line 1063:
```suggestion
    -> [AddressInEra ShelleyEra]
```

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
```diff
@@ -1069,14 +1047,89 @@ updateTemplate (SystemStart start)
     unLovelace :: Integral a => Lovelace -> a
     unLovelace (Lovelace coin) = fromIntegral coin
 
+updateCreateStakedOutputTemplate
+    :: SystemStart
+    -- Genesis delegation (not stake-based):
```
lehins, line 1052: I think it could be good to make those comments into haddock.
```suggestion
    -- | Genesis delegation (not stake-based)
```

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
```diff
@@ -1011,19 +989,19 @@ updateTemplate (SystemStart start)
         shelleyGenesis = template
           { sgSystemStart = start
           , sgMaxLovelaceSupply = fromIntegral $ nonDelegCoin + delegCoin
-          , sgGenDelegs = shelleyDelKeys
-          , sgInitialFunds = Map.fromList
+          , sgGenDelegs = shelleyDelKeys -- TODO should this be a ListMap?
```
lehins, line 992: This should not be a `ListMap`. See: https://github.com/input-output-hk/cardano-ledger/pull/2891#issuecomment-1172541619
```suggestion
          , sgGenDelegs = shelleyDelKeys
```

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
lehins: `cardano-prelude` might be a good place for those functions and tests, unless the overall plan is to stop depending on `cardano-prelude` (at least that is what we are striking for in `cardano-base` and `cardano-ledger` repos)

lehins (commented):
Just a few comments. I can't really say much about the functionality itself, since I am missing a lot of context and I am not terribly familiar with the cli codebase. But at the birds eye view it looks good.

cardano-api/src/Cardano/Api/Key.hs
```diff
@@ -67,6 +69,16 @@ generateSigningKey keytype = do
...
 
+generateInsecureSigningKey :: (Key keyrole, SerialiseAsRawBytes (SigningKey keyrole))
+  => AsType keyrole
+  -> IO (SigningKey keyrole)
+generateInsecureSigningKey keytype = do
+  g <- Random.getStdGen
+  let (bs, _) = Random.genByteString (fromIntegral (deterministicSigningKeySeedSize keytype)) g
+  case deserialiseFromRawBytes (AsSigningKey keytype) bs of
+    Just key -> return key
+    Nothing -> error "Unable to generate insecure key"
```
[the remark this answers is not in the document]
lehins: FYI, it makes me happy to see that `genByteString` is proving to be useful :)

MarcFontaine (approved):
LGTM

cabal.project
```diff
@@ -330,6 +330,18 @@ source-repository-package
   tag: ee59880f47ab835dbd73bea0847dab7869fc20d8
   --sha256: 1lrzknw765pz2j97nvv9ip3l1mcpf2zr4n56hwlz0rk7wq7ls4cm
 
+source-repository-package
```
Jimbo4350, line 333: Can we create an issue for the removal of this?

cardano-api/src/Cardano/Api/Key.hs
```diff
@@ -67,6 +69,16 @@ generateSigningKey keytype = do
     seedSize = deterministicSigningKeySeedSize keytype
 
 
+generateInsecureSigningKey :: (Key keyrole, SerialiseAsRawBytes (SigningKey keyrole))
```
Jimbo4350, line 72: Usually we indent the `::`

cardano-api/src/Cardano/Api/Key.hs
```diff
@@ -67,6 +69,16 @@ generateSigningKey keytype = do
...
+  => AsType keyrole
+  -> IO (SigningKey keyrole)
+generateInsecureSigningKey keytype = do
+  g <- Random.getStdGen
+  let (bs, _) = Random.genByteString (fromIntegral (deterministicSigningKeySeedSize keytype)) g
```
Jimbo4350, line 77: Can we ditch a pair of the parentheses with `$`?

cardano-api/src/Cardano/Api/Key.hs
Jimbo4350, line 80: A better error message would be `"generateInsecureSigningKey: Unable to generate insecure key"`

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
```diff
@@ -683,38 +691,36 @@ runGenesisCreateStaked (GenesisDir rootdir)
...
-        updateTemplate
+      !shelleyGenesis =
+        updateCreateStakedOutputTemplate
           -- Shelley genesis parameters
-          start genDlgs mNonDlgAmount nonDelegAddrs poolMap
-          stDlgAmount delegAddrs stuffedUtxoAddrs template
+          start genDlgs mNonDlgAmount (length nonDelegAddrs) nonDelegAddrs stakePools stake
+          stDlgAmount numDelegations delegAddrs stuffedUtxoAddrs template
+
+  -- shelleyGenesis contains lazy loaded data, so using lazyToJson to serialise to avoid
```
Jimbo4350, line 720: We should be able to delete this comment

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
```diff
@@ -924,31 +911,22 @@ writeBulkPoolCredentials dir bulkIx poolIxs = do
...
-    initialUtxoAddr <- case paySVK of
-      APaymentVerificationKey payVK ->
-        firstExceptT ShelleyGenesisCmdAddressCmdError
-        $ buildShelleyAddress payVK (Just . StakeVerifierKey . VerificationKeyFilePath . VerificationKeyFile $ stakeVKF) nw
-      _ -> left $ ShelleyGenesisCmdUnexpectedAddressVerificationKey payVKF "APaymentVerificationKey" paySVK
-
-    StakeVerificationKey stakeVK <- firstExceptT ShelleyGenesisCmdTextEnvReadFileError
-      . newExceptT
-      $ readFileTextEnvelope (AsVerificationKey AsStakeKey) stakeVKF
+computeInsecureDelegation ::
```
Jimbo4350, line 914: We should add a comment here to indicate that the keys are not cryptographically generated and we don't care because this is for testing purposes. `computeInsecureDelegation` is a strange name. Why not `generateTestDelegation` or something to that effect?

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
```diff
@@ -1229,4 +1282,3 @@ readAlonzoGenesis fpath = do
   lbs <- handleIOExceptT (ShelleyGenesisCmdGenesisFileError . FileIOError fpath) $ LBS.readFile fpath
   firstExceptT (ShelleyGenesisCmdAesonDecodeError fpath . Text.pack)
     . hoistEither $ Aeson.eitherDecode' lbs
-
```
Jimbo4350, line 1232: Is your editor deleting newlines at the end of the file?

cardano-cli/src/Cardano/CLI/Shelley/Run/StakeAddress.hs
```diff
@@ -53,20 +53,21 @@ runStakeAddressCmd (StakeCredentialDeRegistrationCert stakeVerifier outputFp) =
...
-      . newExceptT
-      $ writeFileTextEnvelope skFp (Just skeyDesc) skey
-    firstExceptT ShelleyStakeAddressCmdWriteFileError
-      . newExceptT
-      $ writeFileTextEnvelope vkFp (Just vkeyDesc) vkey
-  where
-    skeyDesc, vkeyDesc :: TextEnvelopeDescr
-    skeyDesc = "Stake Signing Key"
-    vkeyDesc = "Stake Verification Key"
+runStakeAddressKeyGenToFile ::
```
Jimbo4350, line 56: Indent/alignment of `::`?

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
```diff
@@ -1069,14 +1047,89 @@ updateTemplate (SystemStart start)
...
+    -- Genesis delegation (not stake-based):
+    -> Map (Hash GenesisKey) (Hash GenesisDelegateKey, Hash VrfKey)
+    -- Non-delegated initial UTxO spec:
+    -> Maybe Lovelace
+    -> Int
+    -> [AddressInEra ShelleyEra]
+    -- Genesis staking: pools/delegation map & delegated initial UTxO spec:
+    -> [(Ledger.KeyHash 'Ledger.StakePool StandardCrypto, Ledger.PoolParams StandardCrypto)]
+    -> [(Ledger.KeyHash 'Ledger.Staking StandardCrypto, Ledger.KeyHash 'Ledger.StakePool StandardCrypto)]
+    -> Lovelace
```
Jimbo4350, line 1061: Can we add haddocks for these parameters?

Jimbo4350 (changes requested):
A few comments but nothing major

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
```diff
@@ -924,31 +911,22 @@ writeBulkPoolCredentials dir bulkIx poolIxs = do
...
+    stakeVK   <- fmap getVerificationKey $ generateInsecureSigningKey AsStakeKey
 
-    pure Delegation
+    let stakeAddressReference = StakeAddressByValue . StakeCredentialByKey . verificationKeyHash $ stakeVK
+    let initialUtxoAddr = makeShelleyAddress nw (PaymentCredentialByKey (verificationKeyHash paymentVK)) stakeAddressReference
+
+    pure $ force Delegation
       { dInitialUtxoAddr = shelleyAddressInEra initialUtxoAddr
-      , dDelegStaking = Ledger.hashKey stakeVK
+      , dDelegStaking = Ledger.hashKey (unStakeVerificationKey stakeVK)
```
newhoggy, line 927: Use of `unStakeVerificationKey`

cardano-api/src/Cardano/Api/KeysShelley.hs
```diff
@@ -323,8 +323,9 @@ instance HasTypeProxy StakeKey where
 
 instance Key StakeKey where
 
-    newtype VerificationKey StakeKey =
-        StakeVerificationKey (Shelley.VKey Shelley.Staking StandardCrypto)
+    newtype VerificationKey StakeKey = StakeVerificationKey
+      { unStakeVerificationKey :: Shelley.VKey Shelley.Staking StandardCrypto
```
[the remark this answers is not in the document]
newhoggy: I use the field accessor here: https://github.com/input-output-hk/cardano-node/pull/4021/files#r915507713

cardano-api/src/Cardano/Api/Key.hs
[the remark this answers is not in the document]
newhoggy: Done!

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
```diff
@@ -1069,14 +1047,89 @@ updateTemplate (SystemStart start)
...
+
+    distribute :: Integer -> Int -> [AddressInEra ShelleyEra] -> [(AddressInEra ShelleyEra, Lovelace)]
+    distribute funds nAddrs addrs = zipWith (,) addrs (fmap Lovelace (coinPerAddr + rest:repeat coinPerAddr))
+      where coinPerAddr :: Integer
+            coinPerAddr = funds `div` fromIntegral nAddrs
+            rest = coinPerAddr * fromIntegral nAddrs
+
+    mkStuffedUtxo :: [AddressInEra ShelleyEra] -> [(AddressInEra ShelleyEra, Lovelace)]
+    mkStuffedUtxo xs = (, Lovelace minUtxoVal) <$> xs
+      where (Coin minUtxoVal) = Shelley._minUTxOValue $ sgProtocolParams template
```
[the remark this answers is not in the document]
newhoggy: Done

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
newhoggy: Removed comment

cabal.project
newhoggy: I put the functions into `Cardano.CLI.IO.Lazy` for the moment.

cardano-api/src/Cardano/Api/Key.hs
newhoggy: Done

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -2680,8 +2680,8 @@ getByronTxBodyContent (Annotated Byron.UnsafeTx{txInputs, txOutputs} _) =
       txScriptValidity   = TxScriptValidityNone
     }
 
-makeShelleyTransactionBody :: ()
-  => ShelleyBasedEra era
+makeShelleyTransactionBody ::
+     ShelleyBasedEra era
```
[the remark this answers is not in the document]
newhoggy: Moved the `::`

cardano-cli/src/Cardano/CLI/Shelley/Key.hs
```diff
@@ -481,3 +483,8 @@ readVerificationKeyOrHashOrTextEnvFile asType verKeyOrHashOrFile =
       eitherVk <- readVerificationKeyOrTextEnvFile asType vkOrFile
       pure (verificationKeyHash <$> eitherVk)
     VerificationKeyHash vkHash -> pure (Right vkHash)
+
+generatePaymentKeys :: Key keyrole => AsType keyrole -> IO (VerificationKey keyrole, SigningKey keyrole)
```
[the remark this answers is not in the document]
newhoggy: Renamed

cardano-cli/src/Cardano/CLI/Shelley/Run/Address.hs
```diff
@@ -57,37 +58,44 @@ renderShelleyAddressCmdError err =
...
   case cmd of
-    AddressKeyGen kt vkf skf -> runAddressKeyGen kt vkf skf
+    AddressKeyGen kt vkf skf -> runAddressKeyGenToFile kt vkf skf
     AddressKeyHash vkf mOFp -> runAddressKeyHash vkf mOFp
     AddressBuild paymentVerifier mbStakeVerifier nw mOutFp -> runAddressBuild paymentVerifier mbStakeVerifier nw mOutFp
     AddressBuildMultiSig sFp nId mOutFp -> runAddressBuildScript sFp nId mOutFp
     AddressInfo txt mOFp -> firstExceptT ShelleyAddressCmdAddressInfoError $ runAddressInfo txt mOFp
 
-runAddressKeyGen :: AddressKeyType
+runAddressKeyGenToFile :: AddressKeyType
```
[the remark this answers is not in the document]
newhoggy: Moved the `::`

cardano-cli/src/Cardano/CLI/Shelley/Run/Address.hs
```diff
@@ -57,37 +58,44 @@ renderShelleyAddressCmdError err =
...
-    case kt of
-      AddressKeyShelley         -> generateAndWriteKeyFiles AsPaymentKey
-      AddressKeyShelleyExtended -> generateAndWriteKeyFiles AsPaymentExtendedKey
-      AddressKeyByron           -> generateAndWriteKeyFiles AsByronKey
+runAddressKeyGenToFile kt vkf skf = case kt of
+  AddressKeyShelley         -> generateAndWriteKeyFiles AsPaymentKey          vkf skf
+  AddressKeyShelleyExtended -> generateAndWriteKeyFiles AsPaymentExtendedKey  vkf skf
+  AddressKeyByron           -> generateAndWriteKeyFiles AsByronKey            vkf skf
+
+generateAndWriteKeyFiles :: Key keyrole
```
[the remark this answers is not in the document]
newhoggy: Moved the `::`

cardano-cli/src/Cardano/CLI/Shelley/Run/Address.hs
```diff
@@ -57,37 +58,44 @@ renderShelleyAddressCmdError err =
...
+  => AsType keyrole
+  -> VerificationKeyFile
+  -> SigningKeyFile
+  -> ExceptT ShelleyAddressCmdError IO ()
+generateAndWriteKeyFiles asType vkf skf = do
+  uncurry (writePaymentKeyFiles vkf skf) =<< liftIO (generatePaymentKeys asType)
+
+writePaymentKeyFiles :: Key keyrole
```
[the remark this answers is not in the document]
newhoggy: Moved the `::`

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
newhoggy: Done

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
newhoggy: I don't think so.

cardano-cli/src/Cardano/CLI/Shelley/Run/StakeAddress.hs
newhoggy: Moved the `::`

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
newhoggy: I added the haddocks

lehins (approved):
:+1:

cardano-api/src/Cardano/Api/KeysShelley.hs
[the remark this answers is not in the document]
erikd: In general, I am always in favour of added an accessor to a newtype. If its there and no one uses it, its not a big deal (especially with a totally un-ambiguous name like this), but when its needed having it already there is a huge win IMO.

erikd (approved):

PR #4022, opened 2022-06-09
cardano-tracer: remove Cardano.Node prefix

denisshevchenko:
Remove `"Cardano.Node"`-prefix, fix outdated names of some metrics.


deepfire (approved):
Thank you @denisshevchenko !

PR #4030, opened 2022-06-09
Disable development eras configuration flag 

Jimbo4350:
This lets us advertise that we know about the Babbage era and we no longer require the `"TestEnableDevelopmentHardForkEras"` config flag

### DIFF
--- a/cardano-node/src/Cardano/Node/Protocol/Cardano.hs
+++ b/cardano-node/src/Cardano/Node/Protocol/Cardano.hs
@@ -85,7 +85,7 @@ mkSomeConsensusProtocolCardano NodeByronProtocolConfiguration {
                              npcAlonzoGenesisFileHash
                            }
                            NodeHardForkProtocolConfiguration {
-                            npcTestEnableDevelopmentHardForkEras,
+                            -- npcTestEnableDevelopmentHardForkEras,
                             -- During testing of the Alonzo era, we conditionally declared that we
                             -- knew about the Alonzo era. We do so only when a config option for
                             -- testing development/unstable eras is used. This lets us include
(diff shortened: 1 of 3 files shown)


cardano-node/src/Cardano/Node/Protocol/Cardano.hs
```diff
@@ -198,7 +198,7 @@ mkSomeConsensusProtocolCardano NodeByronProtocolConfiguration {
           -- version that this node will declare that it understands, when it
           -- is in the Alonzo era. That is, it is the version of protocol
           -- /after/ Alonzo, i.e. Babbage.
-          alonzoProtVer = ProtVer 6 0,
+          alonzoProtVer = ProtVer 7 0,
```
rdlrt, line 201: why is alonzo era version changed here?? This breaks existing explorers on public testnet that do check on versions...

ashisherc: Yes, this indeed is creating a mess checking block version for specific era

JaredCorduan: This change looks correct to me. We want we the nodes in the Alonzo era that are prepared for the upgrade to signal that they are ready by putting major protocol number 7 in the block header (since 7 corresponds to the next ledger era, Babbage).

rdlrt: The change looks correct for future (babbage era), but it seems to be at the expense of current (future past) - using incorrect / vague version in block header that is kinda unrelated to actual protocol major version.

Thus, the entire mapping of protocol version that so far has been decisive between forks is now different based on the node version (not protocol_major as the block header suggests)?

JaredCorduan: This is actually how it was designed to work from the very beginning, going back to the Shelley ledger spec. See section 13 of https://hydra.iohk.io/job/Cardano/cardano-ledger/shelleyLedgerSpec/latest/download-by-type/doc-pdf/ledger-spec.

The version listed in the block header indicates the latest version of the protocol that the block producer is prepared to upgrade to.

dcoutts (approved):
LGTM, including the retrospective fix for there being an intra-era HF within the Alonzo era (6).

JaredCorduan (approved):

cardano-node/src/Cardano/Node/Protocol/Cardano.hs
rdlrt: Thanks for the link, looks like I have been understanding the value of this block header wrong the whole time 😢  Looks like anyone with similar understanding might need immediate fixes. Can we add a reference to this (if not a short description) to the CIP 274?

JaredCorduan: >  Can we add a reference to this (if not a short description) to the CIP 274?

That is a most excellent idea!

PR #4031, opened 2022-06-09
Update cardano-base, cardano-ledger and consensus

lehins:


cardano-api/src/Cardano/Api/Fees.hs
```diff
@@ -382,7 +387,11 @@ data ScriptExecutionError =
 
        -- | An attempt was made to spend a key witnessed tx input
        -- with a script witness.
-     | ScriptErrorNotPlutusWitnessedTxIn ScriptWitnessIndex
+     | ScriptErrorNotPlutusWitnessedTxIn ScriptWitnessIndex ScriptHash
```
Jimbo4350, line 390: :tada:

cardano-api/src/Cardano/Api/Fees.hs
```diff
@@ -382,7 +387,11 @@ data ScriptExecutionError =
...
+
+       -- | A redeemer was supplied which points to a script hash
```
Jimbo4350, line 392: What does it mean "cannot resolve"? This is vague. The script does not exist in the transaction?

cardano-node/src/Cardano/Tracing/OrphanInstances/Shelley.hs
```diff
@@ -1018,20 +1018,20 @@ instance ToJSON (Alonzo.CollectError StandardCrypto) where
         object
           [ "kind" .= String "PlutusTranslationError"
           , "error" .= case err of
-              Alonzo.ByronInputInContext -> String "Byron input in the presence of a plutus script"
-              Alonzo.ByronOutputInContext -> String "Byron output in the presence of a plutus script"
-              Alonzo.TranslationLogicErrorInput -> String "Logic error translating inputs"
+              Alonzo.ByronTxOutInContext txOutSource ->
+                String $ "Byron output in the presence of a plutus script, output source: " <> show txOutSource
```
Jimbo4350, line 1022: What does this mean? A script is trying to witness a byron tx out?

cardano-node/src/Cardano/Tracing/OrphanInstances/Shelley.hs
```diff
@@ -1018,20 +1018,20 @@ instance ToJSON (Alonzo.CollectError StandardCrypto) where
...
+              Alonzo.TranslationLogicMissingInput txin ->
+                String $ "Logic error translating input: " <> show txin
```
Jimbo4350, line 1024: What does "Logic error" mean?

cardano-node/src/Cardano/Tracing/OrphanInstances/Shelley.hs
```diff
@@ -1018,20 +1018,20 @@ instance ToJSON (Alonzo.CollectError StandardCrypto) where
...
                   [ "kind" .= String "RedeemerPointerPointsToNothing"
                   , "ptr" .= (Api.renderScriptWitnessIndex . Api.fromAlonzoRdmrPtr) ptr
                   ]
-              Alonzo.TranslationLogicErrorDoubleDatum -> String "Logic error double datum"
-              Alonzo.LanguageNotSupported -> String "Language not supported"
-              Alonzo.InlineDatumsNotSupported -> String "Inline datums not supported"
-              Alonzo.ReferenceScriptsNotSupported -> String "Reference scripts not supported"
-              Alonzo.ReferenceInputsNotSupported -> String "Reference inputs not supported"
-              Alonzo.TimeTranslationPastHorizon -> String "Time translation requested past the horizon"
+              Alonzo.LanguageNotSupported lang -> String $ "Language not supported: " <> show lang
```
Jimbo4350, line 1030: :+1:

Jimbo4350 (changes requested):
A few comments

cardano-node/src/Cardano/Tracing/OrphanInstances/Shelley.hs
JaredCorduan: This is a failure to create a transaction context for Plutus, due to the presence of a Byron address in the transaction. It can manifest as either a newly created transaction output, or as an input being spent that is connection to an output in the UTxO that has a Byron address (the case is clarified by `txOutSource`)

cardano-node/src/Cardano/Tracing/OrphanInstances/Shelley.hs
JaredCorduan: We can come up with a better name for the type in a separate ledger PR, and change the prose here. The error indicates that a transaction input does not exist in the UTxO.

In the ledger code, this is indeed a logic error, since we only call the translation function after we have checked that the inputs are valid. So it's a bug if it ever fires. Since we've re-used the translation for the CLI, however, it can happen not as a true logic error. Great catch!

cardano-api/src/Cardano/Api/Fees.hs
JaredCorduan: The helper function attempts to glue a bunch of stuff together. In particular, it tries to take the redeemer pointers, connect them to a script hash, and then figure out what script corresponds to that script hash based on all the information it has in the tx (reference scripts and witnesses outside the body). This error happens if the pointer points to a script hash that we cannot identify.

cardano-node/src/Cardano/Tracing/OrphanInstances/Shelley.hs
Jimbo4350: How about "Cannot construct a Plutus ScriptContext from this transaction due to the a Byron UTxO being created or spent: " <> show ...

cardano-node/src/Cardano/Tracing/OrphanInstances/Shelley.hs
Jimbo4350: So we can ignore this for now and it will be removed/renamed in a future PR in ledger?

Jimbo4350: > that a transaction input does not exist in the UTxO

Why not "Transaction input does not exist in the UTxO"

cardano-api/src/Cardano/Api/Fees.hs
JaredCorduan: works for me!

cardano-node/src/Cardano/Tracing/OrphanInstances/Shelley.hs
JaredCorduan: works for me!

Jimbo4350 (approved):

JaredCorduan:
bors merge

PR #4035, opened 2022-06-10
Updated node and CLI version to 1.35.0

kevinhammond:
These changes update the version number for cardano-node, cardano-cli and other tools to be 1.35.0 (was 1.33.0)


kevinhammond:
I might have tripped over a versioning issue.

I updated `flake.lock` because I assumed I needed to change the version, but nix is preventing  that.  I will check with @disassembler

denisshevchenko (approved):

PR #4051, opened 2022-06-14
Do not require decentralization parameter in protocol parameters

### LINKED ISSUE #3897: [BUG] - Cannot build Alonzo-era transaction with Plutus script in Babbage era
*Internal*

**Area**
*Plutus*    Related to Plutus Scripts (Alonzo).

**Summary**
Building alonzo-era transaction with Plutus script in Babbage era fails with

```
$ cardano-cli transaction build-raw  --alonzo-era --mint-script-file ...

cardano-cli: toAlonzoPParams: Decentralization value required in Alonzo era
  error, called at src/Cardano/Api/ProtocolParameters.hs:1418:45 in cardano-api-1.33.0-G5f0eGKxspf1AYjSZoxnBV:Cardano.Api.ProtocolParameters
```

- OS Name: Fedora 35
- Node version: `vasil-testnet-v1`

### LINKED ISSUE #4033: Make build-raw and build commands able to create alonzo-era tx in Babbage (d=0 in Alonzo d=deprecated in Babbage)
**Internal/External**

**Area**
*Plutus*    Related to Plutus Scripts (Alonzo).

**Summary**
Make `decentralisationParam=0` in the returned values of `query protocol-parameters` cli command in Babbage era

**Expected behavior**
Alonzo-era Plutus transactions should work as default after hard-forking to Babbage. 

**Additional context**
At this moment because the output of the `query protocol-parameters` cli command returns `decentralisationParam=null` in the Babbage era, the Alonzo-era Plutus transactions do not work in the Bababbage era. 

A simple and fast solution could be to set `decentralisationParam=0` (instead of null) in the output of the `query protocol-parameters` cli command

Jimbo4350:
Fixes `build-raw` command and partially resolves #4033 

Closes #3897

### DIFF
--- a/cardano-api/src/Cardano/Api/ProtocolParameters.hs
+++ b/cardano-api/src/Cardano/Api/ProtocolParameters.hs
@@ -147,6 +147,7 @@ data ProtocolParameters =
        --
        -- This is the \"d\" parameter from the design document.
        --
+       -- /Deprecated in Babbage/
        protocolParamDecentralization :: Maybe Rational,
 
        -- | Extra entropy for the Praos per-epoch nonce.
@@ -295,7 +296,7 @@ instance FromJSON ProtocolParameters where
       v <- o .: "protocolVersion"
       ProtocolParameters
         <$> ((,) <$> v .: "major" <*> v .: "minor")
-        <*> o .: "decentralization"
+        <*> o .:? "decentralization"
         <*> o .: "extraPraosEntropy"
         <*> o .: "maxBlockHeaderSize"
         <*> o .: "maxBlockBodySize"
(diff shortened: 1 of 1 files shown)


JaredCorduan (approved):
This seem like a great way to unblock us until we have the time to handle the protocol parameters in an era-generic manor.

dcoutts (approved):
This is a good quick hack, and will be fine for now.

Do we have an open issue for the proper fix? We should not loose track of the proper solution.

CarlosLopezDeLara:
> This is a good quick hack, and will be fine for now.
> 
> Do we have an open issue for the proper fix? We should not loose track of the proper solution.

https://github.com/input-output-hk/cardano-node/issues/4054

PR #4055, opened 2022-06-15
Workbench fixes

deepfire:
1. ligther `startstop` profiles (no stuffed UTxO and no extra delegations)
2. better node startup error handling (prevent halfway-started clusters from leaking processes)
3. new tracing has dropped the `Cardano.Node` prefix, so `locli` had to be updated
4. `-autostay` makefile target variants now actually stay after successful runs
5. fix analysis for short-lived profiles, by not filtering out all data (used to drop first two epochs)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ PROFILE ?= default-${ERA}
 REV     ?= master
 ARGS    ?=
 CMD     ?=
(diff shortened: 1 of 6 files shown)


deepfire:
CI has passed, and this only affects workbench, merging directly.

PR #4059, opened 2022-06-16
update ledger and plutus

JaredCorduan:
The update to ledger brings two small but nice improvements:
* the `SuccessfulPlutusScriptsEvent` event doesn't fire on an empty list anymore
* we do not serialize empty txbody fields anymore

The update to Plutus removes the V2 size limit


Jimbo4350 (approved):

JaredCorduan:
bors merge

PR #4068, opened 2022-06-21
CAD-4540 Add default LMDB limits to node configuration

jorisdral:
The default LMDB limits are now included in the node configuration. These default limits should be suitable for a node in production. A custom `lmdbMapSize` can still be set through manual configuration, but setting this value is optional.

This PR makes slight changes to the command-line flags that were added by #4011. The options for the JSON config haven't changed.

```
Usage: cardano-node run # ...
                          [ --in-memory-ledger-db-backend
                          | --lmdb-ledger-db-backend [--lmdb-mapsize BIN]
                          ]

  Run the node.

Available options:
  # ...
  --in-memory-ledger-db-backend
                           Use the InMemory ledger DB backend. Incompatible with
                           `--lmdb-ledger-db-backend`.
  --lmdb-ledger-db-backend Use the LMDB ledger DB backend. By default, the
                           mapsize (maximum database size) of the backend is set
                           to 16 Gigabytes. Warning: if the database size
                           exceeds the given mapsize, the node will abort.
                           Therefore, the mapsize should be set to a value high
                           enough to guarantee that the maximum database size
                           will not be reached during the expected node uptime.
                           Incompatible with `--in-memory-ledger-db-backend`.
  --lmdb-mapsize BIN       The maximum database size defined as a binary number.
                           The BIN argument must be a number followed by a unit,
                           for example 10Gi for 10 Gibibytes. Note that BIN must
                           be a multiple of the OS page size (in bytes).
```

EDIT: Updated help text in listing after processing PR reviews.

### DIFF
--- a/cardano-node/src/Cardano/Node/Configuration/LedgerDB.hs
+++ b/cardano-node/src/Cardano/Node/Configuration/LedgerDB.hs
@@ -1,8 +1,13 @@
+{-# LANGUAGE NumericUnderscores #-}
+
 module Cardano.Node.Configuration.LedgerDB (
     BackingStoreSelectorFlag(..)
+  , defaultLMDBLimits
   ) where
 
 import           Prelude
+
+import           Ouroboros.Consensus.Storage.LedgerDB.HD.LMDB (LMDBLimits (..))
 
 -- | Choose the LedgerDB Backend
 --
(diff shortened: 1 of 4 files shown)


jorisdral:
These changes should probably squashed

cardano-node/src/Cardano/Node/Configuration/LedgerDB.hs
```diff
@@ -20,3 +25,26 @@ data BackingStoreSelectorFlag =
                      -- default of 12Gi will be used.
   | InMemory
   deriving (Eq, Show)
+
```
dnadales, line 28: :100:

dnadales (approved):
LGTM :pray:

Issue #3688, opened 2022-03-07
[BUG] - Shelley era Tx body in CDDL format is not supported by some CLI commands


mkoura:
*Internal*   if an IOHK staff member.

**Area**
*Other*   Any other topic (Delegation, Ranking, ...).

**Summary**
Shelley era Tx body in CDDL format:
```
cardano-cli transaction build-raw --fee 0 --out-file test_transfer_all_funds_ci0_xrc_estimate_tx.body --tx-in d53a926ce885c6821ec475489be9d0fdade33969c0833e6205333046091a4488#1 --tx-out addr_test1vq98lqkgg36k78axnz5e2g8h44q7mcpsnku9tv7g9qmvwdguhqcu9+1000000000 --invalid-hereafter 28444 --cddl-format --shelley-era
```

It then cannot be e.g. signed:
```
cardano-cli transaction sign --tx-body-file test_transfer_all_funds_ci0_xrc_tx.body --testnet-magic 42 --signing-key-file addr_basic_ci0_1.skey --out-file test_transfer_all_funds_ci0_xrc_tx.signed
```

> Command failed: transaction sign  Error: Failed to decode neither the cli's serialisation format nor the ledger's CDDL serialisation format. TextEnvelope error: test_transfer_all_funds_ci0_xrc_tx.body: TextEnvelope type error:  Expected one of: TxUnsignedByron, TxUnsignedShelley, TxBodyAllegra, TxBodyMary, TxBodyAlonzo Actual: Unwitnessed Tx ShelleyEra
TextEnvelopeCddl error: test_transfer_all_funds_ci0_xrc_tx.body: TextEnvelopeCddl type error:  Expected one of: Witnessed Tx Byron, Witnessed Tx Shelley, Witnessed Tx AllegraEra, Witnessed Tx MaryEra, Witnessed Tx AlonzoEra, Unwitnessed Tx Byron, Unwitnessed Tx Shelley, Unwitnessed Tx AllegraEra, Unwitnessed Tx MaryEra, Unwitnessed Tx AlonzoEra Actual: Unwitnessed Tx ShelleyEra

Same for `calculate-min-fee`:
```
[...]

PR #4071, opened 2022-06-21
Fix cddl format for shelley and byron eras

Jimbo4350:
Fixes: #3688


JaredCorduan (approved):
:+1:

mkoura:
Verified that this fixes https://github.com/input-output-hk/cardano-node/issues/3688 by signing & submitting Shelley-era Tx with CDDL formatted body.

Issue #3951, opened 2022-05-31
tx-generator: add a trace that logs the commit-id on start-up


MarcFontaine:
Similar in the way cardano-node does it

MarcFontaine:
https://github.com/input-output-hk/cardano-node/pull/4083

PR #4083, opened 2022-06-23
log version of tx-generator on startup

MarcFontaine:
Add a version command to the tx-generator and write the tx-generator version to the logs on startup.


MarcFontaine:
#3951

deepfire:
This PR was merged, as part of #4056

Issue #4045, opened 2022-06-14
[FR] - transaction view doesn't show reference inputs


james-iohk:
**Internal/External**
*Internal*   if an IOHK staff member.

**Area**
*Other*   Any other topic (Delegation, Ranking, ...).

**Summary**
`transaction view` works for transactions with reference inputs but doesn't display them in the output.

**Steps to reproduce**
1. Create a transaction body with reference inputs (use `--read-only-tx-in-reference`)
2. Use `transaction view` cmd to inspect the transaction body

**Expected behavior**
Transaction content is displayed including all reference inputs

**System info (please complete the following information):**
- OS Name: 20.04.4 LTS (Focal Fossa)
- Node/Cli version: 07053e4

james-iohk:
Relates to https://github.com/input-output-hk/cardano-node/issues/4039

Issue #4039, opened 2022-06-13
[FR] - Cannot use transaction view to check tx with inline datum


saratomaz:
**Internal/External**
*Internal*   if an IOHK staff member.

**Area**
*Other*   Any other topic (Delegation, Ranking, ...).

**Summary**
Check a tx with inline datum using transaction view fails.

**Steps to reproduce**
```
cardano-cli transaction build 
--tx-in d41343174cf77a962b2b28b52a1d8acc2e936757d33422ef452b33ec50eb9d53#0 
--tx-out addr_test1wqag3rt979nep9g2wtdwu8mr4gz6m4kjdpp5zp705km8wys6t2kla+200000000 
--tx-out-inline-datum-file /home/sara/IOHK/project/cardano-node-tests/cardano_node_tests/tests/data/plutus/typed-42.datum 
--tx-out addr_test1vz5rd5hsead7gcgn6zx6nalqxz6zlvdmg89kswl935dfh8cqn5kcy+2000000 
--change-address addr_test1vrx0w7gndt6gk9svrksmpg23lmwmlcx2w2fre7rk27r8gdcyazwxm 
--out-file test_txout_locking_ci0_ddk_step1_tx.body 
--babbage-era 
--testnet-magic 42

cardano-cli transaction view --tx-body-file test_txout_locking_ci0_ddk_step1_tx.body

cardano-cli: FatalError {fatalErrorMessage = "TODO: Babbage"}
```

**System info (please complete the following information):**

Node version: `master`

james-iohk:
Relates to https://github.com/input-output-hk/cardano-node/issues/4045

CarlosLopezDeLara:
Implemented in https://github.com/input-output-hk/cardano-node/pull/4089

PR #4089, opened 2022-06-23
Update tx view to render inline datums and reference inputs

Jimbo4350:
Resolves: #4045 and #4039


cardano-cli/src/Cardano/CLI/Run/Friendly.hs
```diff
@@ -452,6 +456,10 @@ friendlyAuxScripts = \case
   TxAuxScriptsNone -> Null
   TxAuxScripts _ scripts -> String $ textShow scripts
 
+friendlyReferenceInputs :: TxInsReference build era -> Aeson.Value
```
JaredCorduan, line 459: json = friendly?

JaredCorduan (approved):
LGTM!

erikd (approved):

cardano-cli/src/Cardano/CLI/Run/Friendly.hs
Jimbo4350: Apparently. I was just following the pattern implemented by Yuri. All of it needs to be refactored though.

PR #4095, opened 2022-06-24
Break recursion on old cardano-node versions

jbgi:
(which won't be used by the build anyway), to avoid exponential growth of flake.lock at each pin update.

Alternative solution of #4074.

`flake.lock` now downsized from 236K to 64K.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -16,150 +16,6 @@
         "type": "github"
       }
     },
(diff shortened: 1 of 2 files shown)


newhoggy:
There has been no progress on this PR for a long time.

Closing.

PR #4102, opened 2022-06-27
Fix query protocol-state

newhoggy:
Fixes one of the commands in https://github.com/input-output-hk/cardano-node/issues/3883

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Shelley/Orphans.hs
+++ b/cardano-cli/src/Cardano/CLI/Shelley/Orphans.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TypeFamilies #-}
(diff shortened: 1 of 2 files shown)


newhoggy:
For Alonzo:

```
CARDANO_NODE_SOCKET_PATH=example/main.sock cardano-cli query protocol-state --cardano-mode --testnet-magic 42


{
    "chainDepState": {
        "csLabNonce": {
            "contents": "f0dc1ec610b6be842a892c24fe5b4a59b0b93494fd40177aef444299fea94ac1",
            "tag": "Nonce"
        },
        "csProtocol": [
            {
                "5b34625dfd65a3734e20d4389207d3b50047fcfacb0bb12c554b66bd": 1,
                "c921170865471ecd443ae3d0600749568c67b7e2fd9150f56bd434d0": 1
            },
            {
                "contents": "c272db989e2a582796917d721247fef17707c8b9b2f433bad2701df3465de6c0",
                "tag": "Nonce"
            },
            {
                "contents": "c272db989e2a582796917d721247fef17707c8b9b2f433bad2701df3465de6c0",
                "tag": "Nonce"
            }
        ],
        "csTickn": {
            "ticknStateEpochNonce": {
                "contents": "1779ef55edb987ab38e64c7a8aa7839e44cbdb41ff73a2f3b9e5c06f5f8bb0a2",
                "tag": "Nonce"
            },
            "ticknStatePrevHashNonce": {
                "tag": "NeutralNonce"
            }
        }
    },
    "lastSlot": 42
}
```

newhoggy:
For Babbage:

```
CARDANO_NODE_SOCKET_PATH=example/main.sock cardano-cli query protocol-state --cardano-mode --testnet-magic 42


{
    "candidateNonce": {
        "contents": "bb35f00ef3867227367142df3f296d9fb579701cad7f6eb1a488965a0ed19dc6",
        "tag": "Nonce"
    },
    "epochNonce": {
        "contents": "8745878ee66a004c9476e9f972c70af5b20d32e42831afab0cd3acf0f59a74c6",
        "tag": "Nonce"
    },
    "evolvingNonce": {
        "contents": "4e7db806ca023ffeb6b85cec68c9de5b14aa0c0550033b99a329aad83da8f262",
        "tag": "Nonce"
    },
    "labNonce": {
        "contents": "4fd0b9aea702fa10f216b9e4d2c7a67cd571b0d7b93d07d4654324b5c39680bb",
        "tag": "Nonce"
    },
    "lastEpochBlockNonce": {
        "tag": "NeutralNonce"
    },
    "lastSlot": 269,
    "oCertCounters": {
        "4f6a7786b181dc8192e3643e13b8c5424e5a501d4ba9d803f3d9a8f8": 0,
        "6211474ccb03e8537f810d10520ac37e2822b0ace5bbda346ef6f0aa": 0,
        "84c3e438c497b20a12aa5e7ad394db6aa751ecd2cbad5a5b104540b5": 0
    }
}
```

mkoura:
Verified the behavior in both Alonzo and Babbage, LGTM :+1:

cardano-cli/src/Cardano/CLI/Shelley/Orphans.hs
```diff
@@ -10,41 +10,42 @@
...
-
-import           Control.SetAlgebra as SetAlgebra
-import           Data.Aeson
-import qualified Data.ByteString.Base16 as Base16
-import qualified Data.ByteString.Short as SBS
-import qualified Data.Text.Encoding as Text
-
 import           Cardano.Api.Orphans ()
-
+import           Cardano.Ledger.AuxiliaryData (AuxiliaryDataHash (..))
```
Jimbo4350, line 18: Can you undo the import changes to follow the style in other modules?

cardano-cli/src/Cardano/CLI/Shelley/Orphans.hs
```diff
@@ -58,9 +59,9 @@ deriving newtype instance ToJSON ByronHash
 -- This instance is temporarily duplicated in cardano-config
 
 instance ToJSON (HeaderHash blk) => ToJSON (Tip blk) where
-  toJSON TipGenesis = object [ "genesis" .= True ]
+  toJSON TipGenesis = J.object [ "genesis" .= True ]
```
Jimbo4350, line 62: Why the qualified import?  If you are going to qualify aeson imports we do it via `Aeson`. In this case I think its overkill to qualify `object` as it is fairly obvious already that it comes from the aeson library.

Jimbo4350 (changes requested):
Couple of small change but LGTM otherwise

cardano-cli/src/Cardano/CLI/Shelley/Orphans.hs
newhoggy: I've switched it to `Aeson`.  I'm trying to avoid unqualified imports and at the same time keep the qualified import short so it doesn't wrap.

cardano-cli/src/Cardano/CLI/Shelley/Orphans.hs
newhoggy: Any chance we could use `ImportQualifiedPost` language extension like we do in `plutus-apps` so I could just do a line sort to keep my imports tidy?

Jimbo4350: We would need to make a decision to use `ImportQualifiedPost` in the repo. Having a mix of some modules using it and others not is not ideal. I personally don't see justification for this aesthetic change.

PR #4104, opened 2022-06-27
adding a "make check" to the libsodium and secp256k1 install process

gitmachtl:
doing a "make check" after a "make" is always a good idea to see if all went well so far


gitmachtl:
@KtorZ please check, thx.

newhoggy (approved):

PR #4105, opened 2022-06-28
RTView: MVP

### LINKED ISSUE #4096: RTView: MVP
After #3936 was merged, RTView is _almost_ done as MVP. Now we have to finish it, particularly:

1. Add documentation
2. Polish light/dark theme for UI
3. Add comments in the code

denisshevchenko:
Closes #4096 

MVP for RTView: it's not ready for production yet, but ready for testing by SPOs.


deepfire (approved):
bors r+

cardano-node/src/Cardano/Node/Startup.hs
```diff
@@ -209,3 +209,25 @@ prepareNodeInfo ptcl (SomeConsensusProtocol whichP pForInfo) tc nodeStartTime =
     case tcNodeName tc of
       Just aName -> return aName
       Nothing    -> pack <$> getHostName
+
+-- | This information is taken from 'BasicInfoShelleyBased'. It is required for
+--   'cardano-tracer' service (particularly, for RTView).
+data NodeStartupInfo = NodeStartupInfo {
```
Jimbo4350, line 215: If it is taken from `BasicInfoShelleyBased` why are we essentially duplicating the type?

Jimbo4350 (changes requested):

cardano-node/src/Cardano/Node/Startup.hs
denisshevchenko: @Jimbo4350 

1. `NodeStartupInfo` takes not _all_ fields from `BasicInfoShelleyBased`.
2. `NodeStartupInfo` may pass another data in the future, not only from `BasicInfoShelleyBased`.

Jimbo4350 (approved):

PR #4106, opened 2022-06-28
Fix leadership schedule for current on babbage

newhoggy:


newhoggy:
```
$ cabal test cardano-testnet --enable-tests --test-show-details=direct --test-options '-p Babbage.leadership-schedule'
...
Test suite cardano-testnet-tests: RUNNING...
test/Spec.hs
  Spec
    Babbage
      leadership-schedule: OK (259.17s)
          ✓ leadership-schedule passed 1 test.

All 1 tests passed (259.17s)
Test suite cardano-testnet-tests: PASS
Test suite logged to:
/Users/jky/wrk/iohk/cardano-node/dist-newstyle/build/aarch64-osx/ghc-8.10.7/cardano-testnet-1.35.0/t/cardano-testnet-tests/test/cardano-testnet-1.35.0-cardano-testnet-tests.log
1 of 1 test suites (1 of 1 test cases) passed.
```

newhoggy:
Resolves https://github.com/input-output-hk/cardano-node/issues/3883

Jimbo4350 (dismissed):

cardano-testnet/cardano-testnet.cabal
```diff
@@ -43,6 +43,8 @@ library
                       , hedgehog-extras
                       , http-client
                       , http-types
+                      , lens
```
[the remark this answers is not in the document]
newhoggy: Removed

cardano-testnet/src/Testnet/Cardano.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE BlockArguments #-}
```
Jimbo4350, line 1: Why?

cardano-testnet/src/Testnet/Cardano.hs
```diff
@@ -470,20 +471,45 @@ testnet testnetOptions H.Conf {..} = do
 
   -- Make the pool operator cold keys
   -- This was done already for the BFT nodes as part of the genesis creation
-  forM_ poolNodeNames $ \node -> do
+
+        -- poolVrfVkeys
```
Jimbo4350: Can you elaborate on this more? How was it done?

cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -159,6 +160,8 @@ import qualified Ouroboros.Network.Block
 import qualified Ouroboros.Network.Protocol.ChainSync.Client as CS
 import qualified Ouroboros.Network.Protocol.ChainSync.ClientPipelined as CSP
 import           Ouroboros.Network.Protocol.ChainSync.PipelineDecision
+import Cardano.Ledger.BaseTypes (Nonce)
```
Jimbo4350: `Cardano.Ledger.BaseTypes` should be with the other `Cardano.Ledger.x` modules and the indent is wrong for these modules

cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -1410,8 +1413,8 @@ nextEpochEligibleLeadershipSlots sbe sGen serCurrEpochState ptclState
 -- See Leader Value Calculation in the Shelley ledger specification.
 -- We need the certified natural value from the VRF, active slot coefficient
 -- and the stake proportion of the stake pool.
-isLeadingSlots
-  :: Crypto.Signable v Shelley.Spec.Seed
+isLeadingSlots :: forall v era. ()
```
Jimbo4350: why the `()`?

cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -1474,9 +1480,8 @@ obtainDecodeEpochStateConstraints ShelleyBasedEraBabbage f = f
 
 -- | Return the slots at which a particular stake pool operator is
 -- expected to mint a block.
-currentEpochEligibleLeadershipSlots
-  :: forall era ledgerera .
-     ShelleyLedgerEra era ~ ledgerera
+currentEpochEligibleLeadershipSlots :: forall era ledgerera. ()
```
Jimbo4350, line 1492: Same here, why `()`?

cardano-testnet/src/Test/Assert.hs
```diff
@@ -0,0 +1,52 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Test.Assert
+  ( readJsonLines
+  , assertChainExtended
+  ) where
+
+import Control.Lens ((^.))
```
Jimbo4350: Have a look at the other test modules (in cardano-cli/cardano-node) and follow how the modules are organized there.

cardano-testnet/test/Spec/Cli/Alonzo/LeadershipSchedule.hs
```diff
@@ -0,0 +1,530 @@
...
+{-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
+
+{- HLINT ignore "Redundant return" -}
+{- HLINT ignore "Use let" -}
+
+module Spec.Cli.Alonzo.LeadershipSchedule
+  ( hprop_leadershipSchedule
+  ) where
+
+import Cardano.Api (SerialiseAddress(serialiseAddress), UTxO(UTxO), AlonzoEra)
```
Jimbo4350, line 19: Likewise for these imports. Let's try to get a unified style for the repo.

cardano-testnet/cardano-testnet.cabal
```diff
@@ -123,8 +124,10 @@ test-suite cardano-testnet-tests
                       , tasty-expected-failure
                       , tasty-hedgehog
                       , text
+                      , time
 
   other-modules:
+                        Spec.Cli.Alonzo.LeadershipSchedule
```
Jimbo4350, line 128: It is possible to have a leadership schedule test that is parameterized on the era?

cardano-testnet/src/Testnet/Babbage.hs
```diff
@@ -1,11 +1,12 @@
+{-# LANGUAGE BlockArguments #-}
```
Jimbo4350, line 1: Why?

cardano-testnet/test/Spec/Cli/Babbage/LeadershipSchedule.hs
```diff
@@ -0,0 +1,202 @@
+{-# LANGUAGE BlockArguments           #-}
```
Jimbo4350, line 1: Why?

cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -87,23 +87,22 @@ import qualified Data.Yaml as Yaml
 import           Formatting.Buildable (build)
 import           GHC.Records (HasField (..))
 import           System.FilePath
+import           Control.State.Transition
```
Jimbo4350, line 90: These are not in alphabetical order. Are you using `stylish-haskell` to sort your imports?

Jimbo4350 (changes requested):
Nice :+1:  A few comments. Can you also tidy up the commit history? It's a fairly big PR and it was difficult to follow the changes.

cardano-testnet/testnet/Testnet/Commands/Babbage.hs
```diff
@@ -25,7 +25,35 @@ data BabbageOptions = BabbageOptions
...
+      )
+  <*> OA.option auto
+      (   OA.long "slot-duration"
+      <>  OA.help "Slot duration"
+      <>  OA.metavar "MILLISECONDS"
+      <>  OA.showDefault
+      <>  OA.value (slotDuration defaultTestnetOptions)
+      )
+  <*> OA.option auto
+      (   OA.long "security-param"
```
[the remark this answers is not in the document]
newhoggy: This is currently how all the other testnets are are started up in the same project, so the PR is in keeping witht hose.

I'm partial to changing the design to use the genesis file as the source of truth, but I think we should design this properly as a separate piece of work and not package it up into the same PR.

cardano-testnet/src/Testnet/Babbage.hs
```diff
@@ -174,6 +172,8 @@ testnet testnetOptions H.Conf {..} = do
 
   -- Copy the cost model
 
+  let numPoolNodes = 3 :: Int
```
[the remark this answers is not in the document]
newhoggy: It can be made configurable, but it shouldn't be configurable until it can be tested to confirm it works.  I think this can be a separate PR.  I will add a TODO.

.stylish-haskell.yaml
```diff
@@ -86,7 +86,6 @@ language_extensions:
   - LambdaCase
   - MultiParamTypeClasses
   - MultiWayIf
-  - NoImplicitPrelude
```
newhoggy, line 89: I had to remove this.  It breaks `stylish-haskell` for me:

```
*** cmd start: stylish-haskell -i /Users/jky/wrk/iohk/cardano-node/cardano-testnet/test/Spec/Cli/Babbage/LeadershipSchedule.hs
Unknown extension: "NoImplicitPrelude"
```

cardano-testnet/test/Main.hs
```diff
@@ -5,19 +5,20 @@ module Main
   ) where
 
 import           Prelude
+import           Test.Tasty (TestTree)
+
 -- import qualified Spec.Cli.KesPeriodInfo
```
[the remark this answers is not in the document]
newhoggy: This corresponds to the commented out code below.  We should keep both or delete both.

cardano-testnet/cardano-testnet.cabal
newhoggy: Not yet.  Maybe in the future when we redesign the testnet API.

cardano-api/src/Cardano/Api/LedgerState.hs
newhoggy: `stylish-haskell` wasn't set up.  I've set it up now.

cardano-testnet/src/Testnet/Babbage.hs
```diff
@@ -0,0 +1,386 @@
...
+import qualified Testnet.Conf as H
+
+{- HLINT ignore "Reduce duplication" -}
+{- HLINT ignore "Redundant <&>" -}
+{- HLINT ignore "Redundant flip" -}
+{- HLINT ignore "Redundant id" -}
+{- HLINT ignore "Use let" -}
+
+
+data TestnetOptions = TestnetOptions deriving (Eq, Show)
```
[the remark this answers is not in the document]
newhoggy: Currently there are different functions for setting up different testnets and they take different types because they configurations they take are different.  Unifying them is desirable, but it would take a lot of work.  I think this PR is the wrong place to do it.

cardano-testnet/src/Test/Process.hs
```diff
@@ -130,3 +131,18 @@ assertByDeadlineIOCustom str deadline f = GHC.withFrozenCallStack $ do
       else do
         H.annotateShow currentTime
         failMessage GHC.callStack $ "Condition not met by deadline: " <> str
+
+assertByDeadlineMCustom
```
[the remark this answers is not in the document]
newhoggy: One cannot be substituted for the other because `IO` does not have `MonadTest` instance.

cabal.project
```diff
@@ -330,6 +330,12 @@ source-repository-package
   tag: ee59880f47ab835dbd73bea0847dab7869fc20d8
   --sha256: 1lrzknw765pz2j97nvv9ip3l1mcpf2zr4n56hwlz0rk7wq7ls4cm
 
+source-repository-package
+  type: git
+  location: https://github.com/haskell-works/hw-aeson
```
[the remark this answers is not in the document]
newhoggy: This library is already a dependency via `hedgehog-extras`.  The inclusion here is just because the index state hasn't been updated yet so the latest required version is not available.

cardano-testnet/test/Spec/Cli/Babbage/LeadershipSchedule.hs
newhoggy: I discovered it was an interesting way to use layout in expressions instead of parenthesis.

For example the expression:

```
  H.rewriteYamlFile (tempAbsPath </> "configuration.yaml") . J.rewriteObject
    $ HM.delete "GenesisFile"
    . HM.insert "Protocol" (J.toJSON @String "Cardano")
    ...
    . HM.insert "TestEnableDevelopmentNetworkProtocols" (J.toJSON True)
    . flip HM.alter "setupScribes"
      ( fmap
        ( J.rewriteArrayElements
          ( J.rewriteObject
            ( HM.insert "scFormat"
              ( case nodeLoggingFormat testnetOptions of
                  NodeLoggingFormatAsJson -> "ScJson"
                  NodeLoggingFormatAsText -> "ScText")))))
```

Can be written as:

```
  H.rewriteYamlFile (tempAbsPath </> "configuration.yaml") . J.rewriteObject
    $ HM.delete "GenesisFile"
    . HM.insert "Protocol" (J.toJSON @String "Cardano")
    ...
    . HM.insert "TestEnableDevelopmentNetworkProtocols" (J.toJSON True)
    . flip HM.alter "setupScribes" do
        fmap do
          J.rewriteArrayElements do
            J.rewriteObject do
              HM.insert "scFormat" case nodeLoggingFormat testnetOptions of
                NodeLoggingFormatAsJson -> "ScJson"
                NodeLoggingFormatAsText -> "ScText"
```

This works because `do x y z` is the same as `x y z`.  This is true even without `BlockArguments`.

What `BlockArguments` gives us is that in the expression `f $ x y z` can be rewritten as `f do x y z`.

This gives us two nice things:

* We can use `do` to use layout to group expressions, which means we don't have to match parens anymore.  When we write Haskell, we usually indent sub-expressions to help the reader read the code anyway.  The parentheses are for the compiler because layout is not significant to it inside an expression.  `BlockArguments` allows code to be written such that the mechanism humans use to group expressions and the compiler uses to group expressions can be one and the same.
* The use of `$` is unreliable because it is an operator and operators have a precedence and sometimes it is the wrong precedence.

For example if you tried to write the large expression from before using `$` instead it is actually wrong and doesn't compile:

```
  H.rewriteYamlFile (tempAbsPath </> "configuration.yaml") . J.rewriteObject
    $ HM.delete "GenesisFile"
    . HM.insert "Protocol" (J.toJSON @String "Cardano")
    ...
    . HM.insert "TestEnableDevelopmentNetworkProtocols" (J.toJSON True)
    . flip HM.alter "setupScribes"
      $ fmap
        $ J.rewriteArrayElements
          $ J.rewriteObject
            $ HM.insert "scFormat"
              $ case nodeLoggingFormat testnetOptions of
                  NodeLoggingFormatAsJson -> "ScJson"
                  NodeLoggingFormatAsText -> "ScText"
```

Similarly, `f $ case { x -> y }` can be rewritten without the `$` as `f case {x -> y}`.

And `f $ \x -> y` can be rewritten without the `$` as `f \x -> y`.

Basically there are a number of cases where when we previously used `$`, it can simply be dropped.

cardano-testnet/src/Testnet/Babbage.hs
```diff
@@ -0,0 +1,386 @@
...
+  H.writeFile (tempAbsPath </> "node-spo2/port") "3002"
+  H.writeFile (tempAbsPath </> "node-spo3/port") "3003"
+
+
+  -- Make topology files
+  -- TODO generalise this over the N BFT nodes and pool nodes
+  
+  H.lbsWriteFile (tempAbsPath </> "node-spo1/topology.json") $ J.encode $
+    J.object
+    [ "Producers" .= J.toJSON
```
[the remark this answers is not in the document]
newhoggy: Function names like `object`, `encode` and `decode` are in danger of conflicting, although it doesn't conflict here.

I prefer the use of qualified function names, but I've removed them here.

cardano-testnet/src/Testnet/Cardano.hs
newhoggy: How what was done?  This was testnet was a direct port of the `./scripts/byron-to-alonzo/mkfiles.sh` at the time it was ported.

cardano-api/src/Cardano/Api/LedgerState.hs
newhoggy: This function doesn't exist anymore.

cardano-api/src/Cardano/Api/LedgerState.hs
newhoggy: The reason I put the empty constraint `()` there is because it allows me to begin the next line with a fat arrow.

Code without `()`:

```
currentEpochEligibleLeadershipSlots :: forall era ledgerera.
     ShelleyLedgerEra era ~ ledgerera
  => Ledger.Era ledgerera
```

Code with `()`:

```
currentEpochEligibleLeadershipSlots :: forall era ledgerera. ()
  => ShelleyLedgerEra era ~ ledgerera
  => Ledger.Era ledgerera
```

The second is a little bit more `git` friendly because if I delete the first non-empty constraint, the git sees fewer lines changed.

For example:

```
diff --git a/cardano-api/src/Cardano/Api/LedgerState.hs b/cardano-api/src/Cardano/Api/LedgerState.hs
index bf0674fd1..7115b4a89 100644
--- a/cardano-api/src/Cardano/Api/LedgerState.hs
+++ b/cardano-api/src/Cardano/Api/LedgerState.hs
@@ -1494,7 +1494,6 @@ obtainDecodeEpochStateConstraints ShelleyBasedEraBabbage f = f
 -- | Return the slots at which a particular stake pool operator is
 -- expected to mint a block.
 currentEpochEligibleLeadershipSlots :: forall era ledgerera. ()
-  => ShelleyLedgerEra era ~ ledgerera
   => Ledger.Era ledgerera
   => Consensus.PraosProtocolSupportsNode (Api.ConsensusProtocol era)
   => HasField "_d" (Core.PParams ledgerera) UnitInterval
```

vs

```
diff --git a/cardano-api/src/Cardano/Api/LedgerState.hs b/cardano-api/src/Cardano/Api/LedgerState.hs
index 493594e8b..0be10b221 100644
--- a/cardano-api/src/Cardano/Api/LedgerState.hs
+++ b/cardano-api/src/Cardano/Api/LedgerState.hs
@@ -1494,8 +1494,7 @@ obtainDecodeEpochStateConstraints ShelleyBasedEraBabbage f = f
 -- | Return the slots at which a particular stake pool operator is
 -- expected to mint a block.
 currentEpochEligibleLeadershipSlots :: forall era ledgerera.
-     ShelleyLedgerEra era ~ ledgerera
-  => Ledger.Era ledgerera
+     Ledger.Era ledgerera
   => Consensus.PraosProtocolSupportsNode (Api.ConsensusProtocol era)
   => HasField "_d" (Core.PParams ledgerera) UnitInterval
   -- => Crypto.Signable (Crypto.VRF (Ledger.Crypto ledgerera)) Shelley.Spec.Seed
```

cardano-testnet/src/Test/Assert.hs
newhoggy: I had forgotten we use `stylish-haskell` since moving onto my M1 Macbook where it isn't set up.  😅

Hopefully this is all sorted now?

newhoggy:
@Jimbo4350 I cleaned up all the imports in the git history and also cleaned it up.  Is there anything you want out of the git history?

cardano-testnet/src/Testnet/Babbage.hs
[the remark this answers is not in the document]
Jimbo4350: Ok fair enough. Can you create an issue to track this?

cardano-testnet/testnet/Testnet/Commands/Babbage.hs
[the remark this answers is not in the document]
Jimbo4350: Ok :+1: . Likewise lets create an issue to track this.

cardano-testnet/cardano-testnet.cabal
Jimbo4350: Let's create an issue to track this. We should probably label these issues with a `cardano-tesnet` label.

cardano-testnet/src/Testnet/Babbage.hs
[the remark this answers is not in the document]
Jimbo4350: I think I need to update the style doc because usually we import qualified with `Aeson.x` and I don't want the style to diverge further.

cardano-testnet/src/Testnet/Cardano.hs
Jimbo4350: Ah I now understand the comment. The pool operator keys were created during genesis creation.

cardano-api/src/Cardano/Api/LedgerState.hs
Jimbo4350: Fair point. We can adopt this style :+1:

cardano-testnet/test/Spec/Cli/Babbage/LeadershipSchedule.hs
Jimbo4350: You could write it as follows:
```
        (fmap
          . J.rewriteArrayElements
            . J.rewriteObject
               . HM.insert "scFormat" $ case nodeLoggingFormat testnetOptions of
                                        NodeLoggingFormatAsJson -> "ScJson"
                                        NodeLoggingFormatAsText -> "ScText")
 ```
 The problem I have with `BlockArguments` is that it makes is more difficult to discern between monadic actions and function composition. For this reason I avoid using it.

File: cardano-testnet/src/Test/Assert.hs [hunk 21]
```diff
@@ -0,0 +1,104 @@
...
+
+assertChainExtended :: (H.MonadTest m, MonadIO m)
+  => DTC.UTCTime
+  -> NodeLoggingFormat
+  -> FilePath
+  -> m ()
+assertChainExtended deadline nodeLoggingFormat nodeStdoutFile =
+  H.assertByDeadlineIOCustom "Chain not extended" deadline $ do
+    case nodeLoggingFormat of
+      NodeLoggingFormatAsText -> IO.fileContains "Chain extended, new tip" nodeStdoutFile
```
Jimbo4350, line 63: Out of curiosity, do we want to maintain the parsing of plain text?

cardano-testnet/src/Testnet/Babbage.hs
```diff
@@ -0,0 +1,416 @@
...
+    . HM.insert "RequiresNetworkMagic" (toJSON @String "RequiresMagic")
+    . HM.insert "LastKnownBlockVersion-Major" (toJSON @Int 6)
+    . HM.insert "LastKnownBlockVersion-Minor" (toJSON @Int 0)
+    . HM.insert "TestShelleyHardForkAtEpoch" (toJSON @Int 0)
+    . HM.insert "TestAllegraHardForkAtEpoch" (toJSON @Int 0)
+    . HM.insert "TestMaryHardForkAtEpoch" (toJSON @Int 0)
+    . HM.insert "TestAlonzoHardForkAtEpoch" (toJSON @Int 0)
+    . HM.insert "TestBabbageHardForkAtEpoch" (toJSON @Int 0)
+    . HM.insert "TestEnableDevelopmentHardForkEras" (toJSON True)
+    . HM.insert "TestEnableDevelopmentNetworkProtocols" (toJSON True)
```
Jimbo4350, line 173: Supposedly `TestEnableDevelopmentNetworkProtocols` is deprecated. Can you check with the network team?

cardano-testnet/src/Test/Runtime.hs
```diff
@@ -0,0 +1,116 @@
...
+  , poolNodeSprocket :: Sprocket
+  , poolNodeStdinHandle :: IO.Handle
+  , poolNodeStdout :: FilePath
+  , poolNodeStderr :: FilePath
+  , poolNodeProcessHandle :: IO.ProcessHandle
+  , poolNodeKeys :: PoolNodeKeys
+  }
+
+data PoolNodeKeys = PoolNodeKeys
+  { poolNodeKeysColdVkey :: FilePath
```
Jimbo4350, line 66: It  would be nice to have the actual `cardano-api` types for the keys here but not for this PR (if this is actually possible). Why? Just to separate out as much of the IO as possible.

cardano-testnet/test/Main.hs
```diff
@@ -5,24 +5,33 @@ module Main
...
+      [ H.ignoreOnMacAndWindows "leadership-schedule" Spec.Cli.Alonzo.LeadershipSchedule.hprop_leadershipSchedule
+      ]
+    , T.testGroup "Babbage"
+      [ H.ignoreOnMacAndWindows "leadership-schedule" Spec.Cli.Babbage.LeadershipSchedule.hprop_leadershipSchedule
+      ]
       -- Ignored on Windows due to <stdout>: commitBuffer: invalid argument (invalid character)
       -- as a result of the kes-period-info output to stdout.
       -- TODO: Babbage temporarily ignored due to broken protocol-state query
-      -- H.ignoreOnWindows "kes-period-info" Spec.Cli.KesPeriodInfo.hprop_kes_period_info
+    , H.disabled "kes-period-info" Spec.Cli.KesPeriodInfo.hprop_kes_period_info
```
Jimbo4350, line 34: Do we still need to keep this disabled?

cardano-testnet/test/Spec/Cli/Alonzo/LeadershipSchedule.hs
```diff
@@ -0,0 +1,527 @@
...
+import qualified Test.Assert as H
+import qualified Test.Base as H
+import qualified Test.Process as H
+import qualified Test.Runtime as TR
+import qualified Testnet.Cardano as TC
+import           Testnet.Cardano (TestnetOptions (..), TestnetRuntime (..))
+import qualified Testnet.Conf as H
+import           Testnet.Utils (waitUntilEpoch)
+
+isLinux :: Bool
```
Jimbo4350, line 60: Let's move this to the utils module in `cardano-testnet` because this has already been defined

cardano-testnet/test/Spec/Cli/Alonzo/LeadershipSchedule.hs
```diff
@@ -0,0 +1,527 @@
...
+isLinux = os == "linux"
+
+data LeadershipSlot = LeadershipSlot
```
Jimbo4350, line 63: Likewise for `LeadershipSlot`. There is no reason for there to be duplicate definitions.

cardano-testnet/test/Spec/Cli/Alonzo/LeadershipSchedule.hs
```diff
@@ -0,0 +1,527 @@
...
+
+  H.assertByDeadlineMCustom "Leader schedule is correct" leadershipDeadline $ do
+    leaderSlots <- H.getRelevantLeaderSlots (TR.poolNodeStdout poolNode1) (minimum expectedLeadershipSlotNumbers)
+    maxSlotExpected <- H.noteShow $ maximum expectedLeadershipSlotNumbers
+    maxActualSlot <- H.noteShow $ maximum leaderSlots
+    return $ maxActualSlot >= maxSlotExpected
+
+  leaderSlots <- H.getRelevantLeaderSlots (TR.poolNodeStdout poolNode1) (minimum expectedLeadershipSlotNumbers)
+
+  H.assert $ L.length (expectedLeadershipSlotNumbers \\ leaderSlots) <= 1
```
Jimbo4350, line 527: Can you leave a comment as to why it's acceptable for the difference to be <= 1?

Jimbo4350 (changes requested):
A few comments that need to be address

cardano-testnet/test/Spec/Cli/Babbage/LeadershipSchedule.hs
newhoggy: I've taken your suggestion, as it's really only two places.

Mainly I thought that `BlockArguments` has some really interesting properties and that it is a better `$` and wanted to shared the idea.

At the place I worked previously, we used block argument `do` for error handling in monadic contexts, so perhaps we can revisit this.

cardano-testnet/test/Main.hs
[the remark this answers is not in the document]
newhoggy: Uncommented both and marked them as ignored tests instead.

cardano-testnet/src/Test/Assert.hs
```diff
@@ -0,0 +1,104 @@
...
+      NodeLoggingFormatAsJson -> fileJsonGrep nodeStdoutFile $ \v ->
+                                    Aeson.parseMaybe (Aeson.parseJSON @(LogEntry Kind)) v == Just (LogEntry (Kind "TraceAddBlockEvent.AddedToCurrentChain"))
+
+newtype LogEntry a = LogEntry
+  { data_ :: a
```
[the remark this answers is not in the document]
newhoggy: `data` isn't the only field.  Just the only one I care about right now.

Do you still want this change?

cardano-testnet/src/Test/Assert.hs
newhoggy: Not necessarily, but I'm keeping it so we don't need to do a big bang migration.  Once we're satisfied we don't use it anywhere we can remove it.

cardano-testnet/test/Main.hs
newhoggy: I can't remember the context behind why this was disabled.

I can try enabling this separately after the PR is merged and see what it does.

cardano-testnet/testnet/Testnet/Commands/Babbage.hs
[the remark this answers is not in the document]
newhoggy: https://github.com/input-output-hk/cardano-node/issues/4132

cardano-testnet/cardano-testnet.cabal
newhoggy: https://github.com/input-output-hk/cardano-node/issues/4133

cardano-testnet/test/Spec/Cli/Alonzo/LeadershipSchedule.hs
newhoggy: Done

cardano-testnet/test/Spec/Cli/Alonzo/LeadershipSchedule.hs
newhoggy: Done!

cardano-testnet/test/Spec/Cli/Alonzo/LeadershipSchedule.hs
newhoggy: Done!

cardano-testnet/src/Testnet/Babbage.hs
newhoggy: Deleted.

cardano-testnet/src/Testnet/Babbage.hs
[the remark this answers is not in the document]
newhoggy: https://github.com/input-output-hk/cardano-node/issues/4134

cardano-testnet/test/Spec/Cli/Babbage/LeadershipSchedule.hs
Jimbo4350: > we used block argument do for error handling in monadic contexts

Do you have an example of this I can look at?

cardano-testnet/src/Test/Assert.hs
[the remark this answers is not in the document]
Jimbo4350: Yep, it follows the convention in the repo. If it gets updated we change the field name.

cardano-testnet/src/Test/Base.hs
```diff
@@ -1,12 +1,20 @@
...
+import           Data.Bool (Bool)
+import           Data.Eq (Eq (..))
 import           Data.Function
 import           GHC.Stack (HasCallStack)
+import           System.Info (os)
 
 import qualified Hedgehog as H
 import qualified Hedgehog.Extras.Test.Base as H
 
+
```
Jimbo4350, line 15: We need to decide which we are going to keep. `Test.Base` or `Test.Util`. If you don't want to do this in this PR, can you create an issue to track it?

cardano-testnet/test/Spec/Cli/Babbage/LeadershipSchedule.hs
```diff
@@ -0,0 +1,194 @@
...
+        else do
+          maxActualSlot <- H.noteShow $ maximum leaderSlots
+          return $ maxActualSlot >= maxSlotExpected
+
+    leaderSlots <- H.getRelevantLeaderSlots (TR.poolNodeStdout poolNode1) (minimum expectedLeadershipSlotNumbers)
+
+    H.noteShow_ expectedLeadershipSlotNumbers
+    H.noteShow_ leaderSlots
+
+    -- Next leadership schedule matches reality
```
Jimbo4350, line 193: Sorry but why?

Jimbo4350 (approved):
LGTM! Good work @newhoggy

cardano-testnet/src/Test/Base.hs
newhoggy: https://github.com/input-output-hk/cardano-node/issues/4148

cardano-testnet/test/Spec/Cli/Babbage/LeadershipSchedule.hs
newhoggy: I do, but I'll figure out an appropriate time to share it.  It can wait until some urgent features are done.

cardano-testnet/src/Test/Assert.hs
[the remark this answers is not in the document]
newhoggy: Updated.

cardano-testnet/test/Spec/Cli/Babbage/LeadershipSchedule.hs
newhoggy: Updated comment.

cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -1412,49 +1422,53 @@ nextEpochEligibleLeadershipSlots sbe sGen serCurrEpochState ptclState
...
-  certifiedNaturalValue
-    :: SlotNo
-    -> Consensus.Nonce
-    -> Crypto.SignKeyVRF v
-    -> Crypto.OutputVRF v
-  certifiedNaturalValue slot epochNonce vrfSkey' =
-   Crypto.certifiedOutput
-   $ Crypto.evalCertified () (TPraos.mkSeed TPraos.seedL slot epochNonce) vrfSkey'
+  -> Either LeadershipError (Set SlotNo)
+isLeadingSlotsTPraos slotRangeOfInterest poolid snapshotPoolDistr eNonce vrfSkey activeSlotCoeff' = do
```
nfrisby, line 1436: Nice simplifications! Thanks for discerning which parts of the Consensus API you could use here 👍

cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -1521,16 +1533,21 @@ currentEpochEligibleLeadershipSlots sbe sGen eInfo pParams ptclState
...
+  let setSnapshotPoolDistr :: Map (SL.KeyHash 'SL.StakePool Shelley.StandardCrypto) (SL.IndividualPoolStake Shelley.StandardCrypto)
       setSnapshotPoolDistr = ShelleyAPI.unPoolDistr . ShelleyAPI.calculatePoolDistr
                                 . ShelleyAPI._pstakeSet . obtainIsStandardCrypto sbe
                                 $ ShelleyAPI.esSnapshots cEstate
 
-  relativeStake :: Rational <- maybe (Left $ LeaderErrStakePoolHasNoStake poolid)
-                         (Right . ShelleyAPI.individualPoolStake)
-                         (Map.lookup poolHash setSnapshotPoolDistr)
+  let slotRangeOfInterest = Set.filter
+        (not . Ledger.isOverlaySlot firstSlotOfEpoch (getField @"_d" (toLedgerPParams sbe pParams)))
```
nfrisby, line 1542: I think we (eventually?) won't have `d` at all for Babbage/Praos, right? Will this require an update then?

JaredCorduan: There is no longer an overlay schedule in Praos, so this line could be simplified to just return the set of slots in the given epoch.

(As for removing `d`, it is not in the Babbage era protocol parameters, but the `HasField` instance still [exists](https://github.com/input-output-hk/cardano-ledger/blob/2505b7f103c78ee5a230b3e302e31d9607fffbd5/eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs#L460-L464), sadly, since `d` has its tendrils all throughout the reward calculation and we chose not to redo all that code.)

newhoggy: Since we need to keep it for `TPraos` anyway, I'm happy to keep this code for the moment and be guided by the type errors if/when the `d` parameters get deleted for Babbage onwards.

PR #4109, opened 2022-06-29
New faucet

newhoggy:


cardano-cli/src/Cardano/CLI/Faucet.hs
```diff
@@ -0,0 +1,609 @@
...
+      !(SigningKey ByronKey)
+      -- ^ Byron signing key.
+      !(Maybe (Address ByronAddr))
+      -- ^ An optionally specified Byron address.
+      --
+      -- If specified, both the network ID and derivation path are extracted
+      -- from the address and used in the construction of the Byron witness.
+
+-- | Some kind of Byron or Shelley witness.
+data ByronOrShelleyWitness
```
Jimbo4350, line 75: Please don't duplicate data definitions. Find a way to import them.

Jimbo4350: Ignore this, I just read that the code isn't going to live in the cli anyways

PR #4129, opened 2022-07-04
cardano-node: NodeStartupInfo as datapoint.

denisshevchenko:
There is an important information about the node, such as era name, slot length, epoch length and KES period length. This information is required for RTView, and it's convenient to provide it as a datapoint, because RTView may ask for it not only during start.


deepfire (approved):
Thank you @denisshevchenko !

JaredCorduan (approved):

deepfire:
Obsoleted by https://github.com/input-output-hk/cardano-node/pull/4105

PR #4130, opened 2022-07-04
cardano-api: Drop ToJSON Script (ledger) instance

ch1bo:
This orphan instance is not total and seemingly unused. It makes usage
of the cardano-api library very hard if you do not agree with having all
Alonzo.Script values be serialized as only their hashes.

CC co-author @abailly-iohk

### DIFF
--- a/cardano-api/src/Cardano/Api/Orphans.hs
+++ b/cardano-api/src/Cardano/Api/Orphans.hs
@@ -42,7 +42,6 @@ import           Cardano.Slotting.Time (SystemStart (..))
 
 import qualified Cardano.Crypto.Hash.Class as Crypto
 import qualified Cardano.Ledger.Alonzo.Data as Alonzo
-import qualified Cardano.Ledger.Alonzo.Scripts as Alonzo
 import qualified Cardano.Ledger.Babbage.PParams as Babbage
 import qualified Cardano.Ledger.Babbage.TxBody as Babbage
 import qualified Cardano.Ledger.Coin as Shelley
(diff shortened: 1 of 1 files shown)


ch1bo:
It seems it's used in the cardano-cli though over here in the context of some `DebugLedgerState`: https://github.com/input-output-hk/cardano-node/blob/08e043a62476a69e04a0c5baa97e415e0efe1212/cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs#L1337

If that is the only use for these orphan instances, then I'm happily moving them to `cardano-cli` if that means we don't have them in `cardano-api`.

ch1bo:
See #4138 for an alternative to this, which I would support instead. Still happy to have the discussion about these orphan instances.

Jimbo4350:
> See #4138 for an alternative to this, which I would support instead. Still happy to have the discussion about these orphan instances.

I'd say move it over to the cli in this PR and I can approve @ch1bo

abailly-iohk:
Actually, this instance is used in the node. I got the following error when trying to move it to cardano-cli in #4138:

```
src/Cardano/Node/Protocol/Cardano.hs:133:7: error:
    • No instance for (ToJSON
                         (Cardano.Ledger.Alonzo.Scripts.Script
                            (Cardano.Ledger.Babbage.BabbageEra
                               Cardano.Ledger.Crypto.StandardCrypto)))
        arising from a use of ‘SomeConsensusProtocol’
    • In the expression: SomeConsensusProtocol CardanoBlockType
```

I would say, "done is better than perfect" so let's merge #4138 if no one has strong opinions against it.

ch1bo:
Won't be having time to work on this for the time being and we merged the alternative of fixing the `ToJSON Script` instance to be complete.

PR #4141, opened 2022-07-06
Add utxoCostPerByte protocol parameter

newhoggy:
Resolves https://github.com/input-output-hk/cardano-node/issues/4052
See https://github.com/cardano-foundation/CIPs/pull/265

### DIFF
--- a/cardano-api/gen/Gen/Cardano/Api/Typed.hs
+++ b/cardano-api/gen/Gen/Cardano/Api/Typed.hs
@@ -765,6 +765,7 @@ genProtocolParameters =
     <*> Gen.maybe genNat
     <*> Gen.maybe genNat
     <*> Gen.maybe genNat
+    <*> Gen.maybe genLovelace
 
 genProtocolParametersUpdate :: Gen ProtocolParametersUpdate
 genProtocolParametersUpdate = do
@@ -795,6 +796,7 @@ genProtocolParametersUpdate = do
   protocolUpdateMaxValueSize        <- Gen.maybe genNat
   protocolUpdateCollateralPercent   <- Gen.maybe genNat
   protocolUpdateMaxCollateralInputs <- Gen.maybe genNat
+  protocolUpdateUTxOCostPerByte     <- Gen.maybe genLovelace
   pure ProtocolParametersUpdate{..}
 
 
--- a/cardano-api/src/Cardano/Api/Fees.hs
+++ b/cardano-api/src/Cardano/Api/Fees.hs
@@ -793,10 +793,6 @@ data TxBodyErrorAutoBalance =
        -- parameter, for eras that use this parameter.
      | TxBodyErrorMissingParamMinUTxO
 
-       -- | The 'ProtocolParameters' must provide the value for the cost per
-       -- word parameter, for eras that use this parameter.
-     | TxBodyErrorMissingParamCostPerWord
-
        -- | The transaction validity interval is too far into the future.
        -- See 'TransactionValidityIntervalError' for details.
      | TxBodyErrorValidityInterval TransactionValidityError
@@ -855,9 +851,6 @@ instance Error TxBodyErrorAutoBalance where
 
   displayError TxBodyErrorMissingParamMinUTxO =
       "The minUTxOValue protocol parameter is required but missing"
-
-  displayError TxBodyErrorMissingParamCostPerWord =
-      "The utxoCostPerWord protocol parameter is required but missing"
 
   displayError (TxBodyErrorValidityInterval err) =
       displayError err
@@ -1259,13 +1252,9 @@ calculateMinimumUTxO era txout@(TxOut _ v _ _) pparams' =
 
 data MinimumUTxOError =
     PParamsMinUTxOMissing
-  | PParamsUTxOCostPerWordMissing
   deriving Show
 
 instance Error MinimumUTxOError where
   displayError PParamsMinUTxOMissing =
     "\"minUtxoValue\" field not present in protocol parameters when \
-    \trying to calculate minimum UTxO value."
-  displayError PParamsUTxOCostPerWordMissing =
-    "\"utxoCostPerWord\" field not present in protocol parameters when \
     \trying to calculate minimum UTxO value."
--- a/cardano-api/src/Cardano/Api/ProtocolParameters.hs
+++ b/cardano-api/src/Cardano/Api/ProtocolParameters.hs
@@ -71,7 +71,7 @@ import           Data.Bifunctor (bimap, first)
 import           Data.ByteString (ByteString)
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
-import           Data.Maybe (fromMaybe, isJust, isNothing)
+import           Data.Maybe (fromMaybe, isJust)
 import           Data.String (IsString)
 import           Data.Text (Text)
 import           GHC.Generics
@@ -100,7 +100,6 @@ import qualified Cardano.Ledger.Alonzo.PParams as Alonzo
 import qualified Cardano.Ledger.Alonzo.Scripts as Alonzo
 
 import qualified Cardano.Ledger.Babbage.PParams as Babbage
-import           Cardano.Ledger.Babbage.Translation (coinsPerUTxOWordToCoinsPerUTxOByte)
 
 import           Text.PrettyBy.Default (display)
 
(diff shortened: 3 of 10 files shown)


lehins (dismissed):
Thank you for doing this change, that is exactly what I was advocating for. There are two more things need to be done, I believe, to make this change complete:

1. Json field name needs to be adjusted:

https://github.com/input-output-hk/cardano-node/blob/0cd6878769a10097e7c780e15efb9919fb94f669/cardano-api/src/Cardano/Api/ProtocolParameters.hs#L315

and 

https://github.com/input-output-hk/cardano-node/blob/0cd6878769a10097e7c780e15efb9919fb94f669/cardano-api/src/Cardano/Api/ProtocolParameters.hs#L346

I don't know if it is OK to break backwards compatibility when it comes to json representation, but just renaming the field naturally will break it

2. `ProtocolParametersUpdate` also needs the rename:

https://github.com/input-output-hk/cardano-node/blob/0cd6878769a10097e7c780e15efb9919fb94f669/cardano-api/src/Cardano/Api/ProtocolParameters.hs#L481

Jimbo4350 (changes requested):
We should make this change backwards compatible

cardano-api/src/Cardano/Api/ProtocolParameters.hs
```diff
@@ -473,12 +474,14 @@ data ProtocolParametersUpdate =
...
        protocolUpdateTreasuryCut :: Maybe Rational,
 
-       -- Introduced in Alonzo
+       -- Introduced in Babbage,
 
-       -- | Cost in ada per word of UTxO storage.
+       -- | Cost in ada per byte of UTxO storage.
        --
-       -- /Introduced in Alonzo/
-       protocolUpdateUTxOCostPerWord :: Maybe Lovelace,
```
Jimbo4350, line 481: I think the proper way to do this would not be to modify existing Alonzo fields, but to add a new `protocolUpdateUTxOCostPerByte` field

lehins: See my [comment](https://github.com/input-output-hk/cardano-node/pull/4141#discussion_r915908627), maybe we can get away with renaming. After Babbage is released creating PPUpdates for Alonzo doesn't make much sense, or does it?

In retrospective we should not have renamed it ledger. In particular using the binary flag `17` for perWord and perByte  was a mistake, IMHO. But that ship have sailed.

newhoggy: It might for chains other than `mainnet` if we care about those?

cardano-api/src/Cardano/Api/Fees.hs
```diff
@@ -797,6 +797,10 @@ data TxBodyErrorAutoBalance =
        -- word parameter, for eras that use this parameter.
      | TxBodyErrorMissingParamCostPerWord
 
+       -- | The 'ProtocolParameters' must provide the value for the cost per
+       -- byte parameter, for eras that use this parameter.
+     | TxBodyErrorMissingParamCostPerByte
+
```
newhoggy, line 803: Both `TxBodyErrorMissingParamCostPerWord` and `TxBodyErrorMissingParamCostPerByte` appear to be unused.  Should I delete @Jimbo4350?  Or is a change needed to ensure they are used?

cardano-api/src/Cardano/Api/ProtocolParameters.hs
```diff
@@ -312,7 +312,7 @@ instance FromJSON ProtocolParameters where
         <*> o .: "poolPledgeInfluence"
         <*> o .: "monetaryExpansion"
         <*> o .: "treasuryCut"
-        <*> o .:? "utxoCostPerWord"
```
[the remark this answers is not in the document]
Jimbo4350: It would work but it would be confusing when looking at the `ProtocolParameters` type. It's better to make the changes obvious rather than having it only at the JSON deserialization level.

cardano-api/gen/Gen/Cardano/Api/Typed.hs
```diff
@@ -795,6 +795,7 @@ genProtocolParametersUpdate = do
   protocolUpdateMaxValueSize        <- Gen.maybe genNat
   protocolUpdateCollateralPercent   <- Gen.maybe genNat
   protocolUpdateMaxCollateralInputs <- Gen.maybe genNat
+  protocolUpdateUTxOCostPerByte     <- Gen.maybe genLovelace
```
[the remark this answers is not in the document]
Jimbo4350: For now yeah

cardano-api/src/Cardano/Api/ProtocolParameters.hs
```diff
@@ -1460,7 +1483,8 @@ toAlonzoPParams ProtocolParameters {
                                (Ledger.boundRational protocolParamTreasuryCut)
 
       -- New params in Alonzo:
-    , Alonzo._coinsPerUTxOWord  = toShelleyLovelace utxoCostPerWord
+    , Alonzo._coinsPerUTxOWord  = -- adapted from babbage protocol parameter field
```
Jimbo4350, line 1486: Why the comment? I don't understand.

cardano-api/src/Cardano/Api/ProtocolParameters.hs
```diff
@@ -1506,7 +1530,7 @@ toBabbagePParams ProtocolParameters {
                    protocolParamPoolPledgeInfluence,
                    protocolParamMonetaryExpansion,
                    protocolParamTreasuryCut,
-                   protocolParamUTxOCostPerWord = Just utxoCostPerWord,
+                   protocolParamUTxOCostPerByte = Just utxoCostPerByte,
```
Jimbo4350, line 1533: :+1:

cardano-api/src/Cardano/Api/ProtocolParameters.hs
```diff
@@ -1693,6 +1717,7 @@ fromAlonzoPParams
     , protocolParamMaxValueSize        = Just _maxValSize
     , protocolParamCollateralPercent   = Just _collateralPercentage
     , protocolParamMaxCollateralInputs = Just _maxCollateralInputs
+    , protocolParamUTxOCostPerByte     = Nothing
```
Jimbo4350, line 1720: Adding a comment here explaining this is deprecated in Babbage would be useful

cardano-cli/src/Cardano/CLI/Run/Friendly.hs
```diff
@@ -262,14 +263,16 @@ friendlyProtocolParametersUpdate
         ("monetary expansion" .=) . friendlyRational
     , protocolUpdateTreasuryCut <&> ("treasury expansion" .=) . friendlyRational
     , protocolUpdateUTxOCostPerWord <&>
-        ("UTxO storage cost per unit" .=) . friendlyLovelace
+        ("UTxO storage cost per word" .=) . friendlyLovelace
```
Jimbo4350, line 266: :+1:

cardano-api/src/Cardano/Api/Fees.hs
```diff
@@ -1260,6 +1267,7 @@ calculateMinimumUTxO era txout@(TxOut _ v _ _) pparams' =
 data MinimumUTxOError =
     PParamsMinUTxOMissing
   | PParamsUTxOCostPerWordMissing
+  | PParamsUTxOCostPerByteMissing
```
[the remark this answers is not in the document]
Jimbo4350: You can remove them both :+1:

Jimbo4350 (approved):

Jimbo4350:
LGTM, minor suggestions

cardano-api/src/Cardano/Api/ProtocolParameters.hs
newhoggy: Left over comment from when I previously did the byte to work conversion.  Deleting.

cardano-cli/test/data/golden/alonzo/transaction-view.out
```diff
@@ -18,7 +18,7 @@ update proposal:
   updates:
   - genesis key hash: 1bafa294233a5a7ffbf539ae798da0943aa83d2a19398c2d0e5af114
     update:
-      UTxO storage cost per unit: 194 Lovelace
+      UTxO storage cost per word: 194 Lovelace
```
newhoggy, line 21: Note this golden file has changed.

cardano-api/src/Cardano/Api/Fees.hs
newhoggy: Deleted

cardano-api/src/Cardano/Api/ProtocolParameters.hs
newhoggy: Added comment.

gitmachtl:
tested latest master. ok so the protocol parameters query will introduce the utxoCostPerWord parameter on the switch to alonzo like

gitmachtl:
@Jimbo4350 why wasn't this PR not included in the tag 1.35.2 ?

lehins:
@gitmachtl From what I understand 1.35.2 only includes an important bug fix that is necessary for the upcoming hard fork. All other minor bug fixes and features, which applies to `utxoCostPerByte` param, will be included in the subsequent minor release.

gitmachtl:
@lehins i just posted another one, pretty strange whats going on right now on mainnet: https://github.com/input-output-hk/cardano-node/issues/4228

PR #4152, opened 2022-07-07
Updating changelogs and bumping cabal files

LaurenceIO:
Updating changelogs with changes in 1.35.1, and bumping cabal file versions to 1.35.1


LaurenceIO:
will raise a new one

PR #4155, opened 2022-07-08
CAD-4728  Change the units for the LMDB `mapsize` option to GB and let the parameter be a number

jorisdral:
The LMDB `mapsize` option is now set by providing a "number of gigabytes". The parsing of config files, the parsing of command-line options, and relevant documentation has been updated accordingly.

### DIFF
--- a/cardano-node/src/Cardano/Node/Configuration/LedgerDB.hs
+++ b/cardano-node/src/Cardano/Node/Configuration/LedgerDB.hs
@@ -3,7 +3,7 @@
 
 module Cardano.Node.Configuration.LedgerDB (
     BackingStoreSelectorFlag(..)
-  , Gigabyte
+  , Gigabytes
   , toBytes
   , defaultLMDBLimits
   ) where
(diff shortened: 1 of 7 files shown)


cardano-node/src/Cardano/Node/Configuration/LedgerDB.hs
```diff
@@ -20,17 +24,24 @@ import           Ouroboros.Consensus.Storage.LedgerDB.HD.LMDB (LMDBLimits (..))
...
+                          -- disk space the LMDB database can fill. If not
+                          -- provided, the default of 16GB will be used.
   | InMemory
   deriving (Eq, Show)
 
+newtype Gigabyte = Gigabyte Int
+  deriving stock (Eq, Show)
+  deriving newtype (Read, Aeson.FromJSON)
+
+toBytes :: Gigabyte -> Int
```
dnadales, line 37: I wonder if we should use a `newtype` for `Bytes` so that we make clear what are we converting to.

dnadales (approved):

cardano-node/src/Cardano/Node/Configuration/LedgerDB.hs
jorisdral: I wondered myself too, but I have chosen not to -- we only use the `toBytes` function once where we require an `Int`. I have added a small comment to `toBytes` to clarify its use

```haskell
LMDBBackingStore $ maybe id (\y x -> x { lmdbMapSize = toBytes y }) newLimit defaultLMDBLimits
```

PR #4158, opened 2022-07-09
RTView: fix Mem metric name.

denisshevchenko:
Fix the Mem metric name.


deepfire (approved):
Thank you @denisshevchenko !

PR #4162, opened 2022-07-11
Improve memory usage of query pool params

newhoggy:


newhoggy:
Superceded by https://github.com/input-output-hk/cardano-node/pull/4170

PR #4170, opened 2022-07-12
New query pool-state command

newhoggy:
This obsoletes `query pool-params`.  The command runs sub-second (rather than the `1m30s` the old command takes.  It also allows multiple stake pools to be queried at a time.

* Requires https://github.com/input-output-hk/ouroboros-network/pull/3890
* Requires https://github.com/input-output-hk/cardano-ledger/pull/2908


cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
```diff
@@ -392,6 +393,7 @@ renderQueryCmd cmd =
     QueryStakeSnapshot' {} -> "query stake-snapshot"
     QueryPoolParams' {} -> "query pool-params"
     QueryKesPeriodInfo {} -> "query kes-period-info"
+    QueryPoolState' {} -> "query pool-state"
```
Jimbo4350, line 396: Why are we introducing another query that returns the same thing as the `pool-params` cli command? Surely it would be better to just modify `pool-params` to use your new more efficient query?

Jimbo4350 (dismissed):
Nice work but I think we can further simplify this and avoid adding a new command

cardano-api/src/Cardano/Api/Query.hs
```diff
@@ -768,6 +792,11 @@ fromConsensusQueryResultShelleyBased _ QueryCurrentEpochState q' r' =
     Consensus.GetCBOR Consensus.DebugEpochState -> SerialisedCurrentEpochState r'
     _                                           -> fromConsensusQueryResultMismatch
 
+fromConsensusQueryResultShelleyBased _ QueryPoolState{} q' r' =
+  case q' of
+    Consensus.GetCBOR Consensus.GetPoolState {} -> SerialisedPoolState r'
+    _                                           -> error "moomoo"
```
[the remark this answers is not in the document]
newhoggy: It does.  Left over debugging.  Fixing now.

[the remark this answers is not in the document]
newhoggy: Something that needs improvement here is that when `fromConsensusQueryResultMismatch` fails, I don't know which branch it is that failed because `fromConsensusQueryResultMismatch` is called from so many places.  There might be some opportunity to use `HasCallstack` to get a better experience here.

cardano-api/src/Cardano/Api/Shelley.hs
```diff
@@ -204,8 +204,12 @@ module Cardano.Api.Shelley
     ProtocolState(..),
     decodeProtocolState,
     SerialisedDebugLedgerState(..),
+    CurrentEpochState(..),
```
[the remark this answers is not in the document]
newhoggy: There's tension between keeping things in alphabetical order and keeping related things together and also how ordering should occur between lower and upper case.

It's unlikely to me how to make this tradeoff.  It might be worth agreeing on some rules on how to make this tradeoff.

cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
newhoggy: I experience some confusion because pool params in the ledger refers to something more specific.  What we're actually returning is the pool parameters, future pool parameters and retiring state.

The `query pool-state` command also returns slightly different format allowing for data for multiple stake pools to be returned.

```
{
    "2a20182092016edb6cb818c7e9bde99d03a571a1c724ea6e4c3eb8f3" {
        "futurePoolParams": {..},
        "poolParams": {..},
        "retiring": {..}
    },
    "fc79a939d5986c6565fc5e21bb5dd292261ac5da564c30385b3bb8d4": {
        "futurePoolParams": {..},
        "poolParams": {..},
        "retiring": {..}
    }
}
```

This is the old format:

```
{
    "futurePoolParams": {..},
    "poolParams": {..},
    "retiring": {..}
}
```

cardano-api/src/Cardano/Api/Query.hs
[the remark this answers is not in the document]
Jimbo4350: > Something that needs improvement here is that when fromConsensusQueryResultMismatch fails

Can you paste the failure here?

cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
Jimbo4350: We can rename `pool-params` to `pool-state` and silently accept `pool-params` (via opt-parse `hidden`). Let's modify the original command we ought not to introduce a new one because it basically is returning the same information.

cardano-cli/cardano-cli.cabal
```diff
@@ -140,6 +140,7 @@ library
                       , random
                       , cardano-ledger-shelley
                       , set-algebra
+                      , small-steps
```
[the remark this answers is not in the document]
Jimbo4350: Where is this used? I can't build your branch because the ledger commit apparently does not exist.

cardano-api/src/Cardano/Api/Shelley.hs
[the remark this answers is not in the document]
Jimbo4350: Sure, let's revisit this in the future.

cardano-cli/cardano-cli.cabal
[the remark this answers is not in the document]
newhoggy: Removed.  Thanks!

cardano-api/src/Cardano/Api/Query.hs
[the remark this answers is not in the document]
newhoggy: It just says `fromConsensusQueryResult: internal query mismatch`

cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
newhoggy: Done!

cardano-api/src/Cardano/Api/Query.hs
Jimbo4350, line 798: There is still a cow in the way :cow:

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -167,7 +167,8 @@ renderShelleyQueryCmdError err =
       Text.pack $ "The stake pool associated with: " <> nodeOpCert <> " was not found. Ensure the correct KES key has been " <>
                   "specified and that the stake pool is registered. If you have submitted a stake pool registration certificate " <>
                   "in the current epoch, you must wait until the following epoch for the registration to take place."
-
+    ShelleyQueryCmdDecodeError typeName decoderError ->
```
Jimbo4350, line 170: Rather than propagate `Text` let's rename the error constructor to `ShelleyQueryCmdPoolStateDecodeError` and write a specific error message for the pool state decoding failure.

Jimbo4350 (changes requested):
Two minor fixes

releases.json
```diff
@@ -0,0 +1,1462 @@
+[
```
Jimbo4350, line 1: What is this?

cardano-api/src/Cardano/Api/Query.hs
newhoggy: Fixed.  Thanks!

releases.json
newhoggy: Removed.  Thanks!

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: Done!

Jimbo4350 (approved):
Good work! LGTM

PR #4175, opened 2022-07-13
RTView: a link from tracer's README, quick start

denisshevchenko:
The first feedback from SPO: it's hard to find a link to RTView documentation. Now the link is in `cardano-tracer`'s README.

Later (after official release) the similar link will be placed in `cardano-node`'s README as well.

Also, I've added a Quick Start section for RTView.


reqlez:
Thanks! This will help people get started easier. I had no idea what i was doing so I was trying all sorts of parameters lol

it's not that it was hard to find the RTView / cardano-tracer documentation, it's that the documentation did not have the cardano-node side of the setup examples, now it's there so that's perfect. Thanks for that.

It would be nice to also know, how to setup cardano-node tracer socket file inside the cardano-node's config.json file ( and any extra configuration options available, like Accept/Listen options ), versus running with an argument to the cardano-node command line.

When I first started reading the documentation, it referred to a "socket" but it never mentioned "tracer socket"  so i kept trying to connect to the regular cardano-node socket.

PR #4179, opened 2022-07-13
Optimise query stake snapshot command

newhoggy:
Depends on https://github.com/input-output-hk/ouroboros-network/pull/3898
Resolves https://github.com/input-output-hk/cardano-node/issues/2717

### DIFF
--- a/cardano-api/src/Cardano/Api/Orphans.hs
+++ b/cardano-api/src/Cardano/Api/Orphans.hs
@@ -19,7 +19,7 @@ module Cardano.Api.Orphans () where
 
 import           Prelude
 
-import           Data.Aeson (FromJSON (..), ToJSON (..), object, pairs, (.=))
+import           Data.Aeson (FromJSON (..), ToJSON (..), object, (.=))
 import qualified Data.Aeson as Aeson
 import           Data.Aeson.Types (ToJSONKey (..), toJSONKeyText)
 import           Data.BiMap (BiMap (..), Bimap)
@@ -70,7 +70,6 @@ import qualified Cardano.Ledger.Shelley.Rewards as Shelley
 import qualified Cardano.Ledger.Shelley.RewardUpdate as Shelley
 import           Cardano.Ledger.Val (Val)
 import qualified Ouroboros.Consensus.Shelley.Eras as Consensus
-import qualified Ouroboros.Consensus.Shelley.Ledger.Query as Consensus
 
 -- Orphan instances involved in the JSON output of the API queries.
 -- We will remove/replace these as we provide more API wrapper types
@@ -688,37 +687,3 @@ instance Crypto.Crypto crypto => ToJSON (VMap VB VB (Shelley.KeyHash    'Shelley
 instance Crypto.Crypto crypto => ToJSON (VMap VB VP (Shelley.Credential 'Shelley.Staking   crypto) (Shelley.CompactForm Shelley.Coin)) where
   toJSON = toJSON . fmap fromCompact . VMap.toMap
   toEncoding = toEncoding . fmap fromCompact . VMap.toMap
-
------
-
-instance ToJSON (Consensus.StakeSnapshots crypto) where
-  toJSON = object . stakeSnapshotsToPair
-  toEncoding = pairs . mconcat . stakeSnapshotsToPair
-
-stakeSnapshotsToPair :: Aeson.KeyValue a => Consensus.StakeSnapshots crypto -> [a]
-stakeSnapshotsToPair Consensus.StakeSnapshots
-    { Consensus.ssStakeSnapshots
-    , Consensus.ssMarkTotal
-    , Consensus.ssSetTotal
-    , Consensus.ssGoTotal
-    } = mconcat
-    -- Only output the first pool in order to preserve backwards compatibility of the output
-    -- format.  The output format will have to change to support multiple pools when that
-    -- functionality is added.
-    [ take 1 (Map.elems ssStakeSnapshots) >>= stakeSnapshotToPair
-    , [ "activeStakeMark" .= ssMarkTotal
-      , "activeStakeSet" .= ssSetTotal
-      , "activeStakeGo" .= ssGoTotal
-      ]
-    ]
-
-stakeSnapshotToPair :: Aeson.KeyValue a => Consensus.StakeSnapshot crypto -> [a]
-stakeSnapshotToPair Consensus.StakeSnapshot
-    { Consensus.ssMarkPool
-    , Consensus.ssSetPool
-    , Consensus.ssGoPool
-    } =
-    [ "poolStakeMark" .= ssMarkPool
-    , "poolStakeSet" .= ssSetPool
-    , "poolStakeGo" .= ssGoPool
-    ]
--- a/cardano-api/src/Cardano/Api/Orphans.hs
+++ b/cardano-api/src/Cardano/Api/Orphans.hs
@@ -6,6 +6,7 @@
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
@@ -18,7 +19,7 @@ module Cardano.Api.Orphans () where
 
 import           Prelude
 
-import           Data.Aeson (FromJSON (..), ToJSON (..), object, (.=))
+import           Data.Aeson (FromJSON (..), ToJSON (..), object, pairs, (.=))
 import qualified Data.Aeson as Aeson
 import           Data.Aeson.Types (ToJSONKey (..), toJSONKeyText)
 import           Data.BiMap (BiMap (..), Bimap)
@@ -69,6 +70,7 @@ import qualified Cardano.Ledger.Shelley.Rewards as Shelley
 import qualified Cardano.Ledger.Shelley.RewardUpdate as Shelley
 import           Cardano.Ledger.Val (Val)
 import qualified Ouroboros.Consensus.Shelley.Eras as Consensus
+import qualified Ouroboros.Consensus.Shelley.Ledger.Query as Consensus
 
 -- Orphan instances involved in the JSON output of the API queries.
 -- We will remove/replace these as we provide more API wrapper types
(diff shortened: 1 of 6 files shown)


newhoggy:
Before:

```
$ time CARDANO_NODE_SOCKET_PATH=/Users/jky/wrk/iohk/mainnet/cardano-node.socket cardano-cli query stake-snapshot --cardano-mode --mainnet --stake-pool-id fc79a939d5986c6565fc5e21bb5dd292261ac5da564c30385b3bb8d4

{
    "activeStakeGo": 24456415918186222,
    "activeStakeMark": 24453012636354912,
    "activeStakeSet": 24441961174746539,
    "poolStakeGo": 8595693051,
    "poolStakeMark": 8595693051,
    "poolStakeSet": 8595693051
}
CARDANO_NODE_SOCKET_PATH=/Users/jky/wrk/iohk/mainnet/cardano-node.socket       54.74s user 4.08s system 63% cpu 1:32.24 total
```

After:

```
$ time CARDANO_NODE_SOCKET_PATH=/Users/jky/wrk/iohk/mainnet/cardano-node.socket $(./scripts/bin-path.sh cardano-cli) query stake-snapshot --cardano-mode --mainnet --stake-pool-id fc79a939d5986c6565fc5e21bb5dd292261ac5da564c30385b3bb8d4

{
    "activeStakeGo": 24456415918186222,
    "activeStakeMark": 24453012636354912,
    "activeStakeSet": 24441961174746539,
    "poolStakeGo": 8595693051,
    "poolStakeMark": 8595693051,
    "poolStakeSet": 8595693051
}
CARDANO_NODE_SOCKET_PATH=/Users/jky/wrk/iohk/mainnet/cardano-node.socket       0.01s user 0.01s system 1% cpu 1.014 total
```

cardano-api/src/Cardano/Api/Orphans.hs
```diff
@@ -56,12 +57,13 @@ import qualified Cardano.Ledger.Shelley.Constraints as Shelley
...
 import qualified Cardano.Ledger.Shelley.LedgerState as ShelleyLedger
 import           Cardano.Ledger.Shelley.PParams (PParamsUpdate)
-import qualified Cardano.Ledger.Shelley.RewardUpdate as Shelley
 import qualified Cardano.Ledger.Shelley.Rewards as Shelley
+import qualified Cardano.Ledger.Shelley.RewardUpdate as Shelley
 import qualified Ouroboros.Consensus.Shelley.Eras as Consensus
 
 import           Cardano.Api.Script
 import           Cardano.Api.SerialiseRaw (serialiseToRawBytesHexText)
+import qualified Ouroboros.Consensus.Shelley.Ledger.Query as Consensus
```
Jimbo4350, line 66: this module does not belong here

cardano-api/src/Cardano/Api/Orphans.hs
```diff
@@ -418,3 +420,40 @@ instance Crypto.Crypto crypto => ToJSON (VMap VB VB (Shelley.KeyHash    'Shelley
...
+  toJSON
+    Consensus.StakeSnapshot
+    { Consensus.sMarkPool
+    , Consensus.sSetPool
+    , Consensus.sGoPool
+    , Consensus.sMarkTotal
+    , Consensus.sSetTotal
+    , Consensus.sGoTotal
+    } = object
+    [ "poolStakeMark" .= sMarkPool
```
Jimbo4350, line 436: Can we refactor the key pairs as you did in 4205?

Jimbo4350 (dismissed):
Couple of minor changes

cardano-api/src/Cardano/Api/Orphans.hs
newhoggy: Done!

Jimbo4350 (approved):
LGTM! Please squash the commits.

newhoggy:
Added support for querying of multiple pools.  Note, this changes the JSON structure:

Query for no pools:
```
time CARDANO_NODE_SOCKET_PATH=/Users/jky/wrk/iohk/mainnet/cardano-node.socket $(./scripts/bin-path.sh cardano-cli) query stake-snapshot --cardano-mode --mainnet
{
    "activeStakeGo": 24453012636354912,
    "activeStakeMark": 24525857119151547,
    "activeStakeSet": 24463445965412728,
    "pools": {}
}
```

Query for two pools:
```
CARDANO_NODE_SOCKET_PATH=/Users/jky/wrk/iohk/mainnet/cardano-node.socket       0.01s user 0.00s system 22% cpu 0.054 total
➜  cardano-node git:(newhoggy/optimise-query-stake-snapshot-command) ✗ time CARDANO_NODE_SOCKET_PATH=/Users/jky/wrk/iohk/mainnet/cardano-node.socket $(./scripts/bin-path.sh cardano-cli) query stake-snapshot --cardano-mode --mainnet --stake-pool-id fc79a939d5986c6565fc5e21bb5dd292261ac5da564c30385b3bb8d4 --stake-pool-id 0dedea6dd97bfd6b0d3ddf8e42375266ade6ec6a9c1aceb8efeecdab

{
    "activeStakeGo": 24453012636354912,
    "activeStakeMark": 24525857119151547,
    "activeStakeSet": 24463445965412728,
    "pools": {
        "0dedea6dd97bfd6b0d3ddf8e42375266ade6ec6a9c1aceb8efeecdab": {
            "poolStakeGo": 547234244,
            "poolStakeMark": 547234244,
            "poolStakeSet": 547234244
        },
        "fc79a939d5986c6565fc5e21bb5dd292261ac5da564c30385b3bb8d4": {
            "poolStakeGo": 8595693051,
            "poolStakeMark": 8595693051,
            "poolStakeSet": 8595693051
        }
    }
}
CARDANO_NODE_SOCKET_PATH=/Users/jky/wrk/iohk/mainnet/cardano-node.socket       0.01s user 0.01s system 1% cpu 1.298 total
```

Jimbo4350 (dismissed):
Please squash the commits

cardano-api/src/Cardano/Api/KeysShelley.hs
```diff
@@ -1188,7 +1188,9 @@ instance SerialiseAsBech32 (SigningKey StakePoolKey) where
     bech32PrefixesPermitted _ = ["pool_sk"]
 
 newtype instance Hash StakePoolKey =
-    StakePoolKeyHash (Shelley.KeyHash Shelley.StakePool StandardCrypto)
+    StakePoolKeyHash
```
Jimbo4350, line 1191: The record could exist on the same line as the constructor.

Jimbo4350 (changes requested):
Why don't we implement something similar to:
```
data QueryUTxOFilter =
     -- | /O(n) time and space/ for utxo size n
     QueryUTxOWhole

     -- | /O(n) time, O(m) space/ for utxo size n, and address set size m
   | QueryUTxOByAddress (Set AddressAny)

     -- | /O(m log n) time, O(m) space/ for utxo size n, and address set size m
   | QueryUTxOByTxIn (Set TxIn)
  deriving (Eq, Show)
  ```
  We can then account for the case where we are interested in single or multiple pool stake snapshots.

File: cardano-api/src/Cardano/Api/Orphans.hs [hunk 4]
```diff
@@ -686,3 +688,37 @@ instance Crypto.Crypto crypto => ToJSON (VMap VB VB (Shelley.KeyHash    'Shelley
...
   toJSON = toJSON . fmap fromCompact . VMap.toMap
   toEncoding = toEncoding . fmap fromCompact . VMap.toMap
+
+-----
+
+instance ToJSON (Consensus.StakeSnapshots crypto) where
+  toJSON = object . stakeSnapshotsToPair
+  toEncoding = pairs . mconcat . stakeSnapshotsToPair
+
+stakeSnapshotsToPair :: Aeson.KeyValue a => Consensus.StakeSnapshots crypto -> [a]
```
Jimbo4350, line 698: I think this logic shouldn't be in the JSON instance but at the point of rendering. The JSON instance should only be concerned with rendering JSON. In the cli we should determine how we render depending on if we get a `Map` of length 1 or more.

newhoggy: There's not a good way to do the fact there is a `Map` at all breaks backwards compatibility.

This PR is meant to preserve backwards compatibility.

A follow up PR simplifies this code but breaks backwards compatibility.  https://github.com/input-output-hk/cardano-node/pull/4279

It would be good to get an optimised version of this command merged first before merging the change that breaks compatibility.

cardano-api/src/Cardano/Api/KeysShelley.hs
newhoggy: Done.

Jimbo4350 (approved):

cardano-api/src/Cardano/Api/Orphans.hs
Jimbo4350: I've just re-discovered `data Stakes`. This is what we were rendering. What we should do to maintain backwards compatibility is create a function with type signature `Consensus.StakeSnapshots crypto -> Stakes`.  Imposing our cli logic on a JSON instance seems wrong, which should just be concerned with rendering JSON.

newhoggy: The data doesn't fit into `data Stakes` either because the map can be empty and I would not have anything to fill the pool fields in that case.

I've rendered the JSON manually instead.

cardano-api/src/Cardano/Api/Orphans.hs
```diff
@@ -691,34 +693,34 @@ instance Crypto.Crypto crypto => ToJSON (VMap VB VP (Shelley.Credential 'Shelley
 
 -----
 
-instance ToJSON (Consensus.StakeSnapshots crypto) where
+instance Crypto.Crypto crypto => ToJSON (Consensus.StakeSnapshots crypto) where
```
Jimbo4350, line 696: This instance shouldn't be necessary anymore

cardano-api/src/Cardano/Api/Orphans.hs
```diff
@@ -14,6 +14,8 @@
 
 {-# OPTIONS_GHC -Wno-orphans #-}
 {-# OPTIONS_GHC -Wno-unticked-promoted-constructors #-}
+{-# OPTIONS_GHC -Wno-unused-imports #-}
```
Jimbo4350, line 17: These should also be removable

cardano-api/src/Cardano/Api/Orphans.hs
newhoggy: Removed.

PR #4187, opened 2022-07-15
Fix `TraceNodeIsLeader` JSON parser

Jasagredo:
The log messages are formatted slightly different in each case. In particular:
```
{...,"data":{...,"kind":"TraceAddBlockEvent.AddedToCurrentChain","...},...}
```
versus
```
{"...,"data":{...,"val":{"kind":"TraceNodeIsLeader","slot":85}},...}
```

Also the parser was intended to only get the events with the kind `TraceNodeIsLeader`.

### DIFF
--- a/cardano-testnet/src/Test/Assert.hs
+++ b/cardano-testnet/src/Test/Assert.hs
@@ -10,8 +10,7 @@ module Test.Assert
   , getRelevantLeaderSlots
   ) where
 
-import           Control.Applicative ((<*>))
-import           Control.Monad (Monad (..))
+import           Control.Monad (Monad (..), fail)
 import           Control.Monad.IO.Class (MonadIO)
 import           Control.Monad.Trans.Reader (ReaderT)
 import           Control.Monad.Trans.Resource (ResourceT)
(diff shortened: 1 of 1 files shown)


Jasagredo:
Converted to draft as this doesn't seem to happen on master suprisingly

newhoggy (approved):

Jasagredo:
bors retry

cardano-testnet/src/Test/Assert.hs
```diff
@@ -86,10 +85,11 @@ data TraceNodeIsLeader = TraceNodeIsLeader
...
 instance FromJSON TraceNodeIsLeader where
-  parseJSON = Aeson.withObject "TraceNodeIsLeader" $ \v ->
-    TraceNodeIsLeader
-      <$> v .: "kind"
-      <*> v .: "slot"
+  parseJSON = Aeson.withObject "TraceNodeIsLeader" $ \v -> do
+    k <- v .: "val" >>= (.: "kind")
+    if k == "TraceNodeIsLeader"
+    then TraceNodeIsLeader k <$> (v .: "val" >>= (.: "slot"))
+    else fail "Not the right kind"
```
Jimbo4350, line 92: @Jasagredo @newhoggy What is the failure "Not the right kind" going to indicate to the user? What is not the right kind? The error message says pretty much nothing. 

@Jasagredo can you improve this message to something like `"Expected TraceNodeIsLeader but got: " <> k`

Jasagredo: Sure thing. This is only used in tests though, but yeah, a more informative error message would be better.

Issue #4075, opened 2022-06-22
All balance in --tx-in-collateral is taken. Change should go to --change-address (build command)


CarlosLopezDeLara:
**Summary**

When phase 2 validation fails and NO --tx-out-return-collateral is given, only exact amount of collateral should be used and change should be sent to `--change-address` specified in (`cardano-cli transaction build`) 

**Current behaviour is that it falls back to Alonzo-style of taking all collateral.**

**Steps to reproduce**

1. Create a plutus script transaction that should fail phase2 validation 
2. Give a ` --tx-in-collateral` with more ADA than needed to cover the collateral
3. Do not give a `--tx-out-return-collateral`
4. Sign and submit the transaction
5. All the balance in `-tx-in-collateral` is taken

**Expected behavior**
Only the exact amount of collateral should be taken and change should be sent to `--change-address` when no `--tx-out-return-collateral` is given by the user.

**Related GH issues**
- #3939
____
**Acceptance Criteria**

- [ ] GIVEN that I'm a Dapp developer
WHEN I submit a plutus script that fails phase 2 validation
AND my transaction does not specify `--tx-out-return-collateral`
THEN Only the exact collateral is taken from ` --tx-in-collateral`
AND the excess is sent to `--change-address`

**Definition of Done**

- [ ]  Code peer-reviewed
- [ ]  Builds successfully on CI
- [ ]  Includes documentation and/or examples for the functionality
- [ ]  Log/record changes on Vnext (or similar depending on what we adopt)
- [ ]  Acceptance Criteria met
- [ ]  The new functionality is covered by dev/unit/property tests
- [ ]  Merged to master
- [ ]  Test Engineering Sign-off + E2E automated tests
- [ ]  Released

CarlosLopezDeLara:
Ledger to provide a helper function.

Jimbo4350:
This is not a bug. We are getting too trigger happy with labeling things bugs when they are feature requests!

Jimbo4350:
One caveat is you also need to not provide the `--tx-total-collateral` flag in order to benefit from automatically calculating the return collateral. Specifying either `--tx-total-collateral` or `--tx-total-collateral` will not result in the build command automatically returning the "excess" collateral to your change address.

PR #4198, opened 2022-07-19
Update build command to automatically calculate the total and return collateral values

Jimbo4350:
The build command will now automatically calculate these values and returns the excess collateral to the change address
Resolves: #4075

### DIFF
--- a/cardano-api/src/Cardano/Api/Fees.hs
+++ b/cardano-api/src/Cardano/Api/Fees.hs
@@ -50,7 +50,8 @@ import qualified Data.ByteString as BS
 import           Data.ByteString.Short (ShortByteString)
 import           Data.Map (Map)
 import qualified Data.Map as Map
-import           Data.Maybe (fromMaybe, maybeToList)
+import           Data.Maybe (catMaybes, fromMaybe, maybeToList)
+import           Data.Ratio
 import           Data.Sequence.Strict (StrictSeq (..))
 import           Data.Set (Set)
 import qualified Data.Set as Set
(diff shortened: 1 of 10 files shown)


cardano-api/src/Cardano/Api/Fees.hs
```diff
@@ -1019,10 +1031,92 @@ makeTransactionBodyAutoBalance eraInMode systemstart history pparams
...
+            -- collateral tx inputs to cover the tx
+            let txOuts = catMaybes [ Map.lookup txin utxo' | txin <- collIns]
+                totalCollateralLovelace = mconcat $ map (\(TxOut _ txOutVal _ _) -> txOutValueToLovelace txOutVal) txOuts
+                requiredCollateral@(Lovelace reqAmt) = fromIntegral colPerc * fee
+                totalCollateral = TxTotalCollateral retColSup . fromShelleyLovelace
+                                                              . Ledger.rationalToCoinViaCeiling
+                                                              $ reqAmt % 100
+                enoughCollateral = totalCollateralLovelace * 100 >= requiredCollateral
+                Lovelace amt = totalCollateralLovelace * 100 - requiredCollateral
+                returnCollateral = fromShelleyLovelace . Ledger.rationalToCoinViaCeiling $ amt % 100
```
newhoggy, line 1093: What's the `100` factor for?

cardano-api/src/Cardano/Api/Fees.hs
```diff
@@ -1019,10 +1031,92 @@ makeTransactionBodyAutoBalance eraInMode systemstart history pparams
...
+                                 (TxOut cAddr (lovelaceToTxOutValue $ Lovelace (2^(64 :: Integer)) - 1)
+                                 TxOutDatumNone ReferenceScriptNone)
+                 dummyTotCol = TxTotalCollateral retColSup (Lovelace (2^(32 :: Integer) - 1))
+             in case (txReturnCollateral, txTotalCollateral) of
+                  (rc@TxReturnCollateral{}, tc@TxTotalCollateral{}) -> (rc, tc)
+                  (rc@TxReturnCollateral{},TxTotalCollateralNone) -> (rc, dummyTotCol)
+                  (TxReturnCollateralNone,tc@TxTotalCollateral{}) -> (dummyRetCol, tc)
+                  (TxReturnCollateralNone, TxTotalCollateralNone) -> (dummyRetCol, dummyTotCol)
+
+   calcReturnAndTotalCollateral
```
newhoggy, line 1063: Please add docs and links to any supporting documentation.

newhoggy:
Can you paste/attach the commands/files you used to test?

Jimbo4350:
> Can you paste/attach the commands/files you used to test?

Removing the fields:
```
  --tx-in-collateral "$txinCollateral" \
  --tx-total-collateral 529503 \
```
from the shell script `scripts/babbage/example-babbage-script-usage.sh` is a way to test this functionality is working via a failing plutus script. We need to update the cli's tx pretty printing to include total and return collateral fields. This would be another way to inspect that the automatic balancing is working.

saratomaz (approved):
The PR pass in all our regression tests related to plutus V2.

Also new tests was added to comprove that the collateral amount charged are accorging to the collateral percentage present in protocol parameters and that the change address is now the default address to return the excess of the collaterals amount.

_Related issue on cardano-node-tests:_
https://github.com/input-output-hk/cardano-node-tests/issues/1265

cardano-api/src/Cardano/Api/Fees.hs
Jimbo4350: I updated the comments in the code

cardano-api/src/Cardano/Api/Fees.hs
Jimbo4350: Updated comments in the code

erikd (approved):

PR #4205, opened 2022-07-20
More memory efficient query ledger-state command

newhoggy:
This implements `toEncoding` of various `ToJSON` instances because it allows for more memory efficient lazy encoding.

As part of the change, the output of the query command will no longer be pretty printed.

Before uses about 16.4G:

```
$ time CARDANO_NODE_SOCKET_PATH=/Users/jky/wrk/iohk/mainnet/cardano-node.socket cardano-cli query ledger-state --mainnet > /tmp/ledger-state.bin

CARDANO_NODE_SOCKET_PATH=/Users/jky/wrk/iohk/mainnet/cardano-node.socket       302.63s user 14.17s system 91% cpu 5:47.89 total
```

After uses about 11.5G:

```
$ time CARDANO_NODE_SOCKET_PATH=/Users/jky/wrk/iohk/mainnet/cardano-node.socket $(./scripts/bin-path.sh cardano-cli) query ledger-state --mainnet > ledger-state.bin

CARDANO_NODE_SOCKET_PATH=/Users/jky/wrk/iohk/mainnet/cardano-node.socket       114.72s user 5.11s system 80% cpu 2:27.96 total
```

### DIFF
--- a/cardano-api/src/Cardano/Api/Orphans.hs
+++ b/cardano-api/src/Cardano/Api/Orphans.hs
@@ -1,10 +1,13 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE UndecidableInstances #-}
 
@@ -38,6 +41,7 @@ import qualified Cardano.Ledger.Shelley.PoolRank as Shelley
 import           Cardano.Ledger.UnifiedMap (UnifiedMap)
 import           Cardano.Slotting.Slot (SlotNo (..))
 import           Cardano.Slotting.Time (SystemStart (..))
+import           Control.State.Transition (STS (State))
 
 import qualified Cardano.Binary as CBOR
 import qualified Cardano.Crypto.Hash.Class as Crypto
@@ -56,8 +60,8 @@ import qualified Cardano.Ledger.Shelley.Constraints as Shelley
 import qualified Cardano.Ledger.Shelley.EpochBoundary as ShelleyEpoch
 import qualified Cardano.Ledger.Shelley.LedgerState as ShelleyLedger
 import           Cardano.Ledger.Shelley.PParams (PParamsUpdate)
-import qualified Cardano.Ledger.Shelley.RewardUpdate as Shelley
 import qualified Cardano.Ledger.Shelley.Rewards as Shelley
+import qualified Cardano.Ledger.Shelley.RewardUpdate as Shelley
 import qualified Ouroboros.Consensus.Shelley.Eras as Consensus
 
 import           Cardano.Api.Script
@@ -66,11 +70,14 @@ import           Cardano.Api.Script
 -- We will remove/replace these as we provide more API wrapper types
 
 instance ToJSON (Mary.Value era) where
-  toJSON (Mary.Value l ps) =
-    object
-      [ "lovelace" .= toJSON l
-      , "policies" .= toJSON ps
-      ]
+  toJSON = object . toMaryValuePairs
+  toEncoding = Aeson.pairs . mconcat . toMaryValuePairs
+
+toMaryValuePairs :: Aeson.KeyValue a => Mary.Value crypto -> [a]
+toMaryValuePairs (Mary.Value !l !ps) =
+  [ "lovelace" .= l
+  , "policies" .= ps
+  ]
 
 instance ToJSONKey Mary.AssetName where
   toJSONKey = toJSONKeyText render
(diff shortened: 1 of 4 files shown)


cardano-api/src/Cardano/Api/Orphans.hs
```diff
@@ -5,12 +5,15 @@
...
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE UndecidableInstances #-}
 
 {-# OPTIONS_GHC -Wno-orphans #-}
 {-# OPTIONS_GHC -Wno-unticked-promoted-constructors #-}
+{-# LANGUAGE BangPatterns #-}
```
Jimbo4350, line 15: These pragmas should be above

cardano-api/src/Cardano/Api/Query.hs
```diff
@@ -13,6 +13,8 @@
 -- not export any from this API. We also use them unticked as nature intended.
 {-# LANGUAGE DataKinds #-}
 {-# OPTIONS_GHC -Wno-unticked-promoted-constructors #-}
+{-# LANGUAGE BangPatterns #-}
```
Jimbo4350, line 16: These pragmas should be above and is BlockArguments actually in use here?

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -781,8 +783,8 @@ writeLedgerState :: forall era ledgerera.
 writeLedgerState mOutFile qState@(SerialisedDebugLedgerState serLedgerState) =
   case mOutFile of
     Nothing -> case decodeDebugLedgerState qState of
-                 Left bs -> firstExceptT ShelleyQueryCmdHelpersError $ pPrintCBOR bs
-                 Right ledgerState -> liftIO . LBS.putStrLn $ encodePretty ledgerState
+                Left bs -> firstExceptT ShelleyQueryCmdHelpersError $ pPrintCBOR bs
```
Jimbo4350, line 786: Should be 2 spaces

Jimbo4350 (changes requested):
Nice work :+1: . You have introduce alignment where there wasn't any in the JSON instances. Lets stick to non-alignment. I think we should adopt an approach similar to ledger where we run a script with ormolu so there is no more ambiguity of the formatting.

cardano-api/src/Cardano/Api/Query.hs
```diff
@@ -331,13 +334,31 @@ instance ( IsShelleyBasedEra era
...
+        !nesRu     = Shelley.nesRu newEpochS
+        !nesPd     = Shelley.nesPd newEpochS
+    in Aeson.pairs $ mconcat
+    [ "lastEpoch"             .= nesEL
+    , "blocksBefore"          .= nesBprev
+    , "blocksCurrent"         .= nesBcur
+    , "stateBefore"           .= nesEs
+    , "possibleRewardUpdate"  .= nesRu
+    , "stakeDistrib"          .= nesPd
+    ]
```
lehins, line 360: NewEpochState is fully strict in all its fields, so there are no reasons to force them individually,
also I ment to say it on you previous PRs we should try avoiding duplication:
```suggestion
    $ nesPairs newEpochS
  toEncoding (DebugLedgerState newEpochS) = Aeson.pairs $ mconcat $ nesPairs newEpochS

nesPairs !newEpochS =
    [ "lastEpoch" .= Shelley.nesEL newEpochS
    , "blocksBefore" .= Shelley.nesBprev newEpochS
    , "blocksCurrent" .= Shelley.nesBcur newEpochS
    , "stateBefore" .= Shelley.nesEs newEpochS
    , "possibleRewardUpdate" .= Shelley.nesRu newEpochS
    , "stakeDistrib" .= Shelley.nesPd newEpochS
    ]
```

cardano-api/src/Cardano/Api/Orphans.hs
```diff
@@ -69,8 +72,13 @@ import           Cardano.Api.SerialiseRaw (serialiseToRawBytesHexText)
...
-      [ "lovelace" .= toJSON l
-      , "policies" .= toJSON ps
+      [ "lovelace" .= l
+      , "policies" .= ps
+      ]
+  toEncoding (Mary.Value l ps) =
+    Aeson.pairs $ mconcat
+      [ "lovelace" .= l
+      , "policies" .= ps
       ]
```
lehins, line 82: Avoid duplication, as mentioned in another comment

cardano-api/src/Cardano/Api/Orphans.hs
```diff
@@ -90,58 +98,131 @@ instance ToJSON Mary.AssetName where
...
+  toJSON (Shelley.AccountState tr rs) =
+    object
+      [ "treasury" .= tr
+      , "reserves" .= rs
+      ]
+  toEncoding (Shelley.AccountState !tr !rs) =
+    Aeson.pairs $ mconcat
+      [ "treasury" .= tr
+      , "reserves" .= rs
+      ]
```
lehins, line 110: Avoid duplication as mentioned in another comment, also fields are already strict
```suggestion
  toJSON = object . accountStatePairs
  toEncoding = Aeson.pairs . mconcat . accountStatePairs

 accounStatePairs (Shelley.AccountState tr rs) =
      [ "treasury" .= tr
      , "reserves" .= rs
      ]
```

cardano-api/src/Cardano/Api/Orphans.hs
```diff
@@ -90,58 +98,131 @@ instance ToJSON Mary.AssetName where
...
+        !esNonMyopic    = Shelley.esNonMyopic eState
+    in
+    Aeson.pairs $ mconcat
+      [ "esAccountState"  .= esAccountState
+      , "esSnapshots"     .= esSnapshots
+      , "esLState"        .= esLState
+      , "esPrevPp"        .= esPrevPp
+      , "esPp"            .= esPp
+      , "esNonMyopic"     .= esNonMyopic
+      ]
```
lehins, line 142: Sam thing applies here as well

cardano-api/src/Cardano/Api/Orphans.hs
```diff
@@ -90,58 +98,131 @@ instance ToJSON Mary.AssetName where
...
+      , "delegationState" .= Shelley.lsDPState    lState
+      ]
+  toEncoding lState =
+    let !lsUTxOState  = Shelley.lsUTxOState  lState
+        !lsDPState    = Shelley.lsDPState    lState
+    in
+    Aeson.pairs $ mconcat
+      [ "utxoState"       .= lsUTxOState
+      , "delegationState" .= lsDPState
+      ]
```
lehins, line 160: And here

lehins (dismissed):
We should really try avoiding duplication. If at any point we need to refactor this code we'll have to do double amount of work. And even worse we have more room for mistakes, say if one of the duplicate implementations will get out of sync (eg. change one field name and then forget the other)

cardano-api/src/Cardano/Api/Query.hs
newhoggy: Nope.  Just for temporary type checking purposes.  Removing.

cardano-api/src/Cardano/Api/Query.hs
```diff
@@ -331,13 +333,32 @@ instance ( IsShelleyBasedEra era
...
+  toEncoding = Aeson.pairs . mconcat . toDebugLedgerStatePair
+
+toDebugLedgerStatePair ::
+  ( ShelleyLedgerEra era ~ ledgerera
+  , Consensus.ShelleyBasedEra ledgerera
+  , ToJSON (Core.PParams ledgerera)
+  , ToJSON (Core.PParamsDelta ledgerera)
+  , ToJSON (Core.TxOut ledgerera)
+  , Aeson.KeyValue a
+  ) =>DebugLedgerState era -> [a]
```
lehins, line 346: Missing a space
```suggestion
  ) => DebugLedgerState era -> [a]
```

cardano-api/src/Cardano/Api/Orphans.hs
```diff
@@ -90,58 +97,133 @@ instance ToJSON Mary.AssetName where
...
+
+toLedgerStatePairs ::
+  ( Consensus.ShelleyBasedEra era
+  , ToJSON (Core.TxOut era)
+  , ToJSON (Core.PParamsDelta era)
+  , Aeson.KeyValue a
+  ) => ShelleyLedger.LedgerState era -> [a]
+toLedgerStatePairs lState =
+  let !lsUTxOState = Shelley.lsUTxOState  lState
+      !lsDPState = Shelley.lsDPState    lState
```
lehins, line 158: Some strange misalignment
```suggestion
  let !lsUTxOState = Shelley.lsUTxOState lState
      !lsDPState = Shelley.lsDPState lState
```

lehins (approved):
Looks great, thank you! Very minor syntactic adjustments suggested

cardano-api/src/Cardano/Api/Query.hs
newhoggy: Fixed

cardano-api/src/Cardano/Api/Orphans.hs
newhoggy: Removed all alignment.

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -353,8 +353,9 @@ data TxOut ctx era = TxOut (AddressInEra    era)
                            (TxOutDatum ctx  era)
                            (ReferenceScript era)
 
-deriving instance Eq   (TxOut ctx era)
-deriving instance Show (TxOut ctx era)
+deriving instance Eq      (TxOut ctx era)
+deriving instance Generic (TxOut ctx era)
```
Jimbo4350, line 357: What's this generic instance for?

newhoggy: It was previously used, but no longer.  Removed the generic instance.

newhoggy: Yep, you're right.  This instance is unused.

Jimbo4350 (approved):
Good work! Please squash the commits.

newhoggy:
Squashed

PR #4206, opened 2022-07-20
Fix register-and-delegate-script-staking-address.sh

Jimbo4350:


scripts/babbage/staking-example/register-and-delegate-script-staking-address.sh
```diff
@@ -12,6 +12,7 @@ export CARDANO_NODE_SOCKET_PATH="${CARDANO_NODE_SOCKET_PATH:-example/main.sock}"
 export TESTNET_MAGIC="${TESTNET_MAGIC:-42}"
 export UTXO_VKEY1="${UTXO_VKEY1:-example/utxo-keys/utxo1.vkey}"
 export UTXO_SKEY1="${UTXO_SKEY1:-example/utxo-keys/utxo1.skey}"
+export UTXO_VKEY2="${UTXO_VKEY1:-example/utxo-keys/utxo2.vkey}"
```
newhoggy, line 15: Shouldn't that say `UTXO_VKEY2` instead of `UTXO_VKEY1`

Jimbo4350: Yep thanks :+1:

erikd (approved):

PR #4211, opened 2022-07-21
bump babbage to report it supports 7.1 in block header

disassembler:
Bumping to 7.1 to identify nodes that have updated to 1.35.2


Jimbo4350 (approved):
LGTM

dcoutts (changes requested):
You've changed the protocol version that the node will declare it understands when it's in the Babbage era, but you've not change the protocol version that the node will declare it understands when it's in the Alonzo era.

In the Alonzo era, nodes now need to declare they understand 7.1, not 7.0. They will _also_ declare the same for now when they move into the Babbage era, but that's just for consistency/sanity. It's what they declare now, before the Babbage HF that's really important.

dcoutts (approved):
LGTM

JaredCorduan:
for posterity, let me mention that we are **not** actually moving to protocol version `7.1`. This is a hack, but without the hack we are blind to various versions of the node that are declaring "I'm ready for version 7", but in fact have different behavior.

PR #4231, opened 2022-07-25
tx-generator cleanups

MarcFontaine:
* Merge `createChangeInEra` with `createScriptChange`.
* Make `mkUTxOScript` polymorphic on the era.
* Merge `runBenchmarkInEra` with `runPlutusBenchmark`.
  `runBenchmark` can now create transactions that contain any mix of normal and plutus inputs.
* Change `ToUTxO era = [Lovelace] -> ([TxOut CtxTx era], TxId -> [Fund])` to
  `ToUTxO era = Lovelace -> (TxOut CtxTx era, TxIx -> TxId -> Fund)`
  (Deal with every output of a TX individually. This makes it possible to create TXs with a mix of regular and Plutus outputs).

### DIFF
--- a/bench/tx-generator/src/Cardano/Benchmarking/FundSet.hs
+++ b/bench/tx-generator/src/Cardano/Benchmarking/FundSet.hs
@@ -5,6 +5,8 @@
 {-# Language GeneralizedNewtypeDeriving #-}
 {-# Language MultiParamTypeClasses #-}
 {-# Language RankNTypes #-}
+{-# Language TypeApplications #-}
+{-# Language ScopedTypeVariables #-}
 
 module Cardano.Benchmarking.FundSet
 where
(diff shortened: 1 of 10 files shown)


bench/tx-generator/src/Cardano/Benchmarking/FundSet.hs
```diff
@@ -66,6 +69,22 @@ getFundLovelace (Fund (InAnyCardanoEra _ a)) = case _fundVal a of
   TxOutAdaOnly _era l -> l
   TxOutValue _era v -> selectLovelace v
 
+
```
mgmeier, line 72: Maybe a dummy question... if we don't actually want it, what's the motivation for ruling out era transitions in this context? Would be great if the comment could provide that.

mgmeier (approved):

bench/tx-generator/src/Cardano/Benchmarking/FundSet.hs
MarcFontaine: So there are two 'era's in play: The `era` when the fund is created (this is wrapped an `InAnyCardanoEra`) and the `era` from `getFundWitness :: forall era. IsShelleyBasedEra era => Fund -> Witness WitCtxTxIn era` (That is basically when trying to spend the fund.)
`getFundWitness` must be partial (for example it makes no sense to spend funds in an earlier `era` that that not support a  specific feature). But the way it was done here simply enforced that both `era`s are the same.
That was just lazyness on my side because writing the most general possible function is clunky.
So it does not intentionally rule out era transitions but it is a feature is just not implemented at the moment.

PR #4239, opened 2022-07-28
tx-generator: fix a bug

MarcFontaine:
The bug was introduced in :  #3815.
The number of total output transactions of the tx-generator was used for the number of outputs of a single transaction.
It was not found because the tests were only run for Plutus workloads and the bug was in a non-Plutus code-branch.


mgmeier (approved):

mgmeier:
bors retry

PR #4250, opened 2022-08-01
Optimise query leadership schedule command

newhoggy:
Depends on https://github.com/input-output-hk/cardano-node/pull/4392

Resolves https://github.com/input-output-hk/cardano-node/issues/4325


CarlosLopezDeLara:
Requires https://github.com/input-output-hk/ouroboros-network/pull/3932 to be merged first.

cardano-node/src/Cardano/Node/Tracing/Tracers/P2P.hs
```diff
@@ -970,7 +970,7 @@ instance (Show versionNumber, ToJSON versionNumber, ToJSON agreedOptions)
       ]
   forMachine _dtal (TrConnectionHandlerError e err cerr) =
```
Jimbo4350, line 971: The commit message needs to be changed

cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -1385,9 +1386,10 @@ nextEpochEligibleLeadershipSlots sbe sGen serCurrEpochState ptclState poolid (Vr
                                  $ obtainDecodeEpochStateConstraints sbe
                                  $ decodeCurrentEpochState serCurrEpochState
 
-  let markSnapshotPoolDistr :: Map (SL.KeyHash 'SL.StakePool Shelley.StandardCrypto) (SL.IndividualPoolStake Shelley.StandardCrypto)
-      markSnapshotPoolDistr = ShelleyAPI.unPoolDistr . ShelleyAPI.calculatePoolDistr . ShelleyAPI._pstakeMark
-                                $ obtainIsStandardCrypto sbe $ ShelleyAPI.esSnapshots cEstate
+  let snapshot :: ShelleyAPI.SnapShot Shelley.StandardCrypto
```
Jimbo4350, line 1389: :+1:

cardano-api/src/Cardano/Api/Query.hs
```diff
@@ -246,6 +250,10 @@ data QueryInShelleyBasedEra era result where
     :: Maybe (Set PoolId)
     -> QueryInShelleyBasedEra era (SerialisedPoolState era)
 
+  QueryPoolDistr
```
Jimbo4350, line 253: Let's stick with the current style and name this `QueryPoolDistribution` (as compared to `QueryStakeDistribution`)

cardano-api/src/Cardano/Api/Query.hs
```diff
@@ -406,6 +414,20 @@ decodePoolState
   -> Either DecoderError (PoolState era)
 decodePoolState (SerialisedPoolState (Serialised ls)) = PoolState <$> decodeFull ls
 
+newtype SerialisedPoolDistr era
```
Jimbo4350, line 417: For all instances of `Distr`

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -1248,39 +1248,38 @@ runQueryLeadershipSchedule (AnyConsensusModeParams cModeParams) network
...
+              vrkSkey
+              serCurrentEpochState
+              curentEpoch
+
+        NextEpoch -> do
+          let currentEpochStateQuery = QueryInEra eInMode $ QueryInShelleyBasedEra sbe QueryCurrentEpochState
+              currentEpochQuery = QueryInEra eInMode $ QueryInShelleyBasedEra sbe QueryEpoch
+          tip <- liftIO $ getLocalChainTip localNodeConnInfo
+
+          curentEpoch <- executeQuery era cModeParams localNodeConnInfo currentEpochQuery
```
Jimbo4350, line 1275: Out of curiosity, why can't we just query the pool distribution for the next epoch?

Jimbo4350 (changes requested):
LGTM, some minor changes. Can you also include in the commit message that the enhancement is due to us sending only the pool distribution over the wire vs the entire epoch state?

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: Can you elaborate?  Do you mean don't query for the current epoch but query the pool distribution instead?

newhoggy: I simplified the code so that `curentEpoch` is computed once rather than in each case of `whichSchedule`.

cardano-node/src/Cardano/Node/Tracing/Tracers/P2P.hs
newhoggy: What's the issue with the commit message?

newhoggy: I have removed the commit.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
Jimbo4350: Ignore me, I misread the code

Jimbo4350 (approved):

PR #4258, opened 2022-08-02
Bump block header protocol version

JaredCorduan:
Alonzo will now broadcast 7.2 (even though we are actually moving to 7.0, this is to distinguish from other versions of the node that are broadcasting major version 7). Babbage will now broadcast 7.0, the version that we are actually moving to.

### DIFF
--- a/cardano-node/src/Cardano/Node/Protocol/Cardano.hs
+++ b/cardano-node/src/Cardano/Node/Protocol/Cardano.hs
@@ -198,7 +198,11 @@ mkSomeConsensusProtocolCardano NodeByronProtocolConfiguration {
           -- version that this node will declare that it understands, when it
           -- is in the Alonzo era. That is, it is the version of protocol
           -- /after/ Alonzo, i.e. Babbage.
-          alonzoProtVer = ProtVer 7 1,
+          -- NB: We are not actually transitioning to version 7.2,
+          -- this is a HACK so that we can distinguish between other
+          -- versions of the node that are broadcasting major version 7.
+          -- We intentionally broadcast 7.0 starting in Babbage.
+          alonzoProtVer = ProtVer 7 2,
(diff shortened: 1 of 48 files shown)


cardano-node/src/Cardano/Node/Protocol/Cardano.hs
```diff
@@ -207,7 +212,7 @@ mkSomeConsensusProtocolCardano NodeByronProtocolConfiguration {
           -- version that this node will declare that it understands, when it
           -- is in the Babbage era. Since Babbage is currently the last known
           -- protocol version then this is also the Babbage protocol version.
-          Praos.babbageProtVer = ProtVer 7 1,
+          Praos.babbageProtVer = ProtVer 7 0,
```
Jimbo4350, line 215: Why the bump back to 7.0?

Jimbo4350:
You also have a haddock failure I think due to the line with only `--`

cardano-node/src/Cardano/Node/Protocol/Cardano.hs
JaredCorduan: because this is a hack (see the comment above) that is only needed up until the moment of the hard fork. after the hard fork, I think it is better that the header reflect the truth.

JaredCorduan:
> You also have a haddock failure I think due to the line with only `--`

ah, that's too bad, I thought the white space there was helpful. thanks!

cardano-node/src/Cardano/Node/Protocol/Cardano.hs
JaredCorduan: "this" meaning "putting a protocol version in the block header that we don't actually intended to transition to"

JaredCorduan:
I am closing this PR since we should only be cherry picking onto this branch from master.

PR #4276, opened 2022-08-05
Add tx-mempool command to CLI

LudvikGalois:
This adds 3 commands

`cardano-cli query tx-mempool info` which prints stats for the transaction mempool for the node
`cardano-cli query tx-mempool next-tx` which returns the ID of the next transaction to be processed
`cardano-cli query tx-mempool tx-exists <txid>` which tells you whether or not a transaction is in the mempool

Closes https://github.com/input-output-hk/cardano-node/issues/3774
Closes https://github.com/input-output-hk/cardano-node/issues/4459

### DIFF
--- a/cardano-api/src/Cardano/Api/IPC.hs
+++ b/cardano-api/src/Cardano/Api/IPC.hs
@@ -84,6 +84,7 @@ import           Prelude
 
 import           Data.Void (Void)
 
+import           Data.Aeson (ToJSON, (.=), object, toJSON)
 import           Data.Bifunctor (first)
 import qualified Data.ByteString.Lazy as LBS
 import qualified Data.Map.Strict as Map
@@ -133,6 +134,7 @@ import           Cardano.Api.Modes
 import           Cardano.Api.NetworkId
 import           Cardano.Api.Protocol.Types
 import           Cardano.Api.Query
+import           Cardano.Api.Tx (getTxBody)
 import           Cardano.Api.TxBody
 
 -- ----------------------------------------------------------------------------
@@ -660,6 +662,35 @@ data LocalTxMonitoringResult mode
   | LocalTxMonitoringMempoolSizeAndCapacity
       Consensus.MempoolSizeAndCapacity
       SlotNo -- ^ Slot number at which the mempool snapshot was taken
+
+instance ToJSON (LocalTxMonitoringResult mode) where
+  toJSON result =
+    object $ case result of
+      LocalTxMonitoringTxExists tx slot ->
+          [ "exists" .= True
+          , "txId" .= tx
+          , "slot" .= slot
+          ]
+      LocalTxMonitoringTxDoesNotExist tx slot ->
+          [ "exists" .= False
+          , "txId" .= tx
+          , "slot" .= slot
+          ]
+      LocalTxMonitoringNextTx txInMode slot ->
+          [ "nextTx" .= txId
+          , "slot" .= slot
+          ]
+        where
+          txId = case txInMode of
+            Just (TxInMode tx _) -> Just $ getTxId $ getTxBody tx
+            -- TODO: support fetching the ID of a Byron Era transaction
+            _ -> Nothing
+      LocalTxMonitoringMempoolSizeAndCapacity mempool slot ->
+          [ "capacityInBytes" .= Consensus.capacityInBytes mempool
+          , "sizeInBytes" .= Consensus.sizeInBytes mempool
+          , "numberOfTxs" .= Consensus.numberOfTxs mempool
+          , "slot" .= slot
+          ]
 
 data LocalTxMonitoringQuery mode
   -- | Query if a particular tx exists in the mempool. Note that, the absence
--- a/cardano-cli/ChangeLog.md
+++ b/cardano-cli/ChangeLog.md
@@ -2,7 +2,11 @@
 
 ## vNext
 
-Default to the ledger's CDDL format for transaction body creation by removing flags `--cddl-format` and `--cli-format` from `build` and `build-raw` ([PR 4303](https://github.com/input-output-hk/cardano-node/pull/4303))
+### Features
+
+- Default to the ledger's CDDL format for transaction body creation by removing flags `--cddl-format` and `--cli-format` from `build` and `build-raw` ([PR 4303](https://github.com/input-output-hk/cardano-node/pull/4303))
+
+- Add `query tx-mempool` ([PR 4276](https://github.com/input-output-hk/cardano-node/pull/4276))
 
 ### Bugs
 
--- a/cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
+++ b/cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
@@ -371,6 +371,7 @@ data QueryCmd =
       -- ^ Node operational certificate
       (Maybe OutputFile)
   | QueryPoolState' AnyConsensusModeParams NetworkId [Hash StakePoolKey]
+  | QueryTxMempool AnyConsensusModeParams NetworkId TxMempoolQuery (Maybe OutputFile)
   deriving Show
 
 renderQueryCmd :: QueryCmd -> Text
@@ -388,6 +389,14 @@ renderQueryCmd cmd =
     QueryStakeSnapshot' {} -> "query stake-snapshot"
     QueryKesPeriodInfo {} -> "query kes-period-info"
     QueryPoolState' {} -> "query pool-state"
+    QueryTxMempool _ _ query _ -> "query tx-mempool" <> renderTxMempoolQuery query
+  where
+    renderTxMempoolQuery query =
+      case query of
+        TxMempoolQueryTxExists tx -> "tx-exists " <> serialiseToRawBytesHexText tx
+        TxMempoolQueryNextTx -> "next-tx"
+        TxMempoolQueryInfo -> "info"
+
 
 data GovernanceCmd
   = GovernanceMIRPayStakeAddressesCertificate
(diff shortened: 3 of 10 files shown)


cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
```diff
@@ -348,6 +351,19 @@ renderPoolCmd cmd =
     PoolGetId {} -> "stake-pool id"
     PoolMetadataHash {} -> "stake-pool metadata-hash"
 
+data TxMempoolQuery =
```
Jimbo4350, line 354: Lets put this in the Cardano.CLI.Types module

cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
```diff
@@ -348,6 +351,19 @@ renderPoolCmd cmd =
...
+      TxMempoolQueryTxExists TxId
+    | TxMempoolQueryNextTx
+    | TxMempoolQueryInfo
+  deriving Show
+
+renderTxMempoolQuery :: TxMempoolQuery -> Text
+renderTxMempoolQuery query =
+  case query of
+    TxMempoolQueryTxExists tx -> "tx-exists " <> pack (show tx) -- TODO render this properly
```
Jimbo4350, line 363: It's enough to render the `TxId` here instead of the entire tx. You should be able to use `serialiseToRawBytesHexText`.

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -1010,6 +1012,17 @@ pQueryCmd =
...
 
+    pQueryTxMempool :: Parser QueryCmd
+    pQueryTxMempool = QueryTxMempool
+      <$> pConsensusModeParams
+      <*> pNetworkId
+      <*> pTxMempoolQuery
+      where
+        pTxMempoolQuery :: Parser TxMempoolQuery
+        pTxMempoolQuery = subparser $
+             command "info" (info (pure TxMempoolQueryInfo) (progDesc "Mempool info"))
```
Jimbo4350, line 1023: So generally we implement these as individual parsers (combined with `(<|>)` ) that use `help`  from opt-parse. You can copy and paste the haddocks on  `LocalTxMonitoringQuery` as the help text.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -620,6 +622,52 @@ runQueryPoolState (AnyConsensusModeParams cModeParams) network poolIds = do
...
+        anyE@(AnyCardanoEra era) <- determineEra cModeParams localNodeConnInfo
+        let cMode = consensusModeOnly cModeParams
+        eInMode <- toEraInMode era cMode
+          & hoistMaybe (ShelleyQueryCmdEraConsensusModeMismatch (AnyConsensusMode cMode) anyE)
+        pure $ LocalTxMonitoringQueryTx $ TxIdInMode tx eInMode
+      TxMempoolQueryNextTx -> pure LocalTxMonitoringSendNextTx
+      TxMempoolQueryInfo -> pure LocalTxMonitoringMempoolInformation
+
+  result <- liftIO $ queryTxMonitoringLocal localNodeConnInfo localQuery
+  liftIO $ LBS.putStrLn (renderResult result)
```
Jimbo4350, line 646: We should have the ability to render this JSON to stdout or to a file

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -620,6 +622,52 @@ runQueryPoolState (AnyConsensusModeParams cModeParams) network poolIds = do
...
+  where
+    renderResult result =
+      case result of
+        LocalTxMonitoringTxExists tx slot -> encodePretty $ object
+            [ "exists" .= True
```
Jimbo4350, line 651: Rather than implementing the json logic here, let's create a  `ToJSON` instance for `LocalTxMonitoringResult`. We can still pretty print the result.

Jimbo4350 (commented):
Great work :+1: A few comments

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
LudvikGalois: I'm not completely sure what you wanted here. I think the new version matches the style that's already in the codebase for subcommands. If you want them as flags, I can change that, but since exactly one must always be passed, I think a subcommand is better here.

cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
LudvikGalois: Done

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
LudvikGalois: I've done this, but I'm still unsure of the best way to render the transaction returned by `LocalTxMonitoringNextTx`.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
LudvikGalois: Done

cardano-api/src/Cardano/Api/IPC.hs
```diff
@@ -650,6 +651,30 @@ data LocalTxMonitoringResult mode
...
+          , "txId" .= tx
+          , "slot" .= slot
+          ]
+      LocalTxMonitoringTxDoesNotExist tx slot ->
+          [ "exists" .= False
+          , "txId" .= tx
+          , "slot" .= slot
+          ]
+      LocalTxMonitoringNextTx tx slot ->
+          [ "nextTx" .= (show <$> tx) -- TODO Render this properly
```
Jimbo4350, line 668: It's fine to render a txid here, showing the entire tx is overkill.

cardano-cli/ChangeLog.md
```diff
@@ -1,5 +1,11 @@
 # Changelog for cardano-cli
 
+## vNext
+
+### Features
+
+- Add `query tx-mempool` ([PR 4276](https://github.com/input-output-hk/cardano-node/pull/4276))
```
Jimbo4350, line 7: :+1:

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -1010,6 +1012,25 @@ pQueryCmd =
...
+      <$> pConsensusModeParams
+      <*> pNetworkId
+      <*> pTxMempoolQuery
+      <*> pMaybeOutputFile
+      where
+        pTxMempoolQuery :: Parser TxMempoolQuery
+        pTxMempoolQuery = asum
+          [ subParser "info"
+            (Opt.info (pure TxMempoolQueryInfo) $
+              Opt.progDesc "Ask the server about the current mempool's capacity and sizes")
```
Jimbo4350, line 1026: Change to ask "Ask the node..." because we are querying the node

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -620,6 +622,34 @@ runQueryPoolState (AnyConsensusModeParams cModeParams) network poolIds = do
...
+        let cMode = consensusModeOnly cModeParams
+        eInMode <- toEraInMode era cMode
+          & hoistMaybe (ShelleyQueryCmdEraConsensusModeMismatch (AnyConsensusMode cMode) anyE)
+        pure $ LocalTxMonitoringQueryTx $ TxIdInMode tx eInMode
+      TxMempoolQueryNextTx -> pure LocalTxMonitoringSendNextTx
+      TxMempoolQueryInfo -> pure LocalTxMonitoringMempoolInformation
+
+  result <- liftIO $ queryTxMonitoringLocal localNodeConnInfo localQuery
+  let renderedResult = encodePretty result
+  liftIO $ LBS.putStrLn renderedResult
```
Jimbo4350, line 648: So here if there is no `OutputFile` specified, we then print to stdout. If there is an `OutputFile`, we only write JSON to disk.

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -1010,6 +1012,25 @@ pQueryCmd =
...
+          , subParser "next-tx"
+            (Opt.info (pure TxMempoolQueryNextTx) $
+              Opt.progDesc "Requests the next transaction from the current list")
```
Jimbo4350, line 1029: `...from the mempool"` or `...from the mempool's current list" would be fine

Jimbo4350 (commented):
Extremely minor changes

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
LudvikGalois: That seems straightforward, and I'll make the change, but I'm unsure why this is wanted

When would I want to do something like

```sh
cardano-cli query tx-mempool next-tx --out-file foo
```

instead of

```sh
cardano-cli query tx-mempool next-tx > foo
```

Jimbo4350: What you are saying makes sense but my suggestion mirrors the current behaviour of other cli commands.

newhoggy (approved):

Jimbo4350 (approved):
LGTM, get QA to sign off.

olgahryniuk (approved):

PR #4283, opened 2022-08-08
cardano-tracer: more strictness.

denisshevchenko:
Add more strictness to avoid space leaks.

_There are no changes in code's behavior._


deepfire (approved):
Thank you Denis!

PR #4292, opened 2022-08-10
tx-generator cleanups

MarcFontaine:
* Replace `IO` with `Monad m => m` in some types.
* reintroduce `ToUTxOList`.
* Remove legacy tx-set based `FundSet`. (Use `type FundSet = [ Fund ]` for now.)


mgmeier (approved):

MarcFontaine:
ToDo:
Fix this: With ix-sets removed there needs to be a way to split funds and reuse the change.
(that is : a splitting transaction with two target wallets).

PR #4293, opened 2022-08-10
RTView: extended epoch info.

### LINKED ISSUE #3758: RTView: extend epoch information
Currently, RTView shows only the number of the current epoch. It would be better to show the time as well:

1. When the current epoch started,
2. When it will be finished,
3. Remaining time until it will be finished.

denisshevchenko:
Closes #3758 

Now we display the number of current epoch, the time until it ends and its current progress (pct). Inspired by cardanoscan.io.


deepfire (approved):
:pray: @denisshevchenko

PR #4341, opened 2022-08-16
Rename TestEnableDevelopmentHardForkEras and TestEnableDevelopmentNetworkProtocols

LudvikGalois:
`TestEnableDevelopmentHardForkEras` has been renamed to
`ExperimentalHardForksEnabled `. An error is thrown if
`TestEnableDevelopmentHardForkEras` is used to avoid it silently being set
to `False`.

`TestEnableDevelopmentNetworkProtocols` has been renamed to
`ExperimentalProtocolsEnabled`. An error is thrown if
`TestEnableDevelopmentNetworkProtocols` is used to avoid it silently being set
to `False`.

Closes https://github.com/input-output-hk/cardano-node/issues/4043
Closes https://github.com/input-output-hk/cardano-node/issues/4470


newhoggy:
It won't be enough to just change the code.   There are some tests that use the `TestEnableDevelopmentHardForkEras` configuration that now need to do something else.  Grep over the code to find these occurrences.

LudvikGalois:
> It won't be enough to just change the code. There are some tests that use the `TestEnableDevelopmentHardForkEras` configuration that now need to do something else. Grep over the code to find these occurrences.

Fixed

cardano-node/src/Cardano/Node/Configuration/POM.hs
```diff
@@ -374,8 +385,16 @@ instance FromJSON PartialNodeConfiguration where
              }
 
       parseHardForkProtocol v = do
-        npcTestEnableDevelopmentHardForkEras
-          <- v .:? "TestEnableDevelopmentHardForkEras"
+
+        failOnRemovedFields v
+          [ RemovedField
```
Jimbo4350, line 390: :+1:

cardano-node/src/Cardano/Node/Protocol/Cardano.hs
```diff
@@ -85,12 +85,6 @@ mkSomeConsensusProtocolCardano NodeByronProtocolConfiguration {
                              npcAlonzoGenesisFileHash
                            }
                            NodeHardForkProtocolConfiguration {
-                            -- npcTestEnableDevelopmentHardForkEras,
```
Jimbo4350, line 88: So this comment is still valid. We only need to reflect the name change in the comment.

Jimbo4350 (changes requested):
One minor change

cardano-node/src/Cardano/Node/Protocol/Cardano.hs
Jimbo4350: @LudvikGalois this is not resolved. The comment is still deleted.

LudvikGalois: Because the field has been renamed. A new comment was added

Jimbo4350: Where is the new comment?

LudvikGalois: On line 88, where the old one was https://github.com/input-output-hk/cardano-node/pull/4341/files#diff-8c649864fe4b96d142cfd4c2bb9e32a9510fe737c6ad4e13e04b585ee15a6b74R88

Jimbo4350: The comment being deleted provides context as to what the flag is being used for. Without it newcomers will have a more difficult time understanding the purpose of the flag.

Jimbo4350: `--npcTestEnableDevelopmentProtVer` is not a sufficient replacement for the comment being deleted. Please restore the comment.

LudvikGalois: My mistake, I completely misunderstood what was meant by "the comment"

newhoggy:
In the diff, I see renames to `TestEnableAdvertiseDevelopmentProtVer`, but in the PR description, and change log it notes renames to `TestEnableDevelopmentProtVer`.

Can I confirm the code changes are correct and the change log and PR description need to be updated to match the code?

Jimbo4350 (approved):
LGTM!

disassembler (approved):

Issue #4235, opened 2022-07-26
[BUG] - CLI is not allowing process substitution on certain commands


gitmachtl:
**_Cardano-CLI 1.34.1_**

In the past it was no problem to use a common practice in linux to not write out files directly (if you only need them in ram) and use a process substitution -> file handler instead. 

For some commands in the CLI this is still working like (example):

**Setting up the variables**

`protocolParametersJSON=`
``` json
{
    "collateralPercentage": 150,
    "costModels": {
        "PlutusScriptV1": {
            "addInteger-cpu-arguments-intercept": 205665,
            "addInteger-cpu-arguments-slope": 812,
            "addInteger-memory-arguments-intercept": 1,
            "addInteger-memory-arguments-slope": 1,
    "....etc....etc.......",
    "stakePoolTargetNum": 500,
    "treasuryCut": 0.2,
    "txFeeFixed": 155381,
    "txFeePerByte": 44,
    "utxoCostPerWord": 34480
}
```

``` console
txBodyFile="/home/user/mytransaction.txbody"; txcnt=1; rxcnt=2; magicparam="--mainnet"; etc ...
```

**Executing the command**

```console
cardano-cli transaction calculate-min-fee \
                    --tx-body-file ${txBodyFile} \
                    --protocol-params-file <(echo "${protocolParametersJSON}") \
                    --tx-in-count ${txcnt} \
                    --tx-out-count ${rxcnt} \
                    ${magicparam} ...
```
The part i am talking about is `<(echo "${protocolParametersJSON}")`, **this is working fine for the command above**. 👍 

-----

But in other commands for example for the tx-file content, this is not working anymore:

**Setting up a variable**

`txFileJSON=`
```json
{
    "type": "Witnessed Tx AlonzoEra",
    "description": "Ledger Cddl Format",
    "cborHex": "84a5008282582055c90e5f39ee7e638da909.....64732c204d617274696e203a2d29"
}
```

**Executing the command**

``` console
cardano-cli transaction txid --tx-file <(echo "${txFileJSON}")
cardano-cli transaction submit --tx-file <(echo "${txFileJSON}") --mainnet
```

**Results into the following error** 😢 
``` console
[...]

PR #4342, opened 2022-08-16
Use openFileBlocking for reading signing keys

LudvikGalois:
When reading a key from a file, `readFile` from ByteString was used.
Unfortunately this operation doesn't play nicely with pipes. We now use
`openFileBlocking` so that we can wait for data which comes from a pipe.

Fixes https://github.com/input-output-hk/cardano-node/issues/4101

### DIFF
--- a/cardano-cli/ChangeLog.md
+++ b/cardano-cli/ChangeLog.md
@@ -1,5 +1,11 @@
 # Changelog for cardano-cli
 
+## vNext
+
+### Bugs
+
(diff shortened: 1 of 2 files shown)


newhoggy (approved):

gitmachtl:
will this also fix https://github.com/input-output-hk/cardano-node/issues/4235#issuecomment-1208312560 too?

LudvikGalois:
> will this also fix [#4235 (comment)](https://github.com/input-output-hk/cardano-node/issues/4235#issuecomment-1208312560) too?

No. This only changed to using the (pipe friendly) blocking read in the specific case of reading signing keys.

PR #4350, opened 2022-08-17
cardano-node: sync percent fixed.

### LINKED ISSUE #4345: RTView: sync percent bug
Currently, RTView displays node's sync percentage, but even if the node is fully synced, that percentage shows ~44% and not more.

denisshevchenko:
Closes #4345


deepfire (approved):
Thank you @denisshevchenko !

PR #4230, opened 2022-07-25
Updated ouroboros-network dependency

coot:

PR #4191, opened 2022-07-18
Changes for single relay P2P release.

coot:
Requires:

* [x] input-output-hk/ouroboros-network#3859 to be merged
* [x] input-output-hk/ouroboros-network#3891 to be merged
* [x] update `ouroboros-network` hash once the above PRs are merged

- Removed P2PWarningDevelopmentNetworkProtocols
- Modify P2PWarning message
- Updated ouroboros-network dependency

PR #4359, opened 2022-08-19

newhoggy:
Depends on https://github.com/input-output-hk/cardano-node/pull/4191

### DIFF
--- a/cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
+++ b/cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
@@ -438,12 +438,11 @@ instance HasSeverityAnnotation (ConnectionManagerTrace addr (ConnectionHandlerTr
       TrTerminatedConnection {}               -> Debug
       TrConnectionHandler _ ev'     ->
         case ev' of
-          TrHandshakeSuccess {}               -> Info
-          TrHandshakeClientError {}           -> Notice
-          TrHandshakeServerError {}           -> Info
-          TrError _ _ ShutdownNode            -> Critical
-          TrError _ _ ShutdownPeer            -> Info
-
+          TrHandshakeSuccess {}                     -> Info
+          TrHandshakeClientError {}                 -> Notice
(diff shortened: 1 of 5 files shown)


coot:
Are you aware of #4230?

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
```diff
@@ -1770,7 +1769,7 @@ instance (Show versionNumber, ToJSON versionNumber, ToJSON agreedOptions)
       [ "kind" .= String "HandshakeServerError"
       , "reason" .= toJSON err
       ]
-  toObject _verb (TrError e err cerr) =
+  toObject _verb (TrConnectionHandlerError e err cerr) =
     mconcat
       [ "kind" .= String "Error"
```
mgmeier, line 1774: In PR #4230 this was serialized as `"kind" .= String "ConnectionHandlerError"`. Did you decide to keep the previous serialization / is that change obsolete?

newhoggy: Thanks for the pointing this out.  Fixed now.

deepfire (approved):
Thank you @newhoggy !

coot (changes requested):
The PR #4191 **must** be merged first.  The changes that you're including from `ouroboros-network` enable `NodeToNodeV_10` (p2p). #4191 will be a basis for the initial p2p release, which we want to include as little changes from other components as possible.

nix/workbench/shell.nix
```diff
@@ -88,7 +88,6 @@ in project.shellFor {
   nativeBuildInputs = with pkgs; with haskellPackages; with cardanoNodePackages; [
     cardano-ping
     cabalWrapped
-    db-analyser
```
deepfire, line 91: @newhoggy, can we avoid removing `db-analyser` -- since that breaks the workbench?

deepfire (changes requested):

coot:
Closing in favour of #4392.

PR #4366, opened 2022-08-20
cardano-node: provide default node name

### LINKED ISSUE #4364: RTView: localhost nodes log mis-coalescence
It appears as if `cardano-tracer` does not distinguish between different nodes, when they are all identified with `localhost` -- despite the different ports.

Possibly this is due to lack of necessary identification information.

Repro:

1. `make ci`
2. look at the `tracer` directory of the run -- it only has a single subdirectory (tellingly named `localhost`), despite the cluster having two nodes

denisshevchenko:
Closes #4364

### DIFF
--- a/cardano-node/src/Cardano/Node/Startup.hs
+++ b/cardano-node/src/Cardano/Node/Startup.hs
@@ -11,6 +11,7 @@ import           Prelude
 
 import           Data.Aeson (FromJSON, ToJSON)
 import           Data.Map (Map)
+import           Data.Monoid (Last (..), getLast)
 import           Data.Text (Text, pack)
 import           Data.Time.Clock (NominalDiffTime, UTCTime)
 import           Data.Version (showVersion)
@@ -45,8 +46,9 @@ import           Ouroboros.Network.Subscription.Ip (IPSubscriptionTarget (..))
 
 import           Cardano.Api.Protocol.Types (BlockType (..), protocolInfo)
 import           Cardano.Logging
+import           Cardano.Node.Configuration.POM (NodeConfiguration (..), ncProtocol)
 import           Cardano.Node.Configuration.Socket
-import           Cardano.Node.Protocol.Types (Protocol (..), SomeConsensusProtocol (..))
+import           Cardano.Node.Protocol.Types (SomeConsensusProtocol (..))
 
 import           Cardano.Git.Rev (gitRev)
 import           Paths_cardano_node (version)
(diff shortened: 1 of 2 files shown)


deepfire (approved):
Xie xie, Denis!

cardano-node/src/Cardano/Node/Startup.hs
```diff
@@ -208,7 +210,18 @@ prepareNodeInfo ptcl (SomeConsensusProtocol whichP pForInfo) tc nodeStartTime =
...
+        -- In this case we should form node's name as "host:port", where 'host' and 'port'
+        -- are taken from '--host-addr' and '--port' CLI-parameters correspondingly.
+        let SocketConfig hostIPv4 hostIPv6 port _ = ncSocketConfig nc
+        case (getLast hostIPv4, getLast hostIPv6, getLast port) of
+          (Just ipv4, _,         Just p)  -> return . pack $ show ipv4 <> ":" <> show p
+          (Nothing,   Just ipv6, Just p)  -> return . pack $ show ipv6 <> ":" <> show p
+          (Just ipv4, _,         Nothing) -> return . pack $ show ipv4
+          (Nothing,   Just ipv6, Nothing) -> return . pack $ show ipv6
+          (Nothing,   Nothing,   Just p)  -> do h <- getHostName; return . pack $ show h <> ":" <> show p
+          _ -> pack <$> getHostName -- Nothing is specified, so just take a raw host name.
```
coot, line 224: I think this will be a bit clearer:
```suggestion
        hostName <- case (show <$> hostIPv4) <> (show <$> hostIPv6) of
          Last (Just addr) -> return addr
          Last  Nothing    -> getHostName
        return . pack $ hostName <> maybe "" ((":" ++) . show) (getLast port) 
```

denisshevchenko: IMHO, my version is easier to read... But if you think it's better - I'll commit it.

deepfire: To be frank, I also find @coot 's version clearer.

coot: There are indeed two parts:
* taking out the port part out side of `case` analysis reduces the number of cases by half
* using `Last` monoid reduces the remaining three cases to just two

coot: note: I switched the order of elements in `<>` (we use `Last` rather than `First`).

newhoggy (approved):

PR #4377, opened 2022-08-23
README:  document package localisation

deepfire:
In the README:

1. Add a ToC.
2. Fix the structure, so the ToC actually makes sense.
3. Fix some minor RST syntax issues.
4. Some very minor fixes.
5. Expand the Development section with the workbench usage info.
6. Add `cabal-plan` and `graphmod` to the workbench shell.
7. Makes the node start on Darwin, when metric collection is enabled, by fixing https://github.com/input-output-hk/cardano-node/issues/4380

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -18,7 +18,7 @@
 .. contents:: Contents
 
 *************************
-``cardano-node`` Overview
+Overview of the ``cardano-node`` repository
 *************************
 
 Integration of the `ledger <https://github.com/input-output-hk/cardano-ledger-specs>`_, `consensus <https://github.com/input-output-hk/ouroboros-network/tree/master/ouroboros-consensus>`_,
@@ -41,8 +41,12 @@ Network Configuration, Genesis and Topology Files
 
 The latest supported networks can be found at `<https://hydra.iohk.io/job/Cardano/cardano-node/cardano-deployment/latest-finished/download/1/index.html>`_
 
-How to build
-============
+****
+Obtaining `cardano-node`
+****
+
+Building from source
+====
 
 Documentation for building the node can be found `here <https://docs.cardano.org/getting-started/installing-the-cardano-node>`_.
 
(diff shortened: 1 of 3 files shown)


newhoggy (approved):

newhoggy:
I'm interested to know if the new instructions will work on a Mac (Intel or M1).  I think this should be documented to let people know whether the instructions are worth trying on these platforms.

README.rst
```diff
@@ -413,54 +412,183 @@ Byron vote submission:
...
+
+By default, all binaries originating in the ``cardano-node`` repository are available to ``cabal build`` and ``cabal run``, unless the workbench was entered using one of the pure ``*nix`` modes.  Note that in all cases, the dependencies for the workbench are supplied though Nix and have been built/tested on CI.
+
+**Dependency localisation** -or- *Cabal&Nix for painless cross-repository work*
+----
+
+Normally, to work on ``cardano-node`` dependencies in the context of the node itself, one needs to go through an expensive multi-step process, mandated by the Nix model:
+    - Make changes to a dependency (say ``ouroboros-consensus-cardano`` in the ``ouroboros-network`` repository), commit them & push to Github.
+    - Calculate & bump the pin for that dependency in ``cardano-node``, inside its ``cabal.project``.
+    - Re-enter the Nix shell in ``cardano-node``, suffering full rebuilds of everything.
```
ch1bo, line 461: What's the point of describing this workflow if the actual thing you want to document is below? Also, the language used here is sarcastic and makes the reader question why we are using `nix` if it is resulting in "suffering full rebuilds of everything".

I would suggest to frame the whole section of "Dependency localisation" in a more positive way.

deepfire: > Also, the language used here is sarcastic and makes the reader question why we are using nix if it is resulting in "suffering full rebuilds of everything".

Was not intended to be sarcastic at all -- just recognising a painful reality we all had to deal with.

But perhaps you are right -- I should find a more neutral wording.

jutaro (approved):

mgmeier:
> I'm interested to know if the new instructions will work on a Mac (Intel or M1). I think this should be documented to let people know whether the instructions are worth trying on these platforms.

I've tested both `cabal-plan` and `graphmod` build inputs for workbench successfully on an Intel Mac.

mgmeier (approved):

PR #4395, opened 2022-08-26
Link to 1.35.3 released versions rather than latest master builds

CarlosLopezDeLara:
- Old links point to the latest builds from master, now link to the released version for 1.35.3 
- Rename stake pool operations files to have a more logical sequence. before it was:

1_<Title>.md
10_<Title>.md
2_<Title>.md 
... 

which was not good because 10<Title> shows after 1<Title>. Now it is
 
01_<Title>.md 
02_<Title>.md
...
10_<Title>.md

So that 10, 11 etc show on its right place. =)


kevinhammond (changes requested):
Don't point to a specific Hydra build number, please.  This creates a maintenance headache.  
If Laurence is updating the release notes to point to builds, I suggest you point at the release page instead.

Likewise for the Windows executable.

What's the reason for renaming the files?   [I just want to be sure that nothing breaks!]

Thanks

PR #4404, opened 2022-08-30
Downgrade SimpleScripts to lowest possible version

LudvikGalois:
Fixes: https://github.com/input-output-hk/cardano-node/issues/4261


Jimbo4350:
Subsumed by #4763

PR #4408, opened 2022-08-31
Allow assembling transactions with no witnesses

LudvikGalois:
It's possible to write a script which requires no signatures, for
example

```
{
    "scripts": [
        {
            "keyHash": "df243ee05bbe556970c9f8b48088b90549b8c9e41a3380fdede62d16",
            "type": "sig"
        },
        {
            "keyHash": "452c6cc84a5430b4ff720d5692e0ce7f4d37b1f78a3bea99fdbab240",
            "type": "sig"
        },
        {
            "keyHash": "9b925a6b6462e0c7c48fc69a06d1e72d03da30655fd5af40e988140e",
            "type": "sig"
        }
    ],
    "type": "atLeast",
    "required": 0
}
```

As such, it should be supported by the CLI to assemble a transaction
with no witnesses.

Fixes: https://github.com/input-output-hk/cardano-node/issues/3835
Closes: https://github.com/input-output-hk/cardano-node/issues/4471

### DIFF
--- a/cardano-cli/ChangeLog.md
+++ b/cardano-cli/ChangeLog.md
@@ -10,6 +10,7 @@
 
 - Add `query tx-mempool` ([PR 4276](https://github.com/input-output-hk/cardano-node/pull/4276))
 
+- Allow assembling transactions with no witnesses ([PR 4408](https://github.com/input-output-hk/cardano-node/pull/4408))
 
 ### Bugs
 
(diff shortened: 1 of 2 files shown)


newhoggy (approved):

Jimbo4350 (changes requested):
@LudvikGalois have you confirmed this works on a local testnet?

LudvikGalois:
> @LudvikGalois have you confirmed this works on a local testnet?

I've tested that it works in so far as the command in the ticket works and you can assemble the transaction. I didn't check if it was actually a useful thing to do, and it seems like it's not, since it's impossible to submit that transaction. I'm not even sure how to redeem from a SimpleScript - the docs are out of date.

LudvikGalois:
> @LudvikGalois have you confirmed this works on a local testnet?

I have now confirmed that this allows spending of transactions from script addresses containing a script that requires no signatures.

Jimbo4350 (approved):
LGTM, QA approval needed

PR #4411, opened 2022-09-01
Switch to Cardano haskell package repository

andreabedini:
@michaelpj and I have been working to establish an ~internal~ custom haskell package repository. This simplifes cabal's configuration and improves DX.

The "repository" is a static website that implements the same paths as hackage-server and it is hosted on [GitHub](https://github.com/input-output-hk/cardano-haskell-packages). The repository is built from a static description of its content, with the idea that teams will be able to publish a package with a simple PR, see the repository's [README](https://github.com/input-output-hk/cardano-haskell-packages/blob/main/README.md) for details.

This patch:

- Removes all source-repository-package stanzas (package repository has
  been pre-populated with the exact same versions)
- Removes cabal work-around disabling tests for packages erroneously
  considered local. Cabal now does the right thing
- Removes cabalWrapped, as now unnecessary


cabal.project
```diff
@@ -54,282 +66,10 @@ package cryptonite
   -- generation is dubious. Set the flag so we use /dev/urandom by default.
   flags: -support_rdrand
 
--- ---------------------------------------------------------
--- Disable all tests by default
-
-tests: False
+tests: True
```
andreabedini, line 69: Do we need this? The [documentation](https://cabal.readthedocs.io/en/stable/cabal-project.html#cfg-field-tests) seems to suggest it is not useful:

> Force test suites to be enabled. For most users this should not be needed, as we always attempt to solve for test suite dependencies, even when this value is False; furthermore, test suites are automatically enabled if they are requested as a built target.

michaelpj: Hmm, I think in the past it used to be necessary so cabal wouldn't re-solve for tests when you switched to running tests, perhaps they fixed that. But that's an orthogonal issue I guess.

cabal.project
```diff
@@ -1,5 +1,17 @@
+repository cardano-haskell-package-repo
```
michaelpj, line 1: Comment?

cabal.project
```diff
@@ -1,5 +1,17 @@
...
+  secure: True
+  root-keys:
+    3e0cce471cf09815f930210f7827266fd09045445d65923e6d0238a6cd15126f
+    443abb7fb497a134c343faf52f0b659bd7999bc06b7f63fa76dc99d631f9bea1
+    a86a1f6ce86c449c46666bda44268677abf29b5b2d2eb5ec7af903ec2f117a82
+    bcec67e8e99cabfa7764d75ad9b158d72bfacf70ca1d0ec8bc6b4406d1bf8413
+    c00aae8461a256275598500ea0e187588c35a5d5d7454fb57eac18d9edb86a56
+    d4a35cd3121aa00d18544bb0ac01c3e1691d618f462c46129271bccf39f7e8ee
+
 -- run `nix flake lock --update-input hackageNix` after updating index-state.
```
michaelpj, line 12: update comment to explain how to update the other one. Or maybe a combined comment about both

cabal.project
```diff
@@ -54,282 +66,10 @@ package cryptonite
...
--- Please do not put any `source-repository-package` clause above this line.
-
--- Using a fork until our patches can be merged upstream
-source-repository-package
-  type: git
-  location: https://github.com/input-output-hk/optparse-applicative
-  tag: 7497a29cb998721a9068d5725d49461f2bba0e7a
-  --sha256: 1gvsrg925vynwgqwplgjmp53vj953qyh3wbdf34pw21c8r47w35r
-
-source-repository-package
```
michaelpj, line 162: This one should remain a `source-repository-package`, it's a fork of a Hackage package, and I'm not sure it meets our criteria for being patched in the repo.

cabal.project
```diff
@@ -54,282 +66,10 @@ package cryptonite
...
-    stubs/plutus-ghc-stub
-    word-array
-
-source-repository-package
-  type: git
-  location: https://github.com/input-output-hk/ekg-forward
-  tag: 297cd9db5074339a2fb2e5ae7d0780debb670c63
-  --sha256: 1zcwry3y5rmd9lgxy89wsb3k4kpffqji35dc7ghzbz603y1gy24g
-
-source-repository-package
```
michaelpj, line 320: also this one

cabal.project
michaelpj: We should also remove it from the repo.

nix/haskell.nix
```diff
@@ -52,8 +54,6 @@ haskell-nix.cabalProject' ({ pkgs
       hlint
       ghcid
       haskell-language-server
-      cabalWrapped
```
michaelpj, line 55: also in `nix/workbench/shell.nix`

nix/haskell.nix
```diff
@@ -52,8 +54,6 @@ haskell-nix.cabalProject' ({ pkgs
       hlint
```
michaelpj, line 54: out of the diff above, but I suggest we also take this opportunity to remove all the `packages` arguments to `shellFor`, they shouldn't be needed

cabal.project
andreabedini: :+1:

nix/haskell.nix
andreabedini: :+1:

cabal.project
andreabedini: :+1:

cabal.project
andreabedini: I repeated the comment (adapted)

nix/haskell.nix
andreabedini: Sorted

cabal.project
```diff
@@ -1,5 +1,17 @@
+repository cardano-haskell-package-repo
+  url: https://input-output-hk.github.io/cardano-haskell-package-repo
+  secure: True
+  root-keys:
```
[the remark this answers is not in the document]
andreabedini: I think we won't rely on this very much. Removed for now and using its default value (also from the repo's README).

cabal.project
andreabedini: :+1:

cabal.project
andreabedini: yes, before you had to do `tests: False` and enabled them one by one for the local packages. This is now solved (which is why I left `tests: True` which was the original intent). Now I wonder if one needs to enable them at all.

michaelpj:
@kosyrevSerge there is some code in the workbench scripts that scrapes the `cabal.project` for dependency versions and is now failing:
- https://github.com/input-output-hk/cardano-node/blob/master/nix/workbench/run.sh#L691
- https://github.com/input-output-hk/cardano-node/blob/master/nix/workbench/manifest.sh#L117

We *think* that can just be replaced by using `ghc-pkg field --simple-output <package> version`, which will give the package version. That's what I think we want now that we're fetching released versions from a repository instead. Does that sound right to you?

erikd:
From the sound of the "Internal Hackage repository" its *inernal" and not accessible by people outside IOG. If that is the case, then how are people outside IOG supposed to build our software?

cabal.project
```diff
@@ -1,5 +1,19 @@
--- run `nix flake lock --update-input hackageNix` after updating index-state.
+-- Internal repository for cardano haskell packages, see CONTRIBUTING for more
```
ltouro, line 1:
```suggestion
-- Custom repository for cardano haskell packages, see CONTRIBUTING for more
```

ltouro: Foliage describe itself as a tool to create custom or private Haskell package repositories. Suggest to use the word 'custom' instead of 'internal' per erikd remarks as the repo is indeed public (https://input-output-hk.github.io/cardano-haskell-package-repo/)

erikd:
Even if it is publicly available, we need to look at at how much more difficult (if at all) the new way of accessing packages is in comparison to the old method.

michaelpj:
> From the sound of this "Internal Hackage repository" its internal and not accessible by people outside IOG. If that is the case, then how are people outside IOG supposed to build our software?

It's not. It's publicly accessible. There should be no difference for anyone building our software, this is all understood perfectly well by cabal. The most you need to do is call `cabal update`.

Let's try and avoid the "internal" descriptor. I've been trying to call it just the "Cardano Haskell package repository", which is a bit of a mouthful but hopefully not misleading.

andreabedini:
In https://github.com/input-output-hk/cardano-node/pull/4411/commits/bf585f51ab77cd953bbaba417899fa9ac4cc1a61 I have temporarily disabled parts of the workbench scripts. The script extracted the commit has of some cardano components from the `cabal.project` file, information which is not available anymore. Probably the best equivalent would be either the versions from the install plan (which AFAIU is not available when those scripts run) or just the repos and their index-state. It kinda depends on what the goal of those scripts is.

ch1bo:
> "Cardano Haskell package repository", which is a bit of a mouthful but hopefully not misleading.

We could call it `C`ardano `H`askell `P`ackage re`P`ositor`Y` -> `CHAPPY` :dog: .. unfortunately spelled wrong to be a reference to [this movie](https://en.wikipedia.org/wiki/Chappie_(film))

andreabedini:
Another options is to just drop the "repository": Cardano Haskell Packages. "Repository" is a confusing word here because it has two meaning (as in Hacakge and in git).

andreabedini:
Note to self, I made some cabal file revisions on the repository that end up breaking the build plan for cardano-node :disappointed: Do not bump the index-state until I fix the issue. :pray:

andreabedini:
> Note to self, I made some cabal file revisions on the repository that end up breaking the build plan for cardano-node disappointed Do not bump the index-state until I fix the issue. pray

We run into a bit of a roadblock with haskell.nix not liking cabal file revisions. As an emergency measure we removed them from the repository, so we can unblock this PR.

ch1bo:
> haskell.nix not liking cabal file revisions

Good haskell.nix. Allowing released code to be revised was never a good idea also IMO.

andreabedini:
> Good haskell.nix. Allowing released code to be revised was never a good idea also IMO.

cabal file revisions don't change the code, only the metadata about it so we know what version works with what other version. We can do it in a completely declarative and reproducible way so this should not be a problem.

andreabedini:
@erikd 

> Even if it is publicly available, we need to look at at how much more difficult (if at all) the new way of accessing packages is in comparison to the old method.

See [this repo](https://github.com/andreabedini/test-plutus-tx) for an example downstream consumer. It should work with just `cabal build`. Note that I have `libsecp256k1` installed in my system, I don't remember it is needed or not.

nix/workbench/manifest.sh
```diff
@@ -115,10 +115,8 @@ manifest_git_checkout_state_desc() {
 }
 
 manifest_cabal_project_dep_pin_hash() {
-    local project_file=$1/cabal.project
-    local dep=$2
-    grep "^[ ]*location: .*/${dep}\$" "${project_file}" -A1 \
-        | tail -n-1 | sed 's/^.* tag: //'
+    # FIXME
+    echo 0123456789ABCDEF0123456789ABCDEF01234567
```
deepfire, line 119: Knowing exact gitrevs for key dependencies is useful, since it simplifies discussions around reproduction.

Can we make the `cardanoHaskellPackages` (and I specifically mean the Nix flake) supply the gitrevs for all constituent packages it provides?

Then we could plug in this information here.

abailly-iohk: I maybe missing some context but if you depend on versions, is not even better? The versions are not supposed to be mutable either.

nix/workbench/run.sh
```diff
@@ -709,12 +703,12 @@ legacy_run_manifest() {
...
-      --arg Plutus   $(node_cabal_source_at $node plutus)
-      --arg Crypto   $(node_cabal_source_at $node cardano-crypto)
-      --arg Base     $(node_cabal_source_at $node cardano-base)
-      --arg Prelude  $(node_cabal_source_at $node cardano-prelude)
+      --arg Network       $(node_cabal_source_at $node ouroboros-network)
+      --arg Ledger        $(node_cabal_source_at $node cardano-ledger)
+      --arg Plutus        $(node_cabal_source_at $node plutus)
+      --arg Crypto        $(node_cabal_source_at $node cardano-crypto)
+      --arg Base          $(node_cabal_source_at $node cardano-base)
+      --arg Prelude       $(node_cabal_source_at $node cardano-prelude)
```
deepfire, line 711: Can we avoid this (seemingly) unnecessary change?

abailly-iohk: What are the formatting rules for bash scripts in this repository?

nix/workbench/shell.nix
```diff
@@ -87,7 +85,6 @@ in project.shellFor {
   # These programs will be available inside the nix-shell.
   nativeBuildInputs = with pkgs; with haskellPackages; with cardanoNodePackages; [
     cardano-ping
-    cabalWrapped
```
deepfire, line 90: Ok, this, I think, is the biggest problem I can see in this PR -- this very specifically breaks Cabal-inside-Nix-shell.

I think we agreed in the past we shouldn't break it.

nix/workbench/manifest.sh
michaelpj: I think it would be more appropriate to use the package versions? That should be enough for reproducibility, right? I think @andreabedini had a way to do that with just `ghc-pkg`, so that should be straightforward.

deepfire (changes requested):

nix/workbench/manifest.sh
deepfire: In general, this is going to be at git revision granularity -- since we don't always work with released versions of packages.

Think feature branch development, for example.

So, having gitrevs uniformly available would be a consistent approach.

Having versions _in addition_ -- would be nice as well.

nix/workbench/shell.nix
michaelpj: Can you be specific about what you think is broken? I think it all works.

nix/workbench/manifest.sh
andreabedini: I tried to update this part but I couldn't really do it without understanding its intended use. I don't think there's a way to push the gitrevs through the index. I suggest we stick to versions. @deepfire, just to make ure I understand, what are gitrevs useful for? looking/finding the right source used? We are thinking of adding html pages for the index, they could include a link to the corresponding source (maybe).

andreabedini: > In general, this is going to be at git revision granularity -- since we don't always work with released versions of packages.

right, well, perhaps we can change this script to list both the gitrevs from `source-package-repository` stanzas in `cabal.project` and normal dependencies from the plan. if you add a `source-package-repository` for some in work in progress depedency, they will override the depedency in the repository and show up here.

What do you think?

nix/workbench/shell.nix
andreabedini: cabalWrapped fixes cabal's behaviour of always recompiling source-repository-packages, this does not happen anymore since dependencies are now (mostly) from a repository.

deepfire: @michaelpj, well:

```

Comment (deepfire) [reply nix/workbench/shell.nix:90]: The `cardano-node` `master` branch has this bit of functionality which explains the basic workbench workflow: https://github.com/input-output-hk/cardano-node#workbench-a-local-cluster-playground

Comment (michaelpj) [reply nix/workbench/shell.nix:90]: Oh I see, we just need to provide `cabal` instead. I just assumed it was there!

File: nix/workbench/manifest.sh (hunk 10 above)
Comment (deepfire) [reply nix/workbench/manifest.sh:119]: > I don't think there's a way to push the gitrevs through the index.

@angerman, do you think we have no options here?

Comment (michaelpj) [reply nix/workbench/manifest.sh:119]: Why do you need the gitrevs for _released_ package versions? Getting them when you're using a `source-repository-package` seems essential indeed, but if you are using a package from the repo, surely the version is sufficient?

File: nix/workbench/shell.nix (hunk 12 above)
Comment (deepfire) [reply nix/workbench/shell.nix:90]: If both `cabal` and `cabalWrapped` were there, there'd be a clash, because both of them provide the same executable name.

Comment (deepfire) [reply nix/workbench/shell.nix:90]: @michaelpj, but a question -- what is the motivation for `cabalWrapped` removal, in the first place?

Can this PR be made to work with `cabalWrapped` still there?  How difficult would that be?

File: nix/workbench/manifest.sh (hunk 10 above)
Comment (deepfire) [reply nix/workbench/manifest.sh:119]: @michaelpj, again, this is just simpler when you deal with reproductions and general discussion of changes.

When you have the gitrev available, that's a hard source of truth.  Whereas with git tags it's sometimes not exactly a 1-1 relationship with cabal package versions, since it's managed by humans.

File: nix/workbench/shell.nix (hunk 12 above)
Comment (michaelpj) [reply nix/workbench/shell.nix:90]: It's not needed, because `cabal` just works - like it will for every other project.

Comment (andreabedini) [reply nix/workbench/shell.nix:90]: ok, we can fix this, thanks for letting us know

Comment (michaelpj) [reply nix/workbench/shell.nix:90]: I know you have a slightly different workflow, Serge, I tried to discuss that a bit here: https://docs.google.com/document/d/1PGmrvcNmfLP7HhGZpcgEee6zlHsQ-hozCC2XkjzHcbI/edit#heading=h.q6n5jsdrrm3b

If you find you still use `source-repository-package` lots then `cabalWrapped` might still be useful for you.

Comment (deepfire) [reply nix/workbench/shell.nix:90]: @michaelpj, running node clusters with feature branches is something not specific to the performance team.

I've made a reply in the document: https://docs.google.com/document/d/1PGmrvcNmfLP7HhGZpcgEee6zlHsQ-hozCC2XkjzHcbI/edit?disco=AAAAf9tvVhs

In fact, we specifically want to encourage such use -- precisely to enable people to get direct, early feedback on their code changes.

Comment (deepfire) [reply nix/workbench/shell.nix:90]: In short, I don't think we have evidence that `cabalWrapped` is necessary in this (otherwise great!) PR.

I propose we undo that change and reinstate it.

File: nix/workbench/manifest.sh (hunk 10 above)
Comment (deepfire) [reply nix/workbench/manifest.sh:119]: @andreabedini, so from what @angerman says, the information is available somewhere -- can we make it available in the flake as well?

Comment (angerman) [reply nix/workbench/manifest.sh:119]: @deepfire its in the toml file in the cardano-Haskell-packages repo.

You need to pin the repo through a nix input anyway, and we could then just read the rev from the toml file in nix from there for the package/version; if needed.

File: nix/workbench/shell.nix (hunk 12 above)
Comment (andreabedini) [reply nix/workbench/shell.nix]: I put `cabalWrapped` back. Thanks @deepfire.

File: nix/workbench/run.sh (hunk 11 above)
Comment (andreabedini) [reply nix/workbench/run.sh]: I reverted this change.

File: nix/workbench/manifest.sh (hunk 10 above)
Comment (michaelpj) [reply nix/workbench/manifest.sh:119]: > again, this is just simpler when you deal with reproductions and general discussion of changes.

I don't agree with this. Revisions are long, confusing, don't obviously have an ordering. Versions are much easier to deal with 90% of the time. If you're dealing with something unreleased, sure you need revisions, but I still really don't see how a discussion about released versions could be ambiguous when using versions.

**Comment by deepfire** [comment]
Thank you @michaelpj @andreabedini !

File: nix/workbench/manifest.sh (hunk 10 above)
Comment (deepfire) [reply nix/workbench/manifest.sh:119]: I never said versions are bad to have!

Only suggesting that losing the gitrevs is unfortunate, because gitrevs are precise and not subject to human errors and other deviations.

Having a cabal version, on the other hand, does not guarantee you a gitrev.

Comment (angerman) [reply nix/workbench/manifest.sh:119]: @deepfire while it is true that _hackage.haskell.org_ packages do not have revisions, due to the design of hackage allowing arbitrary source management, and a such just release tarball uploads, `cardano-haskell-packages` _does_ have revisions. So at least for `cardano-haskell-packages` you can go from version number to git revision e.g. [measures-0.1.0.0](https://raw.githubusercontent.com/input-output-hk/cardano-haskell-packages/main/_sources/measures/0.1.0.0/meta.toml);

@andreabedini maybe we should ensure that people don't put _tags_ as revisions in the url, as those may change over time (or resolve them at the point of index constructions).

Comment (andreabedini) [reply nix/workbench/manifest.sh:119]: @angerman At the moment we are pointing to arbitrary tarballs, when I implement a proprer github backend I can do this.

File: cabal.project [hunk 13]
```diff
@@ -54,282 +68,10 @@ package cryptonite
...
-  tag: 4ec92ded05ccf59ba4a874be4b404ac1b6d666b6
-  --sha256: 00fvvaf4ir4hskq4a6gggbh2wmdvy8j8kn6s4m1p1vlh8m8mq514
-
--- Drops an instance breaking our code. Should be released to Hackage eventually.
-source-repository-package
-  type: git
-  location: https://github.com/input-output-hk/flat
-  tag: ee59880f47ab835dbd73bea0847dab7869fc20d8
-  --sha256: 1lrzknw765pz2j97nvv9ip3l1mcpf2zr4n56hwlz0rk7wq7ls4cm
-
```
ch1bo, line 332: :heart_eyes:

cabal.project
```diff
@@ -348,11 +90,21 @@ package snap-server
...
-package cardano-ledger-alonzo-test
-  tests: False
-
 allow-newer:
   *:aeson,
   monoidal-containers:aeson,
   size-based:template-haskell,
   snap-server:attoparsec,
+
+-- Open PR upstream, maintainer unresponsive, hopefully short-lived fork.
```
shlevy, line 99: Might be nice to have documentation saying when downstream repos need these overrides as well.

andreabedini: Oh dear yes.

deepfire (approved):

cabal.project
michaelpj: Happy to add this, just not sure where. Probably this should be in the README. I'll add a section about "using the Node as a Haskell dependency".

nix/workbench/manifest.sh
deepfire: @andreabedini, in the interim it'd be nice to have `cardanoHaskellPackages` provide at least the versions in some kind of a `package-name -> version string` attrset.

michaelpj:
I think this is ready to go. I added a note about using the `cardano-node` packages from elsewhere, and consolidated some development documentation.

Jimbo4350 (changes requested):
I'm generally for this but I will take a closer look after the cli release

cabal.project
ch1bo: Could we get that info on haskell package level? i.e. likely these won't be needed if we only use `cardano-api`?

michaelpj: I don't know the answer, unfortunately.

andreabedini:
@michaelpj I suggest a stronger wording (and a clear policy)

```
--
-- Open PR upstream, maintainer unresponsive, hopefully short-lived fork.
--

-- Note: these dependencies are at project level and not at package level.
-- This means that they won't be part of the dependencies of any package
-- released from this repository and they won't be seen by downstream
-- consumers.
-- It is ok to keep these during development but it is essential that
-- they are not needed by any released package.
```

What do you think?

michaelpj:
I'm not sure if that's realistic. Maybe if we just release a patched version of `ekg-json` to CHaP using our padding scheme we can get past it. I'm unsure about `threepenny-gui`. It would be nice to enforce a "no `source-repository-package`s at release time policy, though.

andreabedini:
> It would be nice to enforce a "no `source-repository-package`s at release time policy, though.

:crossed_fingers:

michaelpj:
@Jimbo4350 if the release is going ahead off the branch, can we move forward with this on master?

CONTRIBUTING.rst
```diff
@@ -0,0 +1,190 @@
...
+----
+
+The Cabal workflow described above only extends to the repository-local packages.  Therefore, ordinarily, to work on ``cardano-node`` dependencies in the context of the node itself, one needs to go through an expensive multi-step process -- with committing, pushing and re-pinning of the dependency changes.
+
+The **dependency localisation** workflow allows us to pick a subset of leaf dependencies of the ``cardano-node`` repository, and declare them *local* -- so they can be iterated upon using the ``cabal build`` / ``cabal run`` of ``cardano-node`` itself.  This cuts development iteration time dramatically and enables effective cross-repo development of the full stack of Cardano packages.
+
+Without further ado (**NOTE**: *the order of steps is important!*):
+
+1. Ensure that your ``cardano-node`` checkout is clean, with no local modifications.  Also, ensure that you start outside the node's Nix shell.
+2. Check out the repository with the dependencies, *beside* the ``cardano-node`` checkout.  You have to check out the git revision of the dependency used by your ``cardano-node`` checkout -- as listed in ``cardano-node/cabal.project``.
```
Jimbo4350, line 135: I don't follow. Checkout the repository with the dependencies "beside" the `cardano-node` checkout?

CONTRIBUTING.rst
```diff
@@ -0,0 +1,190 @@
...
+    - we'll assume the ``ouroboros-network`` repository
+    - ..so a certain parent directory will include checkouts of both ``ouroboros-network`` and ``cardano-node``, at the same level
+    - ..and the git revision checked out in ``ouroboros-network`` will match the version of the ``ouroboros-network`` packages used currently 
+    - Extra point #1:  you can localise/check out several dependency repositories
+    - Extra point #2:  for the dependencies that are not listed in ``cabal.project`` of the node -- how do you determine the version to check out?  You can ask the workbench shell:
+         1. Temporarily enter the workbench shell
+         2. Look for the package version in ``ghc-pkg list``
+         3. Use that version to determine the git revision of the dependency's repository (using a tag or some special knowledge about the version management of said dependency).
+3. Enter the workbench shell, as per instructions in previous sections -- or just a plain Nix shell.
+4. Ensure you can build ``cardano-node`` with Cabal: ``cabal build exe:cardano-node``.  If you can't something else is wrong.
```
Jimbo4350, line 145: This seems unnecessarily complex to simply bump dependencies in the node in order to try them out. Is there a way to generate a `cabal.project` file from CHaP? I could then just use that locally until I'm satisfied the dep bumps build successfully. I do not want to use Nix.

Jimbo4350 (changes requested):
A few comments

CONTRIBUTING.rst
```diff
@@ -0,0 +1,190 @@
+********************************************
+Contributing to the ``cardano-node`` project
+********************************************
+
+The ``cardano-node`` development is primarily based on the Nix infrastructure (https://nixos.org/ ), which enables packaging, CI, development environments and deployments.
```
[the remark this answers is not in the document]
michaelpj: This isn't new: I just moved it from the README. I'll add a link though.

CONTRIBUTING.rst
michaelpj: old content

CONTRIBUTING.rst
michaelpj: old content. using s-r-p works, as noted outside this big workbench section

michaelpj:
@Jimbo4350 squashed and hydra passed.

Jimbo4350 (approved):
LGTM!

cabal.project
erikd: When using `source-repository-package` the directive `tests: True` turns on tests within all `source-repository-package` dependencies, which (especially higher up the stack) just needless runs tests in dependencies for which it should be possible to avoid re-rumning tests. 

In say `db-sync` we have `tests: False` to disable all tests and the specific stanzas to turn on tests for packages in the `db-sync` repo.

michaelpj: > When using source-repository-package the directive tests: True turns on tests within all source-repository-package dependencies, which (especially higher up the stack) just needless runs tests in dependencies for which it should be possible to avoid re-rumning tests.

Is this still true in recent versions of cabal? We have `tests: True` in `plutus` and I've never observed it running or building tests for s-r-ps, and I tried again just now.

PR #4431, opened 2022-09-13
Infer protocol params in transaction build

LudvikGalois:
We now ignore the `--protocol-params-file` in the `transaction build` command. Instead we
fetch it from the node directly. If the user still supplies it, we display a warning, but otherwise
continue.

Fixes https://github.com/input-output-hk/cardano-node/issues/4058

### DIFF
--- a/cardano-cli/ChangeLog.md
+++ b/cardano-cli/ChangeLog.md
@@ -10,9 +10,12 @@
 
 - Add `query tx-mempool` ([PR 4276](https://github.com/input-output-hk/cardano-node/pull/4276))
 
+
 ### Bugs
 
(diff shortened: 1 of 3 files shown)


cardano-cli/ChangeLog.md
```diff
@@ -6,6 +6,8 @@
 
 - Allow reading signing keys from a pipe ([PR 4342](https://github.com/input-output-hk/cardano-node/pull/4342))
 
+- Infer protocol params in `transaction build` ([PR 4431](https://github.com/input-output-hk/cardano-node/pull/4431))
```
Jimbo4350, line 9: I would change this to "Query protocol parameters from the node in the transaction build command"

Jimbo4350 (approved):

PR #4436, opened 2022-09-13
Hkm/ghc924

newhoggy:
This work stream is adding GHC 9.2.4 compatibility _ontop_ of the existing 8.10.7 compatibility. And _only_ that compatibility; there is no intention of changing anything outside of compiler compatibility.


L-as:
I love you.

cabal.project
```diff
@@ -117,6 +117,7 @@ package ouroboros-consensus-cardano-test
 
 package ouroboros-network
   tests: False
+  ghc-options: -XFlexibleContexts
```
angerman, line 120: @newhoggy I think this needs an explanation. I did see the warning; but are we sure that this is the _right_ fix for ouroboros-network? And does this impact 8.10.7 negatively?

nix/haskell.nix
```diff
@@ -91,9 +94,8 @@ haskell-nix.cabalProject' ({ pkgs
       })
       ({ pkgs, ... }: {
         # Use the VRF fork of libsodium
-        packages = lib.genAttrs [ "cardano-crypto-praos" "cardano-crypto-class" ] (_: {
-          components.library.pkgconfig = lib.mkForce [ [ pkgs.libsodium-vrf ] ];
-        });
+        packages.cardano-crypto-class.components.library.pkgconfig = lib.mkForce [[ pkgs.libsodium-vrf pkgs.secp256k1 ]];
```
L-as, line 97: Shouldn't this be fixed once and for all in the relevant packages? There's been a lot of work on pkgconfig stuff in haskell.nix lately, cc @hamishmack

cabal.project
newhoggy: This is just temporary to unblock finding all the other compile errors.  The fixes should eventually go into the repository itself.

PR #4445, opened 2022-09-14
Export toShelleyTxOutAny in Cardano.Api.Shelley and fix function comment of toShelleyTxOutAny.

koslambrou:
Export toShelleyTxOutAny in Cardano.Api.Shelley.

Also fixed function comment of toShelleyTxOutAny.


newhoggy:
Is this still needed?

newhoggy:
No progress for this PR for a long time.

Closing.

PR #4449, opened 2022-09-15
Fix incorrect cardano-node versions in docker-compose.yml

CyberCyclone:
`cardano-node` version in `docker-compose.yml`  is incorrectly set to `1.36.0`. Revert back to the current latest release of `1.35.3-configs`.


newhoggy:
Closing this PR because it is very old.

PR #4450, opened 2022-09-15
Auto-balance multiasset transactions

LudvikGalois:
Previously we gave up when the non-Ada part of a transaction wasn't balanced. We now balance the transaction and correctly update the fee accordingly (since the fee will be higher). We also return an error in the case where the is non-Ada change, but not at least minUTxO change (e.g. in the case where the Ada is already balanced).

Resolves: https://github.com/input-output-hk/cardano-node/issues/3068
Closes: https://github.com/input-output-hk/cardano-node/issues/4453

### DIFF
--- a/cardano-api/src/Cardano/Api/Fees.hs
+++ b/cardano-api/src/Cardano/Api/Fees.hs
@@ -799,11 +799,6 @@ data TxBodyErrorAutoBalance =
        -- | One or more of the scripts were expected to fail validation, but none did.
      | TxBodyScriptBadScriptValidity
 
-       -- | The balance of the non-ada assets is not zero. The 'Value' here is
-       -- that residual non-zero balance. The 'makeTransactionBodyAutoBalance'
-       -- function only automatically balances ada, not other assets.
-     | TxBodyErrorAssetBalanceWrong Value
-
        -- | There is not enough ada to cover both the outputs and the fees.
        -- The transaction should be changed to provide more input ada, or
        -- otherwise adjusted to need less (e.g. outputs, script etc).
@@ -861,13 +856,6 @@ instance Error TxBodyErrorAutoBalance where
 
   displayError TxBodyScriptBadScriptValidity =
       "One or more of the scripts were expected to fail validation, but none did."
-
-  displayError (TxBodyErrorAssetBalanceWrong _value) =
-      "The transaction does not correctly balance in its non-ada assets. "
-   ++ "The balance between inputs and outputs should sum to zero. "
-   ++ "The actual balance is: "
-   ++ "TODO: move the Value renderer and parser from the CLI into the API and use them here"
-   -- TODO: do this ^^
 
   displayError (TxBodyErrorAdaBalanceNegative lovelace) =
       "The transaction does not balance in its use of ada. The net balance "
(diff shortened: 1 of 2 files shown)


Jimbo4350 (approved):
LGTM! Liaise with QA and produce a ticket like discussed in the node team call. Only bors the PR when QA has signed off on it

cardano-api/src/Cardano/Api/Fees.hs
```diff
@@ -971,13 +959,23 @@ makeTransactionBodyAutoBalance eraInMode systemstart history pparams
...
+    -- However, since at this point we know how much non-Ada change to give
+    -- we can use the true values for that.
+
+    let outgoingNonAda = mconcat [filterValue isNotAda v | (TxOut _ (TxOutValue _ v) _ _) <- txOuts txbodycontent]
+    let incomingNonAda = mconcat [filterValue isNotAda v | (TxOut _ (TxOutValue _ v) _ _) <- Map.elems $ unUTxO utxo]
+    let nonAdaChange = incomingNonAda <> negateValue outgoingNonAda
+
+    let changeTxOut = case multiAssetSupportedInEra cardanoEra of
+          Left _ -> lovelaceToTxOutValue $ Lovelace (2^(64 :: Integer)) - 1
+          Right multiAsset -> TxOutValue multiAsset (lovelaceToValue (Lovelace (2^(64 :: Integer)) - 1) <> nonAdaChange)
```
jonathanknowles, line 971: Hi @LudvikGalois  

Just curious: do you have any expectations about how the `makeTransactionBodyAutoBalance` function should behave if the `changeTxOut` value is larger that the maximum value size? (As defined by the `maxValueSize` protocol parameter.)

In `cardano-wallet`, if we detect that a generated (change) output is larger than the maximum size, we partition it into a number of smaller outputs.

LudvikGalois: It's not currently handled here, although it'll be caught when trying to submit. My preference for now is to "give up" and require the user to explicitly make change themselves when building the transaction (my understanding is that this is an edge case in most cases). Manually balancing multiassets is much easier for a human (or script) to do than balancing Ada, because it doesn't need to account for some of what its balancing disappearing as transaction fees.
Multiple change UTxOs is straightforward, but given that we don't know how the user plans to spend it, we might sub-optimally divide it, which could end up costing the user more in fees.

saratomaz:
I am getting an error when I tried to mint a token (the same test works on latest master):

```
cardano-cli transaction build 
--mint-script-file test_minting_and_burning_witnesses_ci0_urn_asset_name_True_multisig.script 
--tx-in 898f28715c1fd838f346c523a6977809c5f9726a54ed4dd3c27418c7a00f4463#0 -
-tx-out addr_test1vqtp42zsxnuaqt4uw8xh565f8s7tfm98w88sakh9xp7fszg7tksef+3000000+5 11d47abca323a319956e447b7efea9abf91b943c05f2adc42375282a.636f75747473636f696e616d6573 
--mint 5 11d47abca323a319956e447b7efea9abf91b943c05f2adc42375282a.636f75747473636f696e616d6573 
--change-address addr_test1vqtp42zsxnuaqt4uw8xh565f8s7tfm98w88sakh9xp7fszg7tksef 
--witness-override 5 
--out-file test_minting_and_burning_witnesses_ci0_urn_asset_name_True_mint_tx.body 
--babbage-era 
--testnet-magic 42
```

> Command failed: transaction build  Error: Negative quantity (-5) in transaction output: TxOutInAnyEra BabbageEra (TxOut (AddressInEra (ShelleyAddressInEra ShelleyBasedEraBabbage) (ShelleyAddress Testnet (KeyHashObj (KeyHash "161aa85034f9d02ebc71cd7a6a893c3cb4eca771cf0edae5307c9809")) StakeRefNull)) (TxOutValue MultiAssetInBabbageEra (valueFromList [(AdaAssetId,18446744073709551615),(AssetId "11d47abca323a319956e447b7efea9abf91b943c05f2adc42375282a" "couttscoinames",-5)])) TxOutDatumNone ReferenceScriptNone)

Jimbo4350:
@newhoggy was https://github.com/input-output-hk/cardano-node/pull/4450#issuecomment-1289598761 resolved?

newhoggy:
> was https://github.com/input-output-hk/cardano-node/pull/4450#issuecomment-1289598761 resolved?

Yes.  Please see: https://input-output-rnd.slack.com/archives/G011N23CEAE/p1679075401311009?thread_ts=1679052340.879369&cid=G011N23CEAE

PR #4460, opened 2022-09-21
tx-generator: reusable API and library

mgmeier:


deepfire (approved):
Thank you @mgmeier !

This is an important first step towards the pure API.

PR #4467, opened 2022-09-22
tx-generator: Fix self-test.

MarcFontaine:
This fixes the benchmarking self-test and criterion benchmark.
'cabal run tx-generator:benchmark:tx-generator-bench' now works.


MarcFontaine:
The benchmark splits the input into 9000 UTxO,  generates 4000 TXs (2 inputs and 2 outputs)
and then discards the TXs.

Generator speed is ca. 200_000 TXs per second.

time                 28.26 ms   (27.55 ms .. 28.87 ms)
                     0.998 R²   (0.997 R² .. 0.999 R²)
mean                 27.73 ms   (27.27 ms .. 28.07 ms)
std dev              890.8 μs   (609.2 μs .. 1.398 ms)
variance introduced by outliers: 10% (moderately inflated)

PR #4481, opened 2022-09-26
bump plutus/base to support secp

disassembler:
This bumps plutus and base dependencies to support SECP.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -180,8 +180,8 @@ source-repository-package
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-base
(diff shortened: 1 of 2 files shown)


michaelpj:
Sorry, I think I gave you the wrong sha! Maybe it was pre-squash? The right sha is d5f6f4a5505094490f61c8d164515adf7c8f46cc

JaredCorduan:
to clarify, `d5f6f4a5505094490f61c8d164515adf7c8f46cc` looks to be the plutus sha that we want, which also uses the same cardano-base sha, `cc049d7c9b9a0129c15b1355fd1dff9e1a1a551c`.

PR #4486, opened 2022-09-27
tx-generator: Use distinct keys/addresses for benchmarking phases

MarcFontaine:
This is useful for debugging the tx-generator.
Hard coded addresses:
`addr_test1vzj7zv9msmdasvy5nc9jhnn2gqvrvu33v5rlg332zdfrkugklxkau` : outputs of the splitting phase (value not Plutus).
`addr_test1vz4qz2ayucp7xvnthrx93uhha7e04gvxttpnuq4e6mx2n5gzfw23z` : final outputs of the benchmarking phase.


bench/tx-generator/src/Cardano/Benchmarking/Compiler.hs
```diff
@@ -250,3 +260,46 @@ newWallet n = do
...
+keyNameGenesisInputFund :: KeyName
+keyNameGenesisInputFund = KeyName "GenesisInputFund"
+
+keyNameTxGenFunds :: KeyName
+keyNameTxGenFunds = KeyName "TxGenFunds"
+
+-- addr_test1vzd3muund27y5nw83vymqj3a83pcuzkkejej6s75e5lfjcc85nc3p
+keyTxGenFunds :: TextEnvelope
+keyTxGenFunds = parseKey "{\n    \"type\": \"PaymentSigningKeyShelley_ed25519\",\n    \"description\": \"Payment Signing Key\",\n    \"cborHex\": \"5820617f846fc8b0e753bd51790de5f5a916de500175c6f5a0e27dde9da7879e1d35\"\n}\n"
+
```
mgmeier, line 278: Why not define the values directly, like:
` = TextEnvelope "PaymentSigningKeyShelley_ed25519" "Payment Signing Key" "5820617f846fc8b0e753bd51790de5f5a916de500175c6f5a0e27dde9da7879e1d35"`

bench/tx-generator/src/Cardano/Benchmarking/Compiler.hs
```diff
@@ -250,3 +260,46 @@ newWallet n = do
...
+-- addr_test1vz4qz2ayucp7xvnthrx93uhha7e04gvxttpnuq4e6mx2n5gzfw23z
+keyBenchmarkDone :: TextEnvelope
+keyBenchmarkDone = parseKey "{\n    \"type\": \"PaymentSigningKeyShelley_ed25519\",\n    \"description\": \"Payment Signing Key\",\n    \"cborHex\": \"582016ca4f13fa17557e56a7d0dd3397d747db8e1e22fdb5b9df638abdb680650d50\"\n}\n"
+
+keyNameCollaterals :: KeyName
+keyNameCollaterals = KeyName "Collaterals"
+
+-- addr_test1vpckd9muw3l4f8ne4uzumy28p0k84rvx48q46kssjkta5ng4v6sfs
+keyCollaterals :: TextEnvelope
+keyCollaterals = parseKey "{\n    \"type\": \"PaymentSigningKeyShelley_ed25519\",\n    \"description\": \"Payment Signing Key\",\n    \"cborHex\": \"58204babdb63537ccdac393ea23d042af3b7c3587d7dc88ed3b66c959f198ad358fa\"\n}\n"
```
mgmeier, line 305: Would it make sense to add a (very brief) human-readable [Note] on how the flow through these wallets is designed during a benchmarking run? I know the names speak for themselves, but it would help creating specific queries for `cardano-cli`.

mgmeier (approved):

bench/tx-generator/src/Cardano/Benchmarking/Compiler.hs
MarcFontaine: That could be done.
This would look like:
```
    (TextEnvelope { teType = TextEnvelopeType "GenesisUTxOSigningKey_ed25519"
                  , teDescription = fromString "Genesis Initial UTxO Signing Key"
                 , teRawCBOR = "X \vl1~\182\201v(\152\250A\202\157h0\ETX\248h\153\171\SI/m\186\242D\228\NAK\182(&\162"})
```
One would have to convert die key to `RawCBOR`.
The way it is now one can just copy contents for the `.skey` file.

bench/tx-generator/src/Cardano/Benchmarking/Compiler.hs
MarcFontaine: Done that.

bench/tx-generator/src/Cardano/Benchmarking/Compiler.hs
mgmeier: The way it's done you still need some manual labour the `.skey` file (escaping the quotes, adding newlines).
You could just use the example I gave above (without field labels), extending the CBOR value to sth. like `(Base16.decodeLenient "5820617f846fc8b0e753bd51790de5f5a916de500175c6f5a0e27dde9da7879e1d35")`.
Every literal's type there has a `IsString` instance.
That way you'd just copy-paste the key file values, and not need to worry about inlining a JSON serialization.

MarcFontaine:
bors  r+

bench/tx-generator/src/Cardano/Benchmarking/Compiler.hs
MarcFontaine: Yes indeet the `Base16` version looks nicer.

PR #4489, opened 2022-09-27
add cicero CI actions

dermetfan:
This adds a tullia task ~~for every hydra job~~ and cicero actions that run them for every push and PR.

Edit: There's only one tullia task now that builds everything.

### DIFF
--- a/nix/tullia.nix
+++ b/nix/tullia.nix
@@ -1,0 +1,100 @@
+/*
+This file defines tullia tasks and cicero actions.
+
+Tullia is a sandboxed multi-runtime DAG task runner with Cicero support.
+Tasks can be written in different languages and are compiled for each runtime using Nix.
+It comes with essential building blocks for typical CI/CD scenarios.
+Learn more: https://github.com/input-output-hk/tullia
+
+Cicero is an if-this-then-that machine on HashiCorp Nomad.
+It can run any event-and-state-driven automation actions
+and hence CI/CD pipelines are a natural fit.
+In tandem with Tullia, an action could be described as
+the rule that describes when a Tullia task is to be invoked.
+Learn more: https://github.com/input-output-hk/cicero
+*/
+
+self: system:
+
+let
+  ciInputName = "GitHub event";
+  repo = "input-output-hk/cardano-node";
+in rec {
+  tasks = let
+    common = {
+      config,
+      ...
+    }: {
+      preset = {
+        # needed on top-level task to set runtime options
+        nix.enable = true;
+
+        github-ci = {
+          enable = config.actionRun.facts != {};
+          inherit repo;
+          sha = config.preset.github-ci.lib.getRevision ciInputName null;
+          clone.enable = false;
+        };
+      };
+
+      # some hydra jobs run NixOS tests
+      env.NIX_CONFIG = ''
+        extra-system-features = kvm
+      '';
+
+      memory = 1024 * 32;
+      nomad.resources.cpu = 10000;
+    };
+
+    # the attribute name in `hydraJobs` for the current system
+    os = {
+      x86_64-linux = "linux";
+      x86_64-darwin = "macos";
+    }.${system};
+
+    flakeUrl = {config, lib, ...}:
+      lib.escapeShellArg "github:${repo}/${config.preset.github-ci.lib.getRevision ciInputName null}";
+  in
+    {
+      "ci/push" = {lib, ...} @ args: {
+        imports = [common];
+
+        command.text = ''
+          nix build -L \
+            ${flakeUrl args}#hydraJobs.${lib.escapeShellArg os}.required \
+            ${flakeUrl args}#hydraJobs.cardano-deployment
+        '';
+      };
+
+      "ci/pr" = {lib, ...} @ args: {
+        imports = [common];
+
+        command.text = ''
+          nix build -L \
+            ${flakeUrl args}#hydraJobsPr.${lib.escapeShellArg os}.required \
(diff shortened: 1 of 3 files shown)


nix/tullia.nix
```diff
@@ -0,0 +1,123 @@
+self: system:
```
angerman, line 1: Can we add some comment here
- what the purpose/intention of this file is?
- what `tullia` is? 
- where to find the corresponding documentation or this file?

I've no read this file a few times, but I still don't think I fully understand why it is I this repo?
There seem to be 20 repo relevant lines from L102 to L122 the rest is ceremony?

nix/tullia.nix
```diff
@@ -0,0 +1,123 @@
...
+    };
+
+  actions = {
+    "cardano-node/ci/push" = {
+      task = "ci/push";
+      io = ''
+        #lib.io.github_push
+        #input: "${ciInputName}"
+        #repo: "input-output-hk/cardano-node"
+        #default_branch: false
```
angerman, line 109: What kind of syntax is this? This is new to me.

angerman (changes requested):

nix/tullia.nix
```diff
@@ -0,0 +1,123 @@
...
+    };
+
+    flakeUrl = {
+      config,
+      lib,
+      ...
+    }: lib.escapeShellArg (
+      if config.actionRun.facts != {}
+      then with config.preset.github-ci; "github:${repo}/${sha}"
+      else "."
```
Pacman99, line 32: How come having a fact means you can refer to the flake with `.`?

nix/tullia.nix
Pacman99: I believe this is cue syntax

nix/tullia.nix
```diff
@@ -0,0 +1,123 @@
...
+        imports = [common];
+        after = [(lib.last (__attrNames ciPrTasks))];
+      };
+    };
+
+  actions = {
+    "cardano-node/ci/push" = {
+      task = "ci/push";
+      io = ''
+        #lib.io.github_push
```
Pacman99, line 106: Will `task = "ci/push";` refer to all tasks starting with `ci/push`? Does this mean the ci/push tasks are run when there is a git push to github? And how come thats part of `io`?

Pacman99: Ohh wait I see, there is a ci/push task defined that is set to run after all the other ci/push/* tasks. But I'm still not sure I understand the "triggering" or run condition mechanism.

nix/tullia.nix
dermetfan: Yes, it's CUE.

nix/tullia.nix
dermetfan: Tullia tasks can be run locally or on Cicero. When no facts are present we know that we are running locally and vice versa. When running locally, the current directory is already bind-mounted into the container, so we don't need to fetch the source  from GitHub.

nix/tullia.nix
dermetfan: Correct. `ci/push` is just an arbitrary name. It will be run when there is a new fact (JSON document) that matches the `io` CUE expression.

nix/tullia.nix
dermetfan: > I've no read this file a few times, but I still don't think I fully understand why it is I this repo?
> There seem to be 20 repo relevant lines from L102 to L122 the rest is ceremony?

As `hydraJobs` in this flake does not follow the flake output schema of `hydraJobs."<attr>"."<system>" = derivation;`, we have to write code specific to this repo in order to generate Tullia tasks from it that can not be stored away in some tullia library.
If this flake is converted to follow the flake output schema later we can likely make this file a bit smaller by writing a generalized hydra jobs to Tullia tasks function in Tullia's library.

Jimbo4350: > There's not much else to say if the reader knows what Tullia and Cicero are

Please assume the reader knows nothing.

Jimbo4350:
> This adds a tullia task for every hydra job and cicero actions that run them for every push and PR.
@dermetfan  can you show where the jobs are translated over?  It's not clear to me.

dermetfan:
@Jimbo4350 [Here](https://github.com/input-output-hk/cardano-node/pull/4489/files#diff-bb99c177bc572780ea9ff28015a48fbef18080f5763796ace98044a6cf23d39eR44-R45): The [`flakeOutputTasks`](https://github.com/input-output-hk/tullia/blob/565de3f76098aa0a3f13a4ef8c11f43eabdb196d/nix/lib.nix#L262-L270) function generates a set of tullia tasks given an arbitrarily nested set that contains derivations.

deepfire (approved):

nix/tullia.nix
angerman: @dermetfan Yes, we don't document `hydraJobs` either. That's not a good precedent. That's actually an example of where we fall massively short.

I think this file should say that the file contains Tulia tasks and Cicero actions. And also provide links as to where one can read more about them. We want to eventually turn this into an open source project, where others need to be able to find out what's going on without having access to IOG slack or other IOG folks readily. It needs to be sufficiently self explanatory.

angerman: Just to stress this again. It doesn't need to contain the full writeup what Tulia and Cicero are. But it needs to have sufficient links to where one can read up on it to make sense of this file without having to informally consult someone else.

dermetfan: I added a short description to the top of this file. Please let me know whether you think that is sufficient.

angerman (approved):
LGTM

PR #4490, opened 2022-09-29
Fix CODEOWNERS errors

Jimbo4350:


LudvikGalois (commented):
Shouldn't the devops on line 46 also be removed?

Jimbo4350:
> Shouldn't the devops on line 46 also be removed?

Yep, thanks. Fixed and pushed

PR #4491, opened 2022-09-29
Separate IO from txBuild and txBuildRaw functions

Jimbo4350:
First step towards condensing modules responsible for reading and validating the things required for transaction construction. The end goal is to expose these modules via `cardano-api` so users can implement their own transaction construction and/or validation functions.

### DIFF
--- a/cardano-api/src/Cardano/Api/Fees.hs
+++ b/cardano-api/src/Cardano/Api/Fees.hs
@@ -1035,8 +1035,10 @@ makeTransactionBodyAutoBalance eraInMode systemstart history pparams
           txTotalCollateral = reqCol
         }
     txbody3 <-
-      first TxBodyError $ -- TODO: impossible to fail now
-        makeTransactionBody finalTxBodyContent
+      first TxBodyError $ -- TODO: impossible to fail now. We need to implement a function
+                          -- that simply creates a transaction body because we have already
+                          -- validated the transaction body earlier within makeTransactionBodyAutoBalance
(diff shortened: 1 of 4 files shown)


cardano-api/src/Cardano/Api/Fees.hs
```diff
@@ -1025,17 +1026,18 @@ makeTransactionBodyAutoBalance eraInMode systemstart history pparams
...
           txFee  = TxFeeExplicit explicitTxFees fee,
           txOuts = accountForNoChange
                      (TxOut changeaddr balance TxOutDatumNone ReferenceScriptNone)
                      (txOuts txbodycontent),
           txReturnCollateral = retColl,
           txTotalCollateral = reqCol
         }
-    return (BalancedTxBody txbody3 (TxOut changeaddr balance TxOutDatumNone ReferenceScriptNone) fee)
+    txbody3 <-
+      first TxBodyError $ -- TODO: impossible to fail now
```
newhoggy, line 1038: I think the TODO comment needs elaboration.  The TODO should say what needs to be done.

Jimbo4350: That was there previously but I'll update it

newhoggy (approved):

cardano-api/src/Cardano/Api/Fees.hs
newhoggy: Thanks!

PR #4494, opened 2022-09-29
Export `TextEnvelopeCddl(..)` from `Cardano.Api`

FinleyMcIlwaine:
This is necessary to be able to refer to results of `deserialiseTxLedgerCddl`


Jimbo4350:
Subsumed by: #4635

PR #4498, opened 2022-09-30
remove JaredCorduan from codeowners

JaredCorduan:
There are now enough members on the node team that can properly review node PRs (better than I can!), so I think it makes sense to remove me from the `CODEOWNERS` file. (Unless folks would prefer I stay on for the occasional PR when folks are on vacation, etc.)


JaredCorduan:
Sounds like folks would prefer I stay on the codeowners file, which I'm happy to do.

Jimbo4350:
Sorry @JaredCorduan, the only way out of the node's CODEOWNERS file is in a box :coffin:

PR #4499, opened 2022-09-30
trace-dispatcher: optional EKG.Store.

denisshevchenko:
This PR makes `EKG.Store` optional for `initForwarding`. Uses updated version of `ekg-forward` library.


cardano-node/src/Cardano/Node/Tracing/API.hs
```diff
@@ -96,7 +96,7 @@ initTraceDispatcher nc p networkMagic nodeKernel p2pMode = do
           -- TODO: check if this is the correct way to use withIOManager
           (forwardSink, dpStore) <- withIOManager $ \iomgr -> do
             let tracerSocketMode = Just . first unSocketPath =<< ncTraceForwardSocket nc
-            initForwarding iomgr trConfig networkMagic ekgStore tracerSocketMode
+            initForwarding iomgr trConfig networkMagic (Just ekgStore) {- FIXME: for compile -} tracerSocketMode
```
denisshevchenko, line 99: This should be changed: we have to add the case when `ekgStore` is `Nothing`.

deepfire (approved):
Looks fine to me, thank you @denisshevchenko

Jimbo4350 (approved):

Issue #4493, opened 2022-09-29
[BUG] - TextEnvelopeCddl error when assembling Shelley-era Tx body in CDDL format


mkoura:
*Internal*

**Area**
*Other*   Any other topic (Delegation, Ranking, ...).

**Summary**
Assembling Shelley-era Tx body in CDDL format with witnesses fails with `TextEnvelopeCddl error`:

> cardano-cli transaction assemble --tx-body-file test_normal_tx_from_script_addr_ci0_dll_False_to_tx.body --out-file test_normal_tx_from_script_addr_ci0_dll_False_to_tx.witnessed --witness-file test_normal_tx_from_script_addr_ci0_dll_False_to_skey0_tx.witness

> Command failed: transaction sign-witness  Error: Failed to decode neither the cli's serialisation format nor the ledger's CDDL serialisation format. TextEnvelope error: test_normal_tx_from_script_addr_ci0_dll_False_to_skey0_tx.witness: TextEnvelope type error:  Expected one of: TxWitnessByron, TxWitnessShelley, TxWitness AllegraEra, TxWitness MaryEra, TxWitness AlonzoEra, TxWitness BabbageEra Actual: TxWitness ShelleyEra
TextEnvelopeCddl error: test_normal_tx_from_script_addr_ci0_dll_False_to_skey0_tx.witness: TextEnvelopeCddl type error:  Expected one of:  Actual: TxWitness ShelleyEra

**Steps to reproduce**
1. Create Shelley-ear Tx body in CDDL format
2. Try to asseble the Tx body and witnesses to form a transaction
3. See error

**Expected behavior**
A Shelley-era Tx can be assembled from body in CDDL format.

**System info:**
- OS Name: Fedora 36
- Node version: 40458eefa87c7c3abd8c6ba542d9e931d8b2ecb8
- CLI version: 40458eefa87c7c3abd8c6ba542d9e931d8b2ecb8

PR #4501, opened 2022-10-03
Fix 4493 bug - TxWitness text envelope format does not roundtrip in Shelley era

Jimbo4350:
Resolves: #4493


erikd (changes requested):
I totally applaud "Fix bug 4493" in the commit message, but the commit message really should briefly say what the bug was.

mkoura (approved):
Verified that it fixes issue #4493

erikd (approved):

PR #4516, opened 2022-10-09
Condense Read and Validation modules in cardano-cli

Jimbo4350:
Two things occur before constructing a tx. You must first read the parts you need and then validate that they can indeed be used in a particular era. We condense these into two module in the cli, with the eventual aim of moving them to cardano-api.


cardano-cli/src/Cardano/CLI/Shelley/Run/Read.hs
```diff
@@ -0,0 +1,741 @@
...
+import           Data.Word
+
+
+import           Cardano.Api
+import           Cardano.Api.Shelley
+
+--TODO: do this nicely via the API too:
+import qualified Cardano.Binary as CBOR
+import           Data.Text (Text)
+--TODO: following import needed for orphan Eq Script instance
```
newhoggy, line 77: This doesn't seem like a TODO.

cardano-cli/src/Cardano/CLI/Shelley/Run/Read.hs
```diff
@@ -0,0 +1,741 @@
...
+import qualified Data.ByteString.Lazy.Char8 as LBS
+import qualified Data.List as List
+import qualified Data.Text as Text
+import           Data.Word
+
+
+import           Cardano.Api
+import           Cardano.Api.Shelley
+
+--TODO: do this nicely via the API too:
```
newhoggy, line 74: Not sure what this TODO means.

cardano-cli/src/Cardano/CLI/Shelley/Run/Read.hs
```diff
@@ -0,0 +1,741 @@
...
+  | MetadataErrorNotAvailableInEra AnyCardanoEra
+
+renderMetadataError :: MetadataError -> Text
+renderMetadataError (MetadataErrorFile fileErr) =
+  Text.pack $ displayError fileErr
+renderMetadataError (MetadataErrorJsonParseError fp jsonErr) =
+  Text.pack $ "Invalid JSON format in file: " <> show fp <>
+              "\nJSON parse error: " <> jsonErr
+renderMetadataError (MetadataErrorConversionError fp metadataErr) =
+  Text.pack $ "Error reading metadata at: " <> show fp <>
```
newhoggy, line 101: Sometimes we say "in file: <fp>" and sometimes we say "at: <fp>".

Is there something different going on?  Or should we use the same language in all cases?

cardano-cli/src/Cardano/CLI/Shelley/Run/Read.hs
```diff
@@ -0,0 +1,741 @@
...
+  | ScriptWitnessErrorExpectedPlutus !FilePath !AnyScriptLanguage
+  | ScriptWitnessErrorReferenceScriptsNotSupportedInEra !AnyCardanoEra
+  | ScriptWitnessErrorScriptData ScriptDataError
+
+renderScriptWitnessError :: ScriptWitnessError -> Text
+renderScriptWitnessError (ScriptWitnessErrorFile err) =
+  Text.pack $ displayError err
+renderScriptWitnessError (ScriptWitnessErrorScriptLanguageNotSupportedInEra (AnyScriptLanguage lang) anyEra) =
+  "The script language " <> Text.pack (show lang) <> " is not supported in the " <>
+  renderEra anyEra <> " era'."
```
newhoggy, line 172: Is the apostrophe are `era` intended?

cardano-cli/src/Cardano/CLI/Shelley/Run/Read.hs
```diff
@@ -0,0 +1,741 @@
...
+                    $ getIsCardanoEraConstraint era' (AnyCardanoEra era')
+    Just _ -> do
+
+      case scriptLanguageSupportedInEra era' anyScriptLanguage of
+        Just sLangInEra ->
+          case languageOfScriptLanguageInEra sLangInEra of
+            SimpleScriptLanguage _v ->
+              -- TODO: We likely need another datatype eg data ReferenceScriptWitness lang
+              -- in order to make this branch unrepresentable.
+              error "readScriptWitness: Should not be possible to specify a simple script"
```
newhoggy, line 271: 👍

cardano-cli/src/Cardano/CLI/Shelley/Run/Read.hs
```diff
@@ -0,0 +1,741 @@
...
+              "\nJSON parse error: " <> jsonErr
+renderScriptDataError (ScriptDataErrorConversion fp sDataJsonErr) =
+  Text.pack $ "Error reading metadata at: " <> show fp <>
+              "\n" <> displayError sDataJsonErr
+renderScriptDataError (ScriptDataErrorValidation fp sDataRangeErr) =
+  Text.pack $ "Error validating script data at: " <> show fp <> ":\n" <>
+              displayError sDataRangeErr
+renderScriptDataError (ScriptDataErrorMetadataDecode fp decoderErr) =
+  Text.pack $ "Error decoding CBOR metadata at: " <> show fp <>
+              " Error: " <> show decoderErr
```
newhoggy, line 331: Language "in file" vs "at"

cardano-cli/src/Cardano/CLI/Shelley/Run/Read.hs
```diff
@@ -0,0 +1,741 @@
...
+
+data CddlError = CddlErrorTextEnv
+                   !(FileError TextEnvelopeError)
+                   !(FileError TextEnvelopeCddlError)
+               | CddlIOError (FileError TextEnvelopeError)
+
+renderCddlError :: CddlError -> Text
+renderCddlError (CddlErrorTextEnv textEnvErr cddlErr) =
+  "Failed to decode neither the cli's serialisation format nor the ledger's \
+  \CDDL serialisation format. TextEnvelope error: " <> Text.pack (displayError textEnvErr) <> "\n" <>
```
newhoggy, line 483: Prefer to not use multiline strings.  Prefer using `<>`.

cardano-cli/src/Cardano/CLI/Shelley/Run/Read.hs
Jimbo4350: We actually don't need to import this anymore. I removed it.

cardano-cli/src/Cardano/CLI/Shelley/Run/Read.hs
Jimbo4350: That's actually a comment from Duncan. I think he wanted to expose the CBOR stuff from ledger in a nicer way.

cardano-cli/src/Cardano/CLI/Shelley/Run/Read.hs
Jimbo4350: Nope just an inconsistency. I wouldn't fix that in this PR however. Open an issue if you think its worth fixing.

cardano-cli/src/Cardano/CLI/Shelley/Run/Read.hs
Jimbo4350: Probably not, I'll remove it.

cardano-cli/src/Cardano/CLI/Shelley/Run/Read.hs
Jimbo4350: See comment above

cardano-cli/src/Cardano/CLI/Shelley/Run/Read.hs
Jimbo4350: Yeah fair enough. We should purge the use of `\`. Can you open an issue for this?

newhoggy: https://github.com/input-output-hk/cardano-node/issues/4585

newhoggy (approved):

cardano-cli/src/Cardano/CLI/Shelley/Run/Validate.hs
```diff
@@ -0,0 +1,359 @@
...
+  displayError (ProtocolParametersNotSupported e) =
+    "Transaction protocol parameters are not supported in " <> Text.unpack (renderEra e)
+
+validateProtocolParameters
+  :: CardanoEra era
+  -> Maybe ProtocolParameters
+  -> Either TxProtocolParametersValidationError (BuildTxWith BuildTx (Maybe ProtocolParameters))
+validateProtocolParameters _ Nothing = return (BuildTxWith Nothing)
+validateProtocolParameters era (Just pparams) =
+    case scriptDataSupportedInEra era of
```
newhoggy, line 315: What's the reason we case match on `scriptDataSupportedInEra` in `validateProtocolParameters`?

PR #4521, opened 2022-10-11
 partially monomorphise & de-generify Env

NadiaYvette:
Store's User case, Action's Set case & branching through Name's cases within Store's Named case are all unfolded. The entire Setters file along with its Tag, Sum etc. data types are removed.


bench/tx-generator/src/Cardano/Benchmarking/Script/Env.hs
```diff
@@ -42,10 +44,12 @@ import           Cardano.Benchmarking.Script.Store
...
 import           Cardano.TxGenerator.Types (TxGenError(..))
 
-data Env = Env { dmap :: DMap Store Identity }
+data Env = Env { dmap :: DMap Store Identity
+               , protoParams :: Maybe ProtocolParameterMode
+               }
 
 emptyEnv :: Env
-emptyEnv = Env { dmap = DMap.empty }
+emptyEnv = Env { dmap = DMap.empty, protoParams = Nothing }
```
deepfire, line 52: Just a very minor nitpick about formatting -- I'd:

0. break the line, between `=` and `Env` aligning the latter naturally.
1. break the line, just before the `,`  aligning the latter with the `{`
2. make the trailing `}` on its own line

...which is how multi-line is done frequently.

bench/tx-generator/src/Cardano/Benchmarking/Script/Env.hs
```diff
@@ -73,12 +77,21 @@ askIOManager = lift RWS.ask
 set :: Store v -> v -> ActionM ()
 set key val = lift $ RWS.modify $ (\e -> e { dmap = DMap.insert key (pure val) (dmap e)})
 
+setProtoParamMode :: ProtocolParameterMode -> ActionM ()
+setProtoParamMode val = lift $ RWS.modify $ (\e -> e { protoParams = pure val })
```
deepfire, line 81: Given that you'll have more of those cases, I'd split this into a combination of:

1. the generic `lift . RWS.modify` helper and have it accept the: 
2. the `Env` modifier for specific cases

bench/tx-generator/src/Cardano/Benchmarking/Script/Env.hs
```diff
@@ -73,12 +77,21 @@ askIOManager = lift RWS.ask
...
+
 get :: Store v -> ActionM v
 get key = do
   lift (RWS.gets $ (\e -> DMap.lookup key $ dmap e)) >>= \case
     Just (Identity v) -> return v
     Nothing -> throwE $ LookupError key
 
+getProtoParamMode :: ActionM ProtocolParameterMode
+getProtoParamMode = do
+  lift (RWS.gets $ (\e -> protoParams e)) >>= \case
```
deepfire, line 91: Here, we can similarly to the setter, extract a generic helper that calls a supplied reader on top of the common `lift . RWS.gets` part.

NadiaYvette: I'll mop up the read accessors in a sweep after finishing off all the Store cases. The set accessors might still be worthwhile vs. open-coding, though lift $ modify lambda isn't too big a deal either.

mgmeier (approved):

mgmeier:
Thank you a lot, Nadia! :)

PR #4522, opened 2022-10-11
Fix the workbench shell

michaelpj:
Using `cabalWrapped` also requires us to mess with `haskell.nix`'s choices for what packages to prepare the shell for.

### DIFF
--- a/nix/workbench/shell.nix
+++ b/nix/workbench/shell.nix
@@ -78,6 +78,15 @@ in project.shellFor {
 
   inherit withHoogle;
 
+  # The workbench shell uses cabalWrapped, which removes the `source-repository-package` stanzas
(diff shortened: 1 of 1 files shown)


deepfire:
> Using `cabalWrapped` also requires us to mess with `haskell.nix`'s choices for what packages to prepare the shell for.

To clarify, this is how it used to work, and that standard behavior was changed by https://github.com/input-output-hk/cardano-node/commit/814b6d8d5ea0822c5460b455b7caba105a9eaab4#diff-65f4ca594d8a70559d53a677a686f5245807d0d094bbbb159e367766bf316ea2L81

deepfire (approved):
Thank you @michaelpj for the fix!

michaelpj:
Yes, to be clear, I messed it up first!

PR #4542, opened 2022-10-18
Bump CHaP again, account for re-release of existing packages

michaelpj:
The packages in CHaP have been re-released to line up with 1.35.3. Unfortunately, these are not compatible with node master. Hence I've had to restore some of the `source-repository-package` stanzas from previously. We will aim to get rid of these as quickly as possible, although this may require doing some additional releases.

This is a pretty major DevX loss, since we're back to `source-repository-package`s, but without this things are actually broken again :(

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -81,8 +81,6 @@ constraints:
     -- that dependency
   , network >= 3.1.1.0
   , HSOpenSSL >= 0.11.7.2
-  , algebraic-graphs < 0.7
-  , protolude < 0.3.1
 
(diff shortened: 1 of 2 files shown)


michaelpj:
@disassembler hydra succeeded

LudvikGalois (approved):

LudvikGalois:
Once I'd cleaned out some stale files in my `~/.cabal/store` I managed to get this working locally.

Jimbo4350 (approved):

deepfire (approved):

PR #4544, opened 2022-10-19
cardano-testnet: add single entrypoint for starting testnets

MarcFontaine:
This PR adds a single entry point for starting Shelley, Babbage and Cardano testnets.

### DIFF
--- a/cardano-testnet/src/Test/Runtime.hs
+++ b/cardano-testnet/src/Test/Runtime.hs
@@ -2,7 +2,6 @@
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DuplicateRecordFields #-}
 {-# LANGUAGE LambdaCase #-}
-{-# LANGUAGE NamedFieldPuns #-}
 
 module Test.Runtime
   ( LeadershipSlot(..)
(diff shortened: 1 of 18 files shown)


cardano-testnet/test/Spec/Cli/Babbage/LeadershipSchedule.hs
```diff
@@ -66,13 +66,13 @@ hprop_leadershipSchedule = H.integration . H.runFinallies . H.workspace "alonzo"
     , poolNodes
     -- , wallets
     -- , delegators
-    } <- TC.testnet testnetOptions conf
+    } <- babbageTestnet testnetOptions conf
```
Jimbo4350, line 69: Same here, we want to call `testnet`

Jimbo4350 (commented):
Nice work. A few comments

cardano-testnet/src/Testnet/Utils.hs
```diff
@@ -83,3 +96,51 @@ queryTip (QueryTipOutput fp) testnetMag execConfig = do
 
 newtype QueryTipOutput = QueryTipOutput { unQueryTipOutput :: FilePath}
 
+startNode ::
```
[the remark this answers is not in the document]
Jimbo4350: Also this function really belongs in `Test.Runtime`

cardano-testnet/src/Test/Runtime.hs
```diff
@@ -58,14 +59,11 @@ data TestnetNode = TestnetNode
   , nodeProcessHandle :: IO.ProcessHandle
   }
 
+newtype TestnetNode = TestnetNode {unTestnetNode :: NodeRuntime}
```
Jimbo4350, line 62: Let's get rid of `TestnetNode`. I don't see a clear benefits of wrapping `NodeRuntime` with it.

MarcFontaine:
This PR is a first step towards
https://github.com/input-output-hk/cardano-node/issues/4134

cardano-testnet/src/Test/Runtime.hs
```diff
@@ -97,31 +98,72 @@ data LeadershipSlot = LeadershipSlot
...
   "text" -> Right NodeLoggingFormatAsText
   s -> Left $ "Unrecognised node logging format: " <> show s <> ".  Valid options: \"json\", \"text\""
+
+allNodes :: TestnetRuntime -> [NodeRuntime]
+allNodes tr = bftNodes tr <> fmap poolRuntime (poolNodes tr)
+
+-- | Start a node, creating file handles, sockets and temp-dirs.
+startNode
+  :: String
+  -- ^ The tempBaseAbsPath
```
Jimbo4350, line 122: Please expand on `tempBaseAbsPath` and `tempAbsPath`. The haddocks don't tell the user anything. Ask @newhoggy if its not immediately obvious what these parameters are doing/generally used for.

cardano-testnet/src/Testnet/Utils.hs
```diff
@@ -26,7 +26,6 @@ import           Hedgehog.Extras.Test.Process (ExecConfig)
 import           Hedgehog.Internal.Property (MonadTest)
 import qualified Test.Process as H
 
-
```
Jimbo4350, line 29: Please undo this diff. It just adds noise to the PR.

cardano-testnet/test/Spec/ShutdownOnSlotSynced.hs
```diff
@@ -54,7 +55,7 @@ hprop_shutdownOnSlotSynced = H.integration . H.runFinallies . H.workspace "chair
           , defaultTestnetNodeOptions
           ]
         }
-  TC.TestnetRuntime { bftNodes = node:_ } <- testnet fastTestnetOptions conf
+  TC.TestnetRuntime { bftNodes = node:_ } <- cardanoTestnet fastTestnetOptions conf
```
Jimbo4350, line 58: This should call `testnet`

Jimbo4350 (changes requested):
Couple minor changes. Happy to approve after.

cardano-testnet/src/Test/Runtime.hs
MarcFontaine: I totally agree that this should be more clear, but if possible,
I would really like to tackle this in a separate PR.
The test infrastructure uses `Testnet.Conf.Conf` which is a set of FilePaths (mix of relative,
absolute) and other config values.
The arguments to `startNode` all come from `Conf` but I'm reluctant to just pass `Conf` because `Conf` contains other additional data.
Basically `tempBaseAbsPath = takeDirectory tempAbsPath`, so one of the two arguments can be safely removed.

Jimbo4350: :+1:

PR #4546, opened 2022-10-20
Reduce exposed modules in cardano-api

Jimbo4350:


newhoggy (approved):

LudvikGalois:
Given that this is a breaking change for downstream users of the API, should there be a note in the changelog about it?

Jimbo4350:
Yep definitely. I'll update the changelog.

deepfire (approved):

MarcFontaine (approved):
LGTM

PR #4547, opened 2022-10-21
Fix the workbench manifest

michaelpj:
This replaces the old code for generating manifest output in the workbench with a new one based on parsing `plan.json`. This gives us information about all components, including the source-repository they come from if they come from one. That gives us a generic way of listing all the component info.

We do need an explicit list of packages, since otherwise we can't distinguish our packages of interest from, say, random Hackage packages.

This doesn't yet fix the manifest report, but should be a step forwards.

### DIFF
--- a/nix/workbench/backend/runner.nix
+++ b/nix/workbench/backend/runner.nix
@@ -76,6 +76,7 @@ in
             nativeBuildInputs = with cardanoNodePackages; with pkgs; [
               bash
               bech32
+              cabal-install
               coreutils
               gnused
               jq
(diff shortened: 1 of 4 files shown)


michaelpj:
cc @andreabedini for interest

deepfire (approved):
bors r+

andreabedini:
@michaelpj I have added cabal to the buildInputs but unfortunately calling `cabal build --dry-run` won't work inside nix. This is because it will need to fetch the srp to make a plan. Can we use haskell.nix or something here?

newhoggy:
I've been informed that this PR should be closed.

PR #4548, opened 2022-10-21
Release 1.35.4

disassembler:
This is initial test of plutus/base bump that this works. Will cut a version of base at 2.0.0.0.1. Will also look at seeing if we can includedb-syhtnesizer from orouboros-network in a minimal change, as well as any CLI improvements that don't require changes to consensus/ledger.


disassembler:
closing in favor of #4508

PR #4549, opened 2022-10-23
Upgrade to cabal-3.8.1.0

newhoggy:


cabal.project
```diff
@@ -183,3 +183,9 @@ source-repository-package
   subdir:
     cardano-prelude
     cardano-prelude-test
+
+source-repository-package
+  type: git
+  location: https://github.com/astanin/moo
+  tag: dbda5e76ac3b4c72c805ec0cdb9bcdff7bb6247d
+  --sha256: 1mdj218hgh7s5a6b9k14vg9i06zxah0wa42ycdgh245izs8nfv0x
```
LudvikGalois, line 191: Is this needed (when using CHaP)?

michaelpj: It is not, there is a patched version in CHaP. Also please don't add more without a _clear_ justification!

newhoggy: When I tried last it the build failed on `moo`.  I'll create another PR without that commit to see if it happens again.

newhoggy: Actually I found the failed build here: https://github.com/input-output-hk/cardano-node/actions/runs/3306791861/jobs/5457943706

newhoggy: I updated the CHaP index state and it works now.

michaelpj (approved):

erikd (approved):

LudvikGalois (approved):

Jimbo4350 (approved):

PR #4550, opened 2022-10-23
Use environment files in Github Actions

newhoggy:
See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


LudvikGalois (approved):

Jimbo4350 (approved):

MarcFontaine (approved):
Looks like the example in GitHub Actions: Deprecating save-state and set-output commands.

PR #4554, opened 2022-10-24
export IsPlutusScriptLanguage

MarcFontaine:


cardano-api/src/Cardano/Api/Script.hs
```diff
@@ -24,6 +24,7 @@ module Cardano.Api.Script (
     PlutusScriptVersion(..),
     AnyScriptLanguage(..),
     AnyPlutusScriptVersion(..),
+    IsPlutusScriptLanguage(..),
```
phadej, line 27: Is `Cardano.Api.Script` a public module? I think it isn't.

https://github.com/input-output-hk/cardano-node/blob/26344329ecb9762f82cf4430fd3a33684a6468e0/cardano-api/cardano-api.cabal#L52-L82

LudvikGalois (approved):

PR #4561, opened 2022-10-25
Trace changes

### LINKED ISSUE #4560: [FR] - The new tracing system does not use severities for `StartupTracer`
[`const Notice`](https://github.com/input-output-hk/cardano-node/blob/master/cardano-node/src/Cardano/Node/Tracing/Tracers.hs?plain=1#L142) is used while we ought to have a proper severities like the old logging system does [here](https://github.com/input-output-hk/cardano-node/blob/master/cardano-node/src/Cardano/Tracing/Startup.hs?plain=1#L27-L34).

coot:
This PR contains various small changes to traces.

- Removed kind field from NodeTo{Node,Client}VersionData
- Fixed a typo
- Changed IP ToJSON instance
- Simplified ToJSON instance of LocalRootPeers
- Use Notice severity for NetworkConfigUpdate
- Added severityStartupTracer

Fixes #4560

### DIFF
--- a/cardano-node/src/Cardano/Node/Startup.hs
+++ b/cardano-node/src/Cardano/Node/Startup.hs
@@ -113,6 +113,16 @@ data StartupTrace blk =
   | BIByron BasicInfoByron
   | BINetwork BasicInfoNetwork
 
+severityStartupTracer :: StartupTrace blk -> SeverityS
+severityStartupTracer (StartupSocketConfigError _) = Error
+severityStartupTracer NetworkConfigUpdate = Notice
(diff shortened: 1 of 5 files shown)


Jimbo4350 (changes requested):
6 commits is overkill for these changes. Please squash the commits to a single commit.

jutaro (approved):
Looks good

coot:
@Jimbo4350 I squashed all the commits.

Jimbo4350 (approved):

PR #4563, opened 2022-10-26
New p2p topology file format

### LINKED ISSUE #4559: [FR] - Simplify p2p topology format
Currently the json format is more verbose than it needs to be, e.g.
```json
{
  "LocalRoots": {
    "groups": [
      { "localRoots":
        { "accessPoints":
            [ { "address": "35.180.225.101"
              , "port": 7776
              }
            ]
        , "advertise": false
        }
      , "valency": 3
      }
    ]
  }
, "PublicRoots": [
    { "publicRoots": {
        "accessPoints": [
          { "address": "relays-new.cardano-mainnet.iohk.io"
          , "port": 3001
          }
        ]
      , "advertise": false
      }
    }
  ]
, "useLedgerAfterSlot": 45000000
}
```

I suggest we simplify this to something like:
```json
{
  "localRoots": [
      { "accessPoints":
            [ { "address": "35.180.225.101"
              , "port": 7776
              }
            ]
      , "advertise": false
      , "valency": 1
      }
    ]
, "publicRoots": [
    { "accessPoints": [
          { "address": "relays-new.cardano-mainnet.iohk.io"
          , "port": 3001
          }
        ]
    , "advertise": false
    }
  ]
, "useLedgerAfterSlot": 45000000
}
```

Note: we ought to support both formats for a while; whenever we match the old format we should log a warning that this format will be only supported for next two major releases.

coot:
Fixes #4559.

### DIFF
--- a/cardano-node/ChangeLog.md
+++ b/cardano-node/ChangeLog.md
@@ -1,4 +1,4 @@
-# Changelog for Cardano node
+# Changelog for cardano-node
 
 ## vNext
 
--- a/cardano-node/ChangeLog.md
+++ b/cardano-node/ChangeLog.md
@@ -1,5 +1,29 @@
 # Changelog for cardano-node
 
+## vNext
+
+### node changes
+
+None
+
+### consensus changes
+
+None
+
+### network changes
+
+- New P2P topology file format, see [issue #4563][#4563] or the [config
+  files][understanding-config-files] documentation.  The old p2p topology
+  format will be supported for next two major releases of the node (the last
+  major version which will support it is `1.37`).
+
+[#4563]: https://github.com/input-output-hk/cardano-node/issues/4563
+[understanding-config-files]: https://github.com/input-output-hk/cardano-node/blob/master/doc/getting-started/understanding-config-files.md
+
+### ledger changes
+
+None
+
 ## 1.35.3 -- August 2022
 
 ### node changes
--- a/cardano-node/ChangeLog.md
+++ b/cardano-node/ChangeLog.md
@@ -1,4 +1,4 @@
-# Changelog for cardano-node
+# Changelog for Cardano node
 
 ## vNext
 
@@ -12,10 +12,10 @@ None
 
 ### network changes
 
-- New P2P topology file format, see [issue #4563][#4563] or the [config
-  files][understanding-config-files] documentation.  The old p2p topology
+- New P2P topology file format. See [issue #4563][#4563] or the [config
+  files][understanding-config-files] documentation. The old P2P topology
   format will be supported for next two major releases of the node (the last
-  major version which will support it is `1.37`).
+  supported major version will be `1.37`).
 
 [#4563]: https://github.com/input-output-hk/cardano-node/issues/4563
 [understanding-config-files]: https://github.com/input-output-hk/cardano-node/blob/master/doc/getting-started/understanding-config-files.md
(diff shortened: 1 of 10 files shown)


cardano-node/src/Cardano/Node/Configuration/TopologyP2P.hs
```diff
@@ -132,77 +138,120 @@ data LocalRootPeersGroup = LocalRootPeersGroup
...
                                              , "useLedgerAfterSlot" .= ul
                                              ]
 
+--
+-- Legacy p2p topology file format
+--
+
+-- | A newtype wrapper which provides legacy 'FromJSON' instances.
+--
+newtype Legacy a = Legacy { getLegacy :: a }
```
Jimbo4350, line 201: I'd prefer you handle the legacy file format in the original type's JSON instances. I'd rather not introduce another type.

cardano-node/src/Cardano/Node/Configuration/TopologyP2P.hs
```diff
@@ -132,77 +138,120 @@ data LocalRootPeersGroup = LocalRootPeersGroup
...
   } deriving (Eq, Show)
 
+-- | Does not use the 'FromJSON' instance of 'RootConfig', so that
+-- 'accessPoints', 'advertise' and 'valency' fields are attached to the same
+-- object.
 instance FromJSON LocalRootPeersGroup where
   parseJSON = withObject "LocalRootPeersGroup" $ \o ->
                 LocalRootPeersGroup
-                  <$> o .: "localRoots"
+                  <$> parseJSON (Object o)
```
Jimbo4350, line 147: I don't understand this. Won't you still end up using the `FromJSON RootConfig` instance?
`parseJSON :: FromJSON a => Value -> Parser a`

Jimbo4350 (changes requested):

cardano-node/src/Cardano/Node/Configuration/TopologyP2P.hs
coot: I want to use `FromJSON RootConfig`, but I don't want to wrap it in an object, I want instead to extend it with `valency`.  The format for `LocalRootPeersGroup` is something like
```json
{ "accessPoints": ...
, "advertise": ...
, "valency": ...
}
```
First two fields are parsed with `RootConfig` instance attached to the same object as the `valency`.

cardano-node/src/Cardano/Node/Configuration/TopologyP2P.hs
coot: This will be removed in `1.37.0` or so, and it makes it very easy to do so. If we'd preserve the original instances, we'd need to have a newtype wrapper for the new ones, and then when we'd remove the legacy format we'd have more work to do.

Jimbo4350:
One last thing, can you add a `vNext` section to cardano-node's change log and add this PR to it (similar to what has been done in cardano-cli and cardano-api)?

bolt12 (approved):

coot:
@Jimbo4350 I added entry in the `ChangeLog`; with links so it makes it easy for people to find what they need to change.

cardano-node/ChangeLog.md
```diff
@@ -1,5 +1,29 @@
 # Changelog for cardano-node
 
+## vNext
+
+### node changes
```
Jimbo4350, line 5: This should be under `network changes`

Jimbo4350 (changes requested):

cardano-node/ChangeLog.md
coot: Done.

Jimbo4350 (approved):

coot:
bors merge

coot:
bors merge

cardano-node/ChangeLog.md
```diff
@@ -1,4 +1,4 @@
-# Changelog for cardano-node
+# Changelog for Cardano node
```
coot, line 1: The library is called `cardano-node`, and it probably makes sense to preserve it.  Cardano Node change log is usually constructed per release and it contains changes from other parts of the system.

olgahryniuk: Okay, this makes sense. I'll change to cardano-node

coot:
bors merge

PR #4568, opened 2022-10-27
Add ToJSON and FromJSON instances for Address

MarcFontaine:


MarcFontaine:
These are the reusable bits from https://github.com/input-output-hk/cardano-node/pull/1888.
The other parts of the functionality from #1888 have landed in master already.

cardano-api/src/Cardano/Api/Address.hs
```diff
@@ -263,6 +263,25 @@ instance SerialiseAddress (Address ShelleyAddr) where
...
+    parseJSON = Aeson.withText "Address" $ \txt ->
+      maybe
+        (fail "Invalid Byron address.")
+        pure
+        (deserialiseAddress AsByronAddress txt)
+
+instance FromJSON (Address ShelleyAddr) where
+    parseJSON = Aeson.withText "Address" $ \txt ->
+      maybe
+        (fail "Invalid Shelley address.")
```
Jimbo4350, line 282: I would update the error messages to "Cardano.Api.Address.FromJSON: Invalid $era address"

Jimbo4350 (approved):
LGTM! One minor change

cardano-api/src/Cardano/Api/Address.hs
MarcFontaine: OK done that

PR #4569, opened 2022-10-27
Expose txScriptValidityToScriptValidity in Cardano.API

berewt:
The function is needed in Marconi and we had to duplicate it at the moment, it would be great if we could simply import it.


Jimbo4350:
Subsumed by #4628

PR #4575, opened 2022-10-28
Improve test output

newhoggy:
Exceptions in tests (for example from partial functions) cause failure without line number which makes the cause difficult to find.  This PR switches to use functions that accurately convey location of failure in the test.

Use `byDeadlineM` instead of `assertByDeadlineM`, which allows specifying a period.  This allows a longer poll time to be used so the test output isn't as spammy.

### DIFF
--- a/cardano-testnet/test/Spec/Cli/KesPeriodInfo.hs
+++ b/cardano-testnet/test/Spec/Cli/KesPeriodInfo.hs
@@ -8,45 +8,43 @@ module Spec.Cli.KesPeriodInfo
   ( hprop_kes_period_info
   ) where
 
-
-import           Prelude
-
 import           Cardano.Api
-import           Cardano.Api.Shelley
-
+import           Cardano.Api.Shelley (PoolId)
 import           Control.Monad (void)
-import qualified Data.Aeson as J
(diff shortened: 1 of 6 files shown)


newhoggy:
Replaces https://github.com/input-output-hk/cardano-node/pull/3596

cardano-testnet/test/Spec/Cli/KesPeriodInfo.hs
```diff
@@ -8,44 +8,41 @@ module Spec.Cli.KesPeriodInfo
...
 import           Cardano.Api
-import           Cardano.Api.Shelley
-
+import           Cardano.Api.Shelley (PoolId)
 import           Control.Monad (void)
-import qualified Data.Aeson as J
-import qualified Data.Map.Strict as Map
 import           Data.Monoid (Last (..))
 import           Data.Set (Set)
-import qualified Data.Set as Set
```
Jimbo4350, line 22: :+1:

Jimbo4350 (commented):
LGTM but there is an hlint error

MarcFontaine (approved):
LGTM

Jimbo4350 (approved):

PR #4584, opened 2022-10-31
cardano-tracer: search in logs, WIP

denisshevchenko:
This is the very first code for the "Search" mechanism in live view logs.


mgmeier:
Will be closed because:
* search feature implementation is incomplete
* targets the `RTView` component, which has been changed to maintenance only

PR #4590, opened 2022-11-01
Windows CI builds

newhoggy:
Resolves https://github.com/input-output-hk/cardano-node/issues/4614


deepfire (approved):
Thank you a lot @newhoggy !

michaelpj:
Can I suggest that we _leave a comment_ explaining why this is doing what it's doing? Otherwise we will lose the knowledge and it will just make our pipelines even more obscure and hard to modify!

.github/workflows/haskell.yml
```diff
@@ -178,7 +179,7 @@ jobs:
           sed -i 's|tests: False|tests: True|g' cabal.project
         fi
 
-        cp .github/workflows/cabal.project.local.ci cabal.project.local
+        cp ".github/workflows/cabal.project.local.ci.$(uname -s)" cabal.project.local
```
newhoggy, line 182: For Windows only, we need to pass extra flags so that `gcc` and `ld` (which `ghc` invokes) can find the libraries and header files.

This is necessary because unfortunately the 3rdparty `msys2` system we use installs packages into `D:/a/_temp/msys64/mingw64` instead of the system `C:\msys64\mingw64`.  This applies also the `secp256k1`.

It's unclear why `gcc` and `ld` invoked by `ghc` is unable to see this alternate installation directory.

.github/workflows/cabal.project.local.ci.MINGW64_NT-10.0-20348
```diff
@@ -0,0 +1,23 @@
...
+  extra-include-dirs: D:/a/_temp/msys64/mingw64/include
+  extra-lib-dirs: D:/a/_temp/msys64/mingw64/lib
+
+package cardano-crypto-class
+  extra-include-dirs: D:/a/_temp/msys64/mingw64/include
+  extra-lib-dirs: D:/a/_temp/msys64/mingw64/lib
+
+package cardano-crypto-praos
+  extra-include-dirs: D:/a/_temp/msys64/mingw64/include
+  extra-lib-dirs: D:/a/_temp/msys64/mingw64/lib
```
newhoggy, line 23: And Haskell packages that fails to build because the system installation directory is missing will need an entry here.

michaelpj:
@coot seems to be addressing the same problems here, perhaps worth comparing notes? https://github.com/input-output-hk/ouroboros-network/pull/4134

.github/workflows/haskell.yml
```diff
@@ -57,6 +57,7 @@ jobs:
           mingw-w64-x86_64-toolchain
           mingw-w64-x86_64-libsodium
           mingw-w64-x86_64-jq
+          mingw-w64-x86_64-openssl
```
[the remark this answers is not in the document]
coot: It seems it worked fine without installing `openssl` in [ouroboros-network](https://github.com/input-output-hk/ouroboros-network/actions/runs/3396596282/jobs/5647834067).

PR #4595, opened 2022-11-02
Reorder module hierarchy

MarcFontaine:


MarcFontaine:
follow up of https://github.com/input-output-hk/cardano-node/issues/4558

Jimbo4350:
:+1:  Nice. Now lets move:
- `cardano-testnet/src/Spec/*` -> `cardano-testnet/test/Test` (rename Spec to Test)
- `cardano-testnet/test` -> `cardano-testnet/util` and we can probable rename `cardano-testnet/src/Test/Util.hs` to `cardano-testnet/src/Test/Ignore.hs` for now

I'll approve after this.

Jimbo4350:
:+1:  I just noticed `:cardano-testnet/src/Testnet/Properties/Cli`. That should also be put in the `cardano-testnet/test/Test/Cli/` dir

MarcFontaine:
> +1 I just noticed `cardano-testnet/src/Testnet/Properties/Cli`. That should also be put in the `cardano-testnet/test/Test/Cli/` dir

There used to be two different `KesPeriodInfo.hs` in the hierarchy (One of them importing the other)..I renamed one of them to `Test.Misc`.

Jimbo4350 (approved):
LGTM! We will need to think of a better name than `Misc` but the hope is as we add more properties this will naturally crystallize out.

Jimbo4350 (approved):

PR #4599, opened 2022-11-03
export RawBytesHexError

MarcFontaine:


MarcFontaine:
https://github.com/input-output-hk/cardano-node/issues/4579

Jimbo4350 (approved):

cardano-api/src/Cardano/Api/SerialiseRaw.hs
```diff
@@ -32,28 +33,32 @@ serialiseToRawBytesHex = Base16.encode . serialiseToRawBytes
...
       ++ message
-    RawBytesHexErrorRawBytesDecodeFail input ->
-      "Failed to deserialise " ++ show input
-      -- TODO(2022-01-26, cblp) show expected output type
+    RawBytesHexErrorRawBytesDecodeFail input asType ->
+      "Failed to deserialise " ++ pretty input ++ " as " ++ (show asType)
+    where
+      pretty bs = case Text.decodeUtf8' bs of
+        Right t -> Text.unpack t
+        Left _ -> show bs
```
Jimbo4350, line 56: Let's also include the error from `decodeUtf8'` e.g "Error decoding: " ++ bs ++ " Error: " ++ decodeErr

MarcFontaine: For me it would be confusing to see an error about UTF8 decoding.
The no-error-case does not even try UTF8 decoding and a correct input is hex-digits only.

Jimbo4350: > For me it would be confusing to see an error about UTF8 decoding.

It would inform the user that what they are trying to decode is not UTF8 encoding. Why is that a bad thing/confusing?

PR #4613, opened 2022-11-04
Update building-the-node-using-nix.md

CarlosLopezDeLara:
Build the cardano-node executable. No default configuration.

### DIFF
--- a/doc/getting-started/building-the-node-using-nix.md
+++ b/doc/getting-started/building-the-node-using-nix.md
@@ -24,15 +24,16 @@ To improve build speed, it is possible to set up a binary cache maintained by IO
 optional**):
 ```
 sudo mkdir -p /etc/nix
-cat <<EOF | sudo tee /etc/nix/nix.conf
-substituters = https://cache.nixos.org https://hydra.iohk.io
-trusted-public-keys = iohk.cachix.org-1:DpRUyj7h7V830dp/i6Nti+NEO2/nhblbov/8MW7Rqoo= hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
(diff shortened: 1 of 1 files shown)


doc/getting-started/building-the-node-using-nix.md
```diff
@@ -44,6 +44,10 @@ or run in in one go:
 nix run github:input-output-hk/cardano-node#mainnet/node
 ```
 
+To build the cardano-node executable:
```
deepfire, line 47: I would have reworded this to underscore -- something like -- "to build just the cardano-node executable"

deepfire: And it doesn't help that line 35 above doesn't really explain what it achieves.

We could reword it too, from:
> Once Nix is installed, log out and then log back in.
>
> After that you can build the full node package with Mainnet configuration as follows:

CarlosLopezDeLara: Amended the commit with your suggestions, thanks Serge!

CarlosLopezDeLara: There goes another amendment! =)

doc/getting-started/building-the-node-using-nix.md
```diff
@@ -24,15 +24,16 @@ To improve build speed, it is possible to set up a binary cache maintained by IO
 optional**):
 ```
 sudo mkdir -p /etc/nix
-cat <<EOF | sudo tee /etc/nix/nix.conf
+cat <<EOF | sudo tee -a /etc/nix/nix.conf
```
disassembler, line 27: Can we fix the caches in this file?;we no longer use iohk.cachix.org. and the substituter should be cache.iog.io now (but the key still has hydra in the name)

disassembler (changes requested):

doc/getting-started/building-the-node-using-nix.md
CarlosLopezDeLara: Ok, I think I understood correctly. I updated the commit.

disassembler (approved):

PR #4615, opened 2022-11-04
bench-data-publish: new tool for publishing benchmark results

mgmeier:
This PR adds `bench/bench-data-publish` to the suite of benchmarking tools.

The new tool is a command-line tool for publishing the result of benchmarking run analysis to a Postgres DB. It additionally contains the DB schema, as well as a definition of views on the data stored within.

1. The schema is designed to only rely on an absolute minimum of data normalization; the `.json`s from a run analysis are stored verbatim in this schema.
2. The views extract all relevant properties and metrics from those objects and expose them to make the data searchable and filterable.

By using a service like PostgREST, a REST API can be served immediately from these views, without an addtional layer of configuration.


deepfire (approved):
This is excellent work, thank you @mgmeier !

Jimbo4350 (changes requested):
This does not belong in `cardano-node`

deepfire:
@mgmeier, what @Jimbo4350 says has merit -- only `*.sql` schemas are coupled to the workbench, the haskell program can come from elsewhere.

mgmeier:
moved to seperate repository: [input-output-hk/cardano-automation](url)

PR #4617, opened 2022-11-04
Restore --cddl-format

newhoggy:
as a hidden switch to avoid breaking 3rd party tooling.  The restore switch has no effect

Resolves https://github.com/input-output-hk/cardano-node/issues/4597


Jimbo4350 (approved):

gitmachtl:
Looks good, the `transaction build/build-raw` command is now working again without an error. We will update the tools to remove the `--cddl-format` calling flag once 1.35.4 is released to be ready for a full removal starting with `1.36.*`

PR #4621, opened 2022-11-07
Keep existing witnesses in cli transaction assemble

MarcFontaine:


MarcFontaine:
https://github.com/input-output-hk/cardano-node/issues/4576

Jimbo4350 (changes requested):
What you have done will work but a clearer way to go about it as as follows:
```
runTxSignWitness
  :: TxBodyFile
  -> ...
```
We should read `TxBodyFile`s and `TxFile`s. This will require a change  the use of `pInputTxOrTxBodyFile`

MarcFontaine:
closed in favor of https://github.com/input-output-hk/cardano-node/pull/4647

PR #4623, opened 2022-11-07
initial release process

disassembler:


RELEASE.md
```diff
@@ -0,0 +1,44 @@
...
+* Rotating release engineer from one of the development teams
+* SRE team supporting deployment and CI/CD
+* Test Engineers that focus on integration tests
+* Performance Engineers running benchmarking tests
+
+# Release Process
+
+This is the release process for node releases
+
+1. Release Squad Lead will work with rotating engineer and dev leads to determine where to cut the releases from ledger, network and node. These will be published via CHaP and will follow the defined process of the team for versioning, etc...
```
abailly-iohk, line 14: Who should be the `Release Squad Lead`?

RELEASE.md
```diff
@@ -0,0 +1,44 @@
...
+2. These will be integrated up the stack to the node to produce a release branch
```
abailly-iohk, line 15: Shouldn't this be done at `main` branch level? eg. update dependencies in discrete step, run integration tests, if it's green and we're happy make it a release branch (or tag)?

RELEASE.md
```diff
@@ -0,0 +1,44 @@
...
+3. Release Squad Lead will work with Test/Performance sub-teams to initiate a testing round. This may be a tagged release or may be tip of branch.
+4. A release candidate will be deployed to preview/preprod by SRE
+5. Community will be notified of release candidate and given a few days to provide feedback
```
abailly-iohk, line 18: Could we make it regularly cadenced, so that people knows every X weeks there's a new preview/rc rolled out?

RELEASE.md
```diff
@@ -0,0 +1,44 @@
...
+6. Release Squad Lead will draft Release notes
+7. Head of Engineering will meet with release team and identify if release should be published as stable or beta (pre-release) in GitHub. stable releases can go all the way to mainnet, beta should only go to preprod and wait for that release to be relabeled as stable or a new stable release to be cut.
+8. Release is published
```
abailly-iohk, line 21: What does this mean? Artifacts are published on docker/binaries registry?

RELEASE.md
```diff
@@ -0,0 +1,44 @@
...
+* Performance Engineers running benchmarking tests
+
+# Release Process
+
+This is the release process for node releases
+
+1. Release Squad Lead will work with rotating engineer and dev leads to determine where to cut the releases from ledger, network and node. These will be published via CHaP and will follow the defined process of the team for versioning, etc...
+2. These will be integrated up the stack to the node to produce a release branch
+3. Release Squad Lead will work with Test/Performance sub-teams to initiate a testing round. This may be a tagged release or may be tip of branch.
+4. A release candidate will be deployed to preview/preprod by SRE
```
abailly-iohk, line 17: This probably implies the rc should be published somewhere too, for people to deploy and download it.

abailly-iohk (commented):
Some questions about details of the proposed process.

RELEASE.md
disassembler: that's difficult, because it requires the integration work to tag an rc. We can't say it will be tagged as an rc on day X because we don't know if the changes will be integrated by day X.

RELEASE.md
disassembler: that's all automated thankfully :)

RELEASE.md
disassembler: that would be me!

RELEASE.md
abailly-iohk: but we are going to integrate those changes through versions, right? eg. bump `ouroboros-consensus` package version to `1.2.3`?

RELEASE.md
```diff
@@ -0,0 +1,44 @@
+# Team
+
+The release team is made up of the following:
+
+* Rotating release engineer from one of the development teams
```
JaredCorduan, line 5: One concern that has come up is a natural asymmetry inside the teams. In particular, the ledger team often makes changes that require integrating in ouroboros-network and node, and hence ledger might be in the best position to do the integration. We (ledger) do keep a changelog, but it is still probably easier for us. If we did use a different procedure, where the team who "caused" the work did the integration, then I would love it if we could also nominate folks on the other teams (for the same duration) for pairing up with the release engineer when they need help. If we name folks explicitly, it makes it easier to ask for help and it makes it easier to justify the time providing help.

RELEASE.md
abailly-iohk: But you could reverse the problem, say that on day X there will be a rc and then let whatever is integrated at that date be deployed?

RELEASE.md
```diff
@@ -0,0 +1,44 @@
...
+3. Release Squad Lead will work with Test/Performance sub-teams to initiate a testing round. This may be a tagged release or may be tip of branch.
+4. A release candidate will be deployed to preview/preprod by SRE
+5. Community will be notified of release candidate and given a few days to provide feedback
+6. Release Squad Lead will draft Release notes
+7. Head of Engineering will meet with release team and identify if release should be published as stable or beta (pre-release) in GitHub. stable releases can go all the way to mainnet, beta should only go to preprod and wait for that release to be relabeled as stable or a new stable release to be cut.
+8. Release is published
+
+# Rotating Release Engineer Role
+
+All sprints are aligned across the node and it's components. At the end of a sprint cycle the new rotationg release engineer is decided on by the leadership team.
```
Jimbo4350, line 25: Spelling error: "rotationg"

RELEASE.md
```diff
@@ -0,0 +1,44 @@
...
+This person's primary duties are integration of new releases of dependencies up the stack to the node. They serve this role until the release is finalized
+according to the above release process (ideally 1 sprint cycle).
+
+# Sub-Teams
+
+## SRE
```
Jimbo4350, line 31: Worth adding what the SRE acronym stands for

Jimbo4350 (approved):

blaggacao:
Would it be possible to enshrine some core tenants about the release artifacts, for example ensuring that an OCI artifact is actually operable (in the well known scenarios) &mdash; whether that is implemented on the binary layer or hacked together in the entrypoint as we did in the past?

This would be an extremely useful guarantee to reduce the pains of operating `cardano-node` (& other components) across the entire ecosystem.

JaredCorduan (approved):

RELEASE.md
```diff
@@ -0,0 +1,67 @@
...
+This is the release process for node releases
+
+1. Release Squad Lead will work with rotating engineer and dev leads to determine where to cut the releases from ledger, network and node. These will be published via CHaP and will follow the defined process of the team for versioning, etc...
+2. These will be integrated up the stack to the node to produce a release branch
+3. Release Squad Lead will work with Test/Performance sub-teams to initiate a testing round. This may be a tagged release, a commit based off of master, or a commit
+based off of a previously released version as a hotfix
+4. A release candidate will be deployed to preview/preprod by SRE
+5. Community will be notified of release candidate and given a few days to provide feedback
+6. Release Squad Lead will draft Release notes
+7. Head of Engineering will meet with release team and identify if release should be published as stable or beta (pre-release) in GitHub. stable releases can go all the way to mainnet, beta should only go to preprod and wait for that release to be relabeled as stable or a new stable release to be cut.
```
lehins, line 21:
```suggestion
7. Head of Engineering will meet with release team and identify if release should be published as stable or beta (pre-release) in GitHub. Atable releases can go all the way to mainnet, beta should only go to preprod and wait for that release to be relabeled as stable or a new stable release to be cut.
```

RELEASE.md
```diff
@@ -0,0 +1,67 @@
...
+
+Performance engineers run benchmarks of the node and report any improvements/regressions between node versions.
+
+# Versioning
+
+The node uses a "pseudo semantic versioning" that takes into account breaking change (e.g. new eras) in the versioning logic. This new versioning
+standard within the node is supported as of `8.0.0` release. The first part of the version always references the max protocol version allowed
+on a stable network with no additional experimental override flags. This first part will remain `8` referencing Babbage era until the next era (Conway)
+is finalized. The second digit is a new release incremented counter. Every release based off the master trunk will increment this number by one. The final
+part should always be `0` *unless* an emergeency hard fork is necessary that cannot wait until the next major release. This should be a rare occurrence going
```
lehins, line 57: A controlled hard fork implies a protocol version and consequently the major version bump. So, shouldn't the last version be bumped only for fixes that do not introduce breaking changes?

I would phrase it something like:
```suggestion
part should always be `0` *unless* an emergency bug fix is necessary that cannot wait until the next major release. This should be a rare occurrence going
```

Regardless of the wording it needs a spellfix:
```suggestion
part should always be `0` *unless* an emergency hard fork is necessary that cannot wait until the next major release. This should be a rare occurrence going
```

lehins (approved):
:+1:

newhoggy (approved):

Jimbo4350 (approved):

PR #4627, opened 2022-11-09
Add fix and test for `foldBlocks`

eyeinsky:
Creating this draft pull request in relation to the `foldBlocks` issue here https://github.com/input-output-hk/cardano-node/issues/4624.

Due to adding genesis hashes to node config file (72e601764f875c6457468c9a8e32e7b1d24df689) cardano-testnet now *does* start (without adding the hashes it exceptioned out as described in the issue), but foldBlocks still doesn't receive ledger states and hangs indefinitely.

(Should I break out the 72e601764f875c6457468c9a8e32e7b1d24df689 to a separate PR so we could merge that?)

### DIFF
--- a/cardano-testnet/src/Testnet/Cardano.hs
+++ b/cardano-testnet/src/Testnet/Cardano.hs
@@ -19,20 +19,26 @@ module Testnet.Cardano
 
 import           Prelude
 import           Control.Monad
-import           Control.Monad.IO.Class (liftIO)
+import           Control.Monad.IO.Class (liftIO, MonadIO)
 import           Data.Aeson ((.=))
 import           Data.ByteString.Lazy (ByteString)
 import           Data.List ((\\))
 import           Data.Maybe
 import           Data.String
+import qualified Hedgehog as H
 import           Hedgehog.Extras.Stock.IO.Network.Sprocket (Sprocket (..))
 import           Hedgehog.Extras.Stock.Time (formatIso8601, showUTCTimeSeconds)
 import           Ouroboros.Network.PeerSelection.LedgerPeers (UseLedgerAfter (..))
 import           Ouroboros.Network.PeerSelection.RelayAccessPoint (RelayAccessPoint (..))
 import           System.FilePath.Posix ((</>))
+import           Control.Monad.Trans.Except
+import qualified Data.ByteString as BS
+import qualified Cardano.Crypto.Hash.Class
+import qualified Cardano.Crypto.Hash.Blake2b
 
 import qualified Cardano.Node.Configuration.Topology as NonP2P
 import qualified Cardano.Node.Configuration.TopologyP2P as P2P
+import           Cardano.Chain.Genesis (readGenesisData, GenesisHash(unGenesisHash))
 import qualified Data.Aeson as J
 import qualified Data.HashMap.Lazy as HM
 import qualified Data.List as L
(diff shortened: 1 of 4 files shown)


eyeinsky:
Cardano testnet now succesfully starts and foldBlocks test (occasionally) succeeds.

I say occasionally because the test is flaky: testnet nodes do start but quite frequently in the test runs they don't start posting anything to the local chain sync protocol and the test fails with a timeout.

The good thing is that the config files now allow the testnet to start, but why the socket is not written to remains to be researched.

Another thing that might be good to pursue is that while monitoring linux processes, the cardano node's start relatively early in the test, but the local chainsync socket starts to receive updates ~20 seconds in -- is that time required, e.g spending it on reaching consensus, or could it be cut much shorter? It would greatly benefit the convenience of testing if that time could be cut shorter.

eyeinsky:
This PR is now ready for review.

In addition to amending the node configuration file, there were two other issues that made this test flaky:
- `foldBlocks` doesn't fold received blocks forever but returns when current block (or error) is reached. Sometimes this happened before any permanent blocks => permanent ledger states have been created (`foldBlocks` only emits permanent ledger states). We run `foldBlocks` in a loop to overcome this.
- The chain is not extended fast enough for the two node/one pool testnet configuration that is created by default, changing the `activeSlotsCoeff` from 0.2 to 0.9 fixes this.

cardano-testnet/src/Testnet/Cardano.hs
```diff
@@ -19,20 +19,26 @@ module Testnet.Cardano
...
 import           Data.List ((\\))
 import           Data.Maybe
 import           Data.String
+import qualified Hedgehog as H
 import           Hedgehog.Extras.Stock.IO.Network.Sprocket (Sprocket (..))
 import           Hedgehog.Extras.Stock.Time (formatIso8601, showUTCTimeSeconds)
 import           Ouroboros.Network.PeerSelection.LedgerPeers (UseLedgerAfter (..))
 import           Ouroboros.Network.PeerSelection.RelayAccessPoint (RelayAccessPoint (..))
 import           System.FilePath.Posix ((</>))
+import           Control.Monad.Trans.Except
```
Jimbo4350, line 34: These are not in order. Can you enable `stylish-haskell` so the imports are ordered correctly.

cardano-testnet/test/Main.hs
```diff
@@ -16,6 +16,7 @@ import qualified Test.Cli.Babbage.LeadershipSchedule
 import qualified Test.Cli.KesPeriodInfo
 import qualified Test.Node.Shutdown
 import qualified Test.ShutdownOnSlotSynced
+import qualified Test.FoldBlocks
```
Jimbo4350, line 19: Also not in order

cardano-testnet/test/Test/FoldBlocks.hs
```diff
@@ -0,0 +1,86 @@
+{-# LANGUAGE OverloadedStrings  #-}
+
+module Test.FoldBlocks where
+
+import Prelude
```
Jimbo4350, line 5: `stylish-haskell` will fix the import formatting

cardano-testnet/test/Test/FoldBlocks.hs
```diff
@@ -0,0 +1,86 @@
...
+      -- until current slot and then quits -- even if there are no
+      -- permanent (= older than the k parameter) blocks created. In
+      -- that case we simply restart `foldBlocks` again.
+      forever $ do
+        let handler _env _ledgerState _ledgerEvents _blockInCardanoMode _ = IO.putMVar lock ()
+        e <- runExceptT (C.foldBlocks configurationFile socketPathAbs  C.QuickValidation () handler)
+        either (throw . FoldBlocksException) (\_ -> pure ()) e
+    link a -- Throw async thread's exceptions in main thread
+
+  _ <- liftIO $ IO.readMVar lock
```
Jimbo4350, line 85: Can you explain further what this test is testing? Specifically wrt the put/readMVar.

Jimbo4350 (changes requested):

cardano-testnet/test/Test/FoldBlocks.hs
eyeinsky: The `MVar` is written to from within the `handler` that is passed to `foldBlocks`, it simply tests that a ledger state is received and the handler is called (which writes to the `lock` and allows the test to finish).

Jimbo4350: Ah right I understand. Can you include this in a comment? I'll approve after.

eyeinsky: Yup, will also add all fixes done here to the other PR as well (the one that's against the `release/1.35` branch).

eyeinsky:
Closing this PR in favor of this one: https://github.com/input-output-hk/cardano-node/pull/4679

PR #4655, opened 2022-11-22
Move cardano-ping from ouroboros-network to cardano-node project

newhoggy:
This moves `cardano-ping` as is.

This has not been integrated with `cardano-cli` because there is an unresolved question regarding whether this tool will work on Windows.

Depends on:
* https://github.com/input-output-hk/cardano-node/pull/4764

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -23,6 +23,7 @@ packages:
     cardano-node
     cardano-node-capi
     cardano-node-chairman
+    cardano-ping
     cardano-submit-api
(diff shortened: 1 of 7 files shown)


Jimbo4350:
Let's not do this. The maintenance burden will be ours and we already have enough on our plate.

newhoggy:
Superseded by https://github.com/input-output-hk/cardano-node/pull/4664

PR #4662, opened 2022-11-24
[CI] Fix haddock builds

angerman:


.github/workflows/haskell.yml
```diff
@@ -102,7 +102,7 @@ jobs:
       run: echo "cabal-store=C:\\cabal\\store" >> $GITHUB_OUTPUT   
 
     - name: Set cache version
-      run: echo "CACHE_VERSION=grFfw8r" >> $GITHUB_ENV
+      run: echo "CACHE_VERSION=grFfw7r" >> $GITHUB_ENV
```
angerman, line 105: We really should hash `env`, and stick that into the cache. Debugging odd failure that started coming up because an env var was changed, and an old cache was used was ... bad.

newhoggy (approved):

angerman:
bors merge

PR #4665, opened 2022-11-25
workbench: small fixes & UX improvements

deepfire:
- fix the `wb analyse variance` invocation
- improve nomenclature of the suboperations
- add top-level aliases for the suboperation, to save typing
- improve the online documentation


fmaste (approved):
Looks great

PR #4671, opened 2022-11-28
Make leadership-schedule test less flaky

newhoggy:
This PR replaces `assertByDeadlineMCustom` with `byDeadlineM`. The latter is better because it catches assertions and retries whereas an assertion in the block of the former will fail the test.

`byDeadlineM` will also take a poll interval so we don't needlessly poll too often which clutters the output.

The test `Babbage.leadership-schedule` is re-enabled on MacOS.

The binding `maxSlotExpected` is moved higher so it is not re-computed unnecessarily.


cardano-testnet/src/Util/Process.hs
```diff
@@ -132,25 +131,11 @@ assertByDeadlineIOCustom str deadline f = GHC.withFrozenCallStack $ do
...
-  success <- f
-  unless success $ do
-    currentTime <- liftIO DTC.getCurrentTime
-    if currentTime < deadline
-      then do
-        liftIO $ IO.threadDelay 1000000
-        assertByDeadlineMCustom str deadline f
-      else do
-        H.annotateShow currentTime
-        failMessage GHC.callStack $ "Condition not met by deadline: " <> str
```
newhoggy, line 156: Replace `assertByDeadlineMCustom` with `byDeadlineM`.  The latter is better because it catches assertions and retries whereas an assertion in the block of the former will fail the test.

newhoggy: `byDeadlineM` will also take a poll interval so we don't needlessly poll too often which clutters the output.

cardano-testnet/src/Testnet/Babbage.hs
```diff
@@ -335,7 +335,7 @@ babbageTestnet testnetOptions H.Conf {..} = do
     return $ PoolNode runtime key
 
   now <- H.noteShowIO DTC.getCurrentTime
-  deadline <- H.noteShow $ DTC.addUTCTime 90 now
+  deadline <- H.noteShow $ DTC.addUTCTime 110 now
```
LudvikGalois, line 338: Why has this been increased? Was it too low?

newhoggy: I suspected it was too low, but I'm thinking maybe it wasn't and my test failures are due to something else.  Will try bumping it back down again.

newhoggy: I've verified I don't need this change.

cardano-testnet/test/Main.hs
```diff
@@ -28,7 +28,7 @@ tests = pure $ T.testGroup "test/Spec.hs"
     --   [ H.ignoreOnMacAndWindows "leadership-schedule" Test.Cli.Alonzo.LeadershipSchedule.hprop_leadershipSchedule
     --   ]
     , T.testGroup "Babbage"
-      [ H.ignoreOnMacAndWindows "leadership-schedule" Test.Cli.Babbage.LeadershipSchedule.hprop_leadershipSchedule
```
newhoggy, line 31: Re-enable leadership schedule test on MacOS.

cardano-testnet/test/Test/Cli/Alonzo/LeadershipSchedule.hs
```diff
@@ -42,15 +42,15 @@ import qualified System.Directory as IO
 import           System.Environment (getEnvironment)
```
Jimbo4350, line 42: So the question is should we bother testing the Alonzo leadership schedule? I can't think of a reason why we should. Did consensus ever give us feedback here?

cardano-testnet/test/Test/Cli/Alonzo/LeadershipSchedule.hs
```diff
@@ -499,15 +498,17 @@ hprop_leadershipSchedule = H.integration . H.runFinallies . H.workspace "alonzo"
...
 
+  maxSlotExpected <- H.noteShow $ maximum expectedLeadershipSlotNumbers
+
   H.assert $ not (L.null expectedLeadershipSlotNumbers)
 
   leadershipDeadline <- H.noteShowM $ DTC.addUTCTime 90 <$> H.noteShowIO DTC.getCurrentTime
 
-  H.assertByDeadlineMCustom "Leader schedule is correct" leadershipDeadline $ do
+  -- Leader schedule is correct
+  H.byDeadlineM 10 leadershipDeadline $ do
```
Jimbo4350, line 508: `"Condition not met by deadline"` is a somewhat opaque error. Can we implement a custom `byDeadlineM` to maintain the custom error messages used by `assertByDeadlineMCustom`

Jimbo4350 (changes requested):

cardano-testnet/test/Test/Cli/Alonzo/LeadershipSchedule.hs
newhoggy: Will require this change: https://github.com/input-output-hk/hedgehog-extras/pull/16

cardano-testnet/test/Test/Cli/Alonzo/LeadershipSchedule.hs
newhoggy: I think I won't get an answer before the New Year, so let's put that in a separate PR.

cardano-testnet/test/Test/Cli/Alonzo/LeadershipSchedule.hs
newhoggy: Updated to include better assertion messages.

cardano-testnet/test/Test/Cli/Alonzo/LeadershipSchedule.hs
newhoggy: I think I won't get an answer before the New Year, so let's put that in a separate PR.

cardano-testnet/test/Test/Cli/Alonzo/LeadershipSchedule.hs
```diff
@@ -499,15 +498,17 @@ hprop_leadershipSchedule = H.integration . H.runFinallies . H.workspace "alonzo"
...
 
+  maxSlotExpected <- H.noteShow $ maximum expectedLeadershipSlotNumbers
+
   H.assert $ not (L.null expectedLeadershipSlotNumbers)
 
   leadershipDeadline <- H.noteShowM $ DTC.addUTCTime 90 <$> H.noteShowIO DTC.getCurrentTime
 
-  H.assertByDeadlineMCustom "Leader schedule is correct" leadershipDeadline $ do
+  -- Leader schedule is correct
+  H.byDeadlineM 10 leadershipDeadline "Wait for a leadership at least as new as the highest one we expect" $ do
```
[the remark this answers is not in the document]
newhoggy: We have to wait until a time when all the slots we expect should exist.  The assertion doesn't do the wait.  Without the wait we would check too early and the slots won't be there.

cardano-testnet/test/Test/Cli/Babbage/LeadershipSchedule.hs
```diff
@@ -86,10 +85,10 @@ hprop_leadershipSchedule = H.integration . H.runFinallies . H.workspace "alonzo"
     , H.execConfigCwd = Last $ Just tempBaseAbsPath
     }
 
-  tipDeadline <- H.noteShowM $ DTC.addUTCTime 180 <$> H.noteShowIO DTC.getCurrentTime
+  tipDeadline <- H.noteShowM $ DTC.addUTCTime 210 <$> H.noteShowIO DTC.getCurrentTime
 
-  H.assertByDeadlineMCustom "stdout does not contain \"until genesis start time\"" tipDeadline $ do
-    H.threadDelay 5000000
+  -- stdout must contain \"until genesis start time\""
```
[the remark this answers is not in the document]
newhoggy: Comment removed.

cardano-testnet/test/Test/Cli/Alonzo/LeadershipSchedule.hs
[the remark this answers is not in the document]
Jimbo4350: `"Wait for a leadership at least as new as the highest one we expect"` is opaque to me. I think here some verbosity is fine because it will make failure diagnosis easier. What about "Wait for chain to surpass all expected leadership slots". And we can add a comment as follows: "We need enough time to pass such that the expected leadership slots generated by the `leadership-schedule` command have actually occurred."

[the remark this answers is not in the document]
newhoggy: Updated!

Jimbo4350 (approved):

PR #4673, opened 2022-11-28
Move implementation inside `runTransactionCmd` to toplevel definitions.

MarcFontaine:


MarcFontaine:
https://github.com/input-output-hk/cardano-node/issues/4672

cardano-cli/src/Cardano/CLI/Shelley/Run/Transaction.hs
```diff
@@ -447,6 +302,215 @@ runTransactionCmd cmd =
 -- Building transactions
 --
 
+runTxBuildCmd
+  :: IsCardanoEra era
+  => CardanoEra era
```
Jimbo4350, line 307: LGTM, however lets use `AnyCardanoEra` here and add some haddock annotations to the parameters where it's not obvious what they are e.g `Maybe Word`.  The haddocks don't have to be overly descriptive, short succinct ones will do.

Jimbo4350 (changes requested):
LGTM but one comment

Jimbo4350 (approved):
LGTM!

MarcFontaine:
https://github.com/input-output-hk/cardano-node/issues/4672

PR #4675, opened 2022-11-28
Replace Data.Map with Data.Map.Strict

### LINKED ISSUE #4648: [FR] - Replace all uses of `Data.Map` with the strict version
**Internal/External**
*Internal*   

There are numerous modules in the `node` code base that uses `Data.Map` when they should be using `Data.Map.Strict` (from `containers`) or `Data.Strict.Map` (from `strict-containers`).

Jimbo4350:
Resolves #4648

### DIFF
--- a/trace-dispatcher/src/Cardano/Logging/Types.hs
+++ b/trace-dispatcher/src/Cardano/Logging/Types.hs
@@ -52,8 +52,8 @@ import qualified Data.Aeson as AE
 import qualified Data.Aeson.Text as AE
 import qualified Data.HashMap.Strict as HM
 import           Data.IORef
-import           Data.Map (Map)
-import qualified Data.Map as Map
+import           Data.Map.Strict (Map)
+import qualified Data.Map.Strict as Map
 import qualified Data.Map.Strict as SMap
 
 import           Data.Text (Text, pack, unpack)
--- a/trace-dispatcher/src/Cardano/Logging/Types.hs
+++ b/trace-dispatcher/src/Cardano/Logging/Types.hs
@@ -54,7 +54,6 @@ import qualified Data.HashMap.Strict as HM
 import           Data.IORef
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
-import qualified Data.Map.Strict as SMap
 
 import           Data.Text (Text, pack, unpack)
 import           Data.Text.Lazy (toStrict)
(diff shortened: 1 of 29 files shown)


erikd (approved):

trace-dispatcher/src/Cardano/Logging/Types.hs
```diff
@@ -52,8 +52,8 @@ import qualified Data.Aeson as AE
 import qualified Data.Aeson.Text as AE
 import qualified Data.HashMap.Strict as HM
 import           Data.IORef
-import           Data.Map (Map)
-import qualified Data.Map as Map
+import           Data.Map.Strict (Map)
+import qualified Data.Map.Strict as Map
 import qualified Data.Map.Strict as SMap
```
newhoggy, line 57: We are importing `Data.Map.Strict` twice here.  Also wondering if we should import `as SMap` to prevent confusion.

erikd: I would actually just suggest replacing all uses of `SMap` with `Map`.

newhoggy: So `LMap` if we ever need to reach for a lazy map?

LudvikGalois: If it ever comes up, personally, I'd suggest importing `Data.Map.Lazy` as `Lazy` unless that clashes with something, because it reads rather nicely.
The underlying datatype is shared between `Data.Map.Lazy` and `Data.Map.Strict`, so simply using `Data.Map.Strict` isn't a magic bullet, since something innocuous like `fmap` won't keep your values in WHNF and you can easily be given "lazy" maps where the values aren't already in WHNF. Honestly, I think it'd be better off as a newtype if you "really" want strictness, but that's probably be a discussion for another time.

erikd: I think `Lazy` is too generic (both `Text` and `ByetString` have lazy versions. I would name is `LazyMap`.

Jimbo4350: I'm changing it to `Map` in this PR. However John and Robert can have the ultimate decision (whilst considering Erik's input) and open a new PR with the relevant changes. This PR should also update the style doc so we can compile a list of how modules should be imported so it's not ambiguous.

PR #4683, opened 2022-11-30
[cardano-api] Derive Eq instance for AcquiringFailure

ch1bo:
This is often used in tests where we would like to assert that we receive one or the other.


cardano-api/ChangeLog.md
```diff
@@ -14,6 +14,8 @@
 
 - **Breaking change** - Reduce exposed modules in cardano-api ([PR4546](https://github.com/input-output-hk/cardano-node/pull/4546))
 
+- **Breaking change** Change return type of `queryNodeLocalState` to new `AcquiringFailure` type.
```
Jimbo4350, line 17: Where is this breaking change?

newhoggy (approved):

PR #4684, opened 2022-11-30
[cardano-api] Export fromConsensusPointHF

ch1bo:
As toConsensusPointHF is also exported


ch1bo:
Not sure if this is intended to be used in the first place? Or do we want to guide users to use `fromConsensusPointInMode`? The implementation of that one seems to be always `fromConsensusPointHF` though.

PR #4693, opened 2022-12-05
Remove `hydra.iohk.io` cache

amesgen:
When accepting this cache locally (which can easily happen accidentally), everything gets super slow as every request timeouts.


amesgen:
Superseded by #4695

PR #4717, opened 2022-12-16
Use cabal-cache

newhoggy:
Use `cabal-cache` to cache `cabal` packages.

This is better than using `actions/cache@v2` because `cabal-cache` is a fine grained cache that caches each built cabal package individually.

This means that if any packages were built successfully in CI, even if CI fails, we can still upload those packages that successfully built.  This is far superior to the situation we have now which is if CI fails we cache nothing.

This is especially useful for those occasions where we update the index state or change our package dependencies (including package dependency bounds) because currently for those cases, a failed build means we cache nothing which hurts iteration times very badly.

The change keeps the build time more reliably tight:

Screenshot 2022-12-21 at 2 12 17 pm

Compare this to the old caching method:

Screenshot 2022-12-21 at 5 17 59 pm

The PR also fixes a bug which mean that MacOS builds weren't cached at all (ie. the `89m` MacOS build time in the old caching method).

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -40,12 +40,6 @@ jobs:
       LD_LIBRARY_PATH: ${{ (matrix.os != 'windows-latest' && '/usr/local/lib') || '' }}
 
     steps:
-    - name: Setup tmate session
-      if: ${{ failure() }}
-      uses: mxschmitt/action-tmate@v3
-      with:
-        limit-access-to-actor: true
-
     - name: "WIN: Install System Dependencies via pacman (msys2)"
       if: runner.os == 'Windows'
       run: |
@@ -61,15 +55,15 @@ jobs:
             automake `
             libtool `
             make
-  
+
      # this seems to break something. It _must_ come after the pacman setup
      # above. It appears as if PATHEXT is set _after_ ghcup install ghc/cabal, and
      # as such we'd need pacman.exe instead.
     - name: Setup Haskell
-      run: |        
+      run: |
         # Use GHCUP to manage ghc/cabal
         ghcup install ghc --set ${{ matrix.ghc }}
-        ghcup install cabal --set 3.6.2.0
+        ghcup install cabal --set 3.8.1.0
 
         ghc --version
         cabal --version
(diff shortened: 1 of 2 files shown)


.github/workflows/haskell.yml
```diff
@@ -96,16 +96,18 @@ jobs:
...
-      run: echo "cabal-store=C:\\cabal\\store" >> $GITHUB_OUTPUT   
+      run: |
+        cabal help user-config
+        cabal_config_file="$(cabal help user-config | grep -A 1 'You can edit the cabal configuration file to set defaults' | tail -n 1 | xargs)"
+
+        if [[ '${{ runner.os }}' != 'Windows' ]]; then
+          echo "cabal-store=$(dirname "$cabal_config_file")/store" | tee -a "$GITHUB_OUTPUT"
+        else
+          echo "cabal-store=C:\\cabal\\store" | tee -a "$GITHUB_OUTPUT"
+        fi
```
newhoggy, line 110: Unify the computation of the `cabal` store directory to a single step.  This makes referencing the cabal store in later steps easier.

This also fixes the problem where on `MacOS`, the `cabal-store=/home/runner/.cabal/store` is incorrect.  This should be `/Users/runner/.cabal/store`.

To avoid this problem in future, on Linux and MacOS, the step queries `cabal` for the some information to compute the location of the `cabal` store.

.github/workflows/haskell.yml
```diff
@@ -186,27 +188,28 @@ jobs:
...
-          cache-${{ env.CACHE_VERSION }}-${{ runner.os }}-${{ matrix.ghc }}
-
-    - uses: actions/cache@v2
-      name: "Cache `dist-newstyle`"
+        region: us-west-2
+        dist-dir: dist-newstyle
+        store-path: ${{ steps.cabal-store.outputs.cabal-store }}
+        threads: 16
+        archive-uri: ${{ secrets.BINARY_CACHE_URI }}
+        skip: "${{ secrets.BINARY_CACHE_URI == '' }}"
```
newhoggy, line 202: This step syncs the cabal store with `S3`, an object store service (think a service for storing files).

The S3 bucket is set to `s3://iohk.cache.haskellworks.io`.

.github/workflows/haskell.yml
```diff
@@ -186,27 +188,28 @@ jobs:
...
-          dist-newstyle
-          !dist-newstyle/**/.git
-        key: cache-dist-${{ env.CACHE_VERSION }}-${{ runner.os }}-${{ matrix.ghc }}-${{ steps.record-deps.outputs.weeknum }}
-        restore-keys: cache-dist-${{ env.CACHE_VERSION }}-${{ runner.os }}-${{ matrix.ghc }}
+        dist-dir: dist-newstyle
+        store-path: ${{ steps.cabal-store.outputs.cabal-store }}
+        threads: 16
+        archive-uri: https://iohk.cache.haskellworks.io
+        skip: "${{ secrets.BINARY_CACHE_URI != '' }}"
+        enable-save: false
```
newhoggy, line 212: It's important to ensure that people who fork this repository can not only successfully build in CI by default, but also have meaning cabal store caching.

Because syncing with S3 requires credentials, we cannot rely on S3 for this.  For this reason a `https` fallback is used.  The `https` server mirrors the content of the S3 bucket.  The `https` cabal store archive is read-only for security reasons.

Users who fork this repository who want to have a writable cabal store archive are encouraged to set up their own S3 bucket.

.github/workflows/haskell.yml
```diff
@@ -288,6 +291,8 @@ jobs:
     # - name: Setup tmate session
     #   if: ${{ failure() }}
     #   uses: mxschmitt/action-tmate@v3
+    #   with:
+    #     limit-access-to-actor: true
```
newhoggy, line 295: Adding this to the commented out `tmate` step for security reasons in case someone decides to use it.

.github/workflows/haskell.yml
Jimbo4350: Can you put this as a comment in the file?

.github/workflows/haskell.yml
Jimbo4350: Likewise

.github/workflows/haskell.yml
Jimbo4350: and this

Jimbo4350 (approved):
Nice :+1: . Just add the comments to the `haskell.yml` file

PR #4719, opened 2022-12-16
Cardano governance documentation

CarlosLopezDeLara:
This article describes the update proposal mechanism for Byron and Shelley+ eras. Demonstrates how to use the CLI to create update proposals for any era.


olgahryniuk (approved):

doc/reference/cardano-governance.md
```diff
@@ -0,0 +1,235 @@
...
+
+Update proposals can be used to introduce changes to the values of specific parameters (e.g., the maximum block size, the maximum transaction size etc.), changes to the validation rules at any level (transaction, block, or blockchain), or even changes at the consensus protocol itself.
+
+Protocol parameters are used to affect the operation of the Cardano protocol. They may be either **updatable** or **non-updatable**. Updatable parameters can be tuned to vary the operation of the block-producing protocol, impacting the proportion of pools that are federated/non-federated, influence of the 'pledge', etc. Non-updatable parameters affect the fundamentals of the blockchain protocol, which includes defining the genesis block, basic security properties etc. [CIP-9](https://cips.cardano.org/cips/cip9/) and [CIP-28](https://cips.cardano.org/cips/cip28/) detail which parameters are updatable and which are not.
```
JaredCorduan, line 7: This terminology is at odds with the terminology that we use in the ledger (the code and the specs). In the ledger, **protocol parameters** are _all_ updatable. Values in the ledger that are hard coded we refer to as **global constants**. It might be fine to have different definitions, I'm just pointing out a potential confusion... I see now that the terminology used here is consistent with those two CIPs, so potential confusion already exists :)

doc/reference/cardano-governance.md
```diff
@@ -0,0 +1,235 @@
...
+
+### Cardano eras
+
+Cardano has come through different development **eras**, each of which introduced new functionalities. Going from one era to the next requires that new features (which might require changes in the ledger, consensus, network, etc) are incorporated into a new node version. For this, an _update proposal_ must be submitted to trigger the update which is handled by the _hardfork combinator_.
```
JaredCorduan, line 11: do you want to mention/link [CIP-59](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0059)? it tries to be clear about what "era" means.

doc/reference/cardano-governance.md
```diff
@@ -0,0 +1,235 @@
...
+```
+In this case, delegate nodes must be running a node version that is compatible with the protocol version 2.0 to endorse the update.
+
+## Update proposals in Shelley+ eras
+
+There are some changes in the update proposal mechanism in Shelley and later eras: there is no distinction between votes and proposals. To 'vote' for a proposal, a _delegate_ submits the exact same update proposal. Update proposals and votes are submitted in a regular transaction. Also, there is no separate endorsement step.
+
+The procedure is as follows:  
+
+![](../reference/images/shelleyup.png)
```
JaredCorduan, line 131: this picture is fantastic!  would it help to mention that the entire epoch has 10k/f many slots?

doc/reference/cardano-governance.md
CarlosLopezDeLara: Perhaps we to amend the CIP-9 so that non-updatable parameters are referred to as global constants?

JaredCorduan: That's a great idea!

doc/reference/cardano-governance.md
```diff
@@ -0,0 +1,235 @@
...
+```
+In this case, delegate nodes must be running a node version that is compatible with the protocol version 2.0 to endorse the update.
+
+## Update proposals in Shelley+ eras
+
+There are some changes in the update proposal mechanism in Shelley and later eras: there is no distinction between votes and proposals. To 'vote' for a proposal, a _delegate_ submits the exact same update proposal. Update proposals and votes are submitted in a regular transaction. Also, there is no separate endorsement step.
+
+The procedure is as follows:  
+
+![](../reference/images/upshelley.png)
```
Jimbo4350, line 131: Twice in your image you say the update proposal is registered. Should the first instance say the update proposal is _submitted_ and the "registered" after a quorum is reached? I haven't looked at the spec to check this but I noticed the use of "registered" twice.

carloslodelar: Update proposal can be registered earlier than 6k/f slots to end of epoch or later than tat. The image tries to show the behavior on both cases.

CarlosLopezDeLara: Update proposals can be registered earlier than 6k/f slots to the end of the epoch, or later. The image tries to show the behavior on both scenarios.

disassembler (approved):

doc/reference/cardano-governance.md
Jimbo4350: Sure but its not clear in the diagram. Preceding with "If the update proposal is registered here" would be clearer as right now is looks like two update proposals are being submitted.
Actually I think the diagram is wrong. The blue and purple lines should be interchanged.

CarlosLopezDeLara: I think the confusion might come from the term valid (from the spec). Valid does not mean "adopted". Valid means that if the update proposal is  registered within the 6k/f slots to the end of the epoch  window, can only start accumulating votes until the next epoch. 

Perhaps I can split this in two diagrams, one for each case 1) earlier than 6k/f 2) later than 6k/f

PR #4722, opened 2022-12-19
Fix protocolVersion for cardano-testnet Babbage genesis.json

james-iohk:
The problem is if using module `Testnet.Babbage` from `cardano-testnet` the protocol version used is actually `0 0`. This means that a Plutus script doing anything useful, that's anything other than resolving as `()`,  will result in the mysterious `MalformedScriptWitnesses` error.


Jimbo4350:
Nice job debugging this

Jimbo4350 (approved):

james-iohk:
closing, see #4729

PR #4729, opened 2022-12-21
fix cardano-testnet Babbage genesis.json

james-iohk:
The problem is if using module `Testnet.Babbage` from `cardano-testnet` the protocol version used is actually `0 0`. This means that a Plutus script doing anything useful, that's anything other than resolving as `()`,  will result in the mysterious `MalformedScriptWitnesses` error.


Jimbo4350 (approved):
Good catch!

PR #4735, opened 2022-12-23
Memoise to pparams

steshaw:


steshaw:
Oops. Opened in error.

PR #4748, opened 2023-01-04
nix: optimize caching by delaying setGitRev and other nix fixes

jbgi:
other simplification/optimization:
  - better filtering of haskell.nix project `src`
  - remove generated project-packages-exes.nix (now at eval time)
  - move rewrite-libs to binary-release.nix: not needed and faster for normal builds
  - use haskell.nix project overlays to simplify access to components and alternative projects config (profiled, etc.).

With this, lots for derivation become much less sensible to rebuild for unrelated changes. It should also prevent lots of now unnecessary cache downloads, making cicero faster.

Also include a fix for hls (fixing `nix develop`).


deepfire (approved):

nix/haskell.nix
```diff
@@ -4,193 +4,259 @@
...
+    inherit inputMap;
+    name = "cardano-node";
+    # We clean-up src to avoid rebuild for unrelated changes:
+    src = haskellLib.cleanSourceWith {
+      src = ../.;
+      name = "cardano-node-src";
+      filter = path: type: let
+        relPath = lib.removePrefix "${../.}/" path;
+      in
+        # excludes top-level directories not part of cabal project (or used by tests):
```
michaelpj, line 30: Is this actually better than the builtin haskell.nix filtering? I thought it filtered things pretty well, so it only pulls in files that are literally mentioned in the cabal file?

nix/haskell.nix
```diff
@@ -4,193 +4,259 @@
...
+        # configureArgs = lib.optionalString stdenv.hostPlatform.isWindows "--disable-tests";
+      ];
+  });
+in project.appendOverlays (with haskellLib.projectOverlays; [
+  projectComponents
+  (final: prev: let inherit (final.pkgs) lib; in {
+    profiled = final.appendModule {
+      modules = [{
+        enableLibraryProfiling = true;
+        packages.cardano-node.components.exes.cardano-node.enableProfiling = true;
```
michaelpj, line 208: Why not just `enableProfiling = true` rather than these specific exes?

nix/pkgs.nix
```diff
@@ -25,7 +25,9 @@ final: prev: with final; {
   };
 
   haskell-language-server = haskell-nix.tool compiler-nix-name "haskell-language-server" {
-    version = "latest";
+    # XXX waiting for https://github.com/haskell/haskell-language-server/issues/3427
+    # before switching back to latest:
```
michaelpj, line 29: it's fixed already

nix/haskell.nix
jbgi: as discussed, this added filtering is mostly for tests that depends on the whole tree (TODO added to fix this).

jbgi:
could some on mac test that cardan-node static binary in
`nix build .#hydraJobs.x86_64-darwin.cardano-node-macos` is still working correcty?

newhoggy (approved):

jbgi:
bors try

PR #4756, opened 2023-01-07
Ignore IntellIJ folder and cabal.project.local*

weebl2000:
Cabal has a tendency to create more files appended with ~x if you run cabal configure. This commit ignores these files, as well as the IntellIJ hidden project configuration directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@
 .nix-shell-cabal.project
 /cabal.project.diff
 /cabal.project.freeze
-/cabal.project.local
+/cabal.project.local*
 /cabal.project.old
 configuration/defaults/simpleview/genesis/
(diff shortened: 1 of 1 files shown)


weebl2000:
> This PR is stale because it has been open 45 days with no activity.

Still useful, not.

newhoggy:
For IntelliJ, it's standard practise to ignore `.idea` from a global `gitignore`.  https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

newhoggy:
What's the reason for ignoring `cabal.project.local*`?

weebl2000:
> What's the reason for ignoring `cabal.project.local*`?

Cabal has a tendency to create more iterations of them if you configure the project and the file already exists.

Issue #4725, opened 2022-12-20
[BUG] - Problem with cardano-cli genesis create-cardano --gen-utxo-keys


CarlosLopezDeLara:
**Internal/External**
*Internal*   

**Area**
cardano-cli 

**Summary**
When running `cardano-cli genesis create-cardano` the option `--gen-utxo-keys` has no effect at all.  

It's supposed to default to zero, however  it always creates one utxo key with the amount in `--supply`
```
  --gen-utxo-keys INT      The number of UTxO keys to make [default is 0].
```

Also, using `--gen-utxo-keys`  with N > 1 does not change anything. Only one key is produced. 

**Steps to reproduce**

1. Get template files from cardano-world
2. Run 

```bash
cardano-cli genesis create-cardano \
--genesis-dir ./ \
--gen-genesis-keys 3 \
--gen-utxo-keys 3 \
--start-time $(date -u -d "now + 2 minutes" +%FT%Tz) \
--supply 30000000000000000 \
--security-param 45 \
--slot-length 100 \
--slot-coefficient 5/100 \
--testnet-magic 42 \
--byron-template template/byron.json \
--shelley-template template/shelley.json \
--alonzo-template template/alonzo.json \
--node-config-template template/config.json
```
3. The folder utxo-keys will contain only 1 set of utxo keys:
```
utxo-keys/
├── byron.000.key
├── shelley.000.skey
└── shelley.000.vkey

0 directories, 3 files
```

**Expected behavior**
It should either be explicit about numUtxoKeys being hardcoded to 1 or generate as many keys as the user sets in `--gen-utxo-keys` 

**System info (please complete the following information):**
```
$ cardano-node --version
cardano-node 1.35.4 - linux-x86_64 - ghc-8.10
git rev ebc7be471b30e5931b35f9bbc236d21c375b91bb
$ cardano-cli --version
cardano-cli 1.35.4 - linux-x86_64 - ghc-8.10
git rev ebc7be471b30e5931b35f9bbc236d21c375b91bb
```

MarcFontaine:
The parameter is ignored in the actual implementation.
There is also a LOC where `1` is hardcode. It looks like this is easy to fix,
but could not find the right template files...
Where exactly did you find the template files ?

CarlosLopezDeLara:
@MarcFontaine I use template files from devops https://github.com/input-output-hk/cardano-world/tree/master/docs/environments/testnet

MarcFontaine:
[...]

PR #4761, opened 2023-01-09
Fix bug in hash computation in 'cardano-cli genesis create-cardano'.

MarcFontaine:
Use value passed with '--gen-utxo-keys' in 'cardano-cli genesis create-cardano'.

Resolves: #4725

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
+++ b/cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TupleSections #-}
@@ -114,6 +115,7 @@ import           Cardano.Slotting.Slot (EpochSize (EpochSize))
 import           Data.Fixed (Fixed (MkFixed))
 import qualified Data.Yaml as Yaml
 import           Text.JSON.Canonical (parseCanonicalJSON, renderCanonicalJSON)
+import qualified Text.JSON.Canonical (ToJSON)
 
 import           Data.ListMap (ListMap (..))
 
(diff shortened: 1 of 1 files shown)


MarcFontaine:
https://github.com/input-output-hk/cardano-node/issues/4725

CarlosLopezDeLara:
Tested. Works perfectly.

disassembler (approved):

Jimbo4350 (changes requested):
Don't merge this yet

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
```diff
@@ -1160,13 +1145,25 @@ updateCreateStakedOutputTemplate
...
+  where
+    content = case genesis of
+       WritePretty a -> LBS.toStrict $ encodePretty a
+       WriteCanonical a -> LBS.toStrict
+          . renderCanonicalJSON
+          . either (error "error parsing json that was just encoded!?") identity
+          . parseCanonicalJSON
+          . canonicalEncodePretty $ a
+
+data WriteFileGenesis where
```
Jimbo4350, line 1164: It would be better to avoid the creation of another data type. We can simply have a function defined in `Cardano.CLI.Byron.Genesis` that does the correct JSON encoding for the Byron genesis and define a function here specifically for the Shelley genesis. Then just call those where appropriate.

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
```diff
@@ -636,7 +621,7 @@ runGenesisCreateCardano (GenesisDir rootdir)
                       (toByronRequiresNetworkMagic network)
     byronBalance = TestnetBalanceOptions
         { tboRichmen = genNumGenesisKeys
-        , tboPoors = 1
+        , tboPoors = genNumUTxOKeys
```
Jimbo4350, line 624: This could be in a commit on its own, which will make it easier to see that you are fixing the previously ignored "genNumUTxOKeys" argument.

Jimbo4350 (changes requested):
Nice, a few comments

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
MarcFontaine: I do not know why Byron was special at first. I just reordered the existing code.
There is a single function that handles writing Byron and Shelley genesis files.
The data type is local and not exported. It makes the arguments of the function more explicit.
The definition of the data type itself is only three lines of code.
Why not create that data type ?

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
MarcFontaine: That single line would not fix it.
@CarlosLopezDeLara found out that the node rejects the generated genesis files.
(I confirmed it).
I still believe that the original version of `cardano-cli genesis create-cardano` was not working.

cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
Jimbo4350: Actually this is fine as its not exported :+1:

Jimbo4350 (approved):

Issue #4261, opened 2022-08-03
[BUG] - SimpleScriptV1 in reference script UTxO is reported as SimpleScriptV2


mkoura:
*Internal*

**Area**
*Other*   Any other topic (Delegation, Ranking, ...).

**Summary**
For a UTxO with SimpleScriptV1, the output of `query utxo` reports the script as SimpleScriptV2.
E.g. for SimpleScriptV1 like
```
{
    "keyHash": "ca42373873e057dee3201a74a2ecd58a1680a319332f915cd28fa38e",
    "type": "sig"
}
```
the query output is
```
$ cardano-cli query utxo --tx-in "eb1f9c60dde3db8281de6c8fc665930119a7f29b21b8f4e2bde9c4823294b32f#0" --testnet-magic 42  --out-file /dev/stdout
{
    "eb1f9c60dde3db8281de6c8fc665930119a7f29b21b8f4e2bde9c4823294b32f#0": {
        "address": "addr_test1vr9yydecw0s90hhryqd8fghv6k9pdq9rryejly2u62868rsj8xlyk",
        "datum": null,
        "datumhash": null,
        "inlineDatum": null,
        "referenceScript": {
            "script": {
                "cborHex": "8200581cca42373873e057dee3201a74a2ecd58a1680a319332f915cd28fa38e",
                "description": "",
                "type": "SimpleScriptV2"
            },
            "scriptLanguage": "SimpleScriptLanguage SimpleScriptV2"
        },
        "value": {
            "lovelace": 4000000
        }
    }
}
```

**Expected behavior**
The script is reported as "SimpleScriptV1".

**System info (please complete the following information):**
- OS Name: Fedora 35
- Node version: aed8e71339cf6c92847fff83fbd92be61e468174
- CLI version: aed8e71339cf6c92847fff83fbd92be61e468174

_____
Acceptance Criteria

- For a UTxO with SimpleScriptV1, the output of `query utxo` reports SimpleScriptV1

PR #4763, opened 2023-01-09
Remove simple script distinction

Jimbo4350:
Resolves: #4261 
- [Timelock](https://github.com/input-output-hk/cardano-ledger/blob/31c0bb1f5e78e40b83adfd1a916e69f47fdc9835/eras/allegra/impl/src/Cardano/Ledger/Allegra/Scripts.hs#L105) scripts are a superset of [MultiSig](https://github.com/input-output-hk/cardano-ledger/blob/31c0bb1f5e78e40b83adfd1a916e69f47fdc9835/eras/shelley/impl/src/Cardano/Ledger/Shelley/Scripts.hs#L70) scripts in that they can also express validation in terms of before or after a particular slot. 

- In `cardano-ledger` the serialization of Timelock and MultiSig scripts are identical. 

Due to these properties, it is not necessary to distinguish between MultiSig and Timelock scripts. We can default to Timelock scripts without issue.

### DIFF
--- a/cardano-api/src/Cardano/Api/Script.hs
+++ b/cardano-api/src/Cardano/Api/Script.hs
@@ -1153,7 +1153,6 @@ instance ToJSON SimpleScript where
 instance FromJSON SimpleScript where
   parseJSON = parseSimpleScript
 
--- TODO: Left off here. You need to comb through cardano-api's property tests concerning simple scripts
 parseSimpleScript :: Value -> Aeson.Parser SimpleScript
 parseSimpleScript v = parseScriptSig v <|>
                       parseScriptBefore v <|>
--- a/cardano-api/src/Cardano/Api/Script.hs
+++ b/cardano-api/src/Cardano/Api/Script.hs
@@ -2,6 +2,7 @@
 {-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GADTs #-}
+{-# LANGUAGE InstanceSigs #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE RankNTypes #-}
@@ -15,18 +16,15 @@
 
 module Cardano.Api.Script (
     -- * Languages
-    SimpleScriptV1,
-    SimpleScriptV2,
+    SimpleScript',
     PlutusScriptV1,
     PlutusScriptV2,
     ScriptLanguage(..),
-    SimpleScriptVersion(..),
     PlutusScriptVersion(..),
     AnyScriptLanguage(..),
     AnyPlutusScriptVersion(..),
     IsPlutusScriptLanguage(..),
     IsScriptLanguage(..),
-    IsSimpleScriptLanguage(..),
 
     -- * Scripts in a specific language
     Script(..),
@@ -65,9 +63,6 @@ module Cardano.Api.Script (
 
     -- * The simple script language
     SimpleScript(..),
-    TimeLocksSupported(..),
-    timeLocksSupported,
-    adjustSimpleScriptVersion,
     SimpleScriptOrReferenceInput(..),
 
     -- * The Plutus script language
@@ -115,6 +110,7 @@ import           Data.ByteString.Short (ShortByteString)
 import qualified Data.ByteString.Short as SBS
 import           Data.Either.Combinators (maybeToRight)
 import           Data.Foldable (toList)
+import           Data.Functor
 import           Data.Scientific (toBoundedInteger)
 import           Data.String (IsString)
 import           Data.Text (Text)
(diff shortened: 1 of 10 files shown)


cardano-api/src/Cardano/Api/Script.hs
```diff
@@ -183,34 +180,32 @@ import           Cardano.Api.Utils (failEitherWith)
...
--- of 'SimpleScriptV1' plus new atomic predicates:
+-- of the original simple script language plus new atomic predicates:
 --
 -- * require the time be before a given slot number
 -- * require the time be after a given slot number
 --
 -- This version of the language was introduced in the 'AllegraEra'.
 --
-data SimpleScriptV2
+-- However we opt for a single type level tag 'SimpleScript\'' as the second version of
```
newhoggy, line 192: From Haddock documentation:

> Nothing special is needed to hyperlink identifiers which contain apostrophes themselves: to hyperlink foo' one would simply type 'foo''. Hyperlinking operators works in exactly the same way.

https://haskell-haddock.readthedocs.io/en/latest/markup.html

cardano-api/src/Cardano/Api/Script.hs
```diff
@@ -446,16 +406,8 @@ instance IsScriptLanguage lang => SerialiseAsCBOR (Script lang) where
...
-            . fromShelleyMultiSig
-          <$> CBOR.decodeAnnotator "Script" fromCBOR (LBS.fromStrict bs)
-
-        SimpleScriptLanguage SimpleScriptV2 ->
-              SimpleScript SimpleScriptV2
-            . (fromAllegraTimelock TimeLocksInSimpleScriptV2
-                                :: Timelock.Timelock StandardCrypto
-                                -> SimpleScript SimpleScriptV2)
+        SimpleScriptLanguage ->
+              SimpleScript .(fromAllegraTimelock :: Timelock.Timelock StandardCrypto -> SimpleScript)
```
newhoggy, line 410: Is the type ascription necessary?

cardano-api/src/Cardano/Api/Script.hs
```diff
@@ -1170,62 +1027,56 @@ fromShelleyBasedScript  :: ShelleyBasedEra era
...
   where
-    go :: SimpleScript SimpleScriptV1 -> Shelley.MultiSig StandardCrypto
+    go :: SimpleScript -> Shelley.MultiSig StandardCrypto
     go (RequireSignature (PaymentKeyHash kh))
                         = Shelley.RequireSignature (Shelley.coerceKeyRole kh)
     go (RequireAllOf s) = Shelley.RequireAllOf (map go s)
     go (RequireAnyOf s) = Shelley.RequireAnyOf (map go s)
     go (RequireMOf m s) = Shelley.RequireMOf m (map go s)
+    go timelock = error $ "toShelleyMultiSig: " <> show timelock <>
+                          " not supported in MultiSig scripts."
```
newhoggy, line 1075: I'm wondering if it might be better not to `error` here.

Using `error` from the CLI is okay, but `error` would cause an application that uses the API to exit.

cardano-api/src/Cardano/Api/Script.hs
```diff
@@ -1303,42 +1151,38 @@ instance ToJSON (SimpleScript lang) where
...
 
-parseSimpleScript :: SimpleScriptVersion lang
-                  -> Value -> Aeson.Parser (SimpleScript lang)
-parseSimpleScript lang v = parseScriptSig          v
-                       <|> parseScriptBefore  lang v
-                       <|> parseScriptAfter   lang v
-                       <|> parseScriptAny     lang v
-                       <|> parseScriptAll     lang v
-                       <|> parseScriptAtLeast lang v
+-- TODO: Left off here. You need to comb through cardano-api's property tests concerning simple scripts
```
newhoggy, line 1157: The TODO could use some more elaboration for the benefit of the developer who will end up implementing it.

cardano-api/src/Cardano/Api/Script.hs
Jimbo4350: So I opted to return an `Either` as you suggested for `toShelleyMultiSig`. However `toShelleyMultiSig` is called in: `toShelleyScript :: ScriptInEra era -> Ledger.Script (ShelleyLedgerEra era)`. I opted to error here instead because this will `error` if someone is running a Shelley era network and tries to use a Timelock. I think the odds of this is extremely low and not worth polluting the code further with an `Either`. Lmk what you think.

cardano-api/src/Cardano/Api/Script.hs
Jimbo4350: Apparently not!

cardano-api/src/Cardano/Api/Script.hs
newhoggy: I think this is okay to merge, but I also that we should create a ticket to remove as many `error` calls as we reasonably can.

newhoggy (approved):

cardano-api/src/Cardano/Api/Script.hs
newhoggy: I think let this pass for now given the low probably of it being a problem, but we should also create a ticket to remove calls to `error` from `cardano-api` as a separate piece of work to be done after the next release.

PR #4764, opened 2023-01-10
Update ouroboros-network

newhoggy:


cardano-cli/cardano-cli.cabal
```diff
@@ -133,6 +133,8 @@ library
                       , ouroboros-consensus-protocol
                       , ouroboros-consensus-shelley
                       , ouroboros-network
+                      , ouroboros-network-api
+                      , ouroboros-network-protocols
```
coot, line 137: Nice, so `cardano-cli` does not depend on `ouroboros-consensus-diffusion` :)

cardano-node/cardano-node.cabal
```diff
@@ -168,10 +168,13 @@ library
                       , ouroboros-consensus ^>= 0.1.0.2
                       , ouroboros-consensus-byron
                       , ouroboros-consensus-cardano
+                      , ouroboros-consensus-diffusion
```
coot, line 171: @nfrisby @dnadales :thinking: maybe we should also have `ouroboros-consensus-api` which exports a stable api and removes the direct dependency of `lib:cardano-node` on `ouroboros-consensus` :bulb: (I have no idea how intrusive that would be).

cardano-tracer/cardano-tracer.cabal
```diff
@@ -140,7 +140,7 @@ library
                       , filepath
                       , mime-mail
                       , optparse-applicative
-                      , ouroboros-network
+                      , ouroboros-network-api
```
coot, line 143: nice :smile:

coot (commented):
LGTM

coot (approved):
LGTM

bolt12:
bump 😄

cabal.project
```diff
@@ -149,17 +150,27 @@ source-repository-package
...
+     ouroboros-network-mock
+     ouroboros-network-protocols
      ouroboros-network-testing
      ouroboros-consensus-cardano-tools
+
+source-repository-package
+   type: git
+   location: https://github.com/input-output-hk/ekg-forward-fork
+   tag: 6cf25fbbaf9815e07f695982da4cfbb1581239c3
+   --sha256: 0qwvx04l9q3h7acb0a3b8iarqcn5bkqhwaplbga5pvryqp3lhn00
```
coot, line 176: Why are we using a fork of our own library?

coot (changes requested):

cabal.project
coot: Also we shouldn't use `srp` any more, if we need a newer version of `ekg-forward` lets release one to `CHaP`.

newhoggy:
This is obsolete.

PR #4765, opened 2023-01-10
cardano-cli | create-staked:  fix UTxO size distribution

deepfire:
1. fix the `create-staked` distribution, where the first stake pool was getting the majority of funds.
2. make `create-staked` print diagnostic output to `stderr`


cardano-cli/src/Cardano/CLI/Shelley/Run/Genesis.hs
```diff
@@ -754,7 +754,7 @@ runGenesisCreateStaked (GenesisDir rootdir)
   writeFileGenesis (rootdir </> "genesis.alonzo.json") alonzoGenesis
   --TODO: rationalise the naming convention on these genesis json files.
 
-  liftIO $ Text.putStrLn $ mconcat $
+  liftIO $ Text.hPutStrLn stderr $ mconcat $
```
Jimbo4350, line 757: Out of curiosity, why this change?

deepfire: To simplify script integration.

Jimbo4350 (approved):

PR #4766, opened 2023-01-10
Check cabal files in CI

newhoggy:


scripts/ci/check-cabal-files.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+for x in $(find . -name '*.cabal' | grep -v dist-newstyle | cut -c 3-); do
```
Jimbo4350, line 3: so this is finding all the cabal files in the cardano-node repo?

Jimbo4350 (approved):
Nice work, LGTM!

PR #4774, opened 2023-01-11
Use -Werror by default only in CI

coot:
If one wants to use `-Werror` in development, one can always set it in
`cabal.project.local` file.


coot:
This is an alternative for #2772.

Jimbo4350 (changes requested):
The issue I have with this is if CI is busted (like it is now), people could potentially merge things with warnings
 
> one can always set it in cabal.project.local file

Likewise if you don't like -Werror

PR #4780, opened 2023-01-12
Stop running CI when the only changes are to docs

Jimbo4350:


.github/workflows/haskell.yml
```diff
@@ -17,6 +20,9 @@ jobs:
   build:
     runs-on: ${{ matrix.os }}
 
+    # Always run on master
+    if: github.event_name != 'create' || github.ref_name == 'master'
```
newhoggy, line 24: What's the `github.event_name != 'create'` for?

Are we saying run only if this is not for a `create` event (which as I understand happens when a branch or tag is created)?

What are we admitting and excluding in the non-`master` case?

angerman: I agree, this conditional here seems somewhat odd.

PR #4785, opened 2023-01-15
Straight line error handling

newhoggy:


cardano-api/src/Cardano/Api/Utils.hs
```diff
@@ -59,6 +61,9 @@ Just x  ?! _ = Right x
 Left  e ?!. f = Left (f e)
 Right x ?!. _ = Right x
 
+throwNothingM :: Monad m => e -> Maybe a -> ExceptT e m a
```
Jimbo4350, line 64: Thoughts on the name `throwOnNothingM`? Also why the M?

Jimbo4350 (commented):
This is a lot better than what we had before :+1:  . I just want discuss the name of `throwNothingM` but happy to approve after.

cardano-api/src/Cardano/Api/Utils.hs
newhoggy: I think the `&` reads okay.  It's a well known operator that I just read it as "and".

For example `f & throwOnNothing` I would read as do `f` and throw on nothing.

I think keeping the `&` is also better because it fits in with other operators like `<&>` that we might want to use in other contexts.

cardano-api/src/Cardano/Api/Convenience/Query.hs
```diff
@@ -63,48 +66,37 @@ queryStateForBalancedTx
...
-              pparams <- ePparams
-              eraHistory <- first AcqFailure eEraHistory
-              systemStart <- first AcqFailure eSystemStart
-              stakePools <- eStakePools
-              Right (utxo, pparams, eraHistory, systemStart, stakePools)
-          Nothing -> return $ Left $ EraConsensusModeMismatch
-                                      (AnyConsensusMode CardanoMode)
-                                      (getIsCardanoEraConstraint era $ AnyCardanoEra era)
+queryStateForBalancedTx era networkId allTxIns = runExceptT $ do
+  SocketPath sockPath <- ExceptT $ first SockErr <$> readEnvSocketPath
```
Jimbo4350, line 70: I'm assuming you're not using `newExceptT` because `ExceptT` is shorter?

newhoggy: Yeah, that's right.

cardano-api/src/Cardano/Api/Utils.hs
```diff
@@ -50,6 +51,7 @@ import           System.Directory (emptyPermissions, readable, setPermissions)
 #endif
 
 import           Cardano.Api.Eras
+import Control.Monad.Trans.Except (ExceptT, throwE)
```
Jimbo4350, line 54: :'(

PR #4787, opened 2023-01-15
New NodeToClientVersionOf typeclass

newhoggy:
This allows code to check what version of the node to client protocol is needed to use the query.  For example:

```
ntcVersionOf $ QueryInEra qeInMode $ QueryInShelleyBasedEra qSbe QueryProtocolParameters
```

Whilst this function can be called directly, it is intended to be called by `queryExpr` eventually so that only queries that are valid for the current protocol are ever sent.

### DIFF
--- a/cardano-api/src/Cardano/Api/Eras.hs
+++ b/cardano-api/src/Cardano/Api/Eras.hs
@@ -58,7 +58,6 @@ import           Data.Type.Equality (TestEquality (..), (:~:) (Refl))
 import           Ouroboros.Consensus.Shelley.Eras as Consensus (StandardAllegra, StandardAlonzo,
                    StandardBabbage, StandardMary, StandardShelley)
 
-
 -- | A type used as a tag to distinguish the Byron era.
 data ByronEra
 
--- a/cardano-api/src/Cardano/Api/IPC/Version.hs
+++ b/cardano-api/src/Cardano/Api/IPC/Version.hs
@@ -1,0 +1,37 @@
+module Cardano.Api.IPC.Version
+  ( NodeToClientVersionOf (..)
+  , MinNodeToClientVersion
+
+  -- *** Error types
+  , UnsupportedNtcVersionError(..)
+  ) where
+
+import           Data.Eq (Eq)
+import           Text.Show (Show)
+
+import           Ouroboros.Network.NodeToClient.Version (NodeToClientVersion (..))
+
+-- | The query 'a' is a versioned query, which means it requires the Node to support a minimum
+-- Node-to-Client version.
+--
+-- Background: The node to client protocol is such that it will disconnect on any
+-- unrecognised queries.  This means that for a Node-to-Client connection, if a query is sent
+-- that was introduced in a Node-to-Client version that is newer than the Node-To-Client version
+-- of the connection, the node will disconnect the client.  The client will not get any
+-- information about why the disconnect happened.  This is a bad user experience for tools
+-- such as the CLI.
+--
+-- To improve the user experience the API needs to prevent the sending of queries that are
+-- newer than the Node-To-Client version of the connection by checking the version of the
+-- query before sending it.  This affords the ability to return 'UnsupportedNtcVersionError',
+-- informing the caller of a Node-To-Client versioning issue.
+--
+-- For maintaining typeclass instances, see the 'NodeToClientVersion' type documentation for
+-- a list of versions and the queries that were introduced for those versions.
+class NodeToClientVersionOf a where
+  nodeToClientVersionOf :: a -> NodeToClientVersion
+
+type MinNodeToClientVersion = NodeToClientVersion
+
+data UnsupportedNtcVersionError = UnsupportedNtcVersionError !MinNodeToClientVersion !NodeToClientVersion
+  deriving (Eq, Show)
--- a/cardano-api/src/Cardano/Api/Query.hs
+++ b/cardano-api/src/Cardano/Api/Query.hs
@@ -110,6 +110,7 @@ import qualified Ouroboros.Consensus.Ledger.Query as Consensus
 import qualified Ouroboros.Consensus.Protocol.Abstract as Consensus
 import qualified Ouroboros.Consensus.Shelley.Ledger as Consensus
 import           Ouroboros.Network.Block (Serialised (..))
+import           Ouroboros.Network.NodeToClient.Version (NodeToClientVersion (..))
 
 import           Cardano.Binary
 import           Cardano.Slotting.Slot (WithOrigin (..))
@@ -129,6 +130,7 @@ import           Cardano.Api.Certificate
 import           Cardano.Api.EraCast
 import           Cardano.Api.Eras
 import           Cardano.Api.GenesisParameters
+import           Cardano.Api.IPC.Version
 import           Cardano.Api.Keys.Shelley
 import           Cardano.Api.Modes
 import           Cardano.Api.NetworkId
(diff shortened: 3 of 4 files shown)


cardano-api/src/Cardano/Api/Eras.hs
```diff
@@ -215,6 +218,9 @@ data AnyCardanoEra where
                    => CardanoEra era    -- and explicit value.
                    -> AnyCardanoEra
 
+instance NtcVersionOf AnyCardanoEra where
+  ntcVersionOf (AnyCardanoEra _) = NodeToClientV_9
```
Jimbo4350, line 222: Why are we defaulting to `NodeToClientV_9` for any era? Shouldn't we be explicit with each era and assign the version as follows:
```
--
data NodeToClientVersion
    = NodeToClientV_9
    -- ^ enabled @CardanoNodeToClientVersion7@, i.e., Alonzo
    | NodeToClientV_10
    -- ^ added 'GetChainBlockNo' and 'GetChainPoint' queries
    | NodeToClientV_11
    -- ^ added 'GetRewardInfoPools` Block query
    | NodeToClientV_12
    -- ^ added 'LocalTxMonitor' mini-protocol
    | NodeToClientV_13
    -- ^ enabled @CardanoNodeToClientVersion9@, i.e., Babbage
    | NodeToClientV_14
    -- ^ added @GetPoolDistr, @GetPoolState, @GetSnapshots
 ```

Comment (Jimbo4350) [reply cardano-api/src/Cardano/Api/Eras.hs:222]: I.e `NodeToClientV_12` for everything up to Alonzo and `NodeToClientV_14` for Babbage? The downside is we have to manually maintain this pairing.

File: cardano-api/src/Cardano/Api/IPC/NtcVersionOf.hs [hunk 2]
```diff
@@ -0,0 +1,21 @@
...
+  , UnsupportedNtcVersionError(..)
+  ) where
+
+import           Data.Eq (Eq)
+import           Text.Show (Show)
+import           Ouroboros.Network.NodeToClient.Version (NodeToClientVersion (..))
+
+-- | The query 'a' is a versioned query, which means it requires the Node to support a minimum
+-- Node-to-Client version.
+class NtcVersionOf a where
```
Jimbo4350, line 15: I think we should expand the name to `NodeToClientVersionOf`

cardano-api/src/Cardano/Api/Query.hs
```diff
@@ -130,6 +131,7 @@ import           Cardano.Api.EraCast
 import           Cardano.Api.Eras
 import           Cardano.Api.GenesisParameters
 import           Cardano.Api.Keys.Shelley
+import           Cardano.Api.IPC.NtcVersionOf (NtcVersionOf (..))
```
Jimbo4350, line 134: Let's keep with the style of the other api related imports.

cardano-api/src/Cardano/Api/IPC/NtcVersionOf.hs
```diff
@@ -0,0 +1,21 @@
...
+
+  -- *** Error types
+  , UnsupportedNtcVersionError(..)
+  ) where
+
+import           Data.Eq (Eq)
+import           Text.Show (Show)
+import           Ouroboros.Network.NodeToClient.Version (NodeToClientVersion (..))
+
+-- | The query 'a' is a versioned query, which means it requires the Node to support a minimum
```
Jimbo4350, line 13: Let's expand this comment to include more about the NodeToClient versioning i.e why do we have it? We can probably also point to documentation in the networking code that explains how the handshaking etc works.

Jimbo4350 (commented):
Nice, a few comments

cardano-api/src/Cardano/Api/Eras.hs
newhoggy: > Why are we defaulting to NodeToClientV_9 for any era?

`NodeToClientV_9` is the oldest era that is still being tracked.  Any queries that exist but are not in the above list above are presume to be from an era older than `NodeToClientV_9`, but we set it to `NodeToClientV_9` because it is long enough ago that we don't care about the precise version anymore.

> The downside is we have to manually maintain this pairing.

Yeah, I believe some form of this might belong in consensus.  It's implemented here for the moment because getting changes in consensus can take a long time.

cardano-api/src/Cardano/Api/IPC/NtcVersionOf.hs
newhoggy: Renamed the type class and moved the module to `Cardano.Api.IPC.Version`

cardano-api/src/Cardano/Api/IPC/NtcVersionOf.hs
newhoggy: More documentation added.

Jimbo4350 (commented):
https://github.com/input-output-hk/cardano-node/pull/4787#discussion_r1072139075

cardano-api/src/Cardano/Api/Eras.hs
newhoggy: That's an interesting point.  Would it be sufficient for `nodeToClientVersionOf` to take an era argument?

newhoggy: Note that we won't know the node's era unless we query for it, so it's not free.

cardano-api/src/Cardano/Api/Eras.hs
```diff
@@ -58,6 +59,7 @@ import           Data.Type.Equality (TestEquality (..), (:~:) (Refl))
 import           Ouroboros.Consensus.Shelley.Eras as Consensus (StandardAllegra, StandardAlonzo,
                    StandardBabbage, StandardMary, StandardShelley)
 
+import           Ouroboros.Network.NodeToClient.Version (NodeToClientVersion (..))
```
Jimbo4350, line 62: No need for the newline above

cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -45,14 +46,14 @@ import           Cardano.Api.Modes
 -- In order to make pipelining still possible we can explore the use of Selective Functors
 -- which would allow us to straddle both worlds.
 newtype LocalStateQueryExpr block point query r m a = LocalStateQueryExpr
-  { runLocalStateQueryExpr :: ContT (Net.Query.ClientStAcquired block point query m r) m a
-  } deriving (Functor, Applicative, Monad, MonadIO)
+  { runLocalStateQueryExpr :: ReaderT NodeToClientVersion (ContT (Net.Query.ClientStAcquired block point query m r) m) a
```
Jimbo4350, line 49: Let's only introduce the `NodeToClientVersionOf` typeclass in this PR

Jimbo4350 (changes requested):

cardano-api/src/Cardano/Api/IPC/Version.hs
```diff
@@ -0,0 +1,36 @@
+module Cardano.Api.IPC.Version
+  ( NodeToClientVersionOf (..)
+  , MinNodeToClientVersion
+
+  -- *** Error types
+  , UnsupportedNtcVersionError(..)
+  ) where
+
+import           Data.Eq (Eq)
+import           Ouroboros.Network.NodeToClient.Version (NodeToClientVersion (..))
```
Jimbo4350, line 10: This doesn't follow the usual import style of `cardano-api`

Jimbo4350 (changes requested):
Still a couple of import changes needed

Jimbo4350 (approved):

Jimbo4350 (changes requested):
Sorry, `instance NodeToClientVersionOf AnyCardanoEra where` need to be updated to reflect the latest versions for the relevant era

newhoggy:
I think there is not a strong relationship between the node to client version and the era.  For example a query could be added during an era, but depending on the circumstances, the added query might work for an earlier era.

cardano-api/src/Cardano/Api/Eras.hs
```diff
@@ -58,7 +58,6 @@ import           Data.Type.Equality (TestEquality (..), (:~:) (Refl))
 import           Ouroboros.Consensus.Shelley.Eras as Consensus (StandardAllegra, StandardAlonzo,
                    StandardBabbage, StandardMary, StandardShelley)
```
Jimbo4350, line 60: Random diff

Jimbo4350 (approved):
LGTM

Jimbo4350 (approved):

PR #4806, opened 2023-01-18
Consolidate two fields of CardanoTestnetOptions

Jimbo4350:
- Consolidate `cardanoBftNodeOptions` and `cardanoNumPoolNodes` records of `CardanoTestnetOptions`
- Implement `testnetMinimumConfigurationRequirements`

### DIFF
--- a/cardano-testnet/src/Parsers/Cardano.hs
+++ b/cardano-testnet/src/Parsers/Cardano.hs
@@ -7,15 +7,16 @@ module Parsers.Cardano
   ) where
 
 import           Prelude
+
 import qualified Data.List as L
 import           Options.Applicative
 import qualified Options.Applicative as OA
 import           Text.Read
 
-import           Testnet.Util.Runtime (readNodeLoggingFormat)
 import           Testnet
 import           Testnet.Cardano
 import           Testnet.Run (runTestnet)
+import           Testnet.Util.Runtime (readNodeLoggingFormat)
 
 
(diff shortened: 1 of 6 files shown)


cardano-testnet/src/Parsers/Cardano.hs
```diff
@@ -25,20 +26,23 @@ data CardanoOptions = CardanoOptions
...
+              <>  OA.value (cardanoNodes defaultTestnetOptions)
+              )
+        <*> OA.option
+              ((`L.replicate` SpoTestnetNodeOptions) <$> auto)
+              (   OA.long "num-pool-nodes"
+              <>  OA.help "Number of pool nodes"
+              <>  OA.metavar "COUNT"
+              <>  OA.showDefault
+              <>  OA.value (cardanoNodes defaultTestnetOptions)
+              )
```
newhoggy, line 45: Perhaps it's better to introduce a `mkCardanoTestnetOptions` function that takes the arguments exactly as we need them so we don't have to do nested parsing like this.

newhoggy: The description still says `num-pool-nodes`, etc.

newhoggy: Will the list argument to `BftTestnetNodeOptions` always be empty?

newhoggy (approved):

PR #4809, opened 2023-01-19
Add ReaderT of NodeToClientVersion to LocalStateQueryExpr

newhoggy:
The `NodeToClientVersion` for a connection is established as soon as the connection is created an never changes.  This makes `ReaderT` a good fit for holding this value so that it can be retrieved by the monadic query expressions.

### DIFF
--- a/cardano-api/src/Cardano/Api/IPC/Monad.hs
+++ b/cardano-api/src/Cardano/Api/IPC/Monad.hs
@@ -13,6 +13,7 @@ import           Control.Applicative
 import           Control.Concurrent.STM
 import           Control.Monad
 import           Control.Monad.IO.Class
+import           Control.Monad.Reader
 import           Control.Monad.Trans.Cont
 import           Data.Bifunctor (first)
 import           Data.Either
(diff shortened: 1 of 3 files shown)


cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -45,14 +46,14 @@ import           Cardano.Api.Modes
...
-  { runLocalStateQueryExpr :: ContT (Net.Query.ClientStAcquired block point query m r) m a
-  } deriving (Functor, Applicative, Monad, MonadIO)
+  { runLocalStateQueryExpr :: ReaderT NodeToClientVersion (ContT (Net.Query.ClientStAcquired block point query m r) m) a
+  } deriving (Functor, Applicative, Monad, MonadReader NodeToClientVersion, MonadIO)
 
 -- | Execute a local state query expression.
 executeLocalStateQueryExpr
   :: LocalNodeConnectInfo mode
   -> Maybe ChainPoint
-  -> (NodeToClientVersion -> LocalStateQueryExpr (BlockInMode mode) ChainPoint (QueryInMode mode) () IO a)
```
newhoggy, line 55: The use of `ReaderT` allows this argument to simplify.  This means that a query can be run like this:

```
executeLocalStateQueryExpr ... $ do
    ...
    expr
    ...
```

instead of like this:

```
executeLocalStateQueryExpr ... $ \ntcVersion -> do
    ...
    expr
    ...
```

We are transitioning from the later code because we want the `ntcVersion` to be checked automatically rather than manually.  The `ntcVersion` argument was passed in via a lambda before because the monadic query expression previously check the version manually which is error prone.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -284,7 +284,8 @@ runQueryTip (AnyConsensusModeParams cModeParams) network mOutFile = do
     CardanoMode -> do
       let localNodeConnInfo = LocalNodeConnectInfo cModeParams network sockPath
 
-      eLocalState <- liftIO $ executeLocalStateQueryExpr localNodeConnInfo Nothing $ \ntcVersion -> do
+      eLocalState <- liftIO $ executeLocalStateQueryExpr localNodeConnInfo Nothing $ do
+        ntcVersion <- ask
```
newhoggy, line 288: This monadic expression checks the node to client version manually.  It will switch to automatic checking in a future PR to be less error prone, but for now, `ntcVersion` is still accessible via `ask`.

PR #4810, opened 2023-01-19
Add Cardano RPC command

mesudip:
I have added a JSON-RPC client feature to `cardano-cli` that allows making RPC calls to any backend. The client uses the `network.http.simple` library for sending HTTP requests and `aeson` for parsing JSON responses.

The RPC methods are not yet decided, so the client currently supports any method specified in the request. The standard for the methods can be added later, once the specification for the RPC is created. Additionally, validation of methods can also be added later to ensure they conform to the specified standard.

```
$ cardano-cli rpc --help
Usage: cardano-cli rpc [--url ARG]
            [-u|--user ARG]
            [-H|--header ARG]
            method
            [argument]

  Call cardano rpc over network

Available options:
  --url ARG                Rpc Address (default: "http://localhost:3345")
  -u,--user ARG            Basic auth "User:Password"
  -H,--header ARG          Rpc extra HTTP headers "HeaderName:HeaderValue"
  method                   Json RPC method
  argument                 RPC method arguments
```

### DIFF
--- a/cardano-cli/cardano-cli.cabal
+++ b/cardano-cli/cardano-cli.cabal
@@ -90,7 +90,8 @@ library
                         Cardano.CLI.Shelley.Run.TextView
                         Cardano.CLI.Shelley.Run.Transaction
                         Cardano.CLI.Shelley.Run.Read
-                        Cardano.CLI.Shelley.Run.Validate
+                        Cardano.CLI.Shelley.Run.Rpc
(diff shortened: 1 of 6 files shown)


Jimbo4350 (changes requested):
Thanks for the PR. I need to discuss with @newhoggy  and @MarcFontaine if cardano-cli is the best place for RPC calls.

PR #4811, opened 2023-01-19
Optimized trace-dispatcher and tracing system

jutaro:
Make namespaces more opwerful, enhance the interface and have all metainformation accessible at runtime for optimizations.


olgahryniuk (approved):

deepfire (approved):
Thank you @jutaro !

deepfire:
So, the problem is that analysis now produces empty results..

deepfire:
`make ci-bench-autonix` passed and analysis data is good, merging by hand.

PR #4816, opened 2023-01-20
Update cardano-cli banner

abailly-iohk:
The previous blurb was pretty outdated and not reflecting what cardano-cli really is about, this is an attempt at being more generic.

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Parsers.hs
+++ b/cardano-cli/src/Cardano/CLI/Parsers.hs
@@ -26,9 +26,9 @@ opts =
   Opt.info (parseClientCommand <**> Opt.helper)
     ( Opt.fullDesc
       <> Opt.header
-      "cardano-cli - utility to support a variety of key\
-      \ operations (genesis generation, migration,\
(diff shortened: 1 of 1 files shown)


Jimbo4350 (approved):

cardano-cli/src/Cardano/CLI/Parsers.hs
```diff
@@ -26,9 +26,9 @@ opts =
   Opt.info (parseClientCommand <**> Opt.helper)
     ( Opt.fullDesc
       <> Opt.header
-      "cardano-cli - utility to support a variety of key\
-      \ operations (genesis generation, migration,\
-      \ pretty-printing..) for different system generations."
```
newhoggy, line 31: While you're here, can you remove multiline literals from these lines and use semigroup concatenation of strings instead?

newhoggy: See https://github.com/input-output-hk/cardano-node/issues/4585

PR #4823, opened 2023-01-21
Straight line code for tx commands

newhoggy:


cardano-cli/src/Cardano/CLI/Shelley/Run/Transaction.hs
```diff
@@ -358,9 +358,8 @@ runTxBuildCmd
                             , localNodeSocketPath = sockPath
                             }
 
-  AnyCardanoEra nodeEra
-    <- firstExceptT (ShelleyTxCmdQueryConvenienceError . AcqFailure)
-         . newExceptT $ determineEra cModeParams localNodeConnInfo
+  AnyCardanoEra nodeEra <- lift (determineEra cModeParams localNodeConnInfo)
+    & onLeft (throwE . ShelleyTxCmdQueryConvenienceError . AcqFailure)
```
Jimbo4350: not a major deal but I'd like to stick with `left` in future

Jimbo4350 (approved):
LGTM!

cardano-cli/src/Cardano/CLI/Shelley/Run/Transaction.hs
newhoggy: I removed `throwE` from the commit history of the PR.

PR #4831, opened 2023-01-23
Refactor `genesis create` command usage in  cardano testnet

Jimbo4350:
Implement the following three modules:
1. Testnet.Commands.Genesis - Houses the refactoring of cli commands concerning genesis generation

2. Testnet.Commands.Governance - Houses the refactoring of cli commands concerning governance 

3. Testnet.Options - Module that will house all the different `TestnetOptions` types. The aim is to only have one `TestnetOptions` type in the future.

### DIFF
--- a/cardano-testnet/src/Testnet/Commands/Governance.hs
+++ b/cardano-testnet/src/Testnet/Commands/Governance.hs
@@ -1,0 +1,33 @@
+
+
+module Testnet.Commands.Governance
+  ( createByronUpdateProposal
+  ) where
+
+import           Prelude
+
+import           Control.Monad.Catch
+import           Control.Monad.IO.Class
+
+import           Testnet.Util.Process (execCli_)
+
+import           Hedgehog.Internal.Property
+
+
+createByronUpdateProposal
+  :: (MonadTest m, MonadCatch m, MonadIO m)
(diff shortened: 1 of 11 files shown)


cardano-testnet/src/Testnet/Commands/Governance.hs
```diff
@@ -0,0 +1,47 @@
...
+
+import           Control.Monad.Catch
+import           Control.Monad.IO.Class
+
+import           Testnet.Util.Process (execCli_)
+
+import           Hedgehog.Internal.Property
+
+
+createByronUpdateProposal
```
newhoggy, line 18: For each of these functions, you will need `HasCallStack` and `GHC.withFrozenCallStack`.

This will ensure than when that command fails, the failure will be marked in the right location in the test (rather than in this file, which is not as useful).

See https://github.com/input-output-hk/cardano-node/blob/newhoggy/update-ouroboros-network/cardano-testnet/src/Testnet/Util/Process.hs#L63-L68 for examples

Jimbo4350: Fixed!

newhoggy (approved):

PR #4833, opened 2023-01-24
Implement ADR-2: Restructure modules for generators

newhoggy:
Rename:

* `Gen.Cardano.*` -> `Test.Gen.Cardano.*`
* `Hedgehog.*` -> `Test.Hedgehog.*`

See https://github.com/input-output-hk/cardano-node/wiki/ADR-2-Module-structure-for-generators


Jimbo4350 (approved):
Good stuff @newhoggy . Thank you for championing the ADR process.

PR #4845, opened 2023-01-25
Use Haskell variables for passing values.

MarcFontaine:
This is a combination of several commits. (They were squashed into one big commit during an elaborate re-base).
Most changes are related to the way filenames work in `cardano-testnet`.

Currently the `cardano-testnet` uses filenames (Strings) to reference objects like keys, certs etc.
Two locations in the code refer to the same object by "magically" using the same filename.
This means that the data flow is implicit and hard to follow.
Old scheme:
```
do
cli_call_1 "filenameX" "filenameY"
cli_call_2 "filenameX" "filenameZ"
cli_call_3 "filenameA" "filenameB"
```
This PR changes the design as follows:
* File names are stored in an opaque `File` datatype (which wraps the file path).
* A wrapper is used for CLI calls. When a CLI call creates a file the wrapper returns the corresponding `File` resource.
* While CLI calls have type `IO ()` the wrapper has type `IO (File,..,File).
* CLI calls don't construct magic Filenames, but instead use the `File` resource that was created earlier.
The new scheme is
```
do
(file1, file2) <- cli_call1 arg1 arg2     -- cli_call1 creates file resources file1 and file2
(fileX, fileY) <- cli_call2 file1 file2    -- cli_call2 uses file resourcs file1 and file2
```
Full transformation to new scheme for filenames is still WIP.

### DIFF
--- a/cardano-testnet/src/Testnet/Babbage.hs
+++ b/cardano-testnet/src/Testnet/Babbage.hs
@@ -32,10 +32,11 @@ import           Testnet.Commands.Genesis
 import qualified Testnet.Conf as H
 import           Testnet.Options
 import qualified Testnet.Util.Assert as H
+import           Testnet.Util.Cli
 import           Testnet.Util.Process (execCli_)
 import           Testnet.Util.Runtime (Delegator (..), NodeLoggingFormat (..), PaymentKeyPair (..),
-                   PoolNode (PoolNode), PoolNodeKeys (..), StakingKeyPair (..), TestnetRuntime (..),
-                   startNode)
+                   PoolNode (PoolNode), PoolNodeKeys (..), StakingKeyPair (..),
+                   TestnetRuntime (..), startNode, getLogDir, TmpPath(..) )
 
 
 {- HLINT ignore "Redundant flip" -}
@@ -124,10 +125,12 @@ babbageTestnet testnetOptions H.Conf {..} = do
     PoolNodeKeys
       { poolNodeKeysColdVkey = tempAbsPath </> "pools" </> "cold" <> show n <> ".vkey"
       , poolNodeKeysColdSkey = tempAbsPath </> "pools" </> "cold" <> show n <> ".skey"
+      , poolNodeKeysVrf = ( File $ tempAbsPath </> "node-spo" <> show n </> "vrf.vkey"
+                          , File $ tempAbsPath </> "node-spo" <> show n </> "vrf.skey"
+                          )
       , poolNodeKeysVrfVkey = tempAbsPath </> "node-spo" <> show n </> "vrf.vkey"
       , poolNodeKeysVrfSkey = tempAbsPath </> "node-spo" <> show n </> "vrf.skey"
-      , poolNodeKeysStakingVkey = tempAbsPath </> "pools" </> "staking-reward" <> show n <> ".vkey"
-      , poolNodeKeysStakingSkey = tempAbsPath </> "pools" </> "staking-reward" <> show n <> ".skey"
+      , poolNodeKeysOperator = undefined
       }
 
   wallets <- forM [1..3] $ \idx -> do
@@ -271,7 +274,7 @@ babbageTestnet testnetOptions H.Conf {..} = do
     ]
 
   poolNodes <- forM (L.zip spoNodes poolKeys) $ \(node,key) -> do
-    runtime <- startNode tempBaseAbsPath tempAbsPath logDir socketDir node
+    runtime <- startNode (TmpPath tempAbsPath) node
         [ "run"
         , "--config", tempAbsPath </> "configuration.yaml"
         , "--topology", tempAbsPath </> node </> "topology.json"
@@ -288,7 +291,7 @@ babbageTestnet testnetOptions H.Conf {..} = do
   deadline <- H.noteShow $ DTC.addUTCTime 90 now
 
   forM_ spoNodes $ \node -> do
-    nodeStdoutFile <- H.noteTempFile logDir $ node <> ".stdout.log"
+    nodeStdoutFile <- H.noteTempFile (getLogDir $ TmpPath tempAbsPath) $ node <> ".stdout.log"
     H.assertChainExtended deadline (babbageNodeLoggingFormat testnetOptions) nodeStdoutFile
 
   H.noteShowIO_ DTC.getCurrentTime
--- a/cardano-testnet/src/Testnet/Cardano.hs
+++ b/cardano-testnet/src/Testnet/Cardano.hs
@@ -242,10 +242,12 @@ cardanoTestnet testnetOptions configuration@H.Conf {tempAbsPath, testnetMagic} =
   byronGenesisHash <- getByronGenesisHash $ tmpDir </> "byron/genesis.json"
   shelleyGenesisHash <- getShelleyGenesisHash $ tmpDir </> "shelley/genesis.json"
   alonzoGenesisHash <- getShelleyGenesisHash $ tmpDir </> "shelley/genesis.alonzo.json"
+  conwayGenesisHash <- getShelleyGenesisHash $ tmpDir </> "shelley/genesis.conway.json"
   H.rewriteYamlFile (tmpDir </> "configuration.yaml") . J.rewriteObject
     $ HM.insert "ByronGenesisHash" byronGenesisHash
     . HM.insert "ShelleyGenesisHash" shelleyGenesisHash
     . HM.insert "AlonzoGenesisHash" alonzoGenesisHash
+    . HM.insert "ConwayGenesisHash" conwayGenesisHash
 
   poolRuntimes <- cardanoTestnetLaunchPoolNodes (TmpPath tmpDir) poolNodeNames
 
@@ -499,12 +501,12 @@ cardanoTestnetShelleyGenesis base (TmpPath tmpDir) testnetMagic testnetOptions s
 
   -- TODO: This is fragile, we should be passing in all necessary
   -- configuration files.
-  let sourceAlonzoGenesisSpecFile = base </> "cardano-cli/test/data/golden/alonzo/genesis.alonzo.spec.json" -- only use of base in this module
+  let sourceAlonzoGenesisSpecFile = base </> "cardano-cli/test/data/golden/alonzo/genesis.alonzo.spec.json" -- use of base in this module
   alonzoSpecFile <- H.noteTempFile tmpDir "shelley/genesis.alonzo.spec.json"
   liftIO $ IO.copyFile sourceAlonzoGenesisSpecFile alonzoSpecFile
 
-  let sourceConwayGenesisSpecFile = base </> "cardano-cli/test/data/golden/conway/genesis.conway.spec.json"
-  conwaySpecFile <- H.noteTempFile tempAbsPath "shelley/genesis.conway.spec.json"
+  let sourceConwayGenesisSpecFile = base </> "cardano-cli/test/data/golden/conway/genesis.conway.spec.json" -- use of base in this module
+  conwaySpecFile <- H.noteTempFile tmpDir "shelley/genesis.conway.spec.json"
   liftIO $ IO.copyFile sourceConwayGenesisSpecFile conwaySpecFile
 
   execCli_
@@ -653,7 +655,7 @@ cardanoTestnetPaymentKeys tmpDir testnetMagic poolNodesN = do
 
   H.createDirectoryIfMissing $ tmpDir </> "addresses"
 
-  wallets <- forM addrs $ \addr -> do
+  forM addrs $ \addr -> do
     let paymentSKey = tmpDir </> "addresses/" <> addr <> ".skey"
     let paymentVKey = tmpDir </> "addresses/" <> addr <> ".vkey"
 
@@ -694,7 +696,6 @@ cardanoTestnetPaymentKeys tmpDir testnetMagic poolNodesN = do
       { paymentSKey
       , paymentVKey
       }
-  return wallets
 
 cardanoTestnetPoolCerts
   :: TmpDir
@@ -820,17 +821,6 @@ cardanoTestnetPoolCerts tmpDir testnetMagic poolNodesN = do
   -- Generated a signed 'do it all' transaction:
   H.assertIO . IO.doesFileExist $ tmpDir </> "tx1.tx"
 
-  -- Add Byron, Shelley and Alonzo genesis hashes to node configuration
-  byronGenesisHash <- getByronGenesisHash $ tempAbsPath </> "byron/genesis.json"
-  shelleyGenesisHash <- getShelleyGenesisHash $ tempAbsPath </> "shelley/genesis.json"
-  alonzoGenesisHash <- getShelleyGenesisHash $ tempAbsPath </> "shelley/genesis.alonzo.json"
-  conwayGenesisHash <- getShelleyGenesisHash $ tempAbsPath </> "shelley/genesis.conway.json"
-  H.rewriteYamlFile (tempAbsPath </> "configuration.yaml") . J.rewriteObject
-    $ HM.insert "ByronGenesisHash" byronGenesisHash
-    . HM.insert "ShelleyGenesisHash" shelleyGenesisHash
-    . HM.insert "AlonzoGenesisHash" alonzoGenesisHash
-    . HM.insert "ConwayGenesisHash" conwayGenesisHash
-
 cardanoTestnetLaunchBftNodes
   :: TmpPath
   -> CardanoTestnetOptions
@@ -890,9 +880,6 @@ cardanoTestnetWaitStartup H.Conf {..} loggingFormat allNodeNames = do
     H.assertChainExtended deadline loggingFormat nodeStdoutFile
 
   H.noteShowIO_ DTC.getCurrentTime
-
-
--- * Generate hashes for genesis.json files
 
 getByronGenesisHash :: (H.MonadTest m, MonadIO m) => FilePath -> m J.Value
 getByronGenesisHash path = do
--- a/cardano-testnet/src/Testnet/Cardano.hs
+++ b/cardano-testnet/src/Testnet/Cardano.hs
@@ -277,8 +277,8 @@ cardanoTestnetByronGenesis
     , [File ByronDelegationKey]
     )
 cardanoTestnetByronGenesis tp testnetMagic testnetOptions configurationTemplate startTime = do
-  let (TmpPath tmpDir) = tp
-  let numBftNodes = cardanoNumBftNodes $ cardanoNodes testnetOptions
+  let TmpPath tmpDir = tp
+      numBftNodes = cardanoNumBftNodes $ cardanoNodes testnetOptions
       bftNodesN = [1 .. numBftNodes]
       poolNodesN = [1 .. cardanoNumPoolNodes $ cardanoNodes testnetOptions]
       bftNodeNames = ("node-bft" <>) . show @Int <$> bftNodesN
@@ -364,6 +364,7 @@ cardanoTestnetByronGenesis tp testnetMagic testnetOptions configurationTemplate 
 
 
     H.writeFile (tmpDir </> node </> "port") (show port)
+
   H.lbsWriteFile (tmpDir </> "byron.genesis.spec.json")
     . J.encode $ defaultByronGenesisJsonValue
 
@@ -396,16 +397,16 @@ cardanoTestnetByronGenesis tp testnetMagic testnetOptions configurationTemplate 
   let
     forallBftNodesMkFile x = forM [1..numBftNodes] (fakeItH . x)
 
-  (genesisKeys :: [File ByronKey])<- forallBftNodesMkFile $ \n -> "byron/genesis-keys.00" <> show @Int (n - 1) <> ".json"
-  (delegationCerts :: [File ByronDelegationCert]) <- forallBftNodesMkFile $ \n -> "byron/delegate-keys.00" <> show @Int (n - 1) <> ".key"
-  (delegationKeys :: [File ByronDelegationKey]) <- forallBftNodesMkFile $ \n -> "byron/delegation-cert.00" <> show @Int (n - 1) <> ".json"
+  genesisKeys <- forallBftNodesMkFile $ \n -> "byron/genesis-keys.00" <> show @Int (n - 1) <> ".json"
+  delegationCerts <- forallBftNodesMkFile $ \n -> "byron/delegate-keys.00" <> show @Int (n - 1) <> ".key"
+  delegationKeys <- forallBftNodesMkFile $ \n -> "byron/delegation-cert.00" <> show @Int (n - 1) <> ".json"
 
   return (genesisKeys, delegationCerts, delegationKeys)
 
 
 {-
 cardanoTestnetByronGenesisExpenditure creates an transaction `tx0.tx'
-which however is never used anywhere ??
+TODO: a test that transmits tx0.tx and checks that it gets accepted.
 -}
 cardanoTestnetByronGenesisExpenditure
   :: CardanoTestnetOptions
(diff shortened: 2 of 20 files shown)


Jimbo4350 (changes requested):
There are a couple issues with this PR that need to be addressed before it is reviewed:
- There are several left over comments that should not be there
- There is a single commit for a relatively large diff

You need to remove comments that are no longer required and break this PR up in to several commits with clear commit messages describing the changes in the given commit. This makes it easier for a reviewer to understand the particular changes you have made.

cardano-testnet/src/Testnet/Cardano.hs
```diff
@@ -1,7 +1,11 @@
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
+{-# OPTIONS_GHC -Wno-unused-do-bind #-}
+{-# OPTIONS_GHC -Wno-unused-matches #-}
+{-# OPTIONS_GHC -Wno-unused-top-binds #-}
```
newhoggy, line 8: Why do we need `no-unused-matches` and `no-unused-top-binds`?

cardano-testnet/src/Testnet/Cardano.hs
```diff
@@ -185,30 +191,104 @@ mkTopologyConfig numNodes allPorts port True = J.encode topologyP2P
...
+       testnetOptions
+       (configurationTemplate configuration)
+       startTime
+{-
+  -- dead code: not checked in any test
+  cardanoTestnetByronGenesisExpenditure testnetOptions configuration byronGenesisKeys
+  cardanoTestnetByronGovernance
+    (L.length $ cardanoBftNodeOptions testnetOptions)
+    testnetMagic
+    tmpDir
```
newhoggy, line 221: Can you elaborate why you mean be "dead code".  There are a number of such references.

cardano-testnet/src/Testnet/Cardano.hs
```diff
@@ -185,30 +191,104 @@ mkTopologyConfig numNodes allPorts port True = J.encode topologyP2P
...
+  -> CardanoTestnetOptions
+  -> FilePath
+  -> DTC.UTCTime
+  -> H.Integration
+    ( [File ByronKey]
+    , [File ByronDelegationCert]
+    , [File ByronDelegationKey]
+    )
+cardanoTestnetByronGenesis tp testnetMagic testnetOptions configurationTemplate startTime = do
+  let (TmpPath tmpDir) = tp
```
newhoggy, line 278: Redundant parentheses.

cardano-testnet/src/Testnet/Cardano.hs
```diff
@@ -268,20 +348,20 @@ cardanoTestnet testnetOptions H.Conf {..} = do
...
     let port = fromJust $ M.lookup node nodeToPort
-    H.lbsWriteFile (tempAbsPath </> node </> "topology.json") $
+    H.lbsWriteFile (tmpDir </> node </> "topology.json") $
       mkTopologyConfig (numBftNodes + cardanoNumPoolNodes (cardanoNodes testnetOptions))
                        allPorts port (cardanoEnableP2P testnetOptions)
 
-    H.writeFile (tempAbsPath </> node </> "port") (show port)
 
-  H.lbsWriteFile (tempAbsPath </> "byron.genesis.spec.json")
+    H.writeFile (tmpDir </> node </> "port") (show port)
```
newhoggy, line 363: I think adding a newline here adds clarity.

cardano-testnet/test/Test/FoldBlocks.hs
```diff
@@ -35,13 +35,12 @@ instance Show FoldBlocksException where
...
 prop_foldBlocks = H.integrationRetryWorkspace 2 "foldblocks" $ \tempAbsBasePath' -> do
-
   -- Start testnet
   base <- HE.noteM $ liftIO . IO.canonicalizePath =<< HE.getProjectBase
-  configTemplate <- H.noteShow $ base </> "configuration/defaults/byron-mainnet/configuration.yaml"
+  configTemplate <- H.noteShow $ base </> "configuration/defaults/byron-mainnet/configuration.yaml" --- outside of the package directory !!
   conf <- HE.noteShowM $
     TN.mkConf (TN.ProjectBase base) (TN.YamlFilePath configTemplate)
-      (tempAbsBasePath' <> "/")
+      (tempAbsBasePath' <> "/") --  '<> "/")'  is needed otherwise test fails !
```
newhoggy, line 43: I know this is old code, but can you find out why?

cardano-testnet/test/Test/FoldBlocks.hs
```diff
@@ -35,13 +35,12 @@ instance Show FoldBlocksException where
 -- that main thread blocks on.
 prop_foldBlocks :: H.Property
 prop_foldBlocks = H.integrationRetryWorkspace 2 "foldblocks" $ \tempAbsBasePath' -> do
-
   -- Start testnet
   base <- HE.noteM $ liftIO . IO.canonicalizePath =<< HE.getProjectBase
-  configTemplate <- H.noteShow $ base </> "configuration/defaults/byron-mainnet/configuration.yaml"
+  configTemplate <- H.noteShow $ base </> "configuration/defaults/byron-mainnet/configuration.yaml" --- outside of the package directory !!
```
newhoggy, line 40: > outside of the package directory !!

Yeah, that's what `base` and `getProjectBase` means.  Is there a way to make this clearer without adding a comment?

cardano-testnet/src/Testnet/Util/Runtime.hs
```diff
@@ -117,22 +122,38 @@ readNodeLoggingFormat = \case
...
+  { unTmpPath :: FilePath
+  } deriving (Eq, Show)
+getTmpRelPath :: TmpPath -> FilePath
+getTmpRelPath (TmpPath fp) = FP.makeRelative (getTmpBaseAbsPath (TmpPath fp)) fp
+getSocketDir :: TmpPath -> FilePath
+getSocketDir fp = getTmpRelPath fp </> "socket"
+getTmpBaseAbsPath :: TmpPath -> FilePath
+getTmpBaseAbsPath (TmpPath fp) = FP.takeDirectory fp
+getLogDir :: TmpPath -> FilePath
+getLogDir (TmpPath fp) = fp </> "logs"
```
newhoggy, line 141: Looks a bit crowded.  Can you spread them out with newlines?

cardano-testnet/src/Testnet/Util/Runtime.hs
```diff
@@ -117,22 +122,38 @@ readNodeLoggingFormat = \case
 allNodes :: TestnetRuntime -> [NodeRuntime]
 allNodes tr = bftNodes tr <> fmap poolRuntime (poolNodes tr)
 
+makeSprocket
+  :: TmpPath
+  -> String
+  -> Sprocket
+makeSprocket tmpPath node
+  = Sprocket (getTmpBaseAbsPath tmpPath)  (getSocketDir tmpPath </> node)
```
newhoggy, line 130: Double space between `)  (` could be one line.

cardano-testnet/src/Testnet/Util/Runtime.hs
```diff
@@ -67,16 +72,16 @@ data NodeRuntime = NodeRuntime
 
 data PoolNode = PoolNode
   { poolRuntime :: NodeRuntime
-  , poolKeys :: PoolNodeKeys
+  , poolKeys :: PoolNodeKeys -- used in Alonzo LeadershipShedule
```
newhoggy, line 75: I'm curious about the comment.  We shouldn't have to care where this data is used.  For example new tests might be added that make use of this.

cardano-testnet/src/Testnet/Util/Runtime.hs
```diff
@@ -67,16 +72,16 @@ data NodeRuntime = NodeRuntime
...
   }
 
 data PoolNodeKeys = PoolNodeKeys
-  { poolNodeKeysColdVkey :: FilePath
+  { poolNodeKeysColdVkey :: FilePath -- == Operator keys -- used in LeadershipShedule
```
newhoggy, line 79: I'm partial to using something more type safe than `FilePath` for each type of file to make usage information easier to find but it doesn't have to be this PR.

cardano-testnet/src/Testnet/Cardano.hs
```diff
@@ -298,98 +378,134 @@ cardanoTestnet testnetOptions H.Conf {..} = do
...
   -- Symlink the BFT operator keys from the genesis delegates, for uniformity
   forM_ bftNodesN $ \n -> do
-    H.createFileLink (tempAbsPath </> "byron/delegate-keys.00" <> show @Int (n - 1) <> ".key") (tempAbsPath </> "node-bft" <> show @Int n </> "byron/delegate.key")
-    H.createFileLink (tempAbsPath </> "byron/delegation-cert.00" <> show @Int (n - 1) <> ".json") (tempAbsPath </> "node-bft" <> show @Int n </> "byron/delegate.cert")
+    H.createFileLink (tmpDir </> "byron/delegate-keys.00" <> show @Int (n - 1) <> ".key") (tmpDir </> "node-bft" <> show @Int n </> "byron/delegate.key")
+    H.createFileLink (tmpDir </> "byron/delegation-cert.00" <> show @Int (n - 1) <> ".json") (tmpDir </> "node-bft" <> show @Int n </> "byron/delegate.cert")
+  let
+    forallBftNodesMkFile x = forM [1..numBftNodes] (fakeItH . x)
+
+  (genesisKeys :: [File ByronKey])<- forallBftNodesMkFile $ \n -> "byron/genesis-keys.00" <> show @Int (n - 1) <> ".json"
```
newhoggy, line 396: Can the typing be done by type applications instead of type ascriptions?

cardano-testnet/src/Testnet/Cardano.hs
```diff
@@ -298,98 +378,134 @@ cardanoTestnet testnetOptions H.Conf {..} = do
...
+    , "--tx", tempAbsPath </> "tx0.tx"
+    , "--wallet-key", tempAbsPath </> "byron/delegate-keys.000.key"
+    , "--rich-addr-from", richAddrFrom
+    , "--txout", show @(String, Int) (txAddr, fundsPerByronAddress)
+    ]
 
-  -- Update Proposal and votes
+{-
+cardanoTestnetByronGovernance creates two update-proposals and two update-votes.
+However they are never transmitted to the chain ?!
```
newhoggy, line 447: These were translated directly from the `mkfiles.sh` script this was based on.  Those scripts were designed so that the update proposals can be submitted manually.

We should consider what we want to do with this in the Haskell code.

cardano-testnet/src/Testnet/Util/Cli.hs
```diff
@@ -0,0 +1,186 @@
...
+
+-- TODO: check no whitespace , slash etc allowed
+-- TODO: use Comment instead of Filepath
+newtype Comment = Comment { unComment :: String}
+  deriving (Show, Eq, Ord )
+
+instance IsString Comment where fromString a = Comment a
+
+instance Semigroup Comment where
+  (Comment a) <> (Comment b) = Comment $ a <> "-" <> b
```
newhoggy, line 61: Redundant parentheses.

cardano-testnet/src/Testnet/Util/Cli.hs
```diff
@@ -0,0 +1,186 @@
...
+import qualified Hedgehog.Extras.Test.File as H (writeFile)
+import           Hedgehog
+import           Hedgehog.Gen
+
+import           Testnet.Util.Process
+
+-- TODO: check no whitespace , slash etc allowed
+-- TODO: use Comment instead of Filepath
+newtype Comment = Comment { unComment :: String}
+  deriving (Show, Eq, Ord )
```
newhoggy, line 56: This will need to be MacOS and Windows friendly.  MacOS filenames are case insensitive.  Windows filenames are case-insensitive and many characters have special meaning.

Perhaps we can call it `Kebab` because it's using kebab-casing and we lock it down do only support lower case alpha-numerics.

cardano-testnet/src/Testnet/Util/Cli.hs
```diff
@@ -0,0 +1,186 @@
...
+data OperatorCounter
+data ByronKey
+data ByronAddress
+data ByronDelegationKey
+data ByronDelegationCert
+
+type TmpDir = FilePath
+type TestnetMagic = Int
+newtype File a = File {unFile :: FilePath}
+  deriving (Show, Eq)
```
newhoggy, line 138: Nice!  Can you add a newline above the `newtype`?

cardano-testnet/src/Testnet/Util/Cli.hs
```diff
@@ -0,0 +1,186 @@
...
+makeFilePath path comment suffix = do
+  almostUnique <- Comment <$> replicateM 4 hexit
+  return $ path </> unComment ( comment <> almostUnique) <.> suffix
+
+data KeyNames = KeyNames
+  { verificationKeyFile :: FilePath
+  , signingKeyFile :: FilePath
+  }
+
+type KeyGen a = H.Integration (File (a VKey), File (a SKey))
```
newhoggy, line 74: The `KeyGen` type alias seems less helpful.

cardano-testnet/src/Testnet/Util/Cli.hs
```diff
@@ -0,0 +1,186 @@
...
+-}
+
+fakeIt :: FilePath -> File a
+fakeIt = File
+
+getVKeyPath ::  (File (a VKey), File (a SKey)) -> FilePath
+getVKeyPath (File a, _ ) = a
+
+getSKeyPath ::  (File (a VKey), File (a SKey)) -> FilePath
+getSKeyPath (_, File a) = a
```
newhoggy, line 160: I think I rather just use `fst` and `snd` and `unFile`.

cardano-testnet/src/Testnet/Util/Cli.hs
```diff
@@ -0,0 +1,186 @@
...
+-- When all uses of fakeItH are removed the cleanup is done.
+fakeItH :: FilePath -> H.Integration (File a)
+fakeItH filePath = do
+  return $ File filePath
+  {-
+  exists <- liftIO $ fileExists filePath
+  if exists
+    then return $ File filePath
+    else error (filePath <> " does not exist.")
+-}
```
newhoggy, line 151: Commented out code.

newhoggy:
I agree with @Jimbo4350 regarding breaking up the code into multiple commits.

cardano-testnet/src/Testnet/Cardano.hs
MarcFontaine: The refactor uses Haskell variables for file resources.
The CLI wrapper functions return those resources.
However the refactor is still WIP some resources are defined but not yet used.
Once everything is don't the warnings should be turned on again.

cardano-testnet/src/Testnet/Cardano.hs
MarcFontaine: Some parts of the code code call `cardano-cli` for example to create a transaction.
However that transaction is never ever transmitted to the testnet or reused.
Also the code is not tested, all the tests pass when the code is disabled.
`cardanoTestnetByronGovernance` creates genesis key, delegation certs and delegation keys,
however this are are not used.

cardano-testnet/test/Test/FoldBlocks.hs
MarcFontaine: I only stumbled upon this because `(tempAbsBasePath' <> "/")` looks strange.
but as the comment says the tests fail without the trailing `<> "/"`.
I'm in the course of reordering the various file paths and this will probably be removed anyway.

cardano-testnet/test/Test/FoldBlocks.hs
MarcFontaine: That comment is a TODO for me.
I think it is better to move copies of those file to the package itself
or maybe there is yet an other way to do this.

cardano-testnet/src/Testnet/Util/Runtime.hs
MarcFontaine: ups yes some blank lines got messed up during the rebase

cardano-testnet/src/Testnet/Util/Runtime.hs
MarcFontaine: I'm about to change the code so I naturally need to know where something is defined and where it is used.
I think I also found stuff that is defined but not not used.
IMHO most of the time, it is not a good idea to add record fields for data that is not used.
Just because the data might be used in the future is not a good reason.

cardano-testnet/src/Testnet/Cardano.hs
MarcFontaine: Thanks for spotting, those types can be inferred and annotation are superfluous.

cardano-testnet/src/Testnet/Cardano.hs
```diff
@@ -298,98 +378,134 @@ cardanoTestnet testnetOptions H.Conf {..} = do
...
-      ] >>= H.writeFile (tempAbsPath </> "byron/genesis-address-00" <> show @Int (n - 1))
-
-  do
-    richAddrFrom <- S.firstLine <$> H.readFile (tempAbsPath </> "byron/genesis-address-000")
-    txAddr <- S.firstLine <$> H.readFile (tempAbsPath </> "byron/address-000")
+  byronAddress0:_ <- forM [1 .. numBftNodes] $ \n -> do
+    skey <- cliKeyGen tmpDir $ "byron/payment-keys.00" <> show @Int (n - 1) <> ".key"
+    cliSigningKeyAddress tmpDir testnetMagic skey $ "byron/address-00" <> show @Int (n - 1)
+
+  byronGensisAddress0:_ <- forM (zip [1 .. numBftNodes] genesisKeys) $ \(n,key) -> do
```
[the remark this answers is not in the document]
MarcFontaine: There is more to be done here..its dead code inside dead code.
`forM [1 .. numBftNodes]` keys are generated but only the first is used.

cardano-testnet/src/Testnet/Cardano.hs
MarcFontaine: Good information.
I should have a look at `mkfiles.sh`. Is it still available ? Is it still in use ?
I'll change that comment to a TODO:
There should be a test that submits the update proposals and checks that they are activated.
It was not clear to me, if that test had been removed or disabled, or if it had never existed.

cardano-testnet/src/Testnet/Util/Cli.hs
MarcFontaine: Interesting. I guess I had written the type signature `fst` copied it to a `snd` line
and then edited that line to become the function definition...

cardano-testnet/src/Testnet/Util/Cli.hs
MarcFontaine: Is there a simple function in the test frame-work that asserts that a file exists ?
Last time I looked at it there was only one which checks for a list of files.

MarcFontaine:
> I agree with @Jimbo4350 regarding breaking up the code into multiple commits.

I break up my code into multiple commits anyway.
However, here it was impossible to rebase it as such, so I squashed the PR into one commit.
Overall this PR is very simple and not extremely long.
Did splitting up the PR again it make easier to review for you ?

cardano-testnet/src/Testnet/Util/Cli.hs
MarcFontaine: I do not know.. I looked up the code `KeyGen` is used five times.

cardano-testnet/src/Testnet/Conf.hs
```diff
@@ -2,43 +2,33 @@
...
   } deriving (Eq, Show)
 
-newtype YamlFilePath = YamlFilePath
+newtype YamlFilePath = YamlFilePath   -- from the project that surounds the cabal package
   { projectBase :: FilePath
   } deriving (Eq, Show)
 
 data Conf = Conf
   { tempAbsPath :: FilePath
-  , tempRelPath :: FilePath
```
Jimbo4350, line 27: I can 100% get behind this simplification :+1:

cardano-testnet/src/Testnet/Babbage.hs
```diff
@@ -124,10 +125,12 @@ babbageTestnet testnetOptions H.Conf {..} = do
...
       { poolNodeKeysColdVkey = tempAbsPath </> "pools" </> "cold" <> show n <> ".vkey"
       , poolNodeKeysColdSkey = tempAbsPath </> "pools" </> "cold" <> show n <> ".skey"
+      , poolNodeKeysVrf = ( File $ tempAbsPath </> "node-spo" <> show n </> "vrf.vkey"
+                          , File $ tempAbsPath </> "node-spo" <> show n </> "vrf.skey"
+                          )
       , poolNodeKeysVrfVkey = tempAbsPath </> "node-spo" <> show n </> "vrf.vkey"
       , poolNodeKeysVrfSkey = tempAbsPath </> "node-spo" <> show n </> "vrf.skey"
-      , poolNodeKeysStakingVkey = tempAbsPath </> "pools" </> "staking-reward" <> show n <> ".vkey"
-      , poolNodeKeysStakingSkey = tempAbsPath </> "pools" </> "staking-reward" <> show n <> ".skey"
+      , poolNodeKeysOperator = undefined
```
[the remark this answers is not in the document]
MarcFontaine: Actually, it got called in one of the tests and
the error message contained the exact Module and line number.
The problem is in the `TestnetRuntime` type, which can either fit `babbage` or `cardano` but not both eras.
(That whole block of code is to be removed in a followup PR)

Jimbo4350:
> @Jimbo4350 @newhoggy Are there any comments or reviews on the underlying idea of the PR ? I see tons of reviews on `hlint` problems and unneeded parenthesis that could easily be fixed in a followup PR.

@MarcFontaine one the node-cli team we do not intentionally merge PRs with hlint problems and other style issues into `master`. You need to fix these first. I am going to separate the `TmpPath` change myself to demonstrate the types of PRs that I'm looking for.

MarcFontaine:
> > @Jimbo4350 @newhoggy Are there any comments or reviews on the underlying idea of the PR ? I see tons of reviews on `hlint` problems and unneeded parenthesis that could easily be fixed in a followup PR.
> 
> @MarcFontaine on the node-cli team we do not intentionally merge PRs with hlint problems and other style issues into `master`. You need to fix these first. I am going to separate the `TmpPath` change myself to demonstrate the type of PRs that I'm looking for.

The job of hlint is best done by CI.
How about this:
1) If there are no technical issues with a PR it gets a review with an approval
2) If it also passes CI (including of cause the linter) it gets merged.
3) We happily discuss "other style issues" as long as it is fun and does not kill productivity 
4) "other style issues" don"t hold back a PR.

Jimbo4350:
> > > @Jimbo4350 @newhoggy Are there any comments or reviews on the underlying idea of the PR ? I see tons of reviews on `hlint` problems and unneeded parenthesis that could easily be fixed in a followup PR.
> > 
> > 
> > @MarcFontaine on the node-cli team we do not intentionally merge PRs with hlint problems and other style issues into `master`. You need to fix these first. I am going to separate the `TmpPath` change myself to demonstrate the type of PRs that I'm looking for.
> 
> The job of hlint is best done by CI. How about this:
> 
>     1. If there are no technical issues with a PR it gets a review with an approval
> 
>     2. If it also passes CI (including of cause the linter) it gets merged.
> 
>     3. We happily discuss "other style issues" as long as it is fun and does not kill productivity
> 
>     4. "other style issues" don"t hold back a PR.

No. This PR can still be broken up and PRs need to be of a certain standard before merging. What you are suggesting will create more churn and make the code less readable. As I said I will demonstrate what I am looking for.

MarcFontaine:
> No. This PR can still be broken up and PRs need to be of a certain standard, both technical and stylistic, before merging. What you are suggesting will create more churn and make the code less readable. As I said I will demonstrate what I am looking for.

The orginal code has been written two years ago and nobody would care about improving it or moving it forward.
I open a PR and try to propose some changes and suddenly people start to merge PR after PR that change right that code. Honestly, I do not care to much about demonstrating anything to anybody..

PR #4848, opened 2023-01-25
Remove hlint from nix required CI

newhoggy:


nix/pkgs.nix
```diff
@@ -11,11 +11,6 @@ final: prev: with final; {
 
   cabal = haskell-nix.cabal-install.${compiler-nix-name};
 
-  hlint = haskell-nix.tool compiler-nix-name "hlint" {
```
jbgi, line 14: removing `hlint` here means `nix develop` will provide the `hlint` version from the nixpkgs pin (currently hlint 3.3.6).

jbgi: (alternatively we could remove the check from nix `required` job and have the gihub check run `nix build '.#checks/hlint'`)

jbgi (approved):

jbgi:
bors merge

PR #4853, opened 2023-01-27
Fix how changed files are selected in stylish-haskell CI check

newhoggy:
The `stylish-haskell` CI had `git diff --name-only HEAD origin/master`, which is wrong because that means added files are not check, but removed files are, which causes `stylish-haskell` to fail because removed files can't be found.  The order needed to be swapped around to `git diff --name-only origin/master HEAD`

### DIFF
--- a/.github/workflows/stylish-haskell.yml
+++ b/.github/workflows/stylish-haskell.yml
@@ -1,4 +1,4 @@
-name: Stylish Haskell
+name: Check Stylish Haskell
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


jbgi (approved):
Ideally we could parametrize the target branch of the PR: https://stackoverflow.com/questions/62331829/how-to-get-the-target-branch-of-the-github-pull-request-from-an-actions

Jimbo4350 (approved):
Nice  catch!

PR #4857, opened 2023-01-30
Remove NoImplicitPrelude

newhoggy:
Remove from `NoImplicitPrelude` from `cardano-api` `cardano-submit-api` `cardano-cli` `cardano-testnet` `cardano-node-chairman` `cardano-node` and `cardano-git-rev`.
Remove dependency on `cardano-prelude` wherever possible.

Removed imported of `Cardano.Prelude` where possible (excluding `bench`).  Remaining occurrences are those that are still required:

* `canonicalDecodePretty`
* `canonicalEncodePretty`
* `cborError`
* `ConvertText (..)`
* `FatalError (..)`
* `maximumDef`

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Shelley/Run.hs
+++ b/cardano-cli/src/Cardano/CLI/Shelley/Run.hs
@@ -4,8 +4,6 @@ module Cardano.CLI.Shelley.Run
   , runShelleyClientCommand
   ) where
 
-import           Cardano.Prelude
-
 import           Cardano.Api
 
 import           Control.Monad.Trans.Except.Extra (firstExceptT)
@@ -24,6 +22,8 @@ import           Cardano.CLI.Shelley.Run.Transaction
                                          -- Block, System, DevOps
 import           Cardano.CLI.Shelley.Run.Genesis
 import           Cardano.CLI.Shelley.Run.TextView
+import           Cardano.Prelude (ExceptT)
+import           Data.Text (Text)
 
 data ShelleyClientCmdError
   = ShelleyCmdAddressError !ShelleyAddressCmdError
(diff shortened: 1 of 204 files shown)


cardano-cli/src/Cardano/CLI/Shelley/Run.hs
```diff
@@ -24,6 +22,8 @@ import           Cardano.CLI.Shelley.Run.Transaction
                                          -- Block, System, DevOps
 import           Cardano.CLI.Shelley.Run.Genesis
 import           Cardano.CLI.Shelley.Run.TextView
+import           Cardano.Prelude (ExceptT)
```
Jimbo4350, line 25: Should be above with `Cardano.Api`

cardano-cli/src/Cardano/CLI/Shelley/Run/Address.hs
```diff
@@ -30,6 +27,9 @@ import           Cardano.CLI.Shelley.Parsers (AddressCmd (..), AddressKeyType (.
 import           Cardano.CLI.Shelley.Run.Address.Info (ShelleyAddressInfoError, runAddressInfo)
 import           Cardano.CLI.Shelley.Run.Read
 import           Cardano.CLI.Types
+import           Cardano.Prelude (MonadIO (..))
```
Jimbo4350, line 30: Should be the first import above with a newline after

cardano-cli/src/Cardano/CLI/Shelley/Run/Key.hs
```diff
@@ -41,6 +39,14 @@ import           Cardano.CLI.Shelley.Key (VerificationKeyTextOrFile (..),
                    VerificationKeyTextOrFileError, readVerificationKeyTextOrFileAnyOf,
                    renderVerificationKeyTextOrFileError)
 import           Cardano.CLI.Types (SigningKeyFile (..), VerificationKeyFile (..))
+import           Cardano.Prelude (MonadIO (..))
```
Jimbo4350, line 42: Same here, should be above

Jimbo4350 (approved):
LGTM, a couple of comments

PR #4866, opened 2023-02-07
Remove usage of nixWrapped.

jbgi:
it was introduced for adding flake flags by default and also as a work around for nix flake show/check: https://github.com/NixOS/nix/issues/4265 But the added complexity is not worth it, given it breaks other workflows with nix run (since the wrapper script does not use `exec`).

Also nowadays everyone (should) have flake activated by default in the nix config.


angerman (approved):
LGTM, removing complexity!

jbgi:
bors retry

PR #4865, opened 2023-02-07
Fix recursion in flake inputs.

andreabedini:
hydra-eval-jobs seems to have trouble evaluating this flake.

  error: cannot update flake input 'cardano-node-workbench/membench' in pure mode

Simplifying the recursion in the flake inputs seems to help.

Fixes #4525

PR #4871, opened 2023-02-09
Bump haskell.nix

andreabedini:
The bump is to support cabal file revisions on CHaP which are necessary to get our packages to build without all those manual constraints we have in cabal.project.

I opened a draft PR to test the waters and warm up the chaces ahead of time since it's relatively big change in haskell.nix. I'll make sure everything works and then mark it as ready.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -163,11 +163,11 @@
     "cabal-34_2": {
       "flake": false,
       "locked": {
-        "lastModified": 1640353650,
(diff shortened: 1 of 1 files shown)


andreabedini:
Rebased on top of #4865 to make the evalution work on hydra.

andreabedini:
Some jobs are failing but they seem to be failing on master as well :shrug: Marking this as ready.

angerman (approved):
LGTM, even though I'm a bit confused by the amount of things that changed int he flake.lock...

PR #4872, opened 2023-02-09
workbench:  `make shell-nix` use Nix-built binaries

deepfire:
This makes the `shell-nix` Makefile target actually use the Nix derivations for the node and components.


deepfire:
This is trivial, tested and only affects the Makefile, skipping CI.

PR #4873, opened 2023-02-09
workbench:  drop cabalWrapped

deepfire:
The end of an era -- the workbench will no longer insist on following the Nix-supplied package DB only -- we're dropping the use of `cabalWrapped`.

### DIFF
--- a/nix/workbench/backend/supervisor.nix
+++ b/nix/workbench/backend/supervisor.nix
@@ -15,7 +15,7 @@ let
     ]
   ++ lib.optionals ( useCabalRun)
     (with haskellPackages; [
-      cabalWrapped
+      cabal-install
       ghcid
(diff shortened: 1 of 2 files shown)


nix/workbench/backend/supervisor.nix
```diff
@@ -15,7 +15,7 @@ let
     ]
   ++ lib.optionals ( useCabalRun)
     (with haskellPackages; [
-      cabalWrapped
+      cabal-install
```
michaelpj, line 18: You should now also drop the explicit `packages` argument in the `haskell.nix` shell function.

deepfire: Oh, yes, right, forgot about that one.

Thank you!

PR #4880, opened 2023-02-13
Detect invalid counter and certificate

newhoggy:
Reports an warning if on-disk certificate counter is `2` or more than the node state certificate counter:

```bash
$ LC_ALL=en_US.iso88591 CARDANO_NODE_SOCKET_PATH=example/node-spo1/node.sock cardano-cli query kes-period-info --testnet-magic 42 --op-cert-file example/node-spo1/opcert.cert
✓ Operational certificate's KES period is within the correct KES period interval
✗ The operational certificate counter too far ahead of the node protocol state counter in the operational certificate at: example/node-spo1/opcert.cert
  On disk operational certificate counter: 4
  Protocol state counter: 2
{
    "qKesCurrentKesPeriod": 0,
    "qKesEndKesInterval": 60,
    "qKesKesKeyExpiry": null,
    "qKesMaxKESEvolutions": 60,
    "qKesNodeStateOperationalCertificateNumber": 2,
    "qKesOnDiskOperationalCertificateNumber": 4,
    "qKesRemainingSlotsInKesPeriod": 7672192,
    "qKesSlotsPerKesPeriod": 129600,
    "qKesStartKesInterval": 0
}
```

The PR also adds colour to the output:

Screenshot 2023-02-13 at 2 53 59 pm

Resolves https://github.com/input-output-hk/cardano-node/issues/4114

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Pretty.hs
+++ b/cardano-cli/src/Cardano/CLI/Pretty.hs
@@ -1,0 +1,44 @@
+module Cardano.CLI.Pretty
+  ( Ann(..),
+    putLn,
+    hPutLn,
+    renderDefault,
+    renderStringDefault,
+  ) where
+
+import           Control.Exception (bracket_)
+import           Control.Monad.IO.Class (MonadIO, liftIO)
+import           Data.Eq (Eq)
+import           Data.Function (($), (.))
+import           Data.String (String)
+import           Prelude (IO)
+import           Text.Show (Show)
+
+import qualified Control.Concurrent.QSem as IO
+import qualified Data.Text.Lazy as TextLazy
+import qualified Data.Text.Lazy.IO as LT
+import qualified Prettyprinter as PP
+import qualified Prettyprinter.Render.Text as PP
+import qualified System.IO as IO
+import qualified System.IO.Unsafe as IO
+
+data Ann = Ann deriving (Eq, Show)
+
+sem :: IO.QSem
+sem = IO.unsafePerformIO $ IO.newQSem 1
+{-# NOINLINE sem #-}
+
+consoleBracket :: IO a -> IO a
+consoleBracket = bracket_ (IO.waitQSem sem) (IO.signalQSem sem)
+
+putLn :: MonadIO m => PP.Doc Ann -> m ()
+putLn = liftIO . consoleBracket . LT.putStrLn . renderDefault
+
+hPutLn :: MonadIO m => IO.Handle -> PP.Doc Ann -> m ()
+hPutLn h = liftIO . consoleBracket . LT.hPutStr h . renderDefault
+
+renderStringDefault :: PP.Doc ann -> String
+renderStringDefault =  TextLazy.unpack . PP.renderLazy . PP.layoutPretty PP.defaultLayoutOptions { PP.layoutPageWidth = PP.Unbounded }
+
+renderDefault :: PP.Doc ann -> TextLazy.Text
+renderDefault =  PP.renderLazy . PP.layoutPretty PP.defaultLayoutOptions { PP.layoutPageWidth = PP.Unbounded }
--- a/cardano-cli/src/Cardano/CLI/Pretty.hs
+++ b/cardano-cli/src/Cardano/CLI/Pretty.hs
@@ -4,6 +4,15 @@ module Cardano.CLI.Pretty
     hPutLn,
     renderDefault,
     renderStringDefault,
+
+    black,
+    red,
+    green,
+    yellow,
+    blue,
+    magenta,
+    cyan,
+    white,
   ) where
 
 import           Control.Exception (bracket_)
@@ -39,4 +48,28 @@ renderStringDefault :: PP.Doc PP.AnsiStyle -> String
 renderStringDefault =  TextLazy.unpack . renderDefault
 
 renderDefault :: PP.Doc PP.AnsiStyle -> TextLazy.Text
-renderDefault =  PP.renderLazy . PP.layoutPretty PP.defaultLayoutOptions-- { PP.layoutPageWidth = PP.Unbounded }
+renderDefault =  PP.renderLazy . PP.layoutPretty PP.defaultLayoutOptions
+
+black :: PP.Doc PP.AnsiStyle -> PP.Doc PP.AnsiStyle
+black = PP.annotate (PP.color PP.Black)
+
+red :: PP.Doc PP.AnsiStyle -> PP.Doc PP.AnsiStyle
+red = PP.annotate (PP.color PP.Red)
+
+green :: PP.Doc PP.AnsiStyle -> PP.Doc PP.AnsiStyle
+green = PP.annotate (PP.color PP.Green)
+
+yellow :: PP.Doc PP.AnsiStyle -> PP.Doc PP.AnsiStyle
+yellow = PP.annotate (PP.color PP.Yellow)
+
+blue :: PP.Doc PP.AnsiStyle -> PP.Doc PP.AnsiStyle
+blue = PP.annotate (PP.color PP.Blue)
+
+magenta :: PP.Doc PP.AnsiStyle -> PP.Doc PP.AnsiStyle
+magenta = PP.annotate (PP.color PP.Magenta)
+
+cyan :: PP.Doc PP.AnsiStyle -> PP.Doc PP.AnsiStyle
+cyan = PP.annotate (PP.color PP.Cyan)
+
+white :: PP.Doc PP.AnsiStyle -> PP.Doc PP.AnsiStyle
+white = PP.annotate (PP.color PP.White)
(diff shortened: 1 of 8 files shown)


cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -527,29 +528,29 @@ runQueryKesPeriodInfo (AnyConsensusModeParams cModeParams) network nodeOpCertFil
...
-            ]
-          )
+        PP.annotate (PP.color PP.Green) "✓" PP.<+> PP.hang 0
+            ( PP.vsep
+              [ "The operational certificate counter agrees with the node protocol state counter"
+              ]
+            )
       OpCertOnDiskCounterTooFarAheadOfNodeState onDiskC nodeStateC ->
-        "✗" PP.<+> PP.hang 0
+        PP.annotate (PP.color PP.Red) "✗" PP.<+> PP.hang 0
```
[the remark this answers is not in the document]
newhoggy: Done!

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -112,6 +112,7 @@ import           Data.Text.Lazy (toStrict)
 import           GHC.Records (HasField)
 import qualified Ouroboros.Consensus.HardFork.History.Qry as Qry
 import qualified Ouroboros.Network.Protocol.LocalStateQuery.Type as LocalStateQuery
+import qualified Prettyprinter.Render.Terminal as PP
```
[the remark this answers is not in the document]
newhoggy: Fixed

cardano-cli/src/Cardano/CLI/Pretty.hs
```diff
@@ -31,14 +29,14 @@ sem = IO.unsafePerformIO $ IO.newQSem 1
...
 
-renderStringDefault :: PP.Doc ann -> String
-renderStringDefault =  TextLazy.unpack . PP.renderLazy . PP.layoutPretty PP.defaultLayoutOptions { PP.layoutPageWidth = PP.Unbounded }
+renderStringDefault :: PP.Doc PP.AnsiStyle -> String
+renderStringDefault =  TextLazy.unpack . renderDefault
 
-renderDefault :: PP.Doc ann -> TextLazy.Text
-renderDefault =  PP.renderLazy . PP.layoutPretty PP.defaultLayoutOptions { PP.layoutPageWidth = PP.Unbounded }
+renderDefault :: PP.Doc PP.AnsiStyle -> TextLazy.Text
+renderDefault =  PP.renderLazy . PP.layoutPretty PP.defaultLayoutOptions-- { PP.layoutPageWidth = PP.Unbounded }
```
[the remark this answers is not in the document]
newhoggy: Fixed!

cardano-testnet/test/Test/Misc.hs
```diff
@@ -21,16 +22,16 @@ prop_op_cert_valid_kes_period opCertFp output =
       OpCertWithinInterval{} -> success
       info@OpCertStartingKesPeriodIsInTheFuture{} ->
         failMessage GHC.callStack
-          $ "Expected OpCertWithinInterval but got: OpCertStartingKesPeriodIsInTheFuture " <> "\n"
-          <> renderOpCertIntervalInformation opCertFp info
+          $ "Expected OpCertWithinInterval but got: OpCertStartingKesPeriodIsInTheFuture\n"
+          <> PP.renderStringDefault (renderOpCertIntervalInformation opCertFp info)
```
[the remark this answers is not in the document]
newhoggy: I've reverted the type change.

MarcFontaine:
What is the motivation/rationale for importing the pretty printer qualified (`import qualified Prettyprinter as PP`) in
[Cardano.CLI.Pretty](https://github.com/input-output-hk/cardano-node/blob/50fe937dd2d1a3c1e96bf1182c43064e3712a53a/cardano-cli/src/Cardano/CLI/Pretty.hs)
This qualified import leads to tons of `PP.` everywhere in the code.
(`grep -o 'PP.' Pretty.hs |wc` says 68 in that single module)
What is the supposed advantage of that style ?

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -506,7 +506,8 @@ runQueryKesPeriodInfo (AnyConsensusModeParams cModeParams) network nodeOpCertFil
    opCertNodeAndOnDiskCounters o@(OpCertOnDiskCounter odc) (Just n@(OpCertNodeStateCounter nsc))
      | odc < nsc = OpCertOnDiskCounterBehindNodeState o n
      | odc > nsc + 1 = OpCertOnDiskCounterTooFarAheadOfNodeState o n
-     | otherwise = OpCertOnDiskCounterMoreThanOrEqualToNodeState o n
+     | odc == nsc + 1 = OpCertOnDiskCounterAheadOfNodeState o n
```
Jimbo4350, line 509: I think this case could be condensed with the following case: `| odc >= nsc + 1 = OpCertOnDiskCounterTooFarAheadOfNodeState o n`  

We can just report how much further ahead it is by rendering `n` and outputting a suitable error message.

newhoggy: This would entail reporting success or failure depending on `n`.  `n + 1` is success and `n + 2` or higher is failure.  Are you happy for the decisioning making to be at the point of rendering?

newhoggy: `OpCertOnDiskCounterTooFarAheadOfNodeState` would have to be renamed as well because "too far ahead" has bad connotations while `n + 1` is actually ok.

Jimbo4350: Ok, lets leave it how you've done it :+1:

Jimbo4350:
> What is the motivation/rationale for importing the pretty printer qualified (`import qualified Prettyprinter as PP`) in [Cardano.CLI.Pretty](https://github.com/input-output-hk/cardano-node/blob/50fe937dd2d1a3c1e96bf1182c43064e3712a53a/cardano-cli/src/Cardano/CLI/Pretty.hs) This qualified import leads to tons of `PP.` everywhere in the code. (`grep -o 'PP.' Pretty.hs |wc` says 68 in that single module) What is the supposed advantage of that style ? It looks like **all** of the qualified uses could just be replaced with unqualified ones. with no other change needed but removing `qualified` from the import.

I also agree with this. We need to have some rules about when we should qualify code. My opinion is to only qualify when we want to avoid name clashes when importing from different libraries e.g `aeson`, `prelude`  etc. I wouldn't bother qualifying imports for things that are defined locally, its too easy to grep for them.

newhoggy:
There is one small advantage, which is when you type `PP.` in VSCode, you get a dropdown list of things exported by `PP`, so when writing code names are more discoverable.

newhoggy:
I have reviewed the `Prettyprinter` module and from the function names, the opportunity for conflict for the moment seems minimal, so I've removed the qualified import.

cardano-cli/src/Cardano/CLI/Pretty.hs
```diff
@@ -0,0 +1,75 @@
...
+    yellow,
+    blue,
+    magenta,
+    cyan,
+    white,
+  ) where
+
+import           Control.Exception (bracket_)
+import           Control.Monad.IO.Class (MonadIO, liftIO)
+import           Data.Function (($), (.))
```
Jimbo4350, line 20: Why import `Data.Function` and `Prelude`?

Jimbo4350:
Is `LC_ALL=en_US.iso88591` a requirement to see the colour?

cardano-cli/src/Cardano/CLI/Pretty.hs
newhoggy: Fixed.  Thanks!

newhoggy:
> Is `LC_ALL=en_US.iso88591` a requirement to see the colour?

Actually, I think this is not colour related.  As I remember it's because the ticks and crosses are unicode characters, which some OSes can't render by default because of the way they are set up.

I use a Mac, so this automatically works without the environment variable.

cardano-cli/src/Cardano/CLI/Pretty.hs
```diff
@@ -0,0 +1,72 @@
+module Cardano.CLI.Pretty
+  ( Ann,
```
Jimbo4350, line 2: Can you fix the export style to follow the usual style in `cardano-cli`?
```
( Ann
, ...
```

Jimbo4350 (approved):
LGTM, just one change

cardano-cli/src/Cardano/CLI/Pretty.hs
newhoggy: Done!

Issue #4433, opened 2022-09-13
[BUG] - Possible corruption of datum by CLI


ArturWieczorek:
**Internal/External**
*Internal*   
**Area**
`cardano-cli` - Tx with datum.

**Summary**
Originally this bug was opened in `cardano-db-sync` repository:
 https://github.com/input-output-hk/cardano-db-sync/issues/1214

The observed errors could be caused by a CLI issue. It looks like Tx is created with the wrong Datum and db-sync does the correct thing of extracting the bytes as they are.

The test is defined here:
https://github.com/input-output-hk/cardano-node-tests/blob/master/cardano_node_tests/tests/test_plutus_v2_spend_raw.py#L375

it fails with:

```
>       assert (
            db_cbor_hex == orig_cbor_hex
        ), "Datum bytes in db-sync doesn't correspond to the original datum"
E       AssertionError: Datum bytes in db-sync doesn't correspond to the original datum
E       assert 'd8799f41ffd8799fd8799fd8799fd8799f581cc279a3fb3b4e62bbc78e288783b58045d4ae82a18867d8352d02775affd8799fd8799fd8799f581c121fd22e0b57ac206fefc763f8bfa0771919f5218b40691eea4514d0ffffffffd87a80ffd87a80ff1a002625a0d8799fd879801a000f4240d8799f1a000fa92effffff' == 'd8798441ffd87982d87982d87982d87981581cc279a3fb3b4e62bbc78e288783b58045d4ae82a18867d8352d02775ad87981d87981d87981581c121fd22e0b57ac206fefc763f8bfa0771919f5218b40691eea4514d0d87a80d87a801a002625a0d87983d879801a000f4240d879811a000fa92e'
```

> We use [https://github.com/input-output-hk/cardano-node-tests/blob/8baf63e9e4c66c02c903f9e[…]e8/cardano_node_tests/tests/data/plutus/typed-finite.datum.cbor](https://github.com/input-output-hk/cardano-node-tests/blob/8baf63e9e4c66c02c903f9e359440f91edf228e8/cardano_node_tests/tests/data/plutus/typed-finite.datum.cbor) which is the datum used in the original reports [1](https://github.com/input-output-hk/cardano-db-sync/issues/1214) and [2](https://github.com/CardanoSolutions/ogmios/commit/3f614c3cf15b6e418fd461a34853b750a3408c4f).
>
> We use this to create the Tx:
> ```
[...]

PR #4886, opened 2023-02-14
Preserve `ScriptData` bytes with `HashableScriptData`

Jimbo4350:
Resolves: #4433

### DIFF
--- a/cardano-api/src/Cardano/Api/ScriptData.hs
+++ b/cardano-api/src/Cardano/Api/ScriptData.hs
@@ -1,11 +1,17 @@
 {-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE InstanceSigs #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 
 module Cardano.Api.ScriptData (
     -- * Script data
+    HashableScriptData,
+    hashScriptDataBytes,
+    getOriginalScriptDataBytes,
+    getScriptData,
+    unsafeHashableScriptData,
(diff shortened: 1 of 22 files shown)


newhoggy (approved):

cardano-api/src/Cardano/Api/ScriptData.hs
```diff
@@ -77,14 +89,35 @@ import           Cardano.Api.Keys.Shelley
...
+      !BS.ByteString -- ^ Original 'ScriptData' bytes
+      !ScriptData
+      deriving (Eq, Show)
+
+instance HasTypeProxy HashableScriptData where
+    data AsType HashableScriptData = AsHashableScriptData
+    proxyToAsType _ = AsHashableScriptData
+
+instance SerialiseAsCBOR HashableScriptData where
+    serialiseToCBOR (HashableScriptData _ sd) = CBOR.serialize' sd
```
JaredCorduan, line 107:
```suggestion
    serialiseToCBOR (HashableScriptData bs _) = bs
```
Maybe I am confused, but is the point of `HashableScriptData` to make sure that if you ever hash a value of `HashableScriptData` that you hash the original bytes? If so, wouldn't you want to make sure that you always retain the original serialization?

I think it would also make sense to hide the first field (`!BS.ByteString`) with a smart constructor so that `HashableScriptData` can't be constructed incorrectly.

Jimbo4350: > Maybe I am confused, but is the point of HashableScriptData to make sure that if you ever hash a value of HashableScriptData that you hash the original bytes? If so, wouldn't you want to make sure that you always retain the original serialization?

Nop you're right, this was a mistake.

JaredCorduan (approved):
looks great to me @Jimbo4350 !

PR #4896, opened 2023-02-17
devops shell: don't include build tools for all hasPkgs

jbgi:
because this lead to duplicates PATH entries for cardano-cli, cardano-node, etc.
 with one having gitrev set and the other not set.


deepfire (approved):
Thank you @jbgi !

PR #4905, opened 2023-02-21
workbench | tx-generator: SECP benchmarking

mgmeier:
This PR expands the benchmarking machinery to enable SECP benchmarking.

- add `plutuscall-*` profile family to workbench
- add `model-*` profile family to workbench
- additional loop calibration strategies for tx-generator, where previously only 'maximize per-tx-budget expenditure' was possible
- make tx-generator emit a Plutus budget summary after loop calibration (via `TraceBenchPlutusBudgetSummary`)
- plus several small convenience improvements

_TODO_: PlutusV2 cost model needs extension; the current one from epoch timeline does not provide enough parameters.


mgmeier:
Closed - duplicate of #4907

PR #4909, opened 2023-02-22
Fix qKesKesKeyExpiry to not always be null

newhoggy:
Note that `qKesKesKeyExpiry` is only an estimate that is usually correct, but it could be wrong due to a hard fork that changes KES parameters.

The fix involves uses an unsafe query interpreter that ignores forecasting horizon.  Therefore a new `Unsafe` `newtype` has been introduced to clearly show all the code that uses the unsafe query interpreter so that it doesn't accidentally get used where it shouldn't.

Resolves https://github.com/input-output-hk/cardano-node/issues/4396

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
+++ b/cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
@@ -448,7 +448,7 @@ runQueryKesPeriodInfo (AnyConsensusModeParams cModeParams) network nodeOpCertFil
       eraHistory <- lift (queryNodeLocalState localNodeConnInfo Nothing eraHistoryQuery)
         & onLeft (left . ShelleyQueryCmdAcquireFailure)
 
-      let eInfo = toEpochInfo eraHistory
+      let eInfo = toTentativeEpochInfo eraHistory
 
 
       -- We get the operational certificate counter from the protocol state and check that
@@ -515,13 +515,13 @@ runQueryKesPeriodInfo (AnyConsensusModeParams cModeParams) network nodeOpCertFil
    opCertNodeAndOnDiskCounters o Nothing = OpCertNoBlocksMintedYet o
 
    opCertExpiryUtcTime
-     :: EpochInfo (Either Text)
+     :: Tentative (EpochInfo (Either Text))
      -> GenesisParameters
      -> OpCertEndingKesPeriod
      -> Maybe UTCTime
    opCertExpiryUtcTime eInfo gParams (OpCertEndingKesPeriod oCertExpiryKesPeriod) =
      let time = epochInfoSlotToUTCTime
-                  eInfo
+                  (tentative eInfo)
                   (SystemStart $ protocolParamSystemStart gParams)
                   (fromIntegral $ oCertExpiryKesPeriod * fromIntegral (protocolParamSlotsPerKESPeriod gParams))
      in case time of
@@ -576,7 +576,7 @@ runQueryKesPeriodInfo (AnyConsensusModeParams cModeParams) network nodeOpCertFil
    createQueryKesPeriodInfoOutput
      :: OpCertIntervalInformation
      -> OpCertNodeAndOnDiskCounterInformation
-     -> EpochInfo (Either Text)
+     -> Tentative (EpochInfo (Either Text))
      -> GenesisParameters
      -> O.QueryKesPeriodInfoOutput
    createQueryKesPeriodInfoOutput oCertIntervalInfo oCertCounterInfo eInfo gParams  =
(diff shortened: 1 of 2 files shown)


cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -1341,8 +1341,16 @@ queryResult eAcq = pure eAcq
 
 toEpochInfo :: EraHistory CardanoMode -> EpochInfo (Either Text)
 toEpochInfo (EraHistory _ interpreter) =
-  hoistEpochInfo (first (Text.pack . show ) . runExcept)
-    $ Consensus.interpreterToEpochInfo interpreter
+  hoistEpochInfo (first (Text.pack . show) . runExcept)
+    $ Consensus.interpreterToEpochInfo (Consensus.unsafeExtendSafeZone interpreter)
+
+newtype Unsafe a = Unsafe { unsafe :: a } deriving (Eq, Show)
```
nfrisby, line 1347: I like this use of a newtype. I think `Unsafe` is a bit too scary/general of a name though. Perhaps `AssumingNoUpcomingHardfork` (or something like that)?

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -1341,8 +1341,16 @@ queryResult eAcq = pure eAcq
...
+
+unsafeToEpochInfo :: EraHistory CardanoMode -> Unsafe (EpochInfo (Either Text))
```
nfrisby, line 1349: Similarly, I'd suggest naming this `toEpochAssumingNoUpcomingHardFork` or something like that.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -1341,8 +1341,16 @@ queryResult eAcq = pure eAcq
 
 toEpochInfo :: EraHistory CardanoMode -> EpochInfo (Either Text)
 toEpochInfo (EraHistory _ interpreter) =
-  hoistEpochInfo (first (Text.pack . show ) . runExcept)
-    $ Consensus.interpreterToEpochInfo interpreter
+  hoistEpochInfo (first (Text.pack . show) . runExcept)
+    $ Consensus.interpreterToEpochInfo (Consensus.unsafeExtendSafeZone interpreter)
```
nfrisby, line 1345: `unsafeExtendSafeZone` should _not_ be called here, right?

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
Jimbo4350: What about `newtype BeyondHorizonEpochInfo = BeyondHorizonEpochInfo (EpochInfo (Either Text))`

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
Jimbo4350: `toBeyondHorizonEpochInfo`?

newhoggy: This is now `toTentativeEpochInfo`.

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: Fixed.  Thanks!

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -1345,9 +1345,25 @@ queryResult eAcq = pure eAcq
...
 toEpochInfo :: EraHistory CardanoMode -> EpochInfo (Either Text)
 toEpochInfo (EraHistory _ interpreter) =
-  hoistEpochInfo (first (Text.pack . show ) . runExcept)
+  hoistEpochInfo (first (Text.pack . show) . runExcept)
     $ Consensus.interpreterToEpochInfo interpreter
 
+-- | A value that is tentative or produces a tentative value if used.  These values
+-- are considered accurate only if some future event such as a hard fork does not
+-- render them invalid.
+newtype Tentative a = Tentative { tentative :: a } deriving (Eq, Show)
```
Jimbo4350, line 1354: What other values do we expect to be tentative besides the `EpochInfo`?

Jimbo4350 (changes requested):

Jimbo4350 (approved):

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -1415,9 +1415,25 @@ queryResult eAcq = pure eAcq
...
+
+-- | Get an Epoch Info that computes tentative values.  The values computed are
+-- tentative because it uses an interpreter that is extended past the horizon.
+-- This interpreter will compute accurate values into the future as long as a
+-- a hard fork does not happen in the intervening time.  Those values are thus
+-- "tentative" because they can change in the event of a hard fork.
```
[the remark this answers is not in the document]
newhoggy: They can change in the event of a hardfork occurring _after_ the horizon.  Normally the interpreter refuses to give an answer beyond the horizon so we construct a different interpreter that ignores the horizon.  

Values before the horizon won't change.

However, because the type allows for values beyond the horizon to change, it is annotated as `Tentative`.

PR #4910, opened 2023-02-23
New --socket-path cli option

newhoggy:
Resolves https://github.com/input-output-hk/cardano-node/issues/1388

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Byron/Run.hs
+++ b/cardano-cli/src/Cardano/CLI/Byron/Run.hs
@@ -8,9 +8,11 @@ module Cardano.CLI.Byron.Run
 
 import           Control.Monad.IO.Class (MonadIO (liftIO))
 import           Control.Monad.Trans.Except (ExceptT)
-import           Control.Monad.Trans.Except.Extra (firstExceptT, hoistEither, left)
+import           Control.Monad.Trans.Except.Extra (firstExceptT, hoistEither, left, onLeft)
+import           Control.Monad.Trans (MonadTrans (..))
 import           Data.Bifunctor (Bifunctor (..))
 import qualified Data.ByteString.Char8 as BS
+import           Data.Function ((&))
 import           Data.Text (Text)
 import qualified Data.Text as Text
 import qualified Data.Text.IO as Text
@@ -74,7 +76,7 @@ runByronClientCommand c =
   case c of
     NodeCmd bc -> runNodeCmd bc
     Genesis outDir params -> runGenesisCommand outDir params
-    GetLocalNodeTip network -> firstExceptT ByronCmdQueryError $ runGetLocalNodeTip network
+    GetLocalNodeTip mNodeSocketPath network -> firstExceptT ByronCmdQueryError $ runGetLocalNodeTip mNodeSocketPath network
     ValidateCBOR cborObject fp -> runValidateCBOR cborObject fp
     PrettyPrintCBOR fp -> runPrettyPrintCBOR fp
     PrettySigningKeyPublic bKeyFormat skF -> runPrettySigningKeyPublic bKeyFormat skF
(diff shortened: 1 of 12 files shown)


cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -1819,6 +1844,19 @@ pOutputFile =
       <> Opt.completer (Opt.bashCompleter "file")
       )
 
+pNodeSocketPath :: Parser (Maybe SocketPath)
+pNodeSocketPath =
+  optional $ fmap SocketPath $
+    Opt.strOption $ mconcat
+      [ Opt.long "node-socket-path"
```
Jimbo4350, line 1851: Let's maintain consistency and leave this as `"socket-path"` as `cardano-node`, `cardano-chairman` and `cardano-submit-api`  parse for `"socket-path"`

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
```diff
@@ -1163,18 +1181,21 @@ instance FromJSON DelegationsAndRewards where
...
   -> VerificationKeyOrHashOrFile StakePoolKey
   -> SigningKeyFile -- ^ VRF signing key
   -> EpochLeadershipSchedule
   -> Maybe OutputFile
   -> ExceptT ShelleyQueryCmdError IO ()
-runQueryLeadershipSchedule (AnyConsensusModeParams cModeParams) network
-                           (GenesisFile genFile) coldVerKeyFile (SigningKeyFile vrfSkeyFp)
-                           whichSchedule mJsonOutputFile = do
-  SocketPath sockPath <- lift readEnvSocketPath & onLeft (left . ShelleyQueryCmdEnvVarSocketErr)
+runQueryLeadershipSchedule
```
Jimbo4350, line 1193: Strange indent here. Did we hit the 100 char limit when you added `mNodeSocketPath`?

Jimbo4350 (changes requested):
Couple of minor changes

cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs
newhoggy: I'm using the convention that arguments indent four spaces whilst the function implementation indents two.  This is to distinguish the two by ensuring they don't have the same indent.

Should I be using some other convention?

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
newhoggy: Done!

cardano-cli/src/Cardano/CLI/Byron/Run.hs
```diff
@@ -40,6 +40,8 @@ import           Cardano.CLI.Byron.Vote
 import           Cardano.CLI.Helpers
 import           Cardano.CLI.Shelley.Commands (ByronKeyFormat (..))
 import           Cardano.CLI.Types
+import           Control.Monad.Trans (MonadTrans (..))
```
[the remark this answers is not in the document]
newhoggy: Moved!

cardano-cli/src/Cardano/CLI/Shelley/Run/Transaction.hs
```diff
@@ -698,8 +705,6 @@ runTxBuild era (AnyConsensusModeParams cModeParams) networkId mScriptValidity
         & onNothing (left (ShelleyTxCmdEraConsensusModeMismatchTxBalance outputOptions
                             (AnyConsensusMode CardanoMode) (AnyCardanoEra era)))
 
-      SocketPath sockPath <- lift readEnvSocketPath & onLeft (left . ShelleyTxCmdSocketEnvError)
```
Jimbo4350, line 701: Why are we not allowing specification of the socket path by environmental variable anymore?

newhoggy: The above code is in the `runTxBuild` function.  This code previously read from the environment variable itself.  However, the caller of `runTxBuild` is `runTxBuildCmd` and that function already reads from the environment variable, so instead that function passes the socket file into `runTxBuild` so that we only read the environment variable once.  Doing so allows the caller to also decide not to read the environment variable, for example when the socket was provided explicitly on the command line.

PR #4912, opened 2023-02-23
Add slotsInEpoch and slotsToEpochEnd to query tip command

newhoggy:
Example output:

```
CARDANO_NODE_SOCKET_PATH=example/node-pool1/node.sock cardano-cli query tip --testnet-magic 42
{
    "block": 768,
    "epoch": 33,
    "era": "Alonzo",
    "hash": "4ca7fd3a9827e74f98e779a8ae333dc6fa3b69832cb11fdbab7af215eb68b1ee",
    "slot": 16814,
    "slotsInEpoch": 314,
    "slotsToEpochEnd": 186,
    "syncProgress": "100.00"
}
```

Resolves https://github.com/input-output-hk/cardano-node/issues/3878

### DIFF
--- a/cardano-cli/ChangeLog.md
+++ b/cardano-cli/ChangeLog.md
@@ -12,6 +12,8 @@
 
 - Allow assembling transactions with no witnesses ([PR 4408](https://github.com/input-output-hk/cardano-node/pull/4408))
 
+- Add `slotInEpoch` and `slotsToEpochEnd` to output of `query tip` command ([PR 4912](https://github.com/input-output-hk/cardano-node/pull/4912))
+
 ### Bugs
 
 - Allow reading signing keys from a pipe ([PR 4342](https://github.com/input-output-hk/cardano-node/pull/4342))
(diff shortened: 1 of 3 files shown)


erikd (approved):

Jimbo4350 (changes requested):
`"slotsInEpoch": 314,` is this describing the current slot in the epoch? It looks like it is. We should change the name so `slotInEpoch` as `slotsInEpoch` sounds like the number of slots per epoch.

cardano-cli/ChangeLog.md
```diff
@@ -12,6 +12,8 @@
 
 - Allow assembling transactions with no witnesses ([PR 4408](https://github.com/input-output-hk/cardano-node/pull/4408))
 
+- Add `slotsInEpoch` and `slotsToEpochEnd` to output of `query tip` command ([PR 4912](https://github.com/input-output-hk/cardano-node/pull/4912))
```
Jimbo4350, line 15: Sp: `slotInEpoch`

Jimbo4350 (approved):
LGTM!

PR #4915, opened 2023-02-24
Tip hash metrics

newhoggy:
Salvaging this PR https://github.com/input-output-hk/cardano-node/pull/2006

Related to https://github.com/input-output-hk/cardano-node/issues/801

### DIFF
--- a/cardano-node/src/Cardano/Tracing/Tracers.hs
+++ b/cardano-node/src/Cardano/Tracing/Tracers.hs
@@ -65,7 +65,7 @@ import           Cardano.BM.Trace (traceNamedObject)
 import           Cardano.BM.Tracing
 
 import           Ouroboros.Consensus.Block (BlockConfig, BlockProtocol, CannotForge,
-                   ConvertRawHash (..), ForgeStateInfo, ForgeStateUpdateError, Header, HeaderHash,
+                   ConvertRawHash (..), ForgeStateInfo, ForgeStateUpdateError, Header,
                    realPointHash, realPointSlot)
 import           Ouroboros.Consensus.BlockchainTime (SystemStart (..),
                    TraceBlockchainTimeEvent (..))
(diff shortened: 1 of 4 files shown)


intricate:
Thanks :grin:

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -1482,23 +1504,30 @@ data ChainInformation = ChainInformation
     -- current chain.
   , fork :: Bool
     -- ^ Was this a fork.
+  , tipBlockHash :: HeaderHash blk
```
Jimbo4350, line 1521: I would argue lets not parameterize `ChainInformation` and instead have the following fields:
```
...
  , tipBlockHash :: Text
  , tipBlockParentHash :: Text
...
```
These are chain hashes that eventually get rendered anyways.

Jimbo4350 (dismissed):

cardano-node/src/Cardano/Tracing/Tracers.hs
newhoggy: Done!

erikd (approved):

cardano-node/src/Cardano/Tracing/Tracers.hs
newhoggy: Done!

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -1466,7 +1492,7 @@ traceInboundGovernorCountersMetrics (OnOff True) (Just ekgDirect) = ipgcTracer
 
 -- | get information about a chain fragment
 
-data ChainInformation = ChainInformation
+data ChainInformation blk = ChainInformation
```
Jimbo4350, line 1495: This is still parameterized

Jimbo4350 (changes requested):
LGTM! You should be able to squash your commit into the first one without issue

PR #4918, opened 2023-02-27
Fix key non extended key for StakeExtendedVerificationKeyShelley_ed25519_bip32 envelope

newhoggy:
* Fix key non extended key for `StakeExtendedVerificationKeyShelley_ed25519_bip32` envelope
* Add ability to generate golden files from test using `CREATE_GOLDEN_FILES=1`
* New `diffFileVsGoldenFile` function

Resolves https://github.com/input-output-hk/cardano-node/issues/4914

### DIFF
--- a/cardano-cli/test/Test/Utilities.hs
+++ b/cardano-cli/test/Test/Utilities.hs
@@ -1,27 +1,87 @@
-module Test.Utilities (diffVsGoldenFile) where
+module Test.Utilities
+  ( diffVsGoldenFile,
+    diffFileVsGoldenFile,
+  ) where
 
-import           Cardano.Prelude (ConvertText (..))
+import           Cardano.Prelude (ConvertText (..), HasCallStack)
 
 import           Control.Monad.IO.Class (MonadIO (liftIO))
 import           Data.Algorithm.Diff (PolyDiff (Both), getGroupedDiff)
 import           Data.Algorithm.DiffOutput (ppDiff)
 import           GHC.Stack (callStack)
+import qualified GHC.Stack as GHC
 import           Hedgehog (MonadTest)
+import qualified Hedgehog.Extras.Test as H
 import           Hedgehog.Extras.Test.Base (failMessage)
+import qualified Hedgehog.Internal.Property as H
+import qualified System.Directory as IO
+import qualified System.Environment as IO
+import qualified System.IO.Unsafe as IO
 
+-- | Whether the test should create the golden files if the file does ont exist.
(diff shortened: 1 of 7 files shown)


cardano-cli/test/Test/Utilities.hs
```diff
@@ -1,27 +1,63 @@
...
+import qualified GHC.Stack as GHC
 import           Hedgehog (MonadTest)
+import qualified Hedgehog.Extras.Test as H
 import           Hedgehog.Extras.Test.Base (failMessage)
+import qualified Hedgehog.Internal.Property as H
+import qualified System.Directory as IO
+import qualified System.Environment as IO
+import qualified System.IO.Unsafe as IO
+
+-- | Whether the test should create the golden files if the file does ont exist.
```
Jimbo4350, line 21: I don't think we should be creating golden files on the fly.  Why? Because a golden file is supposed to serve as a reference against unexpected _future_ changes. If we generate the files on the fly, the file is no longer really a "reference file" because whatever changes we may have made will be present in the "reference" file and our "actual" file. This will then give us a false positive in our test suite if we introduced unintended changes.

cardano-cli/test/Test/Utilities.hs
```diff
@@ -1,27 +1,63 @@
...
+    then do
+      referenceLines <- map toS . lines <$> H.readFile referenceFile
+      let difference = getGroupedDiff actualLines referenceLines
+      case difference of
+        [Both{}] -> pure ()
+        _        -> failMessage callStack $ ppDiff difference
+    else if createFiles
+      then do
+        H.note_ $ "Creating golden file " <> referenceFile
+        H.writeFile referenceFile actualContent
```
Jimbo4350, line 48: So we will never fail in this case! We simply write the `actualContent` to the `referenceFile`.

newhoggy: If the golden file already exists, it will not be overwritten.

Jimbo4350: Ok, I understand what you've done. Can you add a comment that explains this case is simply to conveniently create golden files automatically.

Jimbo4350 (approved):
LGTM, minor comment: https://github.com/input-output-hk/cardano-node/pull/4918/files#r1120108840

cardano-cli/test/Test/Utilities.hs
newhoggy: Comment added.

PR #4919, opened 2023-02-27
Check CLI argument bounds

newhoggy:
The bounds of many CLI arguments are now checked.

```
  --alonzo-era \
  --testnet-magic 42 \
  --change-address 'addr_test1vztpl4gth92fpxh3zth7w6v7mm7wewt89edae2hq3rptu5qa7ahzt' \
  --tx-in 'dee44d80c8b9da8cf20d1fae78a79a03b70b8abf76bfda88fe3651efcb0ab0cd#0' \
  --tx-out 'addr_test1vztpl4gth92fpxh3zth7w6v7mm7wewt89edae2hq3rptu5qa7ahzt+1000000' \
  --out-file tx.body --invalid-before -1

option --invalid-before: SLOT must not be less than 0
...
```

Resolves https://github.com/input-output-hk/cardano-node/issues/4863

### DIFF
--- a/cardano-api/src/Cardano/Api.hs
+++ b/cardano-api/src/Cardano/Api.hs
@@ -751,6 +751,9 @@ module Cardano.Api (
     txInsExistInUTxO,
     notScriptLockedTxIns,
     textShow,
+
+    -- ** CLI option parsing
+    bounded,
   ) where
 
 import           Cardano.Api.Address
--- a/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
+++ b/cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
@@ -1730,31 +1730,28 @@ pSigningKeyFile fdir =
 
 pKesPeriod :: Parser KESPeriod
 pKesPeriod =
-  KESPeriod <$>
-    Opt.option Opt.auto
-      (  Opt.long "kes-period"
-      <> Opt.metavar "NATURAL"
-      <> Opt.help "The start of the KES key validity period."
-      )
+  fmap KESPeriod $ Opt.option (bounded "KES_PERIOD") $ mconcat
+    [ Opt.long "kes-period"
+    , Opt.metavar "NATURAL"
+    , Opt.help "The start of the KES key validity period."
+    ]
 
 pEpochNo :: Parser EpochNo
 pEpochNo =
-  EpochNo <$>
-    Opt.option Opt.auto
-      (  Opt.long "epoch"
-      <> Opt.metavar "NATURAL"
-      <> Opt.help "The epoch number."
-      )
+  fmap EpochNo $ Opt.option (bounded "EPOCH") $ mconcat
+    [ Opt.long "epoch"
+    , Opt.metavar "NATURAL"
+    , Opt.help "The epoch number."
+    ]
 
 
 pEpochNoUpdateProp :: Parser EpochNo
 pEpochNoUpdateProp =
-  EpochNo <$>
-    Opt.option Opt.auto
-      (  Opt.long "epoch"
-      <> Opt.metavar "NATURAL"
-      <> Opt.help "The epoch number in which the update proposal is valid."
-      )
+  fmap EpochNo $ Opt.option (bounded "EPOCH") $ mconcat
+    [ Opt.long "epoch"
+    , Opt.metavar "EPOCH"
+    , Opt.help "The epoch number in which the update proposal is valid."
+    ]
 
 pGenesisFile :: String -> Parser GenesisFile
 pGenesisFile desc =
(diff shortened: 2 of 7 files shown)


cardano-cli/src/Cardano/CLI/Optparse.hs
```diff
@@ -0,0 +1,18 @@
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Cardano.CLI.OptParse
```
Jimbo4350, line 5: Lets not create an entire module for this function. Let's put it in `Cardano.CLI.Helpers`

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -1727,29 +1728,28 @@ pSigningKeyFile fdir =
...
+    ]
 
 pEpochNo :: Parser EpochNo
 pEpochNo =
   EpochNo <$>
-    Opt.option Opt.auto
+    Opt.option (Opt.bounded "EPOCH")
       (  Opt.long "epoch"
-      <> Opt.metavar "NATURAL"
+      <> Opt.metavar "EPOCH"
```
Jimbo4350, line 1742: So I would argue to leave `NATURAL` so users know what kind of value they need to enter and can therefore infer the bounds. The help already says `"The epoch number."` so I don't see the additional value putting `"EPOCH"` as the metavar brings. I don't have a strong opinion but I'm interested to know why you made the change.

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
```diff
@@ -3049,9 +3048,9 @@ pPoolDeposit =
 pEpochBoundRetirement :: Parser EpochNo
 pEpochBoundRetirement =
     EpochNo <$>
-    Opt.option Opt.auto
+    Opt.option (Opt.bounded "EPOCH_BOUNDARY")
       (  Opt.long "pool-retirement-epoch-boundary"
-      <> Opt.metavar "INT"
```
Jimbo4350, line 3054: Same here

cardano-node/src/Cardano/Node/OptParse.hs
```diff
@@ -0,0 +1,18 @@
...
+
+module Cardano.Node.OptParse
+  ( bounded,
+  ) where
+
+import           Control.Monad (when)
+import           Options.Applicative (ReadM, eitherReader)
+import qualified Text.Read as Read
+
+bounded :: forall a. (Bounded a, Integral a, Show a) => String -> ReadM a
```
Jimbo4350, line 13: Since node and cli consumes this, we should expose it from `cardano-api`.

cardano-submit-api/src/Cardano/TxSubmit/CLI/OptParse.hs
```diff
@@ -0,0 +1,18 @@
...
+
+module Cardano.TxSubmit.CLI.OptParse
+  ( bounded,
+  ) where
+
+import           Control.Monad (when)
+import           Options.Applicative (ReadM, eitherReader)
+import qualified Text.Read as Read
+
+bounded :: forall a. (Bounded a, Integral a, Show a) => String -> ReadM a
```
Jimbo4350, line 13: DRY!

Jimbo4350 (dismissed):
Minor comments.

newhoggy:
`bounded` has been moved to `Cardano.Api.Utils`.

cardano-cli/src/Cardano/CLI/Shelley/Parsers.hs
newhoggy: I made the change because if there is a CLI parser error it will say:

```
option --invalid-before: EPOCH must not be less than 0
```

so I changed the `metavar` to match.  Let me know what you think.

Jimbo4350: Fair :+1:

cardano-api/src/Cardano/Api.hs
```diff
@@ -790,3 +793,4 @@ import           Cardano.Api.TxMetadata
 import           Cardano.Api.Utils
 import           Cardano.Api.Value
 import           Cardano.Api.ValueParser
+
```
Jimbo4350, line 796: Random diff?

Jimbo4350 (approved):
LGTM!

cardano-api/src/Cardano/Api.hs
newhoggy: Fixed!

PR #4920, opened 2023-02-27
Update bors.toml to only require hydra

Jimbo4350:

PR #4923, opened 2023-02-28
Remove ci/pr/required-check from bors.toml

newhoggy:


dermetfan:
This could be an alternative to #4920 but bors is not very useful if it checks only cabal files.

Jimbo4350:
> @newhoggy what about requiring a minimal cicero job that would only do the nix build of cardano-node for linux? I think that should avoid most of the failures of cicero build while still ensuring a minimal nix support for cardano-node master.

You can put a cicero job in CI if you like but we are not using it with bors. It's been pretty bad for the past few months merging things into cardano-node with cicero. We are going to use github actions. We can notify devops when other jobs fail but that should not encumber my team.

PR #4938, opened 2023-03-05
Limit concurrency of CI workflows 

newhoggy:
We only want to have one job per PR per workflow running at the same time. This will make all other jobs get cancelled (e.g. when restarting jobs after rebasing)

Based on https://github.com/input-output-hk/hydra/pull/760

### DIFF
--- a/.github/workflows/check-cabal-files.yml
+++ b/.github/workflows/check-cabal-files.yml
@@ -4,6 +4,11 @@ on:
   push:
   create:
 
+# Limit concurrent runs of this workflow within a single PR
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   check-cabal-files:
     runs-on: ubuntu-latest
--- a/.github/workflows/check-git-dependencies.yml
+++ b/.github/workflows/check-git-dependencies.yml
@@ -3,6 +3,11 @@ name: Check git dependencies
 on:
   push:
 
+# Limit concurrent runs of this workflow within a single PR
+concurrency:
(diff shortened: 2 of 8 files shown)


cardano-api/src/Cardano/Api/ProtocolParameters.hs
```diff
@@ -936,33 +935,33 @@ toShelleyPParamsUpdate
     , protocolUpdateTreasuryCut
     } =
     ShelleyPParams {
-      _minfeeA     = noInlineMaybeToStrictMaybe protocolUpdateTxFeePerByte
-    , _minfeeB     = noInlineMaybeToStrictMaybe protocolUpdateTxFeeFixed
-    , _maxBBSize   = noInlineMaybeToStrictMaybe protocolUpdateMaxBlockBodySize
-    , _maxTxSize   = noInlineMaybeToStrictMaybe protocolUpdateMaxTxSize
-    , _maxBHSize   = noInlineMaybeToStrictMaybe protocolUpdateMaxBlockHeaderSize
+      _minfeeA     = maybeToStrictMaybe protocolUpdateTxFeePerByte
```
lehins, line 938: Was this rename `noInlineMaybeToStrictMaybe`->`maybeToStrictMaybe` targeted for this PR?

Does it also mean that you were able to confirm that the bug mentioned in this https://github.com/input-output-hk/cardano-node/pull/4903#issuecomment-1454238434 was fixed?

newhoggy: Sorry, accidentally included this change in the PR.  The change is removed.

newhoggy:
Merged this because I'm working on a lot of CI related PRs and this will speed up that work.

PR #4941, opened 2023-03-05
Combinators for TxBodyContent and related types

newhoggy:
TxBodyContent has a function that constructs one of two default values for `BuiltTx` and `ViewTx`.

```haskell
foo :: TxBodyContent BuildTx BabbageEra
foo = defaultTxBodyContent InBuildTx
  & setTxValidityRange someValidityRange
  & setTxIns someTxIns
  & setTxOuts someTxOuts
  & setTxOuts someMoreTxOuts
```

This is an alternative to https://github.com/input-output-hk/cardano-node/pull/4458.

We don't seem to be able to agree on what the `Semigroup` instance should do out of a number alternatives and if there is confusion with choosing the behaviour, there will likely be confusion using the instance.

The idea behind this alternative is that we shouldn't arbitrarily choose a behaviour for the `Semigroup` instance because there is a high probability people who use our instances might assume it does something else.

This works with both `ViewTx` and `BuildTx`.


erikd (approved):

newhoggy:
Unfortunately I wanted this to work, but type inference fails because of use of GADTs:

```haskell
foo :: TxBodyContent BuildTx BabbageEra
foo = defaultValue { txValidityRange = someValidityRange }

bar :: TxBodyContent BuildTx BabbageEra
bar = defaultValue { txIns = someTxIns, txOuts = someTxOuts }

baz :: TxBodyContent BuildTx BabbageEra
baz = defaultValue { txOuts = someMoreTxOuts }
```

newhoggy:
Unfortunately, this doesn't work as well as I'd like it to.  As you add more fields, it increases the chance that GHC cannot infer the types:

Screenshot 2023-03-06 at 12 24 01 pm

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -1159,6 +1168,16 @@ data BuildTxWith build a where
      ViewTx      ::      BuildTxWith ViewTx  a
      BuildTxWith :: a -> BuildTxWith BuildTx a
 
+instance DefaultValue (BuildTxWith ViewTx a) where
+  defaultValue = ViewTx
+
+instance DefaultValue a => DefaultValue (BuildTxWith BuildTx a) where
+  defaultValue = BuildTxWith defaultValue
```
newhoggy, line 1175: There is no one function that can encapsulate both `DefaultValue (BuildTxWith ViewTx a)` and `DefaultValue (BuildTxWith BuildTx a)` because we use GADTs so we use `DefaultValue` type class.

ch1bo: But why would you need to? Isn't the purpose of the default value to simplify building, i.e. `BuildTx`? I personally don't see a use case for `ViewTx`, maybe you have one? If not, then a simple function `defaultTxBodyContent :: IsCardanoEra era => TxBodyContent BuildTx era` would suffice.

newhoggy: Please see https://github.com/input-output-hk/cardano-node/pull/4941/files#r1126097642 for where `ViewTx` is used.

cardano-api/src/Cardano/Api/DefaultValue.hs
```diff
@@ -0,0 +1,12 @@
+module Cardano.Api.DefaultValue
+  ( DefaultValue(..)
+  ) where
+
+class DefaultValue a where
+  defaultValue :: a
```
lehins, line 6: Why aren't you using [`Deafult`](https://hackage.haskell.org/package/data-default-class-0.1.2.0/docs/Data-Default-Class.html#t:Default) from `data-default-class`? We make use of it extensively in ledger. Seems totally redundant to introduce an exact custom copy of a known concept . It's like redefining `Semigroup` with a different name. Very strange.

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -2364,24 +2474,25 @@ makeTransactionBody =
...
+  :: forall era. IsCardanoEra era
+  => ShelleyBasedEra era
   -> TxScriptValidity era
   -> Ledger.TxBody (ShelleyLedgerEra era)
   -> TxBodyScriptData era
   -> Maybe (Ledger.AuxiliaryData (ShelleyLedgerEra era))
   -> TxBodyContent ViewTx era
 fromLedgerTxBody era scriptValidity body scriptdata mAux =
-    TxBodyContent
+    (defaultTxBodyContent @ViewTx @era)
```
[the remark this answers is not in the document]
ch1bo: Why do you need to use `defaultTxBodyContent` here at all? It's a full record specification, so no need to start with a default version, but the `TxBodyContent` should be just fine.

cardano-api/src/Cardano/Api/DefaultValue.hs
newhoggy: Now using the `Default` class from `data-default-class`.

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -1580,6 +1834,80 @@ data TxBodyContent build era =
...
+    , txExtraKeyWits = def
+    , txProtocolParams = def
+    , txWithdrawals = def
+    , txCertificates = def
+    , txUpdateProposal = def
+    , txMintValue = def
+    , txScriptValidity = def
+    }
+
+setTxIns :: TxIns build era -> TxBodyContent build era -> TxBodyContent build era
```
ch1bo, line 1861: I think you might want to restrict these "Builder functions" to the `BuildTx` case.

Also, we found `addXXX` be more useful on the lists/sets of things. It may very well be that you need to build the transaction step-by-step. Maybe that is something we can provide as a follow-up PR :)

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -1580,6 +1834,80 @@ data TxBodyContent build era =
      }
      deriving (Eq, Show)
 
+instance
+  ( Default (BuildTxWith build (Maybe ProtocolParameters))
+  , IsCardanoEra era
+  ) => Default (TxBodyContent build era) where
+  def = TxBodyContent
```
ch1bo, line 1841: I read an article in the past why `Default` (or rather `def`) is actually not so nice to use after all.. can't find it back, like a commenter on this thread: https://discourse.haskell.org/t/do-people-actually-use-data-default-data-default-default-type-class-in-production/3951

Please consider especially the point that you are writing a library here which already has like 400 dependencies. Adding one more (`data-default-class`), especially downstream to users is not helping .. at least re-export `def` if you decide to go the `Default` way.

To me, as a user, a single `defaultTxBodyContent` would be perfectly fine. Maybe even call it `emptyTxBodyContent` if you want. We don't need any of the other `def` values as the `TxMintNone` etc are good enough and don't introduce "blindness" as `Nothing`s or `def`s would do.

ch1bo (commented):
Left my 2 lovelaces as a user of this library.

cardano-api/src/Cardano/Api/TxBody.hs
lehins: > Default (or rather def) is actually not so nice to use after all

This is not a universally accepted view. It is a subjective opinion of some developers. However, there is nothing wrong with providing specialized functions, like `emptyTxBodyContent`, along with `Default` instances. This would give the user a choice of preference, precisely because not everyone like `Default`.

> Please consider especially the point that you are writing a library here which already has like 400 dependencies.

`data-default-class` is already a transitive dependency, so we are not adding any new dependencies in this PR

lehins: > "It's already a transitive dependency" is exactly how you get into 400 dependencies territory.

Sorry, I don't get it. How do you get to `n` dependencies by not adding a new dependency?

Jimbo4350: > less code to maintain

This is my main contention. I also think defining a `defaultTxBodyContent` value is sufficient. I'd rather not introduce more classes/types if it's not absolutely necessary.

cardano-api/src/Cardano/Api/Script.hs
```diff
@@ -113,6 +113,7 @@ module Cardano.Api.Script (
 import qualified Data.ByteString.Lazy as LBS
 import           Data.ByteString.Short (ShortByteString)
 import qualified Data.ByteString.Short as SBS
+import           Data.Default.Class
```
Jimbo4350, line 116: I don't understand what the benefit is of using `Data.Default.Class` over defining a single default value.

Jimbo4350 (changes requested):

cardano-api/src/Cardano/Api/TxBody.hs
Jimbo4350: > I think you might want to restrict these "Builder functions" to the BuildTx case.

:+1: 
> Also, we found addXXX be more useful on the lists/sets of things.

We should also add these functions as there are probably use cases for both.

cardano-api/src/Cardano/Api.hs
```diff
@@ -160,6 +160,9 @@ module Cardano.Api (
     createAndValidateTransactionBody,
     makeTransactionBody, -- TODO: Remove
     TxBodyContent(..),
+    defaultTxBodyContent,
+    defaultTxFee,
+    defaultTxValidityUpperBound,
     setTxIns,
```
Jimbo4350, line 167: We can potentially add a little heading e.g `-- ** Transaction body builders` or something like that

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -1325,6 +1327,16 @@ data BuildTxWith build a where
      ViewTx      ::      BuildTxWith ViewTx  a
      BuildTxWith :: a -> BuildTxWith BuildTx a
 
+defaultBuildTxWith :: ModeTx build -> a -> BuildTxWith build a
+defaultBuildTxWith inModeTx a =
+  case inModeTx of
+    InViewTx -> ViewTx
```
Jimbo4350, line 1337: We aren't interested in building a `ViewTx` so we can totally ignore it and do away with the `ModeTx a` type.

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -1801,6 +1805,12 @@ defaultTxBodyContent mtx = TxBodyContent
 setTxIns :: TxIns build era -> TxBodyContent build era -> TxBodyContent build era
 setTxIns v txBodyContent = txBodyContent { txIns = v }
 
+modTxIns :: (TxIns build era -> TxIns build era) -> TxBodyContent build era -> TxBodyContent build era
```
Jimbo4350, line 1808: :+1:

Jimbo4350 (dismissed):
Nice work :+1: . We definitely aren't interested in building `ViewTx` transactions, so you can remove support for that.

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -1304,6 +1331,20 @@ data BuildTxWith build a where
...
 
+defaultBuildTxWith :: ModeTx build -> a -> BuildTxWith build a
+defaultBuildTxWith inModeTx a =
+  case inModeTx of
+    InViewTx -> ViewTx
+    InBuildTx -> BuildTxWith a
+
+data ModeTx a where
+  InBuildTx :: ModeTx BuildTx
+  InViewTx :: ModeTx ViewTx
```
ch1bo, line 1342: Could: not mix definitions for ModeTx with InModeTx

ch1bo (approved):
I like it and it's very much what we currently do in the upstream cardano-api. So hydra will be using this 👍

cardano-api/src/Cardano/Api.hs
newhoggy: Done!

cardano-api/src/Cardano/Api.hs
```diff
@@ -160,6 +160,27 @@ module Cardano.Api (
     createAndValidateTransactionBody,
     makeTransactionBody, -- TODO: Remove
     TxBodyContent(..),
+    -- ** Transaction body builders
```
Jimbo4350, line 163: Missing a newline after `TxBodyContent`

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -1727,6 +1770,89 @@ data TxBodyContent build era =
...
+    , txScriptValidity = TxScriptValidityNone
+    }
+
+setTxIns :: TxIns build era -> TxBodyContent build era -> TxBodyContent build era
+setTxIns v txBodyContent = txBodyContent { txIns = v }
+
+modTxIns :: (TxIns build era -> TxIns build era) -> TxBodyContent build era -> TxBodyContent build era
+modTxIns f txBodyContent = txBodyContent { txIns = f (txIns txBodyContent) }
+
+addTxIns :: (TxIn, BuildTxWith build (Witness WitCtxTxIn era)) -> TxBodyContent build era -> TxBodyContent build era
```
Jimbo4350, line 1800: `addTxIn`?

Jimbo4350 (approved):
LGTM! Just a couple naming suggestions.

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -1727,6 +1770,89 @@ data TxBodyContent build era =
...
+setTxInsReference :: TxInsReference build era -> TxBodyContent build era -> TxBodyContent build era
+setTxInsReference v txBodyContent = txBodyContent { txInsReference = v }
+
+setTxOuts :: [TxOut CtxTx era] -> TxBodyContent build era -> TxBodyContent build era
+setTxOuts v txBodyContent = txBodyContent { txOuts = v }
+
+modTxOuts :: ([TxOut CtxTx era] -> [TxOut CtxTx era]) -> TxBodyContent build era -> TxBodyContent build era
+modTxOuts f txBodyContent = txBodyContent { txOuts = f (txOuts txBodyContent) }
+
+addTxOuts :: TxOut CtxTx era -> TxBodyContent build era -> TxBodyContent build era
```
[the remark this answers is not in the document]
newhoggy: Done!

cardano-api/src/Cardano/Api.hs
newhoggy: Done!

cardano-api/src/Cardano/Api/TxBody.hs
newhoggy: Done!

PR #4942, opened 2023-03-06
Create and merge_group triggers in Github Actions

newhoggy:
This will cause merge groups to trigger Github Actions workflows.  See https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#triggering-merge-group-checks-with-github-actions

### DIFF
--- a/.github/workflows/check-cabal-files.yml
+++ b/.github/workflows/check-cabal-files.yml
@@ -2,7 +2,7 @@ name: Check cabal files
 
 on:
   push:
-  create:
+  merge_group:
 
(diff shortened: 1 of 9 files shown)


erikd (approved):

angerman (approved):
I guess this looks good, as we don't need `create` with `push`?

Jimbo4350 (commented):
~Why are we removing the `create` trigger?~

Jimbo4350 (approved):

PR #4943, opened 2023-03-06
GHC 9.2.5 -> 9.2.7

angerman:


.github/workflows/haskell-linux.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ["8.10.7", "9.2.5"]
+        ghc: ["8.10.7", "9.2.7", "9.6.0.20230302"]
```
Jimbo4350, line 19: Are these experimental changes?

Jimbo4350 (changes requested):

.github/workflows/haskell-linux.yml
angerman: 9.2.7 not so much. That's just moving us to the latest (which still contains quite some noticeable improvements and bug fixes 😰)

9.6 is experimental (it's the 9.6.1-rc1) release; the idea is to get a better idea of how much pain it will be to upgrade.

angerman: The 9.6.1-rc1 is now in #4948

erikd (approved):

PR #4947, opened 2023-03-07
Add event.type to concurrency group

newhoggy:
This is necessary because our workflows can be triggered by more than one event type: `push` and `merge_group`.

When a PR is queued to early after a push, the `merge_group` job would cancel the `push` job or vice-versa causing unnecessary build failures.


erikd (approved):

newhoggy:
I rebased onto https://github.com/input-output-hk/cardano-node/pull/4939 because a flaky test aborted the merge.

PR #4955, opened 2023-03-08
Backward compatibility for Prometheus metrics names

jutaro:
With configuration entries
metricsComp = {
            "a" = "b";
            ...
        };
metrics with name a as well appear with name b in Prometheus. The names in the configuration has to be specified as in the original message, and not as in the Prometheus export (Where e.g. a '.' becomes an '_').

### DIFF
--- a/cardano-tracer/src/Cardano/Tracer/Handlers/Metrics/Prometheus.hs
+++ b/cardano-tracer/src/Cardano/Tracer/Handlers/Metrics/Prometheus.hs
@@ -103,7 +103,7 @@ runPrometheusServer tracerEnv (Endpoint host port) = forever $ do
       Just [nodeName] -> do
         liftIO (askNodeId tracerEnv $ decodeUtf8 nodeName) >>= \case
           Nothing -> writeText "No such a node!"
-          Just anId -> writeText =<< liftIO (getMetricsFromNode anId teAcceptedMetrics)
+          Just anId -> writeText =<< liftIO (getMetricsFromNode tracerEnv anId teAcceptedMetrics)
       _ -> writeText "No such a node!"
 
 type MetricName  = Text
@@ -111,10 +111,11 @@ type MetricValue = Text
 type MetricsList = [(MetricName, MetricValue)]
 
 getMetricsFromNode
-  :: NodeId
+  :: TracerEnv
+  -> NodeId
   -> AcceptedMetrics
   -> IO Text
-getMetricsFromNode nodeId acceptedMetrics =
+getMetricsFromNode tracerEnv nodeId acceptedMetrics =
   readTVarIO acceptedMetrics >>=
     (\case
         Nothing ->
@@ -124,7 +125,11 @@ getMetricsFromNode nodeId acceptedMetrics =
     ) . M.lookup nodeId
  where
   getListOfMetrics :: Sample -> MetricsList
-  getListOfMetrics = filter (not . T.null . fst) . map metricsWeNeed . HM.toList
+  getListOfMetrics =
+      metricsCompability
+      . filter (not . T.null . fst)
+      . map metricsWeNeed
+      . HM.toList
 
   metricsWeNeed (mName, mValue) =
     case mValue of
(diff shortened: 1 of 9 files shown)


deepfire (approved):
Jurgen, thank you!

nix/nixos/cardano-tracer-service.nix
```diff
@@ -41,6 +41,11 @@ let serviceConfigToJSON =
           epHost    = "127.0.0.1";
           epPort    = 3200; ## supervisord.portShiftPrometheus
         } // (cfg.prometheus or {});
+        # Just an example for metrics compatibility mapping
+        # metricsComp = {
+        #     "Mempool.TxsInMempool" = "Mempool.TxsInMempool.Mapped";
+        #     "ChainDB.SlotNum" = "ChainDB.SlotNum.Mapped";
```
deepfire, line 47: @jutaro, just one thing -- can we change names in this example to make it more clear -- what metric already exists, and what will be created?   If I didn't have the context, I'd be a bit confused by it.

Also, another question -- can we flip the order?
It's a bit weird that the created metrics are on the right of what reads like an assignment operator.

jutaro: What names to you think make it clearer. I thought adding the ".Mapped" prefix makes it clear.

jutaro: It is a mapping from current name to new name. I think this is pretty straightforward, and reversing it would not make it clearer in my view ?!?

cardano-tracer/src/Cardano/Tracer/Handlers/Metrics/Prometheus.hs
```diff
@@ -143,3 +148,14 @@ getMetricsFromNode nodeId acceptedMetrics =
     . T.replace " " "_"
     . T.replace "-" "_"
     . T.replace "." "_"
+
+  metricsCompability :: MetricsList -> MetricsList
+  metricsCompability metricsList =
```
mgmeier, line 153: did you mean "compatibility"?

cardano-tracer/src/Cardano/Tracer/Handlers/Metrics/Prometheus.hs
```diff
@@ -143,3 +148,14 @@ getMetricsFromNode nodeId acceptedMetrics =
...
+      case metricsComp (teConfig tracerEnv) of
+        Nothing -> metricsList
+        Just mmap -> foldl  (\ accu p'@(mn,mv) -> case M.lookup mn mmap of
+                                                Nothing -> p' : accu
+                                                Just rep -> p' : (rep,mv) : accu)
+                            []
+                            metricsList
```
mgmeier, line 160: Has `metricsList` to be ordered somehow (e.g. identical `mv`s must be subsequent)?
I was wondering about the `fold`...

nix/nixos/cardano-tracer-service.nix
mgmeier: It all hinges on how you read the assignment operator; it's meaning for ("X" = "Y") being 
> "X" has alias "Y"

right?
Without knowing how the resulting `Map Text Text` is applied in code, it's not straightforward; it would be nice to have that clarification in the comment, for someone just wanting to adjust the `.nix` service definition.

mgmeier (approved):

nix/nixos/cardano-tracer-service.nix
deepfire: There is a lot of mathematical baggage, where we define:
- something on the left
- to be equal to something on the right

..which, on the surface might feel like it's imperative PL bullshit, but it actuall runs deeper.

Can you think of a single definition where the "new thing" appears on the right?

deepfire: There are further reasons, which might be silly in this specific context -- but consider this:

- maps are asymmetric -- the thing on the left is unique, the thing on the right not so
- it makes sense to allow multiple aliases for the same source metric
- it doesn't make sense to have a single alias to be related to several source metrics

The status quo allows the strange thing and disallows the silly-but-meaningful thing.

cardano-tracer/src/Cardano/Tracer/Handlers/Metrics/Prometheus.hs
jutaro: Fixed the typo

cardano-tracer/src/Cardano/Tracer/Handlers/Metrics/Prometheus.hs
jutaro: Only one mapping per entry is supported, and then order doesn't matter.

PR #4956, opened 2023-03-08
Guard against overflows in Shelley TxIns

newhoggy:
`makeByronTransactionBody` guard against overflows in the transaction indices, but `makeShelleyTransactionBody` does not.

Add appropriate guards to `makeShelleyTransactionBody`.

Supersedes https://github.com/input-output-hk/cardano-node/pull/3478

### DIFF
--- a/cardano-api/src/Cardano/Api/TxBody.hs
+++ b/cardano-api/src/Cardano/Api/TxBody.hs
@@ -157,7 +157,7 @@ module Cardano.Api.TxBody (
   ) where
 
 import           Control.Applicative (some)
-import           Control.Monad (guard)
+import           Control.Monad (guard, unless)
 import           Data.Aeson (object, withObject, (.:), (.:?), (.=))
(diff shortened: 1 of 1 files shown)


erikd (approved):
Seems legit

Jimbo4350 (approved):
LGTM, but why not just change `TxIn` to `newtype TxIx = TxIx Word16`?

newhoggy:
> LGTM, but why not just change `TxIx` to `newtype TxIx = TxIx Word16`?

Shelley bounds `TxIx` to `Word16`:

https://github.com/input-output-hk/cardano-node/blob/master/cardano-api/src/Cardano/Api/TxBody.hs/#L2570

But Byron bounds `TxIx` to `Word32`:

https://github.com/input-output-hk/cardano-node/blob/master/cardano-api/src/Cardano/Api/TxBody.hs#L3288

I believe changing the type would mean a change in behaviour for Byron.

PR #4957, opened 2023-03-08
Fix merge group cancelling issue

newhoggy:
When more than one PR gets added the the merge queue, the earlier ones might cancel due to our concurrency policy.  This is to prevent that.

See https://github.com/orgs/community/discussions/46757?sort=top#discussioncomment-4980818

### DIFF
--- a/.github/workflows/check-cabal-files.yml
+++ b/.github/workflows/check-cabal-files.yml
@@ -4,9 +4,18 @@ on:
   push:
   merge_group:
 
-# Limit concurrent runs of this workflow within a single PR
+# When pushing branches (and/or updating PRs), we do want to cancel previous
+# build runs. We assume they are stale now; and do not want to spend CI time and
+# resources on continuing to continue those runs. This is what the concurrency.group
+# value lets us express. When using merge queues, we now have to consider
+# - runs triggers by commits per pull-request
+#   we want to cancel any previous run. So they should all get the same group (per PR)
+# - runs refs/heads/gh-readonly-queue/<target branch name> (they should all get their
+#   unique git ref, we don't want to cancel any of the ones in the queue)
+# - if it's neither, we fall back to the run_id (this is a unique number for each
+#   workflow run; it does not change if you "rerun" a job)
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.type }}-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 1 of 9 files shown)


erikd (approved):

angerman:
@newhoggy please make sure to keep notes about this in the knowledge base doc on merge queues. 

Under "common complications" with an _analysis_ and _solution_. This won't be the last time we encounter this 😅

.github/workflows/check-cabal-files.yml
```diff
@@ -6,7 +6,7 @@ on:
 
 # Limit concurrent runs of this workflow within a single PR
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.type }}-${{ github.event.pull_request.number || github.ref }}
+  group: ${{ github.workflow }}-${{ github.event.type }}-${{ startsWith(github.ref, 'refs/heads/gh-readonly-queue/') && github.ref || github.event.pull_request.number || github.run_id }}
```
Jimbo4350, line 9: A comment explaining why we are doing this would be really useful.

Jimbo4350 (changes requested):
Please add comments to the changes otherwise its difficult to discern why we have named the concurrency groups this way.

angerman:
The fact that we have to replicate this over so many files is a bit ... dissatisfying, it feels like this is going to easily lead to divergence? Do we need a GHA linter?

.github/workflows/check-cabal-files.yml
```diff
@@ -6,7 +6,7 @@ on:
 
 # Limit concurrent runs of this workflow within a single PR
```
angerman, line 7: Alright, here's my stab at trying to explain this
```suggestion
# When pushing branches (and/or updating PRs), we do want to cancel previous
# build runs. We assume they are stale now; and do not want to spend CI time and
# resources on continuing to continue those runs. This is what the concurrency.group
# value lets us express. When using merge queues, we now have to consider
# - runs triggers by commits per pull-request
#    we want to cancel any previous run. So they should all get the same group (per PR)
# - runs refs/heads/gh-readonly-queue/<target branch name> (they should all get their unique git ref, we don't want to cancel any of the ones in the queue)
# - if it's neither, we wall back to the run_id (this is a unique number for each workflow run; it does not change if you "rerun" a job)
```

angerman: @newhoggy what do you think? Might need to reference from the other files.

Jimbo4350 (approved):

angerman:
🙈 the replication everywhere. But let's fix that in a different Pr and get this merged to prevent unnecessary merge queue cancellations.

PR #4963, opened 2023-03-10
Fix flake setup

andreabedini:
* Remove cardano-ping from flake.nix since it does not exists anymore in network-mux. It is now part of cardano-cli (`cardano-cli ping`).
* Re-add ouroboros-consensus-cardano-tools to the project to provide db-analyser db-synthesizer, which are part of the flake.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -37,6 +37,9 @@ packages:
     trace-resources
     trace-forward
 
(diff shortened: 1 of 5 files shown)


angerman (approved):
LGTM

PR #4970, opened 2023-03-12
New cardano-api-gen package

newhoggy:
Might resolve https://github.com/input-output-hk/cardano-node/issues/4455


newhoggy:
See https://github.com/input-output-hk/cardano-node/issues/4455#issuecomment-1488787804

PR #4977, opened 2023-03-13
Upgrade to cabal-3.10.1.0

newhoggy:


angerman (approved):
LGTM, but why is the markdown linter failing?

PR #4979, opened 2023-03-13
cardano-testnet: Towards removing the TestnetRuntime data type

MarcFontaine:
The current scheme for tests is:
```
test1 = do
  testnetRuntime <- testnet ...
  test1Core runTime
test2 = do
  testnetRuntime <- testnet ...
  test2Core runTime
```
`testnet ` starts a cluster and returns a `TestnetRuntime` which is passed to the actual test.
Problems:
* `TestnetRuntime` is growing bigger and bigger.
* The same `TestnetRuntime` is passed to every test, even though each test uses only parts of the data.
* It is not easy to track which data is used where.
* There are cases where tests by pass `TestnetRuntime` and just read data from the file system.

This PR is WIP towards removing `TestnetRuntime`.
The idea is tests are called directly inside `testnet`.


cardano-testnet/src/Testnet/Cardano.hs
```diff
@@ -190,8 +195,15 @@ mkTopologyConfig numNodes allPorts port True = J.encode topologyP2P
...
-cardanoTestnet :: CardanoTestnetOptions -> H.Conf -> H.Integration TestnetRuntime
-cardanoTestnet testnetOptions configuration@H.Conf {tempAbsPath, testnetMagic} = do
+data RunTest
+  = StartTestnetOnly
+  | TestShutdownOnSlotSynced
+  | TestStakeSnapshot
+  | TestLeadershipSchedule
+  deriving (Show)
+
+cardanoTestnet :: RunTest -> CardanoTestnetOptions -> H.Conf -> H.Integration TestnetRuntime
```
Jimbo4350, line 205: The vision I have for this function it to simply be concerned with starting a testnet not specific tests. We're heading in the wrong direction here.

PR #4985, opened 2023-03-14
Upgrade to actions/checkout@v3

newhoggy:
There are no functional benefits we require from the new version, but the old version uses a deprecated version of node and will stop working at some point.


angerman (approved):
LGTM, could use a reasoning as to why we want to upgrade, and what the benefits are.

PR #4987, opened 2023-03-15
workbench: support Conway

deepfire:
1. Support Conway
2. Set `SnapshotFrequency` to a reasonable value.


newhoggy (approved):

mgmeier (approved):
Looks good, thanks!

PR #4988, opened 2023-03-15
Conway hard forks on prot-ver 9

newhoggy:
Resolves https://github.com/input-output-hk/cardano-node/issues/4990


mkoura:
Tested that it fixes https://github.com/input-output-hk/cardano-node/issues/4978 :+1:

Issue #4061, opened 2022-06-17
[BUG] - min required UTxO reported by CLI is not correct


mkoura:
*Internal*

**Area**
*Other*   Any other topic (Delegation, Ranking, ...).

**Summary**
In Babbage, both `transaction build` and `transaction calculate-min-required-utxo` report 857690 Lovelace to be minimum required UTxO for Lovelace-only output with no datum. However I am able to create UTxO with just 849070 Lovelace (using `transaction build-raw`).

**Expected behavior**
CLI reports precise minimum required UTxO.

**System info:**
- OS Name: Fedora 35
- Node version: 1.35.0-rc3

gitmachtl:
Is this on vasil-dev? doing my own calculation with the method used in alonzo-era i get 999920 lovelaces. using the cli inbuilt calculation method i get 961130 lovelaces. wondering where the 849070 are coming from. using also node 1.35.0-rc3.

gitmachtl:
Now that the transaction calculate-min-required-utxo this is a bit strange, because how to query the real minimum value with this function? In the past i used --tx-out ADDRESS+0 because of the word boundary it didn't make a difference in the outcome and the result was the minUTXO that must be used. But now that the calculation is done down to the bytelevel, the result of a minUTXO request depends on the amount you're doing the request for. Thats strange. Examples:

JaredCorduan:
The build command might be overestimating the size needed, perhaps using `maxBound` somewhere.

The min value depends (linearly) on the number of bytes of the output. You cannot therefore figure it out precisely by substituting zero for the value you want. See this CIP: https://github.com/cardano-foundation/CIPs/pull/265/files

gitmachtl:
> But I guess the suggestion is for the CLI to take an output with zero ada, and compute the fix point?

Yes exactly, like its now working in the alonzo-era. Otherwise external tools would need to do at least two or three requests with different lovelaces values to get the amount that really works for sending it out with that value.

Jimbo4350:
> In Babbage, both transaction build and transaction calculate-min-required-utxo report 857690 Lovelace to be minimum required UTxO for Lovelace-only output with no datum. However I am able to create UTxO with just 849070 Lovelace (using transaction build-raw).

So this is complaining about the change output. We purposely overestimate the size of the change output because we want to avoid recursively solving for the tx fee (i.e evaluate tx fee, put fee in tx, evaluate the fee again, put the fee in the body etc). This is expected.

[...]

PR #5013, opened 2023-03-23
Integrate latest ledger dependencies

lehins:
The goal of this PR is to integrate latest changes in the cardano-ledger, which accrued over the period of the last few months. Most notable breaking changes in ledger are:

* Addition of Conway era
* Implementation of versioned CBOR serialization 
* Switching to lenses for manipulating of PParams
* Changes to CostModel serialization
* Many more ...

Besides integrating ledger this PR also:
* Accounts for recent changes in consensus
* Changes representation of `CostModel`
* Fixes a bug where only metadata from TxAuxData was hashed upon transaction body creation with `createTransactionBody`
* Reduces a whole lot of duplication
* Moves all orphan instances for types defined in cardano-base and cardano-ledger
* Fixes issues with `calculateMinimumUTxO` funciton:
  * Proper fix for minutxo computation: #4061
  * Makes that function total


cardano-api/src/Cardano/Api/TxMetadata.hs
```diff
@@ -106,21 +106,18 @@ instance HasTypeProxy TxMetadata where
 
 instance SerialiseAsCBOR TxMetadata where
     serialiseToCBOR =
-          CBOR.serialize'
-        . (Shelley.Metadata :: Map Word64 Shelley.Metadatum -> Shelley.Metadata ())
-        -- The Shelley (Metadata era) is always polymorphic in era,
-        -- so we pick the unit type.
+          -- This is a workaround. There is a tiny chance that serialization could change
+          -- for Metadata in the future, depending on the era it is being used in. For now
+          -- we can pretend like it is the same for all eras starting with Shelley
```
newhoggy, line 111: How would we replace this work-around in the future?

cardano-api/src/Cardano/Api/ProtocolParameters.hs
```diff
@@ -779,21 +775,11 @@ fromAlonzoPrices Alonzo.Prices{Alonzo.prSteps, Alonzo.prMem} =
 -- Script cost models
 --
 
-newtype CostModel = CostModel (Map Text Integer)
+newtype CostModel = CostModel [Integer] --(Map Text Integer)
```
newhoggy: I think we can delete this comment from the history.

cardano-api/src/Cardano/Api/ProtocolParameters.hs
```diff
@@ -907,33 +893,34 @@ toLedgerProposedPPUpdates era =
   . Map.mapKeysMonotonic (\(GenesisKeyHash kh) -> kh)
   . Map.map (toLedgerPParamsUpdate era)
 
-
+-- TODO: Stop using partial function and switch to the Either String variant
```
newhoggy, line 894: Will this be done as part of this PR?

cardano-api/src/Cardano/Api/ProtocolParameters.hs
```diff
@@ -1401,262 +1252,184 @@ unbundleProtocolParams (BundleAsShelleyBasedProtocolParameters _ pp _) = pp
...
+                -> Either String (PParams (Ledger.AlonzoEra crypto))
+toAlonzoPParams
+    protocolParameters@ProtocolParameters {
+      protocolParamDecentralization
+    , protocolParamUTxOCostPerWord
+    } = do
+  ppAlonzoCommon <- toAlonzoCommonPParams protocolParameters
+  -- QUESTION? This is strange, why do we need to construct Alonzo Tx with Babbage PParams?
+  -- This feels to me like an issue with the api design, as there should never be such an
+  -- incosistency, because PParams affect the validity of the transaction.
```
newhoggy, line 1390: `s/incosistency/inconsistency`

Who is this question for?

cardano-api/src/Cardano/Api/ProtocolParameters.hs
```diff
@@ -1401,262 +1252,184 @@ unbundleProtocolParams (BundleAsShelleyBasedProtocolParameters _ pp _) = pp
...
+         -- in the api's 'ProtocolParameter' type. If we don't do this
+         -- we won't be able to construct an Alonzo tx using the Babbage
+         -- era's protocol parameter because our only other option is to
+         -- error.
+         Nothing -> Right minBound
+         Just dParam -> boundRationalEither "D" dParam
+  -- This is the correct implementation that should be the used instead:
+  -- d <- requireParam "protocolParamDecentralization"
+  --                   (boundRationalEither "D")
+  --                   protocolParamDecentralization
```
newhoggy, line 1403: Why is the correct implementation in a comment?

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
```diff
@@ -456,7 +456,7 @@ instance HasSeverityAnnotation (ConnectionManagerTrace addr (ConnectionHandlerTr
       TrShutdown                              -> Info
       TrConnectionExists {}                   -> Info
       TrForbiddenConnection {}                -> Info
-      TrImpossibleConnection {}               -> Info
+      -- TrImpossibleConnection {}               -> Info
```
newhoggy: Are these constructors gone?  Should we just delete these lines?

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
```diff
@@ -1901,11 +1901,11 @@ instance (Show addr, Show versionNumber, Show agreedOptions, ToObject addr,
...
-      TrImpossibleConnection connId ->
-        mconcat
-          [ "kind" .= String "ImpossibleConnection"
-          , "connectionId" .= toJSON connId
-          ]
+      -- TrImpossibleConnection connId ->
+      --   mconcat
+      --     [ "kind" .= String "ImpossibleConnection"
+      --     , "connectionId" .= toJSON connId
+      --     ]
```
newhoggy: Are these constructors gone?  Should we just delete these lines?

cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -1510,46 +1510,48 @@ nextEpochEligibleLeadershipSlots sbe sGen serCurrEpochState ptclState poolid (Vr
 
   -- Then we get the "mark" snapshot. This snapshot will be used for the next
   -- epoch's leadership schedule.
-  CurrentEpochState cEstate <- first LeaderErrDecodeProtocolEpochStateFailure
-                                 $ obtainDecodeEpochStateConstraints sbe
-                                 $ decodeCurrentEpochState serCurrEpochState
+  -- CurrentEpochState cEstate <- first LeaderErrDecodeProtocolEpochStateFailure
+  --                                $ obtainDecodeEpochStateConstraints sbe
+  --                                $ decodeCurrentEpochState serCurrEpochState
```
newhoggy, line 1515: What's the status of this commented out code?

cardano-api/ChangeLog.md
```diff
@@ -42,6 +42,8 @@
 
 - Fix minUTxO calculation in `calculateMinimumUTxO` function in `cardano-api`
 
+- **Breaking change** - `calculateMinimumUTxO` no longer fails, it is a total computation.
```
newhoggy, line 45: 👍

cardano-cli/test/data/golden/shelley/transaction-view.out
```diff
@@ -66,7 +66,7 @@ update proposal:
       pool influence: 82/83
       pool registration deposit: 83 Lovelace
       pool retirement epoch boundary: 84
-      protocol version: '85.86'
```
newhoggy, line 69: What is this change?  How did we go from `85` to `8`?

newhoggy:
LGTM

cardano-cli/test/data/golden/shelley/transaction-view.out
lehins: Very easy. Protocol version `85` is no longer possible to construct. Currently the highest possible major protocol version is `10`.
* `8` is current version in Babbage
* `9` will be `Conway`
* `10` is the version when upgrade `Conway`

We enforce this validity with `Version` type.

cardano-api/src/Cardano/Api/ProtocolParameters.hs
lehins: I suspect this will have much ramification in downstream uses, so I would suggest to do it in a follow up PR. However, if really necessary I could squeeze it in into this one

cardano-api/src/Cardano/Api/ProtocolParameters.hs
lehins: To whoever wrote this comment. Which seems like @Jimbo4350 in this commit: bbb8a5702eeef990462befdba85d52cc79602744
```
         -- The decentralization parameter is deprecated in Babbage
         -- so we default to 0 if no decentralization parameter is found
         -- in the api's 'ProtocolParameter' type. If we don't do this
         -- we won't be able to construct an Alonzo tx using the Babbage
         -- era's protocol parameter because our only other option is to
         -- error.
```

cardano-api/src/Cardano/Api/ProtocolParameters.hs
lehins: Because I would like to get the answer for the above question first. I don't want to break some behavior downstream that potentially relies on the incorrect implementation

cardano-api/src/Cardano/Api/Certificate.hs
```diff
@@ -401,28 +404,28 @@ fromShelleyPoolParams :: Shelley.PoolParams StandardCrypto
...
-    , Shelley._poolMD
+      Shelley.ppId
+    , Shelley.ppVrf
+    , Shelley.ppPledge
+    , Shelley.ppCost
+    , Shelley.ppMargin
+    , Shelley.ppRewardAcnt
+    , Shelley.ppOwners
+    , Shelley.ppRelays
+    , Shelley.ppMetadata
```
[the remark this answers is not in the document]
lehins: That makes two of us :grinning:

Jimbo4350 (commented):
Thank you @lehins for propagating the changes from ledger. I just have a couple comments
1. I think the `ProtocolParameter` related improvments that are not directly related to ledger changes should be in a separate PR.
2. I'd prefer to not use lenses if we don't have to. We're not updating the fields of deeply nested structures and I think the lenses further obfuscate the code vs record accessors which immediately give me information/context about the data structure I am updating.

There are many good refactors here :+1: . We should definitely organize a call next week so you can take us through the changes.

cardano-api/src/Cardano/Api/Genesis.hs
```diff
@@ -39,27 +42,26 @@ shelleyGenesisDefaults =
...
-        , Ledger._maxBHSize = 1100                  -- TODO: compute from crypto
-        , Ledger._maxBBSize = 64 * 1024             -- max 64kb blocks
-        , Ledger._maxTxSize = 16 * 1024             -- max 16kb txs
-        , Ledger._eMax      = 18
-        , Ledger._minfeeA   = 1                     -- The linear factor for the minimum fee calculation
-        , Ledger._minfeeB   = 0                     -- The constant factor for the minimum fee calculation
-        }
+        emptyPParams
+        & ppDL         .~ maxBound
+        & ppMaxBHSizeL .~ 1100                  -- TODO: compute from crypto
```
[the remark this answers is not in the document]
lehins: > I'd rather not use lenses unless we have to

Sorry to be a bearer of bad news, but we have to. PParams were redesigned in ledger and field accessors are one type deeper and are no longer part of the public api.

cardano-api/src/Cardano/Api/ProtocolParameters.hs
```diff
@@ -942,180 +929,148 @@ toShelleyPParamsUpdate
...
-      protocolUpdateProtocolVersion
-    , protocolUpdateDecentralization
+    } = do
+  a0 <- mapM (boundRationalEither "A0") protocolUpdatePoolPledgeInfluence
+  rho <- mapM (boundRationalEither "Rho") protocolUpdateMonetaryExpansion
+  tau <- mapM (boundRationalEither "Tau") protocolUpdateTreasuryCut
+  protVer <- mapM mkProtVer protocolUpdateProtocolVersion
+  let ppuCommon =
+        emptyPParamsUpdate
+        & ppuMinFeeAL     .~
```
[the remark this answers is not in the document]
lehins: Be it is at may, lenses allowed me to remove tons of duplication and reduce this module alone by 20%. 

In any case, we've switched to lenses for all of the type families in ledger in order to gain the ability to work on types across multiple eras. The plan is also to completely hide field accessors and constructors in the future version to reduce maintenance burden.

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -3963,28 +3869,24 @@ makeShelleyTransactionBody era@ShelleyBasedEraBabbage
...
+           & L.updateTxBodyL              .~ convTxUpdateProposal era txUpdateProposal
+           & L.reqSignerHashesTxBodyL     .~ convExtraKeyWitnesses txExtraKeyWits
+           & L.mintTxBodyL                .~ convMintValue txMintValue
+           & L.scriptIntegrityHashTxBodyL .~ convPParamsToScriptIntegrityHash
                                              era txProtocolParams redeemers datums languages
-           , Babbage.adHash = convAuxiliaryDataToHash txAuxData
-           , Babbage.txnetworkid = SNothing
-           })
+           -- TODO Babbage: support optional network id in TxBodyContent
+           -- & L.networkIdTxBodyL .~ SNothing
```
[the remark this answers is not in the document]
lehins: I am not its author. It was there before. I have no idea what it means and why NetworkId is not supported by cardano-api

cardano-api/src/Cardano/Api/Fees.hs
```diff
@@ -1330,20 +1330,20 @@ calculateMinimumUTxO era txout@(TxOut _ v _ _) bpp =
     ShelleyBasedEraAlonzo ->
       let lTxOut = toShelleyTxOutAny era txout
           babPParams = unbundleLedgerShelleyBasedProtocolParams era bpp
-          minUTxO = L.getMinCoinTxOut babPParams lTxOut
-          val = fromShelleyLovelace minUTxO
+          lTxOutWithMinCoin = L.setMinCoinTxOut babPParams lTxOut
```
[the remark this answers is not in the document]
lehins: It sure does :grin:

cardano-api/src/Cardano/Api/Genesis.hs
```diff
@@ -29,7 +32,7 @@ import           Ouroboros.Consensus.Shelley.Node (ShelleyGenesis (..), emptyGen
 -- * 'sgMaxLovelaceSupply' must be at least the sum of the 'sgInitialFunds'
 --   but more if you want to allow for rewards.
 --
-shelleyGenesisDefaults :: ShelleyGenesis crypto
+shelleyGenesisDefaults :: Crypto crypto => ShelleyGenesis crypto
```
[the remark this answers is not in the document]
lehins: Sounds good with me

lehins:
> I think the ProtocolParameter related improvements that are not directly related to ledger changes should be in a separate PR.

They are all related to the ledger changes, so they can't be extracted into a separate PR. PParams and PParamsUpdate have been completely redone in the latest version of ledger. Here is the relevant PR: https://github.com/input-output-hk/cardano-ledger/pull/3242

> I'd prefer to not use lenses if we don't have to.

We have to. Every single field in almost every ledger type family is now to be manipulated by a dedicated lens.

> We're not updating the fields of deeply nested structures and I think the lenses further obfuscate the code vs record accessors which immediately give me information/context about the data structure I am updating.

In the matter of fact we are. This is how you'd add inputs to  a transaction:
```haskell
> let addInputs tx ins = tx & bodyTxL . inputsTxBodyL <>~ ins
> :t addInputs
addInputs
  :: EraTx era => Tx era -> Set (TxIn (EraCrypto era)) -> Tx era
```
It is important to note that this function here **will work for all eras.** This is the power of lenses from ledger, not only that they allow to modify nested structure.

Moreover. Both `Tx` and `TxBody` are nested structures themselves for every era. We historically provided pattern synonyms with record fields, that allowed us to pretend like they are not nested structures. But his is a maintenance burden and it doe snot help us with avoiding duplication. Lenses do. For example, I'd love to see an alternative approach that will let you compute a total output of any transaction in any era:
```haskell
> let txTotalAda tx = fold $ fmap (\txOut -> txOut ^. coinTxOutL) (tx ^. bodyTxL . outputsTxBodyL)
> :t txTotalAda
txTotalAda :: EraTx era => Tx era -> Coin
```

Same applies to PParams, for example here is a lens that will let you set and get protocol version from PParams for any era:
```haskell
> :t ppProtocolVersionL 
ppProtocolVersionL :: EraPParams era => Lens' (PParams era) ProtVer
```

If a field was added at a later era then it will guarded by a specific type class that allows using the lens from specific era onwards, eg `ppCostModelsL` can be used in ALonzo, Babbage..., but not with PParams from eras prior to Alonzo:
```haskell
> :t ppCostModelsL 
ppCostModelsL
  :: AlonzoEraPParams era => Lens' (PParams era) CostModels
```

Before that we were using with PParams the hacky `getField @"_protocolVersion" pp` approach, but it was too ugly and limited in functionality. It only allowed for accessing, not modifying. And with respect to other types, like `Tx`, `TxBody`, `TxOut` we had no way to deal in an era agnostic fashion.

Long story short, lenses are the designated way to deal with ledger types that can change across eras.


> There are many good refactors here +1 . We should definitely organize a call next week so you can take us through the changes.

Thank you. Sounds good with me.

cardano-api/src/Cardano/Api/TxMetadata.hs
newhoggy: Cool.  Can you add that to the comment?

lehins: Sure. Will do.

newhoggy:
FWIW, I think lenses are way easier than type families provided the lens API is discoverable.

In your example, how easy is it to find all the accessors for the expression:`tx ^. bodyTxL`?

For example you have `outputsTxBodyL`.  What other fields are there?

cardano-api/src/Cardano/Api/ProtocolParameters.hs
newhoggy: Does that resolve your question @lehins ?

lehins:
> In your example, how easy is it to find all the accessors for the expression: `tx ^. bodyTxL`?
> For example you have `outputsTxBodyL`. What other fields are there?

Those are great questions! We have a proper `cardano-ledger-api` in the works. For now we export all of the lenses, but there will be a lot more exported later with proper documentation and examples. For instance here are all the lenses  that are available for `TxBody`: https://github.com/input-output-hk/cardano-ledger/blob/302e46f4d83e5ad8dafec68130ea5b3c07942f52/libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx/Body.hs#L12-L56

cardano-api/src/Cardano/Api/ProtocolParameters.hs
lehins: It sure does. Thank you. Removed the hacky implementation in favor of the correct one

cardano-api/src/Cardano/Api/ProtocolParameters.hs
```diff
@@ -755,13 +751,13 @@ instance FromJSON ExecutionUnitPrices where
         <*> o .: "priceMemory"
 
 
-toAlonzoPrices :: ExecutionUnitPrices -> Maybe Alonzo.Prices
+toAlonzoPrices :: ExecutionUnitPrices -> Either String Alonzo.Prices
```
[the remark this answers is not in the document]
lehins: I totally support this idea, however I suggest we do it as a separate follow up PR for two reasons:
* This PR is already pretty complex
* Right now those errors are local to this module only, but with changes that will expose those failures downstream,  as described in this comment: https://github.com/input-output-hk/cardano-node/pull/5013#discussion_r1154582800,  it would make even more sense to add types for each individual failure,

cardano-api/src/Cardano/Api/Genesis.hs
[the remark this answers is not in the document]
Jimbo4350: Can you expose the field accessors in the future?

cardano-api/src/Cardano/Api/ProtocolParameters.hs
lehins: Turns out removing the hack and making `d` parameter required breaks some tests. I've reverted the hack back and we can tackle it at a later PR

cardano-api/src/Cardano/Api/TxMetadata.hs
lehins: Done

cardano-api/src/Cardano/Api/Orphans.hs
```diff
@@ -1,695 +1,19 @@
...
-  ( Aeson.KeyValue a
-  , Crypto.Crypto crypto
-  ) => ShelleyLedger.PState crypto -> [a]
-toPStatePair pState =
-  let !pParams = Shelley._pParams pState
-      !fPParams = Shelley._fPParams pState
-      !retiring = Shelley._retiring pState
-  in  [ "pParams pState" .= pParams
-      , "fPParams pState" .= fPParams
-      , "retiring pState" .= retiring
```
lehins, line 496: List all changed fields in the changelog

cardano-api/src/Cardano/Api/TxMetadata.hs
```diff
@@ -106,21 +106,27 @@ instance HasTypeProxy TxMetadata where
...
+          --
+          -- Versioned cbor works only when we have protocol version available during
+          -- [de]serialization. The only two ways to fix this:
+          --
+          -- - Paramterize TxMetadata with era. This would allow us to get protocol version
+          --   from the type level
+          --
+          -- - Change SerialiseAsCBOR interface in such a way that it allows major
+          --   protocol version be supplied as an argument.
+          CBOR.serialize' CBOR.shelleyProtVer
```
lehins, line 121:
```suggestion
          CBOR.serialize' maxBound
```

cardano-api/src/Cardano/Api/Tx.hs
```diff
@@ -257,15 +246,20 @@ instance IsCardanoEra era => SerialiseAsCBOR (Tx era) where
...
-serialiseShelleyBasedTx :: ToCBOR tx => tx -> ByteString
-serialiseShelleyBasedTx = CBOR.serialize'
-
-deserialiseShelleyBasedTx :: FromCBOR (CBOR.Annotator tx)
-                          => (tx -> tx')
+serialiseShelleyBasedTx :: forall ledgerera .
+                           L.EraTx ledgerera
+                        => L.Tx ledgerera
+                        -> ByteString
+serialiseShelleyBasedTx = CBOR.serialize' (L.eraProtVerLow @ledgerera)
```
lehins, line 253: Potentially simplify
```suggestion
serialiseShelleyBasedTx = Plain.serialize'
```

cardano-api/src/Cardano/Api/Genesis.hs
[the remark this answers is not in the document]
lehins: Done

cardano-api/src/Cardano/Api/Orphans.hs
lehins: Done

cardano-api/src/Cardano/Api/TxBody.hs
```diff
@@ -2967,17 +2949,16 @@ fromLedgerTxReturnCollateral era txbody =
...
-    -> (( Ledger.TxOut (ShelleyLedgerEra era) ~ BabbageTxOut (ShelleyLedgerEra era)
-      , CC.Crypto (ShelleyLedgerEra era) ~ StandardCrypto
-      , BabbageEraTxBody (ShelleyLedgerEra era)
-      ) => a)
+    -> ((L.EraCrypto (ShelleyLedgerEra era) ~ StandardCrypto
+        , L.BabbageEraTxBody (ShelleyLedgerEra era)
+        ) => a)
     -> a
   obtainCollateralReturnHasFieldConstraint TxTotalAndReturnCollateralInBabbageEra f = f
   obtainCollateralReturnHasFieldConstraint TxTotalAndReturnCollateralInConwayEra f = f
```
lehins, line 2964: Remove HasField

Jimbo4350 (approved):
Thanks @lehins . I'm happy with the state of the PR. Once @newhoggy is also happy feel free to merge.

cardano-api/src/Cardano/Api/ProtocolParameters.hs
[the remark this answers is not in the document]
lehins: Create an issue for this https://github.com/input-output-hk/cardano-node/issues/5092

cardano-api/src/Cardano/Api/ProtocolParameters.hs
lehins: Created a ticket for this: https://github.com/input-output-hk/cardano-node/issues/5092

cardano-api/src/Cardano/Api/ProtocolParameters.hs
lehins: Turned this comment into an issue: https://github.com/input-output-hk/cardano-node/issues/5093

cardano-api/src/Cardano/Api/ProtocolParameters.hs
lehins: Here is the ticket about it: https://github.com/input-output-hk/cardano-node/issues/5093

newhoggy (approved):

deepfire (approved):

PR #5024, opened 2023-03-27
plutus ledger events tool

JaredCorduan:
I have added a new tool to the `cardano-client-demo` package. It uses `foldBlocks` to gather Plutus debugging information about scripts in the chain history. It was helpful recently while tracking down a ledger bug which caused a script in the chain history to fail, since it let us inspect the context when it ran successfully.

This is the second tool that I have added to `cardano-client-demo`. I think that we should consider making a new repository for `foldBlock` tools.

---

The help command describes tool like this:

```
$ cabal run plutus-events -- --help
Up to date
Usage: plutus-events (-c|--conf FILEPATH) (-s|--socket FILEPATH) 
                     [--slot NATURAL] [--hash SCRIPTHASH]

  This is a tool which provides debugging information for any Plutus script
  included in the chain history. The debugging information includes the script,
  the execution units, the script context, and the cost model. This tool
  requires a running node. The cardano-ledger repository provides a function
  'debugPlutus' in thecardano-ledger-alonzo package which will replay the script
  and provide the Plutus logs. The debugging information is first CBOR encoded
  and then base 64 encoded, but note that 'debugPlutus' expects the input to be
  base 64 encoded.

Available options:
  -c,--conf FILEPATH       node configuration file
  -s,--socket FILEPATH     node socket
  --slot NATURAL           Restrict to the given slot numbers. If not used, no
                           slot restrictions are imposed. This field can be
                           supplied multiple times.
  --hash SCRIPTHASH        Restrict to the given script hashes. If not used, no
                           script hash restrictions are imposed. This field can
                           be supplied multiple times.

```

### DIFF
--- a/cardano-client-demo/PlutusEvents.hs
+++ b/cardano-client-demo/PlutusEvents.hs
@@ -1,0 +1,147 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE GADTs        #-}
+
+import           Cardano.Api
+import           Cardano.Api.Shelley
+import           Cardano.Ledger.Alonzo.TxInfo (PlutusDebug(..), PlutusDebugLang(..))
+import           Cardano.Ledger.BaseTypes (ProtVer(..))
+import           Cardano.Ledger.Binary (serialize')
+import           Cardano.Ledger.Language (SLanguage(..))
+import           Control.Monad (when)
+import           Control.Monad.Trans.Except (runExceptT)
+import qualified Data.ByteString.Base64 as B64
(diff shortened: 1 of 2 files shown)


newhoggy:
Are these tools distributed with `cardano-node` & `cardano-cli`?  Or are they mostly for demo purposes?

JaredCorduan:
> This PR is stale because it has been open 45 days with no activity.

That's fair, github-action-bot, but I still think it's a good tool. Perhaps I'll rebase it soon here...

PR #5026, opened 2023-03-28
Remove dup of index-state, no longer needed with latest haskell.nix

jbgi:
as of https://github.com/input-output-hk/haskell.nix/pull/1903 haskell.nix (the nix part) does not parse the index-state so the duplication which was done for haskell.nix is no longer needed.

There is also a change to use hls from github (hence with access to the cabal.project) instead of hackage (which was breaking regularly), required due to https://github.com/input-output-hk/haskell.nix/pull/1909

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -12,8 +12,6 @@ repository cardano-haskell-packages
 
 -- See CONTRIBUTING for information about these, including some Nix commands
 -- you need to run if you change them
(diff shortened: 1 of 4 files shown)


newhoggy:
Conflicts.

deepfire (approved):
Thank you @jbgi !

PR #5036, opened 2023-03-30
Default Cardano.Testnet to NodeLoggingFormatAsJson

MarcFontaine:
Change `hprop_shutdownOnSlotSynced`  to parse JSON logs.

### DIFF
--- a/cardano-testnet/test/Test/ShutdownOnSlotSynced.hs
+++ b/cardano-testnet/test/Test/ShutdownOnSlotSynced.hs
@@ -1,6 +1,5 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TypeApplications #-}
 
 module Test.ShutdownOnSlotSynced
   ( hprop_shutdownOnSlotSynced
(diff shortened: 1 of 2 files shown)


cardano-testnet/test/Test/ShutdownOnSlotSynced.hs
```diff
@@ -57,15 +58,24 @@ hprop_shutdownOnSlotSynced = H.integrationRetryWorkspace 2 "shutdown-on-slot-syn
     H.cat (nodeStdout node)
     H.cat (nodeStderr node)
   mExitCodeRunning === Right ExitSuccess
+
   logs <- H.readFile (nodeStdout node)
-  slotTip <- case find (isInfixOf "Closed db with immutable tip") (reverse (lines logs)) of
```
newhoggy, line 61: This code is asserting that the node has (or will soon exit)

Is the new code asserting on the same thing?

cardano-testnet/test/Test/ShutdownOnSlotSynced.hs
```diff
@@ -57,15 +58,24 @@ hprop_shutdownOnSlotSynced = H.integrationRetryWorkspace 2 "shutdown-on-slot-syn
...
-    Nothing -> H.failMessage callStack "Could not find current tip in node's log."
```
newhoggy, line 62: The string `"Could not find current tip in node's log."` is probably a copy-paste error.  I think it should say `"Could not find close DB message"`

newhoggy:
There is a `TypeApplications` unused language pragma `hlint` warning.

cardano-testnet/test/Test/ShutdownOnSlotSynced.hs
MarcFontaine: I updated the PR to look for `TraceOpenEvent.ClosedDB` as well.

PR #5037, opened 2023-03-30
Bench nomad sub-backends

fmaste:
- The Nomad backend contains two local backends, `nomadpodman` using the podman task driver and `nomadexec` using the exec task driver, and one cloud backend `nomadcloud` using IOG infrastructure. This PR makes the distinction and usage between the three of them more clear.
- Created a new workbench subcommand (`wb nomad`) to reuse the `wb publish` code for fetching the credentials (`wb nomad vault`) and make sub-backend agnostic the code that manages the life-cycle of Nomad Agents (`wb nomad server|client start|stop`).
- Now when starting a new cloud deployments everything is automated


nix/workbench/backend/nomad/podman.sh
```diff
@@ -0,0 +1,275 @@
...
+      # nomadcloud  (IOG Nomad Agents and Amazon S3 with credentials from Vault)
+      echo 'nomadpodman'
+    ;;
+
+    # Generic backend sub-commands, shared code between Nomad sub-backends.
+
+    describe-run )
+      backend_nomad describe-run         "$@"
+    ;;
+
```
deepfire, line 25: @fmaste, you can simplify this passthrough very significantly -- you can see how the top-level `wb` does it, at the end of the file, when handling aliases.

nix/workbench/wb
```diff
@@ -294,6 +299,7 @@ main() {
         chaindb      | c )        chaindb             "$@";;
         genesis      | g )        genesis             "$@";;
         manifest     | m )        manifest            "$@";;
+        nomad        | n )        wb_nomad            "$@";;
```
deepfire, line 302: :joy:

nix/workbench/backend/nomad/podman.sh
fmaste: It was simpler, the way you recommend, but I used this form to debug and I kept it

PR #5042, opened 2023-03-31
Update network internals to use oops

newhoggy:
This PR is intentionally a minimal change with the intention of showing how `oops` is used without impacting the rest of the code base.

There is no breaking change as far as API exports is concerned.

In this PR, instead of implementing the networking code directly in `queryNodeLocalState` where it previously was, the networking code is implemented in `queryNodeLocalState_`.  The `queryNodeLocalState_` function uses `oops` style exception handling.  The `queryNodeLocalState` preserves its original type signature and merely delegates to `queryNodeLocalState_`.

The code is written this way to show an important point about `oops`.  The `oops` library is **un-opinionated**.  Unlike many effect systems and some other error handling libraries, `oops` was intentionally designed to interoperate with code that does not use `oops`.  The way it does this is to provide the combinators (in this case `runOopsInEither`) to wrap an `oops` enabled function into a function has has no `oops` influence in its type-signature.

This means that it is possible to write our internals in `oops` and have a public interface that has no `oops` exports whatsoever if that is what we want to do.  By choosing `oops` we are not committed to requiring or recommending our users use `oops`.

Whether or not we do that remains our decision to do now, later or never.

The PR also demonstrates another point, which is that adopting `oops` does not require us to significantly restructure our code and if we wanted to restructure our code because `oops` offers a better way to do it, then it can be done at our own discretion and on our own schedule.

I have worked with libraries before where this was not the case and it was a bad experience because small refactoring in one part of the code triggers a domino of refactoring until nearly all of the code-base use that library.  Some effect libraries 
have that bad quality.  `oops` is intentionally designed to not be like that.

This commit is PR is intentionally small do demonstrate this point.

This PR is the first commit of [a larger PR](https://github.com/input-output-hk/cardano-node/pull/4777) that rewrites much more of the code to use `oops`.  That PR is also broken into lots of commits, every one of which is a small change and compiles - again underscoring that introducing a small amount of `oops` does only just that and does not precipitate much more.

The reason I wanted to stress these important points is because I wanted to show that we can experimentally introduce `oops` into our codebase with minimal risk.  If at any point we find that `oops` is not as good as good an experience as we like, we can pause, think about why it is and fix or back out problematic changes.

It may be the case that `oops` represents to us an unfamiliar way to do error handling.  The hope is that _the only_ barrier to its adoption is a temporary non-familiarity that would go away with knowledge sharing and practise and that the flexibility and usability it affords us would sell itself as we use it.

### DIFF
--- a/cardano-api/src/Cardano/Api/IPC.hs
+++ b/cardano-api/src/Cardano/Api/IPC.hs
@@ -57,6 +57,7 @@ module Cardano.Api.IPC (
     QueryInEra(..),
     QueryInShelleyBasedEra(..),
     queryNodeLocalState,
+    queryNodeLocalState_,
 
     -- *** Local tx monitoring
     LocalTxMonitorClient(..),
@@ -80,18 +81,24 @@ module Cardano.Api.IPC (
     NodeToClientVersion(..),
 
     UnsupportedNtcVersionError(..),
+
+    -- ** Error types
+    AcquireFailure(..),
   ) where
-
-import           Data.Void (Void)
-
-import           Data.Aeson (ToJSON, object, toJSON, (.=))
-import qualified Data.ByteString.Lazy as LBS
-import qualified Data.Map.Strict as Map
 
 import           Control.Concurrent.STM (TMVar, atomically, newEmptyTMVarIO, putTMVar, takeTMVar,
                    tryPutTMVar)
 import           Control.Monad (void)
+import           Control.Monad.IO.Class
+import           Control.Monad.Oops (CouldBe, Variant, runOopsInEither)
+import qualified Control.Monad.Oops as OO
+import           Control.Monad.Trans.Except
 import           Control.Tracer (nullTracer)
+import           Data.Aeson (ToJSON, object, toJSON, (.=))
+import qualified Data.ByteString.Lazy as LBS
+import           Data.Function ((&))
+import qualified Data.Map.Strict as Map
+import           Data.Void (Void)
 
 import           Ouroboros.Consensus.Shelley.Ledger.SupportsProtocol ()
 import qualified Ouroboros.Network.Block as Net
@@ -586,17 +593,26 @@ queryNodeLocalState :: forall mode result.
                     -> Maybe ChainPoint
                     -> QueryInMode mode result
                     -> IO (Either AcquiringFailure result)
-queryNodeLocalState connctInfo mpoint query = do
-    resultVar <- newEmptyTMVarIO
-    connectToLocalNode
-      connctInfo
+queryNodeLocalState connctInfo mpoint query =
+  runOopsInEither $ queryNodeLocalState_ connctInfo mpoint query
+
+queryNodeLocalState_ :: forall e mode result. ()
+  => e `CouldBe` AcquiringFailure
+  => LocalNodeConnectInfo mode
+  -> Maybe ChainPoint
+  -> QueryInMode mode result
+  -> ExceptT (Variant e) IO result
+queryNodeLocalState_ connectInfo mpoint query = do
+    resultVar <- liftIO newEmptyTMVarIO
+    liftIO $ connectToLocalNode
+      connectInfo
       LocalNodeClientProtocols {
         localChainSyncClient    = NoLocalChainSyncClient,
         localStateQueryClient   = Just (singleQuery mpoint resultVar),
         localTxSubmissionClient = Nothing,
         localTxMonitoringClient = Nothing
       }
-    atomically (takeTMVar resultVar)
+    liftIO (atomically (takeTMVar resultVar)) & OO.onLeft OO.throw
   where
     singleQuery
       :: Maybe ChainPoint
--- a/cardano-api/src/Cardano/Api/IPC/Monad.hs
+++ b/cardano-api/src/Cardano/Api/IPC/Monad.hs
@@ -1,20 +1,29 @@
 {-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE GeneralisedNewtypeDeriving #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeOperators #-}
 
 module Cardano.Api.IPC.Monad
   ( LocalStateQueryExpr
   , executeLocalStateQueryExpr
+  , executeLocalStateQueryExpr_
   , queryExpr
   , determineEraExpr
+
+  , NodeToClientVersionOf (..)
   ) where
 
 import           Control.Concurrent.STM
 import           Control.Monad
 import           Control.Monad.IO.Class
+import           Control.Monad.Oops (CouldBe, Variant, runOopsInEither)
 import           Control.Monad.Reader
 import           Control.Monad.Trans.Cont
 import           Control.Monad.Trans.Except (ExceptT (..), runExceptT)
+import qualified Data.Variant as DV
 
 import           Cardano.Ledger.Shelley.Scripts ()
 import qualified Ouroboros.Network.Protocol.LocalStateQuery.Client as Net.Query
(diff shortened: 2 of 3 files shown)


cardano-api/src/Cardano/Api/IPC.hs
```diff
@@ -57,6 +57,7 @@ module Cardano.Api.IPC (
     QueryInEra(..),
     QueryInShelleyBasedEra(..),
     queryNodeLocalState,
+    queryNodeLocalState_,
```
newhoggy, line 60: We export `queryNodeLocalState_` so it can be used, but we do not use them yet.

All functions that end with `_` will by convention be support `oops` style error handling.  The function of the same name, but with the ending underscore is a non-`oops` wrapper that is kept for backwards compatibility or provide a way for developers to use our library without needing to deal with `oops`.

cardano-api/src/Cardano/Api/IPC.hs
```diff
@@ -586,17 +592,26 @@ queryNodeLocalState :: forall mode result.
                     -> Maybe ChainPoint
                     -> QueryInMode mode result
                     -> IO (Either AcquiringFailure result)
-queryNodeLocalState connctInfo mpoint query = do
-    resultVar <- newEmptyTMVarIO
-    connectToLocalNode
-      connctInfo
+queryNodeLocalState connctInfo mpoint query =
+  runOopsInEither $ queryNodeLocalState_ connctInfo mpoint query
```
newhoggy, line 596: `oops` provides combinators to make wrapping `oops` code into non-`oops` code easy.

The ease of conversion is two ways, but we only show one direction here.

Here the [`runOopsInEither`](https://hackage.haskell.org/package/oops-0.2.0.0/docs/Control-Monad-Oops.html#v:runOopsInEither) function accepts an `oops` function that can only throw one error and returns a new function that returns an `Either` with that error on the LHS.

cardano-api/src/Cardano/Api/IPC.hs
```diff
@@ -586,17 +592,26 @@ queryNodeLocalState :: forall mode result.
...
+
+queryNodeLocalState_ :: forall e mode result. ()
+  => e `CouldBe` AcquiringFailure
+  => LocalNodeConnectInfo mode
+  -> Maybe ChainPoint
+  -> QueryInMode mode result
+  -> ExceptT (Variant e) IO result
```
newhoggy, line 603: The type-signature of an `oops`-enabled function is very similar to the one that is not.  There are only two differences:

* The return value has `ExceptT (Variant e)` instead of `ExceptT AcquiringFailure`.
* The function has an additional constraint `e ``CouldBe`` AcquiringFailure`

The `Variant e` means the function could return `0` or more errors.  To find out what errors the function can return, look at the `CouldBe` constraints.  Each one of those constraint is an error the function can raise.

The `oops` style code has some advantages over the non-oops style code.  We no longer have to worry about the ordering of errors compared to code like `ExceptT x (ExceptT y ...` or `Either x (Either y ...)`.  We don't have worry that it isn't compatible with `ExceptT y (Except x ..` or `ExceptT y`.

This advantage does not play out here yet.  Only in a future commit.  Other than that, there is very little novel about this type signature.  It uses type application which Haskell developers should be familiar with and class constraints.  The only novelty is we are using here is the use of constraints to specified what errors are thrown.

It's important to note that specifying errors in constraints is **useful**.  That's because we want to not worry about the order of errors and constraints is exactly the Haskell feature that offers this.  The use of constraints here is not an example of gratuitous abstraction, it is exactly the abstraction we need.

cardano-api/src/Cardano/Api/IPC.hs
```diff
@@ -586,17 +592,26 @@ queryNodeLocalState :: forall mode result.
...
+    liftIO $ connectToLocalNode
+      connectInfo
       LocalNodeClientProtocols {
         localChainSyncClient    = NoLocalChainSyncClient,
         localStateQueryClient   = Just (singleQuery mpoint resultVar),
         localTxSubmissionClient = Nothing,
         localTxMonitoringClient = Nothing
       }
-    first toAcquiringFailure <$> atomically (takeTMVar resultVar)
+    liftIO (atomically (takeTMVar resultVar)) & OO.onLeft (OO.throw . toAcquiringFailure)
```
newhoggy, line 614: A note-worthy thing here is that aside from having to use `oops` functions to do error handling, the and the use of `liftIO`, which Haskell developers already ought to know, the implementation of the function is structurally the same.  `oops` does not force us to structure code in a particular way.

Moreover.  `oops` offers combinators that are already familiar.  `onLeft` already exists in `transformers-except` and we use it liberally.  `throw` is a familiar function in Haskell exception libraries.  `oops` tries to use familiar names for its combinators.

cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -52,44 +59,56 @@ executeLocalStateQueryExpr
   -> Maybe ChainPoint
   -> LocalStateQueryExpr (BlockInMode mode) ChainPoint (QueryInMode mode) () IO a
   -> IO (Either AcquiringFailure a)
-executeLocalStateQueryExpr connectInfo mpoint f = do
-  tmvResultLocalState <- newEmptyTMVarIO
+executeLocalStateQueryExpr connectInfo mpoint f =
+  runOopsInEither $ executeLocalStateQueryExpr_ connectInfo mpoint (lift f)
```
newhoggy, line 63: Again straightforward calling of `oops` code from non-`oops` code.

cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -52,44 +59,56 @@ executeLocalStateQueryExpr
...
+
+-- | Execute a local state query expression.
+executeLocalStateQueryExpr_
+  :: forall e mode a . ()
+  => e `CouldBe` AcquiringFailure
+  => LocalNodeConnectInfo mode
+  -> Maybe ChainPoint
+  -> ExceptT (Variant e) (LocalStateQueryExpr (BlockInMode mode) ChainPoint (QueryInMode mode) () IO) a
+  -> ExceptT (Variant e) IO a
```
newhoggy, line 72: This function is slightly more interesting that the previous one.

We declare that `e ` could be `AcquiringFailure`.  There is nothing preventing `e` from also being some other error.  The function body doesn't throw anything other than `AcquiringFailure`, however, the function `f` can.  Any errors that `f` throws will propagate to the caller of `executeLocalStateQueryExpr_` for free.  Non-`oops` code will struggle to propagate it without introducing a convoluted type signature.

How about the non-`oops` wrapper?  That wrapper returns `IO (Either AcquiringFailure a)` and the function `f` it accepts is `LocalStateQueryExpr (BlockInMode mode) ChainPoint (QueryInMode mode) () IO a`.  It has no capacity to propagate errors.  It can only return values.

cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -52,44 +59,56 @@ executeLocalStateQueryExpr
...
       { localChainSyncClient    = NoLocalChainSyncClient
-      , localStateQueryClient   = Just $ setupLocalStateQueryExpr waitResult mpoint tmvResultLocalState ntcVersion f
+      , localStateQueryClient   = Just $ setupLocalStateQueryExpr_ waitResult mpoint tmvResultLocalState ntcVersion f
       , localTxSubmissionClient = Nothing
       , localTxMonitoringClient = Nothing
       }
     )
 
-  first toAcquiringFailure <$> atomically waitResult
+  ExceptT . return =<< liftIO (atomically waitResult)
```
newhoggy, line 88: Again an important feature of `oops` is it does not require us to restructure our code.  The `oops` version is structurally very similar to the non-`oops` one.

cardano-api/src/Cardano/Api/IPC/Monad.hs
```diff
@@ -52,44 +59,56 @@ executeLocalStateQueryExpr
...
+    { Net.Query.recvMsgAcquired = runContT (runReaderT (runLocalStateQueryExpr (runExceptT f)) ntcVersion) $ \result -> do
+        atomically $ putTMVar resultVar' result
         void $ atomically waitDone -- Wait for all protocols to complete before exiting.
         pure $ Net.Query.SendMsgRelease $ pure $ Net.Query.SendMsgDone ()
 
     , Net.Query.recvMsgFailure = \failure -> do
-        atomically $ putTMVar resultVar' (Left failure)
+        atomically $ putTMVar resultVar' (Left (DV.throw (toAcquiringFailure failure)))
         void $ atomically waitDone -- Wait for all protocols to complete before exiting.
         pure $ Net.Query.SendMsgDone ()
```
newhoggy, line 113: This function is rewritten to `oops` without preserving a wrapper function because it is not exported and no-one calls it.  If we need one in the future, it is easy to write.

Jimbo4350 (changes requested):
Let's split the IPC modules in two, one with the existing implementation and one with `oops`.

newhoggy:
> Let's split the IPC module in two, one with the existing implementation and one with oops.

To clarify, there is only one implementation in this PR and it is written in `oops` and ends in an underscore.

The function that was the existing implementation was in `queryNodeLocalState`, but that is not a thing wrapper that delegates to the `oops` implementation `queryNodeLocalState_`.

The type signature of `queryNodeLocalState` is unchanged and has no `oops` types in it so all the existing code that doesn't know about `oops` can call it.

Do you mean `queryNodeLocalState_` function and `queryNodeLocalState` wrapper should go into different modules?

PR #5048, opened 2023-04-03
use TH-generated plutus scripts

NadiaYvette:
Using TH-generated scripts helps keep flake dependencies down & avoids having to go through a different team for changes to the scripts.


bench/plutus-scripts-bench/plutus-scripts-bench.cabal
```diff
@@ -76,6 +77,8 @@ library
   -- Non-IOG dependencies
   ------------------------
   build-depends:
+    , text
     , bytestring
     , serialise
     , template-haskell
+    , filepath
```
deepfire, line 84: We try to keep the package and module lists sorted.

bench/plutus-scripts-bench/src/Cardano/Benchmarking/PlutusScripts.hs
```diff
@@ -10,39 +10,40 @@ module Cardano.Benchmarking.PlutusScripts
...
-getAllScripts ::
-     [(String, ScriptInAnyLang)]
+getAllScripts :: [BenchScript]
 getAllScripts =
-  [ (normalizeModuleName CustomCall.scriptName, asAnyLang CustomCall.scriptSerialized)
-  , (normalizeModuleName ECDSA.scriptName     , asAnyLang ECDSA.scriptSerialized)
-  , (normalizeModuleName Loop.scriptName      , asAnyLang Loop.scriptSerialized)
-  , (normalizeModuleName Schnorr.scriptName   , asAnyLang Schnorr.scriptSerialized)
-  ]
+  [ CustomCall.script, ECDSA.script, Loop.script, Schnorr.script ]
```
deepfire, line 35: This will have quicker to perceive diffs, if we break this line to list-entry-per-line format:
```suggestion
  [ CustomCall.script
  , ECDSA.script
  , Loop.script
  , Schnorr.script
  ]
```

bench/plutus-scripts-bench/src/Cardano/Benchmarking/PlutusScripts.hs
```diff
@@ -10,39 +10,40 @@ module Cardano.Benchmarking.PlutusScripts
...
-  = fst <$> getAllScripts
+  = psName <$> getAllScripts
 
 findPlutusScript ::
      String
   -> Maybe ScriptInAnyLang
-findPlutusScript
-  = (`lookup` getAllScripts)
+findPlutusScript s
+  = listToMaybe [psScript t | t <- getAllScripts, last (split (=='.') . pack $ psName t) == pack (takeBaseName s)]
```
deepfire, line 46: I think this might be easier to read:

```suggestion
  = listToMaybe [ psScript t
                         | t <- getAllScripts
                               , last (split (=='.') . pack $ psName t) == pack (takeBaseName s) ]
```

deepfire: Also, further, `last (split (=='.') . pack $ psName t) == pack (takeBaseName s)` might be worth binding to a local function name -- to hint at intended meaning.

bench/plutus-scripts-bench/src/Cardano/Benchmarking/PlutusScripts/CustomCall.hs
```diff
@@ -7,24 +7,24 @@
 
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 
-module Cardano.Benchmarking.PlutusScripts.CustomCall
-  ( scriptName
-  , scriptSerialized
-  ) where
+module Cardano.Benchmarking.PlutusScripts.CustomCall (script) where
```
deepfire, line 10: :heart:

deepfire, line 10: :heart:

deepfire: The API change here is very nice.

bench/plutus-scripts-bench/src/Cardano/Benchmarking/ScriptAPI.hs
```diff
@@ -0,0 +1,17 @@
...
+  where
+
+import           Prelude as Haskell (String)
+import           Cardano.Api (ScriptInAnyLang)
+
+data BenchScript
+  = BenchScript
+    { psName   :: String
+    , psScript :: ScriptInAnyLang
+    }
```
deepfire, line 14: I apologise for not raising this matter sooner, since it'd save work, but I can't fail to mention this either.

The choice of name here is a bit confusing -- if taken with a bit of a wider context "bench script" can mean different things in benchmarking automation ecosystem.

A bit more specificity, like, perhaps, but not necessarily, `PlutusBenchScript` or `PlutusWorkloadScript` would be really useful.

bench/plutus-scripts-bench/src/Cardano/Benchmarking/ScriptAPI.hs
```diff
@@ -0,0 +1,17 @@
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+module Cardano.Benchmarking.ScriptAPI
+  (BenchScript, psName, psScript, mkBenchScript)
```
deepfire, line 4: Again, to reach for easier diff reading time later:

```suggestion
  ( BenchScript
  , psName
  , psScript
  , mkBenchScript
  )
```

bench/tx-generator/src/Cardano/TxGenerator/Setup/Plutus.hs
```diff
@@ -27,14 +27,19 @@ import           Cardano.Ledger.Alonzo.TxInfo (exBudgetToExUnits)
...
 
 
 type ProtocolVersion = (Int, Int)
 
 
-readPlutusScript :: FilePath -> IO (Either TxGenError ScriptInAnyLang)
-readPlutusScript fp
+readPlutusScript :: Either String FilePath -> IO (Either TxGenError ScriptInAnyLang)
+readPlutusScript (Left s)
+  = return
```
deepfire, line 39: Very minor, but `pure` in stead of `return` is more suggestive.

bench/tx-generator/src/Cardano/TxGenerator/Setup/Plutus.hs
```diff
@@ -27,14 +27,19 @@ import           Cardano.Ledger.Alonzo.TxInfo (exBudgetToExUnits)
...
+  $ maybe (Left . TxGenError $ "readPlutusScript: " ++ s ++ " not found.")
+            Right (findPlutusScript s)
```
deepfire, line 41: This might be more readable, since it:
1. lowers the arity of the functions used from 3 to 2.
2. puts the expected case ahead of the error case.

```suggestion
  $ (Right <$> findPlutusScript s)
     & fromMaybe (Left . TxGenError $ "readPlutusScript: " ++ s ++ " not found.")
```

nix/workbench/profile/prof1-variants.jq
```diff
@@ -276,7 +276,7 @@ def all_profile_variants:
    ({ generator:
       { plutus:
           { type:                       "LimitSaturationLoop"
-          , script:                     "v1/loop.plutus"
+          , script:                     { "Left": "Loop.hs" }
```
deepfire, line 279: I'd question whether the `.hs` suffix really belong in the script names.

We want the script name to be present in the reports -- and the report generator (and the rest of the workbench, by the way) mostly treats profile definitions as abstract, where all names are clean and usable directly.

If the report generator had to do the `.hs` suffix stripping, that'd be a pretty severe violation of the abstraction.

The same goes for the `Left`/`Right` wrappers -- they'd be unnatural to deal with.

So what I suggest -- since we're unlikely to use the Plutus scripts directly -- they all depend on the script-specific `tx-generator` knowledge to run them -- to just:

1. drop the `Left` wrappers from the profile definitions
2. unconditionally add the `Left` at the point where we pass the value to `tx-generator`

So it'd be a Nix-only change.

deepfire (approved):
@NadiaYvette  -- this is excellent work, thank you!

bench/plutus-scripts-bench/src/Cardano/Benchmarking/ScriptAPI.hs
```diff
@@ -18,3 +20,14 @@ data PlutusBenchScript
...
+prepareScriptName :: String -> String
+prepareScriptName
+  = head
+  . dropWhile (=="hs")
+  . map unpack
+  . reverse
+  . split (=='.')
+  . last
+  . split (=='/')
+  . pack
```
deepfire, line 33: https://hackage.haskell.org/package/filepath-1.4.100.3/docs/System-FilePath.html#v:dropExtension can simplify this a lot.

nix/workbench/profile/prof1-variants.jq
```diff
@@ -311,7 +311,7 @@ def all_profile_variants:
    ({ generator:
       { plutus:
           { type:                       "LimitTxPerBlock_8"
-          , script:                     { "Left": "SchnorrSecp256k1Loop.hs" }
+          , script:                     "SchnorrSecp256k1Loop"
```
deepfire, line 314: Nadial, :heart: !

mgmeier (approved):
Thank you very much @NadiaYvette, great work.

PR #5057, opened 2023-04-05
Define Functor instance for FileError

newhoggy:


carbolymer (approved):

cardano-api/src/Cardano/Api/Error.hs
```diff
@@ -1,5 +1,6 @@
 {-# LANGUAGE ExistentialQuantification #-}
 {-# LANGUAGE GADTSyntax #-}
+{-# LANGUAGE DeriveFunctor #-}
```
Jimbo4350, line 3: Why isn't your `stylish-haskell` not automatically sorting this?

Jimbo4350 (approved):

cardano-api/src/Cardano/Api/Error.hs
newhoggy: I temporarily disabled it to make a PR that changes `bench` code and forgot to turn it back on for this PR.

PR #5061, opened 2023-04-05
Remove --stake-address option from stake-address build

newhoggy:
The option `--stake-address` does not make sense for `stake-address build` command.


carbolymer (approved):

Jimbo4350 (approved):
Thanks :+1:

PR #5067, opened 2023-04-06
Add script and json file to generate man pages

happystaking:
The manpages.sh script will iterate through the array in the mandefinitions.json file. It will use pandoc to convert each input (an existing markdown document) into a man page. The generated man page will be gzipped and copied to the specified directory into the corresponding section.

After the man pages have been generated the script will trigger a mandb update.


scripts/manpages.sh
```diff
@@ -0,0 +1,103 @@
...
+#
+# Usage: ./manpages.sh [<target>]
+#
+# $1 - Where to put the generated man pages. The default is /usr/share/man and
+#      the man<#> subdirectories will be created by this script if needed.
+
+set -o pipefail
+
+if [ "$EUID" -ne 0 ]
+  then echo "Please run this script as root"
```
newhoggy, line 12: I have concerns about this script needing to be run as root.

What step in the script requires this?

happystaking: This was required in order to write the man pages to /usr/share/man and to run mandb.
However since you mentioned it, it's probably a good idea to enable the man pages to be set up in the user's home directory aswell. 
Today's commit enables this; so when running as a normal user it will setup the man pages only for this user in their home directory and when running as root the man pages will be set up system-wide for all users.

PR #5085, opened 2023-04-12
Expose SerialiseAsRawBytesError in Cardano.Api

ch1bo:
The function `deserialiseFromRawBytes` was changed from `Maybe` to `Either SerialiseAsRawBytesError`, but the error type was not exposed in the `Cardano.API`.


ch1bo:
This is even more important when defining `SerialiseAsRawBytes` instances. Without the data constructor, instances cannot create the `SerialiseAsRawBytesError` and will be partial by definition.

newhoggy (approved):

PR #5094, opened 2023-04-12
Workflow dispatch to take inputs

newhoggy:
Workflow dispatch needs to take inputs because the nightly trigger provides them.


newhoggy:
Force merging to test if this works.

newhoggy:
This was forced merged because it could only be tested on `master`.

Nightly builds are now confirmed to be working.

PR #5096, opened 2023-04-14
Expose LocalTxSubmissionClient data constructor

ch1bo:
Any user of the LocalNodeClientProtocols IPC interface will need to construct this type. Hence it needs to be exported.

### DIFF
--- a/cardano-api/ChangeLog.md
+++ b/cardano-api/ChangeLog.md
@@ -153,133 +153,136 @@
   ([PR 4021](https://github.com/input-output-hk/cardano-node/pull/4021))
 
 - SPO on-chain poll commands adjustments
-  [PR 5132](https://github.com/input-output-hk/cardano-node/pull/5132)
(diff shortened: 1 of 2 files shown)


carbolymer (approved):

newhoggy:
Also can you tell me more about your use case that you should need to use `LocalTxSubmissionClient` directly?

ch1bo:
> Also can you tell me more about your use case that you should need to use LocalTxSubmissionClient directly?

Of course. In https://github.com/input-output-hk/hydra we are connecting to a local cardano-node to submit transactions using the ouroboros node-to-client protocol. [Lately](https://github.com/input-output-hk/hydra/blob/6ae48d3cba6b709bdc1edcb90eadb847c4d388a1/hydra-node/src/Hydra/Chain/Direct.hs#L337) I tried to use the `IPC`module from `cardano-api`, instead of `ouroboros-network-framework` and protocols directly.

newhoggy (approved):

newhoggy:
I pushed a commit to fix the compile error.

PR #5097, opened 2023-04-14
workbench:  updates

deepfire:
- allow profiling mode control via `WB_PROFILING_MODE` make variable -- which defaults to `time`, can be set to `space`
- analysis: eliminate logfile prefiltering
- add `forge-stress-large` profiles, with three times the useful data
- more analysis fixes


bench/locli/src/Cardano/Command.hs
```diff
@@ -50,7 +50,7 @@ data ChainCommand
   |        ReadMetaGenesis  (JsonInputFile  RunPartial) (JsonInputFile  Genesis)
   |        WriteMetaGenesis TextOutputFile              TextOutputFile
 
-  |        Unlog            (JsonInputFile (RunLogs ())) Bool [LOAnyType]
+  |        Unlog            (JsonInputFile (RunLogs ())) Bool (Maybe [LOAnyType])
```
mgmeier, line 53: ❤️

bench/locli/src/Cardano/Unlog/LogObject.hs
```diff
@@ -99,11 +99,16 @@ readLogObjectStream f okDErr anyOks =
...
-    filter ((\case
+    filter ((case loAnyLimit of
+              Nothing -> \case
+                LOAny{} -> False
+                _       -> True
+              Just constraint -> \case
                 LOAny laty obj ->
-                  if elem laty anyOks then True else
+                  if elem laty constraint
+                  then True else
```
mgmeier, line 109: not just `elem laty anyOks || error (printf "...")` ? ;-)

mgmeier (approved):
Thank you, much needed.

bench/locli/src/Cardano/Unlog/LogObject.hs
deepfire: The idea is that having an explicit `True` sort of rhymes with other case branches.

PR #5103, opened 2023-04-15
Add two digesting datapoints.

mtullsen:
We add a concept of a "digesting" datapoint: its purpose is to transform a sequence of events (such as we get with "logging tracers") into a datapoint (with its stateless protocol) from which one can reconstruct the sequence of events on the datapoint client side.

Two new "digesting" datapoints are added to the tracing system.
  - ChainSync.Client  
  - BlockFetch.Client These datapoints turn the logging tracers of the same name into "digesting datapoints".

Issues:
- See PR #5104 for the implementation of a client for these digesting datapoints.  Due to the stateless nature of the datapoint mini-protocol, the client (/server) must receive (/send) a higher than needed number of packets in order to be able to not lose events. One possible extension to datapoints could alleviate this
  - The ability to configure the initialization of datapoints.  E.g.,
    ```
     off | on: Parameters -- I.e., Maybe DatapointParameter   
    ```

### DIFF
--- a/cardano-node/src/Cardano/Node/Tracing/Tracers.hs
+++ b/cardano-node/src/Cardano/Node/Tracing/Tracers.hs
@@ -8,13 +8,17 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
 
+{-# OPTIONS_GHC -Wno-orphans #-}
+
 module Cardano.Node.Tracing.Tracers
   ( mkDispatchTracers
   ) where
 
 import           Codec.CBOR.Read (DeserialiseFailure)
 import           Data.Proxy (Proxy (..))
+import           Data.Time.Clock (NominalDiffTime, UTCTime, addUTCTime, getCurrentTime)
 
+                 
 import           Cardano.Logging
 
 import           Cardano.Node.Tracing.Formatting ()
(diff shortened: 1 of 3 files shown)


deepfire (commented):
We need to benchmark the impact of this on the node, before we can merge.

Thank you @mtullsen !

cardano-node/src/Cardano/Node/Tracing/Tracers.hs
```diff
@@ -650,3 +666,75 @@ mkDiffusionTracersExtra configReflection trBase trForward mbTrEKG _trDataPoint t
...
+      now <- getCurrentTime
+      let cutoffTime = addUTCTime (negate period) now 
+      return $ (now,a)
+             : take (max'-1) (takeWhile ((> cutoffTime) . fst) as)
+
+mkDefaultDigestTracer :: Trace IO ([(UTCTime, a)]) -> IO (Trace IO a)
+mkDefaultDigestTracer = mkDigestTracer digestDefaultTime digestDefaultLen
+
+
+-- FIXME: turn these into configuration parameters.
```
jutaro, line 732: We can add this to the namespace based configuration, then it can be configured based on namespaces up to individual messages. However, then these digest tracers have to be integrated into the trace-dispatcher library. Currently we have the concept of frequency limiting. This seams to me somehow similar. But up till now its not working together with DataPoints.

jutaro (commented):
I think its better to talk about this first, before merging. I'm invited to a meeting tomorrow, this maybe an opportunity.

PR #5108, opened 2023-04-17
CIP-1694 make space for DRep certificates

newhoggy:
[CIP-1694](https://github.com/JaredCorduan/CIPs/blob/voltaire-v1/CIP-1694/README.md) introduces the idea of also delegating DReps in addition to stake pools.

The code currently use names that assume we can only delegate to stake pools from a stake-address.

This PR renames them to specifically refer to stake pools so that it is easier to add support for DReps later.

### DIFF
--- a/cardano-cli/src/Cardano/CLI/Run/Friendly.hs
+++ b/cardano-cli/src/Cardano/CLI/Run/Friendly.hs
@@ -315,7 +315,7 @@ friendlyCertificate =
       StakeAddressDeregistrationCertificate credential ->
         "stake address deregistration"
           .= object [friendlyStakeCredential credential]
-      StakeAddressDelegationCertificate credential poolId ->
+      StakeAddressPoolDelegationCertificate credential poolId ->
         "stake address delegation"
           .= object [friendlyStakeCredential credential, "pool" .= poolId]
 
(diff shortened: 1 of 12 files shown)


cardano-cli/src/Cardano/CLI/Shelley/Commands.hs
```diff
@@ -115,7 +116,7 @@ data StakeAddressCmd
   | StakeRegistrationCert StakeIdentifier OutputFile
   | StakeCredentialDelegationCert
       StakeIdentifier
-      (VerificationKeyOrHashOrFile StakePoolKey)
+      DelegationTarget
```
newhoggy, line 119: `DelegationTarget` is a wrapper type around `VerificationKeyOrHashOrFile StakePoolKey`.

cardano-cli/src/Cardano/CLI/Run/Friendly.hs
```diff
@@ -315,7 +315,7 @@ friendlyCertificate =
       StakeAddressDeregistrationCertificate credential ->
         "stake address deregistration"
           .= object [friendlyStakeCredential credential]
-      StakeAddressDelegationCertificate credential poolId ->
+      StakeAddressPoolDelegationCertificate credential poolId ->
```
newhoggy, line 318: Rename `StakeAddressDelegationCertificate` to `StakeAddressPoolDelegationCertificate` so that we can introduce `StakeAddressDRepDelegationCertificate` later.

erikd (approved):

Jimbo4350 (approved):
LGTM!

carbolymer (approved):

PR #5121, opened 2023-04-18
Disallow use of undefined function

newhoggy:


newhoggy:
This may not be needed.

PR #5128, opened 2023-04-19
Create defaultYamlValue for cardano-testnet

Jimbo4350:
Previously we were reading configuration files from disk and then modifying the configuration value. This is not ideal as one could not reasonably expect the on disk configuration to be equivalent to what the node's actual configuration is.


cardano-testnet/src/Testnet/Cardano.hs
```diff
@@ -27,14 +27,18 @@ import           Control.Monad
 import           Control.Monad.IO.Class (MonadIO, liftIO)
 import           Control.Monad.Trans.Except
 import qualified Data.Aeson as J
+import qualified Data.Aeson.Key as Key
+import           Data.Aeson.KeyMap (KeyMap)
+import qualified Data.Aeson.KeyMap as Aeson
 import qualified Data.ByteString as BS
-import           Data.ByteString.Lazy (ByteString)
+import qualified Data.ByteString.Lazy as LB
```
newhoggy, line 34: Checking over the repository.  There are examples of `LB` and `LBS` being used, but `LBS` is used more.  Perhaps we can standardise on that?

cardano-testnet/src/Testnet/Cardano.hs
```diff
@@ -207,64 +212,13 @@ cardanoTestnet testnetOptions H.Conf {..} = do
...
-          . HM.insert "TestAllegraHardForkAtEpoch" (J.toJSON @Int 0)
-          . HM.insert "TestMaryHardForkAtEpoch" (J.toJSON @Int 0)
-          . HM.insert "LastKnownBlockVersion-Major" (J.toJSON @Int 4)
-
-        Alonzo -> id
-          . HM.insert "TestShelleyHardForkAtEpoch" (J.toJSON @Int 0)
-          . HM.insert "TestAllegraHardForkAtEpoch" (J.toJSON @Int 0)
-          . HM.insert "TestMaryHardForkAtEpoch" (J.toJSON @Int 0)
-          . HM.insert "TestAlonzoHardForkAtEpoch" (J.toJSON @Int 0)
-          . HM.insert "LastKnownBlockVersion-Major" (J.toJSON @Int 6)
```
newhoggy, line 239: This code allowed us to choose which eras to hard fork to.  Are we dropping this support?

cardano-testnet/src/Testnet/Cardano.hs
```diff
@@ -700,15 +654,17 @@ cardanoTestnet testnetOptions H.Conf {..} = do
   H.assertIO . IO.doesFileExist $ tempAbsPath </> "tx1.tx"
 
   -- Add Byron, Shelley and Alonzo genesis hashes to node configuration
+  -- TODO: These genesis filepaths should not be hardcoded. Using the cli as a library
+  -- rather as an executable will allow us to get the genesis files paths in a more
+  -- direct fashion.
```
newhoggy, line 659: If we test using the library rather than the executable, it would mean that we aren't testing the CLI parsing code.  Is this definitely the direction we want to go?

If we do head down this way, there is something we need to watch out for.  Some CLI does error handling with `error` which doesn't give us a clear interface with which to catch errors and handle them.  We should probably be more rigorous with how we error in the command implementation and not use `error` if we go down this way so tests can handle those cases.

cardano-testnet/src/Testnet/Options.hs
```diff
@@ -30,3 +42,112 @@ defaultTestnetOptions = BabbageTestnetOptions
...
+    , ("TestMaryHardForkAtEpoch", Aeson.Number 0)
+    , ("TestAlonzoHardForkAtEpoch", Aeson.Number 0)
+    , ("TestEnableDevelopmentHardForkEras", Aeson.Bool True)
+    , ("EnableP2P", Aeson.Bool False)
+    , ("setupScribes", setupScribes)
+    , ("rotation", rotationObject)
+    , ("defaultScribes", defaultScribes)
+    , ("setupBackends", Aeson.Array $ Vector.fromList ["KatipBK"])
+    , ("defaultBackends", Aeson.Array $ Vector.fromList ["KatipBK"])
+    , ("options", Aeson.object mempty)
```
newhoggy, line 82: It might be useful to have comments here about where these values come from or why these particular values were chosen if that information is available.

cardano-testnet/src/Testnet/Util/Runtime.hs
```diff
@@ -144,22 +145,34 @@ startNode tempBaseAbsPath tempAbsPath logDir socketDir node nodeCmd = do
...
-               IO.cwd = Just tempBaseAbsPath})
-    =<<
-      H.procNode
-        (nodeCmd
-           <>
-             [ "--socket-path", IO.sprocketArgumentName sprocket
-             , "--port", portString
-             ])
+
+  createProcessNode <- H.procNode (nodeCmd <>
```
newhoggy, line 149: Suggest breaking after the `<-` like below and using `mconcat` instead of `<>`.

cardano-testnet/src/Testnet/Cardano.hs
Jimbo4350: Definitely, I'll switch to LBS here and follow up with a PR to change LB -> LBS

cardano-testnet/test/Test/Cli/KesPeriodInfo.hs
```diff
@@ -42,11 +42,9 @@ hprop_kes_period_info :: Property
 hprop_kes_period_info = H.integrationRetryWorkspace 2 "kes-period-info" $ \tempAbsBasePath' -> do
   H.note_ SYS.os
   base <- H.note =<< H.evalIO . IO.canonicalizePath =<< H.getProjectBase
-  configurationTemplate
-    <- H.noteShow $ base </> "configuration/defaults/byron-mainnet/configuration.yaml"
-
   conf@Conf { tempBaseAbsPath, tempAbsPath }
-    <- H.noteShowM $ mkConf (ProjectBase base) (YamlFilePath configurationTemplate)
+    -- TODO: Move yaml filepath specification into individual node options
+    <- H.noteShowM $ mkConf (ProjectBase base) (YamlFilePath "")
```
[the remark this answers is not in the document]
Jimbo4350: It means we are not reading the input yaml file. What do you have in mind instead?

cardano-testnet/src/Testnet/Cardano.hs
Jimbo4350: We aren't. I was going to do a follow up PR implementing a function that will take an era and produce the "correct" yaml configuration value. I can do it here as well however.

cardano-testnet/src/Testnet/Cardano.hs
Jimbo4350: > If we test using the library rather than the executable, it would mean that we aren't testing the CLI parsing code. Is this definitely the direction we want to go?

I think so. We can separately test individual commands to see if they parse correctly but its not needed to launch a testnet and should simplify things.

Jimbo4350: > Some CLI does error handling with error which doesn't give us a clear interface with which to catch errors and handle them. We should probably be more rigorous with how we error in the command implementation and not use error if we go down this way so tests can handle those cases.

Good point :+1:

cardano-testnet/test/Test/Cli/KesPeriodInfo.hs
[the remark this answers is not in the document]
newhoggy: Perhaps `Nothing` is more intuitive.

cardano-testnet/src/Testnet/Cardano.hs
newhoggy: Either that or add a TODO so we don't forget.

cardano-testnet/src/Testnet/Util/Runtime.hs
```diff
@@ -144,22 +145,34 @@ startNode tempBaseAbsPath tempAbsPath logDir socketDir node nodeCmd = do
...
+            , IO.std_err = IO.UseHandle hNodeStderr
+            , IO.cwd = Just tempBaseAbsPath
+            }
 
   H.noteShowM_ $ H.getPid hProcess
 
+  -- The node process can fail on startup, e.g while
+  -- parsing the configuration yaml file. If we don't fail
+  -- when stderr is populated, we end up having to wait for the
+  -- timeout to expire before we can see the error.
```
newhoggy, line 168: Maybe if we had a command or function that only parses the configuration file and we can call that before starting the node?

cardano-testnet/src/Testnet/Cardano.hs
```diff
@@ -207,64 +212,13 @@ cardanoTestnet testnetOptions H.Conf {..} = do
...
-          . HM.insert "TestAllegraHardForkAtEpoch" (J.toJSON @Int 0)
-          . HM.insert "TestMaryHardForkAtEpoch" (J.toJSON @Int 0)
-          . HM.insert "TestAlonzoHardForkAtEpoch" (J.toJSON @Int 0)
-          . HM.insert "LastKnownBlockVersion-Major" (J.toJSON @Int 6)
+  H.createDirectoryIfMissing_ logDir
+
 
   -- We're going to use really quick epochs (300 seconds), by using short slots 0.2s
   -- and K=10, but we'll keep long KES periods so we don't have to bother
   -- cycling KES keys
```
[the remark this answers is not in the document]
Jimbo4350: This comment exists by the shelley genesis creation below (which is where it should be). I'll delete this comment.

cardano-testnet/src/Testnet/Util/Runtime.hs
Jimbo4350: It seems wrong/not ideal to parse the configuration file twice. We should propagate the exceptions thrown by node.

cardano-testnet/src/Testnet/Conf.hs
```diff
@@ -30,16 +29,27 @@ data Conf = Conf
...
+mkConf :: ProjectBase -> Maybe YamlFilePath -> FilePath -> Maybe Int -> H.Integration Conf
+mkConf (ProjectBase base') mConfigTemplate tempAbsPath' maybeMagic = do
+  testnetMagic' <- H.noteShowIO $ maybe (IO.randomRIO (1000, 2000)) return maybeMagic
+  tempBaseAbsPath' <- H.noteShow $ FP.takeDirectory tempAbsPath'
+  tempRelPath' <- H.noteShow $ FP.makeRelative tempBaseAbsPath' tempAbsPath'
+  socketDir' <- H.createSubdirectoryIfMissing tempBaseAbsPath' $ tempRelPath' </> "socket"
+  logDir' <- H.createDirectoryIfMissing $ tempAbsPath' </> "logs"
+  let configTemplate = case  mConfigTemplate of
+                         Just (YamlFilePath yamlFilePath) -> Just yamlFilePath
+                         Nothing -> Nothing
```
newhoggy, line 45: Should be able to do `fmap` or `<$>` instead of pattern match.

cardano-testnet/src/Testnet/Options.hs
```diff
@@ -30,3 +42,142 @@ defaultTestnetOptions = BabbageTestnetOptions
...
+    [
+    -- TODO: Remnants from the Byron era that we can actually eliminate
+    -- and hardcode.
+      ("ApplicationName", "cardano-sl")
+    , ("ApplicationVersion", Aeson.Number 1)
+
+    -- ^ The consensus protocol to use
+    , ("Protocol", "Cardano")
+
+    -- ^ Socket path of the node
```
newhoggy, line 59: `-- ^` is haddock and it usually is a comment for the line above the comment, not the one below.

We should just `--` instead.

cardano-testnet/src/Testnet/Options.hs
```diff
@@ -30,3 +42,142 @@ defaultTestnetOptions = BabbageTestnetOptions
...
+    -- ^ Logging related
+    , ("setupScribes", setupScribes)
+    , ("rotation", rotationObject)
+    , ("defaultScribes", defaultScribes)
+    , ("setupBackends", Aeson.Array $ Vector.fromList ["KatipBK"])
+    , ("defaultBackends", Aeson.Array $ Vector.fromList ["KatipBK"])
+    , ("options", Aeson.object mempty)
+    ] ++ tracers
+ where
+  -- ^ Various tracers we can turn on or off
```
newhoggy, line 114: In this case use `-- |`

cardano-testnet/test/Test/ShutdownOnSlotSynced.hs
```diff
@@ -49,6 +48,8 @@ hprop_shutdownOnSlotSynced = H.integrationRetryWorkspace 2 "shutdown-on-slot-syn
         }
   TestnetRuntime { bftNodes = node:_ } <- testnet fastTestnetOptions conf
 
+  -- Read from stderr handle and fail if anything is present
```
newhoggy, line 51: Is this a TODO?  There's no code under here.

Jimbo4350: Left over comment

newhoggy (approved):

newhoggy (approved):

PR #5129, opened 2023-04-19
Bench nomad placement

fmaste:
Workbench: Placement of benchmark clusters nodes when using Nomad

- Include region data inside `node-specs.json`
- Check if the selected backend is compatible with the nodes' regions
- Add placement properties (`namespace`, `region`, `class`, `distinct`) to Nomad Job description
- Fixes to the download of nodes' files (add debug files)
- Show entrypoints' stdout and stderr when a Nomad Allocation fails
- Check that the Vault token is not expired
- Fix: use same naming strategy for supervisord programs and its directories and files
- Nomad backend small cleanups


nix/workbench/backend/nomad/cloud.nix
```diff
@@ -23,6 +23,19 @@ let
     ]
   ;
 
+  validateNodeSpecs = { nodeSpecsValue }:
+    let
+      # SRE is using these 3 Nomad "datacenters" (how they are called in Nomad)
+      datacenters = [ "eu-central-1" "eu-west-1" "us-east-2" ];
```
deepfire, line 29: @fmaste, we need SRE to enable support of the original locations -- which were originally picked by benchmarking:

https://github.com/input-output-hk/cardano-node/blob/master/bench/cardano-topology/cardano-topology.hs#L367-L369

You might need to work with @mgmeier to negotiate this extension with SRE.

PR #5132, opened 2023-04-20
SPO on-chain poll commands adjustments

KtorZ:
# Description

Follow-up from a conversation with @dcoutts, I proceeded with the following changes:

- [x] The option to prove with the VRF secret key has been removed, the only option now is to use the cold key
- [x] The witness is no longer included in the metadata themselves, but expected to be part of the standard transaction witness set (requires extra signatories on the submitting transaction)

I have also adjusted [the tutorial](https://hackmd.io/@KtorZ/cip-0094-tutorial) accordingly. 

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
  See [Runnings tests](https://github.com/input-output-hk/cardano-node/wiki/Running-tests) for more details
- [x] Code is linted with `hlint`.  See `.github/workflows/check-hlint.yml` for to get the `hlint` version
- [x] Code is formatted with `stylish-haskell`.  See `.github/workflows/stylish-haskell.yml` for to get the `stylish-haskell` version
- [x] Self-reviewed the diff

---

# History

- :round_pushpin: **Remove VRF signing option for producing SPO on-chain poll answers**
    And also, remove the poll witnesses altogether from metadata.
  Since we'll now only rely on cold key signing, we can rely on the
  existing witness mechanism of transactions (combined with the
  'required extra signatories' field).

  This slightly adjust the 'verifyPollAnswer' function to now work
  directly from a transaction and return the actual signatories.
  Consumers can then verify whether signatories meet their expected
  criteria.

- :round_pushpin: **Re-implement command-line handler for 'governance verify-poll'**
    As well as finalize the metadata extraction logic on the API's side.
  This is mostly the same thing as before, but it now takes a signed
  transaction file for verification instead of a standalone metadata.

  ```
  ❯ cardano-cli governance verify-poll --poll-file polls/basic.json --tx-file verify/valid
  Found valid poll answer, signed by:
  [
      "f8db28823f8ebd01a2d9e24efb2f0d18e387665770274513e370b5d5"
  ]
  ```

  ```
  ❯ cardano-cli -- governance verify-poll --poll-file polls/basic.json --tx-file verify/none
  Command failed: governance verify-poll  Error: No answer found in the provided transaction's metadata.
  ```

  ```
  ❯ cardano-cli -- governance verify-poll --poll-file polls/basic.json --tx-file verify/malformed
  Command failed: governance verify-poll  Error: Malformed metadata; couldn't deserialise answer: An error occured while decoding GovernancePollAnswer.2.
  Error: missing mandatory field
  ```

  ```
  ❯ cardano-cli -- governance verify-poll --poll-file polls/basic.json --tx-file verify/invalid
  Command failed: governance verify-poll  Error: Invalid answer (42) not part of the poll.
  Accepted answers:
  0 → yes
  1 → no
  ```

  ```
  ❯ cardano-cli -- governance verify-poll --poll-file polls/basic.json --tx-file verify/mismatch
  Command failed: governance verify-poll  Error: Answer's poll doesn't match provided poll (hash mismatch).
  ```

- :round_pushpin: **Use more meaningful (phantom) types for files around governance commands.**
  
- :round_pushpin: **Rework integration tests for governance commands**
    - [x] Cover new error cases for 'verify'
  - [x] Create full-blown test transactions for 'verify'
  - [x] Slightly re-organised and re-structure data folder
  - [x] Remove now-unnecessary old files

  ```
  ❯ tree cardano-cli/test/data/golden/shelley/governance
  .
  ├── answer
  │   └── basic.json
  ├── cold.sk
  ├── cold.vk
  ├── create
  │   ├── basic.json
  │   └── long-text.json
  ├── polls
  │   ├── basic.json
  │   └── long-text.json
  └── verify
      ├── invalid
      ├── malformed
      ├── mismatch
      ├── none
      └── valid
  ```


cardano-cli/test/Test/Golden/Shelley/Governance/VerifyPoll.hs
```diff
@@ -1,64 +1,96 @@
...
-golden_shelleyGovernanceVerifyPollCold :: Property
-golden_shelleyGovernanceVerifyPollCold = propertyOnce $ do
-  pollFile <- noteInputFile "test/data/golden/shelley/governance/poll.json"
-  metadataFile <- noteInputFile "test/data/golden/shelley/governance/answer-cold.json"
+  liftIO (readVerificationKeyOrTextEnvFile AsStakePoolKey vkFile) >>= \case
+    Left{} ->
+      H.failure
+    Right vk -> do
+      let expected = prettyPrintJSON $ serialiseToRawBytesHexText <$> [verificationKeyHash vk]
+      H.assert $ expected `BSC.isInfixOf` stdout
```
dcoutts, line 44: Yeah, better now that we can assert we're getting the right key.

dcoutts (approved):
This is great. Thanks @KtorZ, especially for helping us get this feature in on short notice.

A couple general comments:

I think it could be clearer that the final step of answer verification is that one needs to verify who is answering the poll, and that this step is manual. It's not enough to just run the command, it merely checks that someone answered the poll. I don't have a very strong opinion on where that should be clearer. The options I suppose are the cli output "and now do this", any readme/docs or in the CIP. Your original PR has a nice tutorial section but I didn't see that end up in any permanent docs. It'd be nice to capture that. That'd be a good place to call it out.

On a similar point, I don't see any way to verify poll questions. As an SPO I don't want to be tricked into signing some hash and index by anyone. I want to know the poll question came from the CF or whatever. I guess this is perhaps something to address in the CIP first rather than here in this PR. In an implementation perhaps it could be done automagically by checking the signer is a delegated governance key. Or at a minimum doing what the verify poll command does for answers and print out the vk hash of who proposed the poll question.

dcoutts:
> Your original PR has a nice tutorial section but I didn't see that end up in any permanent docs. It'd be nice to capture that.

Oops, missed the fact that you've got a tutorial at https://hackmd.io/@KtorZ/cip-0094-tutorial

KtorZ:
Regarding the verification of the poll answers, I am not _too worried_ about that because this a step mainly taken by power users (eg: the CF). I originally introduced this command because of the VRF option and the lack of general support for VRF in other ecosystems. Now that this solely rely on Ed25519 keys I don't think this verify command has much benefits and I don't think we'll even be using it when processing results (but simply replicate the same logic on scripts consuming that data). 

Having said that, it wouldn't hurt to have one or two extra lines in the cli output (or at minima in the tutorial) to highlight the expected verification steps. 

Regarding the verification of the question, that's a fair point. I expect in practice to be enough ceremony around questions with proper announcements, bells and whistle to make the authenticity of a survey sufficient obvious. We should however encourage people, in the tutorial at the very least, to check the signatories and and control that a poll was emitted from a transaction signed by a genesis delegate. This is something we can certainly provide from a CF perspective (genesis key hashes are publicly known anyway and part of the genesis configuration). I expect IOG to also provide adequate disclosure of their genesis verification key (hashes) should you decide to also conduct your own survey through this method.

cardano-api/src/Cardano/Api/Governance/Poll.hs
```diff
@@ -427,15 +312,10 @@ verifyPollAnswer poll answer witness = do
       , invalidAnswerAcceptableAnswers
       }
 
-  unless isValid $
-    Left ErrGovernancePollInvalidWitness
+  pure [ witVKeyHash wit | (ShelleyKeyWitness _ wit) <- getTxWitnesses tx ]
```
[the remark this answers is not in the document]
KtorZ: That's a good point. We could potentially omit normal unrelated payment keys this way. Ideally, I would print those hashes as bech32 `poolId` but we can't actually have any guarantee that these are indeed `poolId`. Yet, it would also make it clear that one has to check against the existing distribution of pools to assess their validity and non-existing pool ids would be discarded.

cardano-api/src/Cardano/Api/Governance/Poll.hs
```diff
@@ -412,10 +296,11 @@ renderGovernancePollError err =
...
 verifyPollAnswer
   :: GovernancePoll
-  -> GovernancePollAnswer
-  -> GovernancePollWitness
-  -> Either GovernancePollError ()
-verifyPollAnswer poll answer witness = do
+  -> InAnyCardanoEra Tx
+  -> Either GovernancePollError [KeyHash 'Witness StandardCrypto]
+verifyPollAnswer poll (InAnyCardanoEra _era tx) = do
+  answer <- extractPollAnswer (getTxBody tx)
```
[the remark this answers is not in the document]
KtorZ: But I love functions :D ! I honestly find it to read better than a nested pattern match.

cardano-cli/test/Test/Golden/Shelley/Governance/VerifyPoll.hs
```diff
@@ -1,64 +1,96 @@
...
+  txFile <- noteInputFile "test/data/golden/shelley/governance/verify/valid"
+  vkFile <- VerificationKeyFilePath . File <$>
+    noteInputFile "test/data/golden/shelley/governance/cold.vk"
 
-  void $ execCardanoCLI
+  stdout <- BSC.pack <$> execCardanoCLI
     [ "governance", "verify-poll"
     , "--poll-file", pollFile
-    , "--metadata-file", metadataFile
+    , "--tx-file", txFile
```
newhoggy, line 36: Should we be extra explicit and make this `--signed-tx-file`?

cardano-cli/test/Test/Golden/Shelley/Governance/VerifyPoll.hs
```diff
@@ -1,64 +1,96 @@
...
     ]
 
-golden_shelleyGovernanceVerifyPollCold :: Property
-golden_shelleyGovernanceVerifyPollCold = propertyOnce $ do
-  pollFile <- noteInputFile "test/data/golden/shelley/governance/poll.json"
-  metadataFile <- noteInputFile "test/data/golden/shelley/governance/answer-cold.json"
+  liftIO (readVerificationKeyOrTextEnvFile AsStakePoolKey vkFile) >>= \case
+    Left{} ->
+      H.failure
```
Jimbo4350, line 41: You should render the failure with `failWith Nothing . Text.unpack $ displayError e.`

cardano-cli/test/Test/Golden/Shelley/Governance/VerifyPoll.hs
```diff
@@ -1,64 +1,96 @@
...
     , "--poll-file", pollFile
-    , "--metadata-file", metadataFile
+    , "--tx-file", txFile
     ]
 
-golden_shelleyGovernanceVerifyPollVrfTempered :: Property
-golden_shelleyGovernanceVerifyPollVrfTempered = propertyOnce $ do
-  pollFile <- noteInputFile "test/data/golden/shelley/governance/poll.json"
-  metadataFile <- noteInputFile "test/data/golden/shelley/governance/answer-vrf-tempered.json"
+  either (const H.success) (const H.failure) result
```
Jimbo4350, line 57: Would be good to render the failure here too.

cardano-cli/test/Test/Golden/Shelley/Governance/VerifyPoll.hs
```diff
@@ -1,64 +1,96 @@
...
+  pollFile <- noteInputFile "test/data/golden/shelley/governance/polls/basic.json"
+  txFile <- noteInputFile "test/data/golden/shelley/governance/verify/none"
+
+  result <- tryExecCardanoCLI
+    [ "governance", "verify-poll"
+    , "--poll-file", pollFile
+    , "--tx-file", txFile
+    ]
+
+  either (const H.success) (const H.failure) result
```
Jimbo4350, line 70: And here

cardano-cli/test/Test/Golden/Shelley/Governance/VerifyPoll.hs
```diff
@@ -1,64 +1,96 @@
...
+  txFile <- noteInputFile "test/data/golden/shelley/governance/verify/invalid"
 
   result <- tryExecCardanoCLI
     [ "governance", "verify-poll"
     , "--poll-file", pollFile
-    , "--metadata-file", metadataFile
+    , "--tx-file", txFile
     ]
 
   either (const H.success) (const H.failure) result
```
Jimbo4350, line 96: And here

Jimbo4350 (commented):
Rendering the failures in the property tests will make it easier for us to debug failures if they ever arise.

cardano-api/src/Cardano/Api/Governance/Poll.hs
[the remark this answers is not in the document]
newhoggy: Should we report all witness but group the pool ones separately?

```
{ "pollWitnesses":
  { "pools": [ ... ]
  , "other": [..]
  }
}
```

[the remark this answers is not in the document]
KtorZ: > @Jimbo4350 You could use executeQueryCardanoMode with the QueryStakePools query to return the pool ids 

I'll push back on that one for three reasons:

1. It is slow.
    Few seconds to minutes, so this won't scale well when we need to verify thousands of answers in a row. 
    
2. It makes those commands no longer 'static'
    This introduces a dependency on a running cardano-node, which makes it particularly painful as go we go from a relatively easy-to-use static command even usable in an air-gapped environment, to something a lot heavier with a full node. 
    
3. Verification is done against a fixed snapshot  
    We need to verify pool ids against a known snapshot performed ahead of time. So even if some pool ids is registered _now_, doesn't make it valid at the moment the verification is performed. Snapshots are produced externally and thus, we can reliably do that verification in this step (unless we provide the full snapshot to the command but I am not looking into that. It's way easier for every ends as currently implemented).
    
> @newhoggy Should we report all witness but group the pool ones separately?

Not a huge fan. It requires an extra data structure which feels oddly arbitrary for that purpose. Plus, we really have no use of the "other" witnesses and we can't be sure that "pools" are indeed pools. So I think it's best to only go with @dcoutts' suggestion and (a) return only extra signatories, (b) fails when there's no extra signatories as we always necessarily expect at least one.

KtorZ:
@dcoutts @Jimbo4350 @newhoggy. Review feedback incorporated. 

- [x] `verifyPollAnswer` now only returns the list of extra signatories, and fails with a proper error when there's none.  (cc @dcoutts)

- [x] The command's output of `answer-poll` now also indicates that the transaction carrying the answer must be signed with a valid key and hint to the cold key. I've left the message relatively "flexible" so that we can easily introduce proxy keys later without having to change the command-line (cc @dcoutts)

- [x] Integration tests now use `failWith` & `displayError` when relevant. (cc @Jimbo4350)

- [x] Option to `verify-poll` renamed to `--signed-tx-file` (cc @newhoggy)

newhoggy:
I was cleaning up the golden tests and something came up whilst I was doing it.

I noticed the negative test case for `governance answer-poll` was option `3` which is curious because option `2` also is a failure case.

I then tried `0` which is a success case.

The behaviour is that the numbering is `0` based but the test suggests you were going for `1` based.

I wanted to ask to be sure the `0` base was actually the intention.

PR #5147, opened 2023-04-24
Remove dependency on plutus-example.

NadiaYvette:
This removes the  plutus-example dependency & flake input.


deepfire (approved):
Thank you @NadiaYvette !

PR #5154, opened 2023-04-25
Bench refactoring

fmaste:
Remove unused code and duplicate code, plus a variable rename that touches many files


deepfire (approved):
Thank you @fmaste

PR #5155, opened 2023-04-26
Workbench: make shell pure

andreabedini:
# Description

This change makes the workbench shell `pure`. Nix will do its best to strip off the current environment and will give the user a pristine shell, where the only environment variables are set by nix. This prevents unintentional mingling of system- and nix-provided libraries.

Tested with:

- make shell
- start-cluster
- wb a
- wb a std 2023-04-26-08-40-28344-default-bage-sup
- wb a compare 2023-04-26-08-40-28344-default-bage-sup 2023-04-26-08-40-28344-default-bage-sup

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ workbench-internals-walkthrough:
 ## Base targets:
 ##
 shell:                                           ## Nix shell, (workbench from /nix/store), vars: PROFILE, CMD, RUN
-	nix-shell -A 'workbench-shell' --max-jobs 8 --cores 0 --show-trace --argstr profileName ${PROFILE} --argstr backendName ${BACKEND} ${ARGS} ${if ${CMD},--command "${CMD}"} ${if ${RUN},--run "${RUN}"}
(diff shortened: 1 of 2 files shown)


mgmeier (approved):
Great, thanks @andreabedini

NadiaYvette:
> Great, thanks @andreabedini

This fixes the LD_LIBRARY_PATH build issue I was having. I see something unusual with ps not being found in my path when I try to run start-cluster, though I'm not sure if that's my end again or not.

angerman:
> > Great, thanks @andreabedini
> 
> This fixes the LD_LIBRARY_PATH build issue I was having. I see something unusual with ps not being found in my path when I try to run start-cluster, though I'm not sure if that's my end again or not.

This is probably related. `--pure` ensures that there is no leakage from the `env` prior to the `nix-shell` call into the shell. So it would ingore any `LD_LIBRARY_PATH` values, as well as `PATH` values.

Do check the `PATH` var _inside_ the shell, and check outside of the shell where `ps` lives (e.g. with `which ps`).

andreabedini:
Closing for now, we wil revisit another time.

PR #5156, opened 2023-04-26
Generate Plutus bytecode as backup.

NadiaYvette:
This creates some executables to serve as backups in case the next version bump has tension between ghc versions & the Plutus libs.


deepfire (approved):
Thank you Nadia, this is great.

PR #5163, opened 2023-04-27
fix `musl` build -- try: extra GHC options for `gen-plutus`

mgmeier:
Let's see if that's the reason for the musl failures, pt.2

### DIFF
--- a/bench/plutus-scripts-bench/plutus-scripts-bench.cabal
+++ b/bench/plutus-scripts-bench/plutus-scripts-bench.cabal
@@ -84,7 +84,8 @@ library
     , text
 
 executable gen-plutus
-  import:               project-config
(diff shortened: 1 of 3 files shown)


bench/plutus-scripts-bench/plutus-scripts-bench.cabal
```diff
@@ -84,7 +84,8 @@ library
     , text
 
 executable gen-plutus
-  import:               project-config
+  import:               common-definitions
+                      , project-config
```
angerman, line 88: I think we might want to consider this as a separate addition? @mgmeier @NadiaYvette 
The rest of this PR are part of #5160

angerman (changes requested):
See inline comments and #5160

PR #5159, opened 2023-04-27
Handle Nothing

angerman:
Sometimes we need to.

PR #5164, opened 2023-04-27
Give errmsg in lieu of irrefutable match

NadiaYvette:
Serge noted it would be better practice to handle the match failure & give an explicit errmsg.

This should be minor.


angerman:
This is the same as #5159, which is now also part of #5160

PR #5174, opened 2023-05-01
Update install.md

CarlosLopezDeLara:
Installation instruction update. Use latest libsodium  from https://github.com/input-output-hk/iohk-nix/commit/ceb181d109a427e22184928221b85f313d1594ab

### DIFF
--- a/doc/getting-started/install.md
+++ b/doc/getting-started/install.md
@@ -120,7 +120,7 @@ Download and install libsodium:
 ```bash
 git clone https://github.com/input-output-hk/libsodium
 cd libsodium
-git checkout 66f017f1
+git checkout a5baeccc
 ./autogen.sh
 ./configure
 make
@@ -140,25 +140,6 @@ the executable is linked with the right `libsodium.so` file (which you can
 check by running `ldd`), the running binary might still use the wrong library.
 You can check this by running `pldd`. If the `pldd` shows that the running executable
 is using the wrong library, run `ldconfig`.
-
-##### Using the ported `c` code
-
-In order to avoid having to install the custom version, `cardano-crypto-praos`
-defines a `cabal` flag that makes use of C code located
-[here](https://github.com/input-output-hk/cardano-base/tree/master/cardano-crypto-praos/cbits).
-To enable this code, one has to add the following code in the
-`cabal.project.local` file:
-
-```bash
-package cardano-crypto-praos
-  flags: -external-libsodium-vrf
-```
-
-For this to work, `libsodium` has to be in the system. For Ubuntu, this is achieved by:
-
-```bash
-sudo apt install libsodium-dev
-```
 
 #### Installing Secp256k1
 


disassembler (approved):

Jimbo4350 (approved):
:+1:

olgahryniuk (approved):

reqlez:
I spent some time today trying to figure this out, only for somebody to link me this PR. I guess I was correct that the new sodium fork should be used from the new "audited" branch. By the way there is another commit on there after this one, that says "for backwards compatibility" but I'm guessing that's not needed?

However, my question is around this:
https://github.com/input-output-hk/cardano-node/commit/797289144ad87d494603f19b301e6dbf93310f81

The official documentation, states that it's "okay" to use the ported cbits with the " package cardano-crypto-praos   flags: -external-libsodium-vrf " option with the release version of libsodium ( libsodium/stable v1.0.18 ).

Is this still correct information, to be used in production? Or should this be removed as a recommendation as part of this install.md update?

angerman:
@CarlosLopezDeLara I'm not sure. It was updated in https://github.com/input-output-hk/iohk-nix/pull/522. We now have the situation where the cardano-node 8 release notes reference what was merged in https://github.com/input-output-hk/iohk-nix/pull/520, yet, iohk-nix we use in our nix builds, has https://github.com/input-output-hk/iohk-nix/pull/522.

The difference is effectively only https://github.com/input-output-hk/libsodium/commit/dbb48cce5429cb6585c9034f002568964f1ce567 commit. Are we certain we don't need the backwards compatibility anymore?

angerman:
@iquerejeta maybe you got some input here :-)

CarlosLopezDeLara:
As mentioned in [Discord](https://discord.com/channels/826816523368005654/915340581411106916/1103006156579942420) We're updating the docs to remove mentioning the external VRF hack. Here's the official position regarding using it in production:

In summary, it's not known to be wrong, rather it's not known to be ok, and since this is critical to security, the cautionary principle says don't use it in production.
The libsodium library has certain supported and documented ways to build and use the library. In particular it involves building the whole library using libsodiums own build system. Turning off the external-libsodium-vrf flag (and thus using the bundled C code instead) does neither of these things: it only builds little bits of libsodium and does not use the libsodium build system. So from the perspective of the libsodium developers it is not a supported configuration. For that reason it's not something we can currently recommend to use in production

reqlez:
What about secp ? Is recommendation to still use unreleased version?  https://github.com/bitcoin-core/secp256k1/releases/tag/v0.3.1  release versions now include schnorrsig by default it seems. In fact I built 8.0.0 with v0.3.1 minting blocks on preview right now with it.  Also seen on github devs are using it also?

doc/getting-started/install.md
```diff
@@ -120,7 +120,7 @@ Download and install libsodium:
 ```bash
 git clone https://github.com/input-output-hk/libsodium
 cd libsodium
-git checkout 66f017f1
+git checkout a5baeccc
```
iquerejeta, line 123: We should use the latest commit used in iohk-nix 

https://github.com/input-output-hk/iohk-nix/blob/master/overlays/crypto/libsodium.nix#L10

iquerejeta (changes requested):

CarlosLopezDeLara:
closing in favor of https://github.com/input-output-hk/cardano-node/pull/5181

zhp1254:
> What about secp ? Is recommendation to still use unreleased version? https://github.com/bitcoin-core/secp256k1/releases/tag/v0.3.1 release versions now include schnorrsig by default it seems. In fact I built 8.0.0 with v0.3.1 minting blocks on preview right now with it. Also seen on github devs are using it also?

does cardano support secp256k1 transaction?

iquerejeta:
Hi @zhp1254 . Yes, secp256k1 transactions are supported. Here you can find an example https://github.com/input-output-hk/antaeus/blob/main/e2e-tests/test/Spec/Builtins/SECP256k1.hs

PR #5200, opened 2023-05-06
Split `cardano-cli` test files into separate directories

newhoggy:
# Description

This restructures tests components into their separate subfolders: `test/cardano-cli-golden` and `test/cardano-cli-test`.

Additionally moves `Test.OptParse` module to new `cardano-cli:test-lib` package component which lives in `test/cardano-cli-test-lib`.

The separation is useful because this means there is an easy way to run only the golden tests via `cabal test cardano-cli:cardano-cli-golden` command.

The introduction of `cardano-cli:test-lib` allows `Test.OptParse` to be used by `cardano-testnet` for its golden tests.  This is needed in another PR https://github.com/input-output-hk/cardano-node/pull/5254.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff

# Note on CI
for you.


cardano-cli/cardano-cli.cabal
```diff
@@ -174,28 +174,39 @@ executable cardano-cli
                       , optparse-applicative-fork
                       , transformers-except
 
+library test-lib
```
Jimbo4350, line 177: Why create a library for a single module?

cabal.project
```diff
@@ -13,7 +13,7 @@ repository cardano-haskell-packages
 -- See CONTRIBUTING for information about these, including some Nix commands
 -- you need to run if you change them
 index-state:
-  , hackage.haskell.org 2023-04-29T00:00:00Z
```
disassembler, line 16: normally, I wouldn't suggest this, but given we want to get 8.0.0 out and we don't want any issues coming up, I'ld like to hold this one back a couple days until we get the tag because of the hackage bump.

carbolymer (approved):

cabal.project
newhoggy: Instead of bumping the index state, I will duplicate the module I wanted to share.  This will be deduplicated again after the release tag.

cardano-cli/cardano-cli.cabal
newhoggy: So that the module can be shared by `cardano-cli-golden` and `cardano-cli-test`.  Currently that module is compiled twice.

Jimbo4350: What was the reason behind splitting the test suite into `cardano-cli-golden` and `cardano-cli-test`? Creating a library for the purpose of sharing a module seems like the wrong way forward. Why not have a single test suite instead?

newhoggy:
I've renamed `golden/files/golden` to `cardano-cli-golden/files/golden` to make it clearer what the the directory above files is.  The change has the added benefit that compile errors are more precise.  For example compile errors will now look like this:

```
cardano-cli-golden/Test/Golden/Key.hs:22:1: error:
    Parse error: module header, import declaration
    or top-level declaration expected.
   |
22 | xx
   | ^^
```

So it's more obvious this golden test that failed belongs to cardano-cli-golden and not some other package/component.  This also means in an IDE when clicking the file, it opens the correct file always.

cardano-cli/cardano-cli-golden/Test/Golden/Byron/SigningKeys.hs
```diff
@@ -25,21 +25,21 @@ import           Test.OptParse
 
 prop_deserialise_legacy_signing_Key :: Property
 prop_deserialise_legacy_signing_Key = propertyOnce $ do
-  legSkeyBs <- H.evalIO $ LB.readFile "test/data/golden/byron/keys/legacy.skey"
+  legSkeyBs <- H.evalIO $ LB.readFile "cardano-cli-golden/files/golden/byron/keys/legacy.skey"
```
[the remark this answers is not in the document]
newhoggy: The files for `cardano-cli-golden` and `cardano-cli-test` should not mingle.

If we do move `cardano-cli-golden` to `test/cardano-cli-golden` I would want to move what's currently in `test` to `test/cardano-cli-test`.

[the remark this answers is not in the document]
newhoggy: Done!

PR #5210, opened 2023-05-09
Bench meetup nomad

fmaste:
Many fixes for the workbench Nomad backend
- ctrl+c was not working as it should
- Update and fix the Nomad Podman backend
- A local run was not possible without a network interface with a default route
- Fail as soon as any job fails
- Better error reporting, including colors


mgmeier (approved):
Thank you @fmaste. LGTM, and test worked like a charm.

PR #5214, opened 2023-05-10
Add GHA to upload artifacts

hamishmack:
Automate generating release binaries for musl/mac/win on every tag to cardano-node repository (similar to how our docker containers work).

### DIFF
--- a/.github/workflows/release-upload.yaml
+++ b/.github/workflows/release-upload.yaml
@@ -1,0 +1,39 @@
+name: Post Release Upload
+
+on:
+  release:
+    types:
+      - published
+  push:
+    tags:
+      - '**'
+
+jobs:
+  upload:
+    name: "Upload Artifacts"
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - linux
+          - win64
+          - macos
+
+    runs-on: ${{ matrix.platform == 'macos' && 'macos-latest' || 'ubuntu-latest' }}
+    steps:
+      - name: Install Nix with good defaults
+        uses: cachix/install-nix-action@v20
+        with:
+          extra_nix_config: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release-upload.yaml
```diff
@@ -0,0 +1,39 @@
+name: Post Release Upload
```
angerman, line 1: We need a few lines of documentation at the top here explaining what the reason for this workflow is.
What do we want, why do we want it, how do we achieve it? All three questions should be answered.

angerman (approved):

hamishmack:
An example of workflow running is here https://github.com/input-output-hk/cardano-node/actions/runs/4933488682

.github/workflows/release-upload.yaml
```diff
@@ -0,0 +1,82 @@
...
+        uses: actions/download-artifact@v3
+        with:
+          name: cardano-node-macos
+      - name: Download linux
+        uses: actions/download-artifact@v3
+        with:
+          name: cardano-node-linux
+      - name: Release
+        uses: softprops/action-gh-release@v1
+        with:
```
jbgi, line 81: `draft = true`
```suggestion
        with:
          draft = true
```

jbgi (commented):
This looks good, but I don't understand why the release binaries don't appear on the release page : https://github.com/input-output-hk/cardano-node/releases/tag/hkm/test-tag-8

jbgi:
> This looks good, but I don't understand why the release binaries don't appear on the release page : https://github.com/input-output-hk/cardano-node/releases/tag/hkm/test-tag-8

never mind, it works but the release was deleted.

jbgi (approved):

jbgi:
Wondering though, do we need to upload the artifacts to the build (as double-zip, so not very practical)? Could we just upload to the GH release?

angerman:
> Wondering though, do we need to upload the artifacts to the build (as double-zip, so not very practical)? Could we just upload to the GH release?

I think we should do _just_ release.

I'll discuss a few security concerns with @disassembler tomorrow.
- do we trust GH runners? Should we have shared runners? 
- do we trust the release action? Should we fork it, audit it?
- do we want additional backups on S3?
- do we want to instantiate only on the runners and force realization from a trusted store?

jbgi (approved):
also need to remove unofficial nix cache.

jbgi:
>     * do we want to instantiate only on the runners and force realization from a trusted store?

we could use something like `nix --builders "" --max-jobs 0` to enforce that the binaries come from our trusted cache.

PR #5217, opened 2023-05-10
Replace remaining `FilePath` use in `cardano-api`

newhoggy:
# Description

After this change, `cardano-api` is using `File` types exclusively.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/cardano-api/src/Cardano/Api/LedgerState.hs
+++ b/cardano-api/src/Cardano/Api/LedgerState.hs
@@ -55,7 +55,7 @@ module Cardano.Api.LedgerState
   -- * Node Config
   , NodeConfig(..)
   -- ** Network Config
-  , NodeConfigFile (..)
+  , NodeConfigFile
   , readNodeConfig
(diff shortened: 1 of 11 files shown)


cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -77,6 +77,15 @@ module Cardano.Api.LedgerState
   -- ** Environment
   , Env(..)
   , genesisConfigToEnv
+
+  , ByronGenesisConfig
```
Jimbo4350, line 81: Let's move these types to `Cardano.Api.Genesis` and export them via `Cardano.Api` instead.

newhoggy: Done.  Some additional types that were defined in this module also had to move with them.

JaredCorduan (approved):

PR #5231, opened 2023-05-11
Use a c-stub for the placeholder.

angerman:
We do this because c (without `const`) can't know whether or not that string will be modified. Hence it will end up in the .data section not .rodata.


angerman:
note, this has not been tested yet. I didn't get around to test this yet due to time constraints.

cleverca22 (approved):

cleverca22:
tested and confirmed working

erikd (approved):
Needs to be squashed, but LGTM.

angerman:
Sigh, we don't have squashing on GH. Alright. Manual then.

PR #5239, opened 2023-05-12
workbench: add an AWS version of ci-bench for the Nomad cloud backend

fmaste:
- Added an `exit_code` file for every supervisord program because of supervisord limitations to fetch the exit codes
- Proper handling of supervisord programs startup failures, only waiting for the socket to appear does not catch program failures
- Keep checking that the `generator` has not exited with a non-zero exit code
- Add/rename specific `cardano-world-qa-*` shells because they are specific for cardano-world's Nomad Server and also add sensible resources defaults for their profiles
- Better output logs and error checking for genesis deployment and stdout and stderr files
- Fix for the workbench's cluster termination timeout that was not working properly
- Fix for the cluster run counter that was misleading due to time drift


mgmeier (approved):
Thank you @fmaste LGTM!

PR #5249, opened 2023-05-16
nix: default ghc to version 9.2.7

jbgi:


deepfire:
We'll need to be switching GHC versions from an external repository import (`cardano-ops`).

I'd rather suggest we parametrise the compiler here, instead of chaning the default:

https://github.com/input-output-hk/cardano-node/blob/master/nix/haskell.nix#L14-L25

PR #5259, opened 2023-05-18
Add zw3rk substituter/key to flake.nix

iburzynski:
# Description

Currently if users build cardano-node via `nix build` and accept the flake config, Nix will still build ghc-8.10.7 from source unless the zw3rk cache has previously been added to `nix.conf`, since it is no longer cached at `cache.iog.io`. Adding the zw3rk cache and its trusted public key to `nixConfig` in `flake.nix` will allow trusted users to fetch all dependencies from the binary caches and avoid adding additional entries to `nix.conf`. 

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -2,8 +2,8 @@
   description = "Cardano Node";
 
   nixConfig = {
-    extra-substituters = [ "https://cache.iog.io" ];
-    extra-trusted-public-keys = [ "hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ=" ];
+    extra-substituters = [ "https://cache.iog.io" "https://cache.zw3rk.com"];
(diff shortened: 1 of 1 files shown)


angerman:
I'm afraid this won't happen. It's not an official cache.

iburzynski:
Okay, in that case can you explain why IOG's fork of GHC 8.10.7 is not available at the official cache? Without the second cache the official one doesn't serve much purpose at all, since users still end up building GHC from source. This is a constant source of frustration.

angerman:
> Okay, in that case can you explain why IOG's fork of GHC 8.10.7 is not available at the official cache? Without the second cache the official one doesn't serve much purpose at all, since users still end up building GHC from source. This is a constant source of frustration.

That it's not is a defects, and we'll try to get this resolved asap.

iburzynski:
Maybe also worth noting that the zw3rk cache is already in the `plutus-apps` flake: https://github.com/input-output-hk/plutus-apps/blob/main/flake.nix

PR #5262, opened 2023-05-21
Add reforwarding option to cardano-tracer (in support of CNSA)

mtullsen:
# Description

We add a new reforwarding service in cardano-tracer.  It creates a new socket in which it serves the three tracing miniprotocols.  Currently,
 - reforwards trace messages to the new socket, optionally filtering trace messages. (via the config file)
 - it does not reforward EKG to the new socket.
 - it serves the datapoint protocol, but with an empty datapoint store.

PR #5263, opened 2023-05-21
Galois/cardano tracer cnsasink

mtullsen:
# Description

Adds a simple proof-of-concept CNSA analysis to cardano-tracer.

- Creates a datapoint for consolidating all sampling nodes: captures last completed blockfetch time for each of the sampling nodes.

This serves as a POC for review, not intended to be merged.

This should be rewritten as a new executable (not in iohk repo) that uses cardano-tracer as a library.


mgmeier:
Thanks for demonstrating the use case with this POC.
All relevant commits that enable re-forwarding traces to custom application logic built with `cardano-tracer` (the library) are contained in #5262.

PR #5265, opened 2023-05-21
Fix `cardano-testnet` CLI golden tests

newhoggy:
# Description

The tests were actually previously testing the `cardano-cli` which is already tested.

This PR:

* changes the tests to test `cardano-testnet` CLI instead.
* adds the `help` command to `cardano-testnet` which is used by the CLI golden tests to enumerate all the commands that need to be tested.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff

# Note on CI
for you.


carbolymer (approved):

Jimbo4350 (approved):

cardano-testnet/src/Testnet/Parsers.hs
```diff
@@ -29,19 +33,24 @@ data CardanoTestnetCommands
   | StartCardanoTestnet CardanoOptions
   | StartShelleyTestnet ShelleyOptions
   | GetVersion VersionOptions
+  | forall a. Help ParserPrefs (ParserInfo a) HelpOptions
```
Jimbo4350, line 36: It's not obvious to me what we gain by making this an existential type. I think this is premature abstraction.

newhoggy: Done!

PR #5267, opened 2023-05-22
tx-generator documentation

NadiaYvette:
This documentation meant to cover the important types and API functions and modules of the transaction generator. Some familiarisation with haddock was needed to get started and the Action type was chosen as a starting point on the basis of having had prior experience working with it. Ultimately, this is meant to contribute to an exportable API so other things can generate transaction streams for their own use e.g. analyses involving ledgers, direct examination of or analysis of streams of generated transactions etc. Feedback on the content or style of the documentation can also be taken onboard to revise it.


NadiaYvette:
The documentation updates got folded into a different branch and I forgot this thing was here. It should be deleted, and its branch, too.

PR #5285, opened 2023-05-30
Bench nomad healthchecks

fmaste:
Workbench enhancement tailored for cloud/Nomad deployments
 
- Add a healthcheck service that test nodes' connections and monitors runs
- Deploy the generator only inside the Nomad Task that will run it
- Make in which node the generators runs a parameter
- Use the same name for all nodes' db folders
- Fixes for the `exit_code` file of the supervisord programs
- Run supervisord unbuffered and do not rotate log files
- Misc fixes, mostly for cloud runs with an `explorer` node that was not previously used in the workbench

The healthcheck is backend agnostic, uses the same code for the `supervisord` backend and the `nomad` backend. Added as an extra supervisord "program" and controlled using `supervisorctl` (workbench clusters use the use the "supervisord" model).
The initial idea was to use Nomad's builtin `service -> check` functionality but it won't be 100% compatible/interchangeable with local runs using the `supervisord` backend and critical business logic, like when to start it/how to control it, will be delegated to Nomad. Plus, Nomad needs `consul` to configure "check"s and that means an extra dependency for local runs.

### DIFF
--- a/nix/workbench/backend/supervisor.sh
+++ b/nix/workbench/backend/supervisor.sh
@@ -186,6 +186,27 @@ EOF
         echo -n $state_dir/$node_name/node.socket
         ;;
 
+    start-healthchecks )
+        local usage="USAGE: wb backend $op RUN-DIR"
+        local dir=${1:?$usage}; shift
+
+        while test $# -gt 0
+        do case "$1" in
+               --* ) msg "FATAL:  unknown flag '$1'"; usage_supervisor;;
+               * ) break;; esac; shift; done
+
+        ls -l $dir/{tracer/tracer,node-{0,1}/node}.socket || true
+        if ! supervisorctl start healthcheck
+        then progress "supervisor" "$(red fatal: failed to start) $(white healthcheck)"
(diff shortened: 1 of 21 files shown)


mgmeier (approved):
Looks great @fmaste, thank you!

nix/workbench/backend/supervisor.sh
```diff
@@ -147,7 +147,9 @@ EOF
         local usage="USAGE: wb backend $op RUN-DIR"
         local dir=${1:?$usage}; shift
 
-        if ! supervisord --config  "$dir"/supervisor/supervisord.conf $@
+        # Avoid buffer related problems with stdout and stderr disabling buffering
+        # https://docs.python.org/3/using/cmdline.html#envvar-PYTHONUNBUFFERED
+        if ! PYTHONUNBUFFERED=TRUE supervisord --config  "$dir"/supervisor/supervisord.conf $@
```
mgmeier, line 152: ❤️

mgmeier (approved):

PR #5320, opened 2023-06-14
Some cleanup

jasagredo:
# Description

- `Setup.hs` is not needed if it is the default.
- All the local packages `-Werror` can be collapsed into program options.
- `tests: True` is redundant as per the cabal reference:

> furthermore, test suites are automatically enabled if they are
> requested as a built target.

### DIFF
--- a/Setup.hs
+++ b/Setup.hs
@@ -1,2 +1,0 @@
-import           Distribution.Simple
-main = defaultMain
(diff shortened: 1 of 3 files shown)


newhoggy (approved):

carbolymer (approved):

jasagredo:
How is it even possible that this PR, on top of master, has so many errors? It doesn't change any logic cc @newhoggy @carbolymer

jasagredo:
Rebased on top of that one :+1:

mgmeier (approved):
LGTM.

PR #5353, opened 2023-06-21
Turn on some traces by default

coot:


dnadales (approved):
I wonder if we should document somewhere how to turn these tracers off in the config file, because it does not seem obvious.

PR #5356, opened 2023-06-22
cardano-node top level error handler

coot:
The top level handler for diffusion errors should dispatch one of the
methods.

In particular it must rethrow async exceptions, and also prevents from
printing `DiffusionError user interrupt` when `SIGINT` was received.

### DIFF
--- a/cardano-node/src/Cardano/Node/Handlers/TopLevel.hs
+++ b/cardano-node/src/Cardano/Node/Handlers/TopLevel.hs
@@ -73,8 +73,8 @@ toplevelExceptionHandler prog = do
     hSetBuffering stderr LineBuffering
     catches prog [
         Handler rethrowAsyncExceptions
-      , Handler rethrowExitCode -- TODO this is possibly not needed anymore
-      , Handler rethrowDiffusionErrorExceptionInLinkedThreadExitSuccess
+      , Handler rethrowExitCode
+      , Handler handleDiffusionError
       , Handler handleSomeException
       ]
   where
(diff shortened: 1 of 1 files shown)


cardano-node/src/Cardano/Node/Handlers/TopLevel.hs
```diff
@@ -95,30 +95,40 @@ toplevelExceptionHandler prog = do
...
-              | Just exitCode <- fromException eInner
-              -> throwIO @ExitCode exitCode
-            _ -> throwIO full
-        _ -> throwIO full
+        Network.DiffusionError e
+          | Just e' <- fromException e
+          -> rethrowAsyncExceptions e'
+
+          | Just e' <- fromException e
+          -> rethrowExitCode e'
```
bolt12, line 107: I don't get these guards, aren't they the same? What are you trying to achieve here?

bolt12 (changes requested):

cardano-node/src/Cardano/Node/Handlers/TopLevel.hs
coot: They are different because the `rethrowAsyncExceptions` and `rethrowExitCode` force  `e'` to have different type (`SomeAsyncException` vs `ExitCode`).

bolt12 (approved):

cardano-node/src/Cardano/Node/Handlers/TopLevel.hs
newhoggy: Perhaps `fromException` can take type arguments to make this difference explicit?

newhoggy (approved):

Jimbo4350 (approved):

PR #5373, opened 2023-06-30
Bench refactoring

fmaste:
Cleanups on the nix side of the profile outputs, as used all around the workbench

Removes unused and duplicate code to make room for future changes. The profile and backend data was moved around in different attribute sets and now everything is contained in a single Nix file


NadiaYvette:
I like how it removes more lines than it adds.

mgmeier (approved):
Thank you @fmaste . LGTM!

PR #5377, opened 2023-06-30
Bench Nomad placement

fmaste:
Changes that allow the pinning of nodes to the same machines across Nomad Cloud runs plus more realistic runs (or as similar as possible to cardano ops)

- New workbench subcommand to create an index of available Nomad Clients
- Based on this Nomad Clients index, nodes are placed on an orderly manner by constraining AWS instance type, availability zone, instance id, public ip address and mac address. If Nomad Clients change the job fails.
- Use public IP/routing between nodes when using cardano world perf class
- Add new profiles that allow to run short tests on cardano world perf class
- Better documentation and cleanups of the Nomad Job creation

### DIFF
--- a/nix/workbench/profile/prof1-variants.jq
+++ b/nix/workbench/profile/prof1-variants.jq
@@ -600,6 +600,14 @@ def all_profile_variants:
   , { name: "default"
     , desc: "Default, as per nix/workbench/profile/prof0-defaults.jq"
     }
+  , $cardano_world_qa *
+    { name: "default-cw-qa"
+    , desc: "Default, but on Cardano World QA"
+    }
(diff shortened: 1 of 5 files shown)


nix/workbench/profile/prof1-variants.jq
```diff
@@ -45,10 +45,6 @@ def all_profile_variants:
       { utxo:                              (0.5 * $M)
       , delegators:                        (0.1 * $M)
       }
-    , generator:
-      { tps:                               1
-      , tx_count:                          10
```
mgmeier, line 50: Thanks!

mgmeier (approved):
LGTM. Thanks you @fmaste !

PR #5384, opened 2023-07-05
Bench nomad ssh

fmaste:
- When using the Nomad Cloud backend a custom OpenSSH server that can run inside the Nomad Task isolated namespace is added.
- When SSH is enabled logs are fetched directly from the Nomad Clients using rsync (the Nomad Server was the intermediary when doing `nomad alloc fs`).
- Removed cluster start/stop logic from the scenarios. Now the scenarios only start/stop cluster components, nodes/generator/tracers/healthchecks.
- Added functions `start-tracers`, `stop-all` and `fetch-logs` that were implicit when calling `start-cluster` and/or `stop-cluster`.
- Now the exit trap explicitly calls `stop-all` and `fetch-logs` that are either extra steps or not implicit when using the Nomad backend.
- Now the healthcheck service only emits warnings for no blocks / empty blocks timeouts. Plus other fixes.
- Make, only the wait loop, more resilient to networking errors.
- A fix for an error where a cluster may unintentionally run in "qa" class Nomad Nodes.
- Changes that make the workbench Mac compatible
- Clean ups to make more verbose the object like inheritance of Nomad sub-backends scripts.
- Towards a new SRE-built Nomad cluster: Changes the make the Nomad Cloud backend work with different Nomad clusters.

### DIFF
--- a/nix/workbench/backend/backend.sh
+++ b/nix/workbench/backend/backend.sh
@@ -38,21 +38,29 @@ backend() {
 local op=${1:-$(usage_backend)} # No need to shift -- backends will use the op.
 
 case "${op}" in
-    is-running )                 backend_$WB_BACKEND "$@";;
+    # Prepare functions
     setenv-defaults )            backend_$WB_BACKEND "$@";;
     allocate-run )               backend_$WB_BACKEND "$@";;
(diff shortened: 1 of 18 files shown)


mgmeier (approved):
Great work @fmaste

PR #5390, opened 2023-07-14
Make flake evaluation resilient to missing self.rev

andreabedini:
The attribute self.rev is missing also when flake inputs are overridden,
and it's highly desirable that the flake's evaluation is able to succeed
with overridden inputs.

The workbench-run-supervisor-ci-test-bage has its own logic around
checking self.rev, which does not use the top-level overlay; and the
system-tests script aborts the evaluation when self.rev is missing.

This change makes workbench-run-supervisor-ci-test-bage use pkgs.gitrev
as the rest of the flake and moves the checking self.rev to inside
system-tests script.


angerman (approved):
LGTM. Deferring the error to runtime.

yvan-sraka (approved):

PR #5407, opened 2023-07-21
Fix printing of NaN when validating chunks.

jasagredo:
Closes input-output-hk/ouroboros-consensus#247

### DIFF
--- a/cardano-node/src/Cardano/Tracing/OrphanInstances/Consensus.hs
+++ b/cardano-node/src/Cardano/Tracing/OrphanInstances/Consensus.hs
@@ -675,10 +675,10 @@ instance ( ConvertRawHash blk
         ImmDB.ChunkValidationEvent e -> case e of
           ImmDB.StartedValidatingChunk chunkNo outOf ->
                "Validating chunk no. " <> showT chunkNo <> " out of " <> showT outOf
-            <> ". Progress: " <> showProgressT (max (chunkNoToInt chunkNo - 1) 0) (chunkNoToInt outOf) <> "%"
+            <> ". Progress: " <> showProgressT (chunkNoToInt chunkNo) (chunkNoToInt outOf + 1) <> "%"
           ImmDB.ValidatedChunk chunkNo outOf ->
                "Validated chunk no. " <> showT chunkNo <> " out of " <> showT outOf
-            <> ". Progress: " <> showProgressT (chunkNoToInt chunkNo) (chunkNoToInt outOf) <> "%"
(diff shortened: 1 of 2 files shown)


jorisdral:
What's the context for this PR? Is this fixing a UTxO HD specific problem?

jasagredo:
I have observed this only on UTxO-HD but I think it is something that happens even without UTxO-HD (we didn't change how the tracers for validating chunks are emitted). Perhaps I could point this to master

jorisdral:
> Perhaps I could point this to master

I think that would make sense

cardano-node/src/Cardano/Tracing/OrphanInstances/Consensus.hs
```diff
@@ -674,11 +674,17 @@ instance ( ConvertRawHash blk
...
+            | chunkNo == outOf ->
+                "Validating last chunk no. " <> showT chunkNo <> " out of " <> showT outOf <> "."
+            | otherwise ->
+                   "Validating chunk no. " <> showT chunkNo <> " out of " <> showT outOf
+                <> ". Progress: " <> showProgressT (chunkNoToInt chunkNo) (chunkNoToInt outOf + 1) <> "%"
           ImmDB.ValidatedChunk chunkNo outOf ->
-               "Validated chunk no. " <> showT chunkNo <> " out of " <> showT outOf
-            <> ". Progress: " <> showProgressT (chunkNoToInt chunkNo) (chunkNoToInt outOf) <> "%"
+            | chunkNo == outOf ->
+                "Validated last chunk no. " <> showT chunkNo <> " out of " <> showT outOf <> "."
```
jorisdral, line 684: Do we need these cases? It might be nice to show the `100%` progress bar instead of this message

jorisdral (approved):

cardano-node/src/Cardano/Tracing/OrphanInstances/Consensus.hs
jasagredo: Not sure I understand, this should print "Validated chunk 100 out of 100. Progress: 100%"

jasagredo: Actually no. You are right. Let me amend it.

jasagredo: Done

jorisdral (approved):

carbolymer (approved):

PR #5409, opened 2023-07-21
Bump network packages version

bolt12:
# Description

Compatibility with ouroboros-network-0.8.2.0

Added WarmValency option to the Topology file

Now the Topology allows the user to not only specify the HotValency of
its node's local root peers, via 'valency' ('hotValency' is now also
valid) field, but also is able to specify the WarmValency, via
'warmValency' field. This field is optional and defaults to the value
set in the 'valency'/'hotValency' field. The 'warmValency' value set
should be greater than or equal to the one specified in
'valency'/'hotValency' otherwise 'valency'/'hotValency' will be
truncated to this value. We recommend users to set 'warmValency' value to
'hotValency' + 1 in order to keep at least 1 backup peer to be promoted
to hot in case something happens.

Check https://github.com/input-output-hk/ouroboros-network/issues/4565
for more context on this WarmValency addition.

### DIFF
--- a/bench/locli/locli.cabal
+++ b/bench/locli/locli.cabal
@@ -120,7 +120,7 @@ library
                       , optparse-generic
                       , ouroboros-consensus
                       -- for Data.SOP.Strict:
-                      , ouroboros-network ^>= 0.8.1.0
+                      , ouroboros-network ^>= 0.8.2.0
                       , ouroboros-network-api
                       , process
                       , quiet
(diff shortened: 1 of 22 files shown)


carbolymer (approved):

disassembler (approved):

NadiaYvette (approved):
These changes are straightforward & innocuous.

PR #5437, opened 2023-08-15
Optimize Mux tracer

jutaro:
1. Implement a forMachine json representation.
2. Implement a forHuman representation, which is based on the formatting library, which has a printf like functionality, but don't use Haskell Strings, but works with Text.

### DIFF
--- a/cardano-node/src/Cardano/Node/Tracing/Tracers/Diffusion.hs
+++ b/cardano-node/src/Cardano/Node/Tracing/Tracers/Diffusion.hs
@@ -4,7 +4,9 @@
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+
 
 
 {-# OPTIONS_GHC -Wno-orphans  #-}
@@ -12,9 +14,11 @@
 module Cardano.Node.Tracing.Tracers.Diffusion
   () where
 
+
 import           Cardano.Logging
 import           Data.Aeson (Value (String), (.=))
 import           Data.Text (pack)
+import           Formatting
 import           Network.Mux (MuxTrace (..), WithMuxBearer (..))
 import           Network.Mux.Types
 import           Network.TypedProtocol.Codec (AnyMessageAndAgency (..))
@@ -65,7 +69,7 @@ instance LogFormatting MuxTrace where
     forMachine _dtal (MuxTraceRecvHeaderEnd MuxSDUHeader { mhTimestamp, mhNum, mhDir, mhLength }) = mconcat
       [ "kind" .= String "MuxTraceRecvHeaderStart"
       , "msg"  .=  String "Bearer Receive Header End"
-      , "timestamp" .= String (showT (unRemoteClockModel mhTimestamp))
+      , "timestamp" .= String (showTHex (unRemoteClockModel mhTimestamp))
       , "miniProtocolNum" .= String (showT mhNum)
       , "miniProtocolDir" .= String (showT mhDir)
       , "length" .= String (showT mhLength)
@@ -74,7 +78,7 @@ instance LogFormatting MuxTrace where
       [ "kind" .= String "MuxTraceRecvDeltaQObservation"
       , "msg"  .=  String "Bearer DeltaQ observation"
       , "timeRemote" .=  String (showT ts)
-      , "timeLocal" .= String (showT (unRemoteClockModel mhTimestamp))
+      , "timeLocal" .= String (showTHex (unRemoteClockModel mhTimestamp))
       , "length" .= String (showT mhLength)
       ]
     forMachine _dtal (MuxTraceRecvDeltaQSample d sp so dqs dqvm dqvs estR sdud) = mconcat
@@ -102,7 +106,7 @@ instance LogFormatting MuxTrace where
     forMachine _dtal (MuxTraceSendStart MuxSDUHeader { mhTimestamp, mhNum, mhDir, mhLength }) = mconcat
       [ "kind" .= String "MuxTraceSendStart"
       , "msg"  .= String "Bearer Send Start"
-      , "timestamp" .= String (showT (unRemoteClockModel mhTimestamp))
+      , "timestamp" .= String (showTHex (unRemoteClockModel mhTimestamp))
       , "miniProtocolNum" .= String (showT mhNum)
       , "miniProtocolDir" .= String (showT mhDir)
       , "length" .= String (showT mhLength)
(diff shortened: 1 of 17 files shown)


mgmeier (approved):
Excellent, thank you @jutaro

cardano-node/src/Cardano/Node/Tracing/Tracers/Diffusion.hs
```diff
@@ -53,6 +61,274 @@ instance MetaTrace tr => MetaTrace (WithMuxBearer peer tr) where
...
+      , "msg"  .= String "TCPInfo"
+      , "len"  .= String (showT len)
+      ]
+#endif
+
+    forHuman MuxTraceRecvHeaderStart =
+      "Bearer Receive Header Start"
+    forHuman (MuxTraceRecvHeaderEnd MuxSDUHeader { mhTimestamp, mhNum, mhDir, mhLength }) =
+      sformat ("Bearer Receive Header End: ts:" % prefixHex % "(" % shown % ") " % shown % " len " % int)
+        (unRemoteClockModel mhTimestamp) mhNum mhDir mhLength
```
coot, line 255: `mhNum` has type `newtype MiniProtocolNum = MiniProtocolNum Word16` if you unpack it you could use `int` formatter or (`"MiniProtocol " % int)`, it might be even faster.

cardano-node/src/Cardano/Node/Tracing/Tracers/Diffusion.hs
```diff
@@ -140,6 +416,8 @@ instance MetaTrace MuxTrace where
     severityFor (Namespace _ ["StartedOnDemand"]) _       = Just Debug
     severityFor (Namespace _ ["Terminating"]) _           = Just Debug
     severityFor (Namespace _ ["Shutdown"]) _              = Just Debug
+    severityFor (Namespace _ ["Stopping"]) _               = Just Debug
```
coot, line 419:
```suggestion
    severityFor (Namespace _ ["Stopping"]) _              = Just Debug
```
just indentation

cardano-node/src/Cardano/Node/Tracing/Tracers/Diffusion.hs
```diff
@@ -195,6 +473,10 @@ instance MetaTrace MuxTrace where
       "Started on demand."
     documentFor (Namespace _ ["Terminating"])           = Just
       "Terminating."
+    documentFor (Namespace _ ["Stopping"])              = Just
+      "Mux shutdown."
+    documentFor (Namespace _ ["Stopped"])              = Just
```
coot, line 478:
```suggestion
    documentFor (Namespace _ ["Stopped"])               = Just
```

cardano-tracer/cardano-tracer.cabal
```diff
@@ -280,6 +280,9 @@ test-suite cardano-tracer-test
   type:                 exitcode-stdio-1.0
   default-extensions:   OverloadedStrings
 
+  if os(windows)
+    buildable:          False
```
coot, line 284: Are you sure it's the right thing to exclude Windows? Is it possible to trace without `cardano-tracer`?  This is imporant for full node wallets.

cardano-tracer/src/Cardano/Tracer/Handlers/RTView/Notifications/Send.hs
```diff
@@ -16,6 +16,8 @@ import qualified Data.Text as T
 import           Data.Time.Clock (UTCTime)
 import           Data.Time.Format (defaultTimeLocale, formatTime)
 
+import           Cardano.Logging (showT)
```
coot, line 19: Where `showT` was coming from previously? Maybe `Cardano.Tracer.Utils`?

cardano-tracer/src/Cardano/Tracer/Handlers/RTView/Update/EKG.hs
```diff
@@ -12,11 +12,12 @@ import qualified Data.Map.Strict as M
 import           Data.Text (intercalate, isPrefixOf)
 import           Graphics.UI.Threepenny.Core (UI, liftIO)
 
+import           Cardano.Logging (showT)
+
 import           Cardano.Tracer.Environment
 import           Cardano.Tracer.Handlers.Metrics.Utils
 import           Cardano.Tracer.Handlers.RTView.UI.Utils
 import           Cardano.Tracer.Types
-import           Cardano.Tracer.Utils
```
coot, line 19: Probably, because here `Cardano.Tracer.Utils` is removed bug `Cardano.Logging` is added.

Should it be still exported from `Cardano.Tracer.Utils`?

cardano-node/src/Cardano/Node/Tracing/Tracers/Diffusion.hs
```diff
@@ -53,6 +61,274 @@ instance MetaTrace tr => MetaTrace (WithMuxBearer peer tr) where
     metricsDocFor ns = metricsDocFor (nsCast ns :: Namespace tr)
     allNamespaces = map nsCast (allNamespaces :: [Namespace tr])
 
+instance LogFormatting MuxTrace where
+    forMachine _dtal MuxTraceRecvHeaderStart = mconcat
+      [ "kind" .= String "MuxTraceRecvHeaderStart"
+      , "msg"  .= String "Bearer Receive Header Start"
```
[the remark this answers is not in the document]
jutaro: Kind will be removed shortly, as any message will be uniquely identified by its namespace. That's why we need it.

cardano-node/src/Cardano/Node/Tracing/Tracers/Diffusion.hs
```diff
@@ -53,6 +61,274 @@ instance MetaTrace tr => MetaTrace (WithMuxBearer peer tr) where
     metricsDocFor ns = metricsDocFor (nsCast ns :: Namespace tr)
     allNamespaces = map nsCast (allNamespaces :: [Namespace tr])
 
+instance LogFormatting MuxTrace where
+    forMachine _dtal MuxTraceRecvHeaderStart = mconcat
+      [ "kind" .= String "MuxTraceRecvHeaderStart"
```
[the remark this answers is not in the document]
jutaro: See above, kinds are only legacy fields, and no need to change it now.

cardano-node/src/Cardano/Node/Tracing/Tracers/Diffusion.hs
```diff
@@ -53,6 +61,274 @@ instance MetaTrace tr => MetaTrace (WithMuxBearer peer tr) where
...
+      , "msg"  .= String "Bearer Receive Header Start"
+      ]
+    forMachine _dtal (MuxTraceRecvHeaderEnd MuxSDUHeader { mhTimestamp, mhNum, mhDir, mhLength }) = mconcat
+      [ "kind" .= String "MuxTraceRecvHeaderStart"
```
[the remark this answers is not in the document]
jutaro: Thanks for finding this, will correct this in my next pull request

cardano-node/src/Cardano/Node/Tracing/Tracers/Diffusion.hs
```diff
@@ -53,6 +61,274 @@ instance MetaTrace tr => MetaTrace (WithMuxBearer peer tr) where
...
+      , "msg"  .= String "Handshake Server end"
+      ]
+    forMachine dtal (MuxTraceHandshakeClientError e duration) = mconcat
+      [ "kind" .= String "MuxTraceHandshakeClientError"
+      , "msg"  .= String "Handshake Client Error"
+      , "duration" .= String (showT duration)
+      -- Client Error can include an error string from the peer which could be very large.
+      , "error" .= if dtal >= DDetailed
+                      then show e
+                      else take 256 $ show e
```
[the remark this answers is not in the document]
jutaro: Then we can add detailed in the standard configuration for this message.

PR #5438, opened 2023-08-16
Split release upload into multiple jobs

angerman:
# Description

This changes the release upload workflow to pull all three assets separately (this should help in debugging what's wrong) uses a consistent naming scheme for the jobs, and might even be faster.


.github/workflows/release-upload.yaml
andreabedini, line 1: Can we maybe reduce the duplication with a matrix strategy?
We could split the workflow into two jobs: one to pull with `strategy: matrix: arch: [linux, macos, win64]` and one that only does `upload-assets`. It should work I think?

angerman: I'm not certain we can actually _name_ jobs, which we'd then need for the `needs` clause.

And shared workflows, which I think might work, are quite intransparent in the GH UI.

angerman:
> If you add a `workflow_dispatch:` dispatch event, you can test this from your branch (I think).

I don't think this works while the workflow is on a branch.

andreabedini (approved):
LGTM (as long as it works 🙃)

angerman:
> LGTM (as long as it works 🙃)

Lol. And how do we figure that out?

andreabedini:
Do you know why the CI failing? I haven't looked into it myself.

angerman:
> Do you know why the CI failing? I haven't looked into it myself.

Looks like "Upload to ghci.io / Upload to ghcr.io (push)" ran too early.

hamishmack (approved):

andreabedini:
Right. The failsafe is that it fails rather than building everything itself. The idea was that if this happens, it's harmless, and the one can simply rerun the same job later.

.github/workflows/release-ghcr.yaml
```diff
@@ -12,7 +12,34 @@ env:
...
+    steps:
+    - name: Waiting for ci/hydra-build:required to complete
+      run: |
+        while [[ true ]]; do
+          # Get the list of check runs for the latest commit
+          check_runs_url="https://api.github.com/repos/$GITHUB_REPOSITORY/commits/$GITHUB_SHA/check-runs"
+          check_runs=$(curl --silent --header "Authorization: token ${{ secrets.GITHUB_TOKEN }}" --header "Accept: application/vnd.github.v3+json" $check_runs_url)
+  
+          # Filter to get the specific check run you're interested in and its conclusion
+          conclusion=$(echo "$check_runs" | jq -r '.check_runs[] | select(.name == "ci/hydra-build:required") | .conclusion')
```
andreabedini, line 27:
```suggestion
          conclusion=$(gh api repos/$GITHUB_REPOSITORY/commits/$$GITHUB_SHA/check-runs --jq '.check_runs[] | select(.name == "ci/hydra-build:required") | .conclusion')
```
should work the same. Do we get to see the pending checks?

PR #5446, opened 2023-08-25
Shutdown stdoutThread gracefully

jasagredo:
# Description

Instead of silencing a `BlockedIndefinitelyOnMVar`, this PR makes sure the thread listening in the stdoutTracer thread exits gracefully.

# Checklist

- [ ] ~New tests are added if needed and existing tests are updated.  These may include:~
- [ ] ~Any changes are noted in the `CHANGELOG.md` for affected package~ (internal change)
- [ ] ~The version bounds in `.cabal` files are updated~ (N/A)
- [x] Self-reviewed the diff

### DIFF
--- a/cardano-node/src/Cardano/Node/Tracing/API.hs
+++ b/cardano-node/src/Cardano/Node/Tracing/API.hs
@@ -86,7 +86,9 @@ initTraceDispatcher nc p networkMagic nodeKernel p2pMode = do
     EKG.registerGcMetrics ekgStore
     ekgTrace <- ekgTracer (Left ekgStore)
 
-    stdoutTrace <- standardTracer
+    -- We don't shutdown the trace layer in the node so we can ignore the
+    -- shutdown action
+    (stdoutTrace, _shutdown) <- standardTracer
 
     -- We should initialize forwarding only if 'Forwarder' backend
     -- is presented in the node's configuration.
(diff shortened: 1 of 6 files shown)


cardano-node/src/Cardano/Node/Tracing/API.hs
```diff
@@ -86,7 +86,9 @@ initTraceDispatcher nc p networkMagic nodeKernel p2pMode = do
     EKG.registerGcMetrics ekgStore
     ekgTrace <- ekgTracer (Left ekgStore)
 
-    stdoutTrace <- standardTracer
+    -- We don't shutdown the trace layer in the node so we can ignore the
+    -- shutdown action
+    (stdoutTrace, _shutdown) <- standardTracer
```
coot, line 91: Should we ignore it? Note that the thread that is running it, is left stranded - it will be killed by RTS, which is not ideal.  Would it be fine to use it in `finally` handler of the main thread? The only concern is to not shut it down to early while still some messages are generated.

coot (commented):
That's definitely a move in the right direction!

I'd prefer a different pattern:
```
withStandardTracer :: (Trace m FormattedMessage -> m a) -> m a
withStandardTracer k = do
  (tracer, shutdown) <- standardTracer
  k tracer `finally` shutdown
```
This way the shutdown is an internal detail, and it stops leaking the logging thread as well (although `standardTracer` should use `withAsync` instead of `async` to be really safe).

I see though that the txgenerator doesn't want to use it.  Is there a way to just not start it at all instead of shuting it down?

jasagredo:
This PR becomes a Draft as the performance team will look into a better way to solve this.

PR #5449, opened 2023-08-29
Isolate BS trace and downgrade severity of rejected txs

jasagredo:
# Description

- Downgrade the severity of `MempoolRejectedTx` to `Info`
- Isolate the backing store tracer into its own tracer that can be switched off.


mgmeier (approved):
Looks great, thank you @jasagredo

PR #5460, opened 2023-09-06
Fix deadlock issue with stdout tracer in tx-generator

jutaro:
Problem:
The root issue we encountered with the tx-generator was the creation of a new stdout backend for every tracer instance. This approach, however, conflicts with the fundamental requirement of having only one stdout backend per application. Consequently, this practice led to unnecessary garbage collection of these backends and resulted in the dreaded "BlockedIndefinitelyOnMVar" exception.

Solution:
To rectify this issue, the proposed solution involves constructing and utilizing only a single backend for all tracers. The introduction of a shutdown handler, in this context, becomes redundant. It's important to note that a shutdown handler wouldn't actually address the underlying problem but would merely mask it.

Alternative Consideration:
While it is conceivable to enforce a single instance from the trace-dispatcher library, this approach would necessitate the use of unsafePerformIO. We've chosen not to adopt this method due to concerns related to its contravariant interface.

This pull request aims to implement the solution by ensuring a single backend, thus resolving the issue and enhancing the stability of the tx-generator.

In addition we adopted the tracing of tx-generator to use the new interface, where it was possible. So the fix itself can be found in function initTxGenTracers.

### DIFF
--- a/bench/tx-generator/src/Cardano/Benchmarking/GeneratorTx.hs
+++ b/bench/tx-generator/src/Cardano/Benchmarking/GeneratorTx.hs
@@ -1,5 +1,4 @@
 {-# LANGUAGE FlexibleContexts #-}
-{-# LANGUAGE PackageImports #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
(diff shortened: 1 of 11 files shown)


mgmeier (approved):
LGTM, thank you!

PR #5472, opened 2023-09-18
Remove some workarounds

michaelpj:
# Description

The simd issue seems to only appear when using the external interpreter. We are only using the external interpereter for `plutus-core` in order to work around a weird profiling issue. The weird profiling issue seems to be gone, so we can just delete both workarounds.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -100,9 +100,6 @@ write-ghc-environment-files: always
 package snap-server
   flags: +openssl
 
-package bitvec
-   flags: -simd
-
 -- IMPORTANT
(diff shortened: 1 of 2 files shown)


Jimbo4350 (approved):

angerman:
We do use ext int for all our cross pipelines.

michaelpj:
Okay, now it fails somewhere apparently unrelated. I don't know what's going on, I guess this will have to stay for now.

PR #5480, opened 2023-09-21
Trace Configuration Trace Message Enhancement

jutaro:
This pull request aims to enhance our codebase by adding trace messages that provide valuable insights into the effective trace configuration used during program execution. Trace messages are essential for debugging and monitoring our application, and this enhancement will make it easier for developers and operators to understand and troubleshoot any issues related to tracing.

### DIFF
--- a/trace-dispatcher/src/Cardano/Logging/ConfigurationParser.hs
+++ b/trace-dispatcher/src/Cardano/Logging/ConfigurationParser.hs
@@ -1,8 +1,6 @@
-{-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE LambdaCase #-}
 
 {-# OPTIONS_GHC -Wno-orphans #-}
 
@@ -13,16 +11,16 @@ module Cardano.Logging.ConfigurationParser
   , configToRepresentation
   ) where
 
+import           Control.Applicative ((<|>))
 import           Control.Exception (throwIO)
 import qualified Data.Aeson as AE
 import           Data.ByteString (ByteString)
 import qualified Data.ByteString.Char8 as BS
 import           Data.List (foldl')
 import qualified Data.Map.Strict as Map
-import           Data.Maybe (isJust)
-import           Data.Text (Text, split, intercalate)
+import           Data.Maybe (catMaybes, listToMaybe)
+import           Data.Text (Text, intercalate, split)
 import           Data.Yaml
-import           GHC.Generics
 
 import           Cardano.Logging.Types
 
@@ -52,9 +50,9 @@ instance AE.ToJSON ConfigRepresentation where
   toJSON ConfigRepresentation{..} = object
     [ "TraceOptions"                  .= traceOptions
     , "TraceOptionForwarder"          .= traceOptionForwarder
-    , "traceOptionNodeName"           .= traceOptionNodeName
+    , "TraceOptionNodeName"           .= traceOptionNodeName
     , "TraceOptionPeerFrequency"      .= traceOptionPeerFrequency
-    , "traceOptionResourceFrequency"  .= traceOptionResourceFrequency
+    , "TraceOptionResourceFrequency"  .= traceOptionResourceFrequency
     ]
 
 -- | In the external configuration representation for configuration files
@@ -65,7 +63,7 @@ data ConfigOptionRep = ConfigOptionRep
     , backends :: Maybe [BackendConfig]
     , maxFrequency :: Maybe Double
     }
-  deriving (Eq, Ord, Show, Generic)
+  deriving (Eq, Ord, Show)
 
 instance AE.FromJSON ConfigOptionRep where
   parseJSON (Object obj) = ConfigOptionRep
@@ -107,18 +105,10 @@ readConfigurationWithDefault fp defaultConf = do
         (if (not . Map.null) (tcOptions fileConf)
           then tcOptions fileConf
           else tcOptions defaultConf)
-        (if isJust (tcForwarder fileConf)
-            then tcForwarder fileConf
-            else tcForwarder defaultConf)
-        (if isJust (tcNodeName fileConf)
-            then tcNodeName fileConf
-            else tcNodeName defaultConf)
-        (if isJust (tcPeerFrequency fileConf)
-            then tcPeerFrequency fileConf
-            else tcPeerFrequency defaultConf)
-        (if isJust (tcResourceFrequency fileConf)
-            then tcResourceFrequency fileConf
-            else tcResourceFrequency defaultConf)
+        (tcForwarder fileConf <|> tcForwarder defaultConf)
+        (tcNodeName fileConf <|> tcNodeName defaultConf)
+        (tcPeerFrequency fileConf <|> tcPeerFrequency defaultConf)
+        (tcResourceFrequency fileConf <|> tcResourceFrequency defaultConf)
 
 -- | Parse the byteString as external representation and converts to internal
 -- representation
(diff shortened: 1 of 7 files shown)


trace-dispatcher/src/Cardano/Logging/ConfigurationParser.hs
```diff
@@ -19,53 +20,108 @@ import qualified Data.ByteString.Char8 as BS
...
+-- all options for a namespace are part of a record
+data ConfigOptionRep = ConfigOptionRep
+    { severity :: Maybe SeverityF
+    , detail :: Maybe DetailLevel
+    , backends :: Maybe [BackendConfig]
+    , maxFrequency :: Maybe Double
+    }
+  deriving (Eq, Ord, Show, Generic)
+
+instance AE.FromJSON ConfigOptionRep where
```
mgmeier, line 70: Is there any reason why `deriving (FromJSON, ToJSON)` isn't used? I can't really spot a difference in the handcrafted instances.

trace-dispatcher/src/Cardano/Logging/ConfigurationParser.hs
```diff
@@ -19,53 +20,108 @@ import qualified Data.ByteString.Char8 as BS
...
+    parseJSON _ = mempty
+
+instance AE.ToJSON ConfigRepresentation where
+  toJSON ConfigRepresentation{..} = object
+    [ "TraceOptions"                  .= traceOptions
+    , "TraceOptionForwarder"          .= traceOptionForwarder
+    , "traceOptionNodeName"           .= traceOptionNodeName
+    , "TraceOptionPeerFrequency"      .= traceOptionPeerFrequency
+    , "traceOptionResourceFrequency"  .= traceOptionResourceFrequency
+    ]
```
mgmeier, line 58: The capitalization of object keys doesn't align between `FromJSON` and `ToJSON`, is that intentional?
If there's a `Generic` instance for everything (not sure), one could used derived instances, or for capital initials something like:
```
instance FromJSON ConfigRepresentation where
    parseJSON = genericParseJSON defaultOptions {fieldLabeModifier =  <capitalize first letter>}
```

trace-dispatcher/src/Cardano/Logging/ConfigurationParser.hs
```diff
@@ -19,53 +20,108 @@ import qualified Data.ByteString.Char8 as BS
...
+        (if isJust (tcNodeName fileConf)
+            then tcNodeName fileConf
+            else tcNodeName defaultConf)
+        (if isJust (tcPeerFrequency fileConf)
+            then tcPeerFrequency fileConf
+            else tcPeerFrequency defaultConf)
+        (if isJust (tcResourceFrequency fileConf)
+            then tcResourceFrequency fileConf
+            else tcResourceFrequency defaultConf)
+
```
mgmeier, line 122: Is it an option to use `((<|>))` here? It would enhance readability, replacing conditional + `isJust` fields like so:
```
(tcForwarder fileConf <|> tcForwarder defaultConf)
```

trace-dispatcher/src/Cardano/Logging/ConfigurationParser.hs
```diff
@@ -95,57 +151,74 @@ parseRepresentation bs = transform (decodeEither' bs)
...
+        Just dtl -> [ConfDetail dtl]
+      ++
+      case backends of
+        Nothing -> []
+        Just bcks -> [ConfBackend bcks]
+      ++
+      case maxFrequency of
+        Nothing -> []
+        Just lim -> [ConfLimiter lim]
+
```
mgmeier, line 172: Could you use sth. like `maybeToList (ConfDetail <$> maxDetail)` here?

trace-dispatcher/src/Cardano/Logging/ConfigurationParser.hs
```diff
@@ -95,57 +151,74 @@ parseRepresentation bs = transform (decodeEither' bs)
...
+                          ConfBackend _ -> True
+                          _ -> False) opts of
+                      ConfBackend back : _ -> Just back
+                      _ -> Nothing
+      , maxFrequency = case filter (\case
+                          ConfLimiter _ -> True
+                          _ -> False) opts of
+                      ConfLimiter freq : _ -> Just freq
+                      _ -> Nothing
+      }
```
mgmeier, line 220: Could one use `listToMaybe [d | ConfDetail d <- opts]` here?

trace-dispatcher/src/Cardano/Logging/TraceDispatcherMessage.hs
```diff
@@ -91,6 +96,9 @@ instance LogFormatting TraceDispatcherMessage where
         [ "kind" .= String "TracerConsistencyWarnings"
         , "errors" .= String ((pack . show) errs)
         ]
+  forMachine _dtl (TracerInfoConfig tc) = mconcat
+        [ "conf" .= toJSON tc
+        ]
```
mgmeier, line 101: ❤️

trace-dispatcher/src/Cardano/Logging/TraceDispatcherMessage.hs
```diff
@@ -147,9 +160,11 @@ instance MetaTrace TraceDispatcherMessage where
     documentFor (Namespace _ ["TracerConsistencyWarnings"]) = Just $ mconcat
       [ "Tracer consistency check found errors."
       ]
+    documentFor (Namespace _ ["InfoConfig"]) = Just $ mconcat
```
mgmeier, line 163: The name is without `Tracer...` here, yet it provides a `TracerConfig` json. Maybe we could rename it to sth. like `TracerConfigInfo` to keep the pattern? Or is there a reason to set it apart?

trace-dispatcher/src/Cardano/Logging/Utils.hs
```diff
@@ -44,15 +44,16 @@ runInLoop action localSocket prevDelayInSecs =
 uncurry3 :: (a -> b -> c -> d) -> (a, b, c) -> d
 uncurry3 f (a,b,c) = f a b c
 
+-- | map over second element of a tuple
 mapSnd :: (a -> b) -> (c, a) -> (c, b)
```
mgmeier, line 48: Just a veeeeeeery minor thing: There is `Data.Bifunctor.second` in `base`....

trace-dispatcher/src/Cardano/Logging/ConfigurationParser.hs
NadiaYvette: Is it possible that mappend from the Monoid instance for Maps could do similar for the first of these, or is there a possibility of absent configuration options that would be invalid to override with the defaults?

trace-dispatcher/src/Cardano/Logging/ConfigurationParser.hs
NadiaYvette: toOptionRepresentation looks like Map.mapKeys (intercalate ".") . Map.map fromOptions could avoid needing the helper function, if you like the idea. (Map.mapKeysMonotonic might be applicable too.)

trace-dispatcher/src/Cardano/Logging/ConfigurationParser.hs
mgmeier: Using `(<>)` extends to a set union of keys, which changes semantics, as you pointed out.

trace-dispatcher/src/Cardano/Logging/ConfigurationParser.hs
jutaro: Corrected the capitalisation. Thanks for pointing this out. It is only one unused Generic instance here, which I remove.

trace-dispatcher/src/Cardano/Logging/ConfigurationParser.hs
jutaro: The maybe fields are only printed, when the value is not Nothing. This is hand coded, so that I used handcrafted instances here. I think this is only possible with Template Haskell automatically, but in my opinion the handwritten stuff is preferred.

trace-dispatcher/src/Cardano/Logging/ConfigurationParser.hs
jutaro: Clever. I changed it

trace-dispatcher/src/Cardano/Logging/TraceDispatcherMessage.hs
jutaro: Changed

trace-dispatcher/src/Cardano/Logging/ConfigurationParser.hs
jutaro: @mgmeier changed. @NadiaYvette sorry, but I like my simple fold, which is easier for me to understand

trace-dispatcher/src/Cardano/Logging/Utils.hs
jutaro: Removed

trace-dispatcher/src/Cardano/Logging/ConfigurationParser.hs
jutaro: Clever, I changed it

mgmeier (approved):
Thanks @jutaro LGTM!

PR #5481, opened 2023-09-22
workbench: nomad value profile

fmaste:
# Description

Changes needed to bring the Nomad cloud backend in line with cardano-ops to finally be able to switch. The focus is on the `value` (52 nodes + explorer) profile.

- Nomad cloud profiles now use as suffix `-nomadperf` and `-nomadcwqa` for the P&T exclusive Nomad cluster and the `qa` class Nomad clients on Cardano World respectively.
- On Cardano World regions are an `affinity` not a `constraint`, we try to only use what is available.
- A `value-nomadperf` and a `value-oldtracing-nomadperf` profile were added. The latter is not directly comparable for benchmarks due to emergent performance but as similar as possible to what `cardano-ops` runs.
- Topology for the value profile was directly imported from `cardano-ops`, it's currently not being build by `cardano-topology`.
- Extra check: After deployment an ex-post `node-specs.json` like file and a `topology.json` like file are constructed (saved inside the `RUN/nomad` directory) using only what was actually deployed because files like individual `topology.json` use Nomad templates to dynamically fill IP addresses during Job startup. These two files are compared with the profile definition ones to look for differences.
- Extra check: Forced nodes to bind to its specific AWS private ip during cloud runs.
- Extra reproducibility constraint: Added as placement constraints kernel versions and CPU models. Also means the kernel version and CPU model used are logged for later use.
- When deploying sort Nomad clients first by name so if a Nomad client gets redeployed, replaced by a new one with the same name, only that Task is placed in a different EC2 machine.
- Added a warning message if the healthcheck of the "explorer" node fails. This node is the only one that keeps a healthcheck running and right now it only checks for blocks.
- Keep Nomad jobs running, no automatic stop, during cloud runs on P&T exclusive Nomad cluster.
- Only start the sandboxed ssh servers before fetching the logs, not always running 
- Dynamic sleep times when waiting for the pool to stop. Do not query the Nomad server too often during cloud runs while being less susceptible to network outages.
- Made sure supervisor does not run in unbuffered mode.
- Simplified and abstracted the differences between Nomad filesystem isolation modes, the relatives directories used by Nomad Jobs, and make room to support the `raw_exec` plugin.
- Refactored the log fetching logic to remove duplicate downloads and be more resilient.
- Fixed bug that deployed incorrect topology files. Inside the workbench (does not affect cardano-ops runs), individual nodes' topology files may be swapped. AFAIK this only affected profiles with different regions, non-local, all "loopback" regions profiles are not affected.
- `wait_all` and `wait_kill_em_all` functions for async like programming in bash.
- Other small fixes.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [x] Self-reviewed the diff

# Note on CI
for you.


mgmeier (approved):
Great work, thanks a lot @fmaste

PR #5485, opened 2023-09-27
Bootstrap Peers

bolt12:
# Description

This PR integrates: https://github.com/input-output-hk/ouroboros-network/pull/4555

### DIFF
--- a/cardano-node/src/Cardano/Node/Configuration/TopologyP2P.hs
+++ b/cardano-node/src/Cardano/Node/Configuration/TopologyP2P.hs
@@ -30,7 +30,9 @@ import           Cardano.Node.Startup (StartupTrace (..))
 import           Cardano.Node.Types
 import           Cardano.Tracing.OrphanInstances.Network ()
 import           Ouroboros.Network.NodeToNode (PeerAdvertise (..))
-import           Ouroboros.Network.PeerSelection.LedgerPeers (UseLedgerAfter (..))
+import           Ouroboros.Network.PeerSelection.Bootstrap (UseBootstrapPeers (..))
+import           Ouroboros.Network.PeerSelection.LedgerPeers.Type (UseLedgerPeers (..))
+import           Ouroboros.Network.PeerSelection.PeerTrustable (PeerTrustable (..))
 import           Ouroboros.Network.PeerSelection.RelayAccessPoint (RelayAccessPoint (..))
 import           Ouroboros.Network.PeerSelection.State.LocalRootPeers (HotValency (..),
                    WarmValency (..))
@@ -53,7 +55,7 @@ data NodeSetup = NodeSetup
   , nodeIPv4Address :: !(Maybe NodeIPv4Address)
   , nodeIPv6Address :: !(Maybe NodeIPv6Address)
   , producers       :: ![RootConfig]
-  , useLedger       :: !UseLedger
+  , useLedger       :: !UseLedgerPeers
   } deriving (Eq, Show)
 
 instance FromJSON NodeSetup where
@@ -63,7 +65,7 @@ instance FromJSON NodeSetup where
                   <*> o .:  "nodeIPv4Address"
                   <*> o .:  "nodeIPv6Address"
                   <*> o .:  "producers"
-                  <*> o .:? "useLedgerAfterSlot" .!= UseLedger DontUseLedger
+                  <*> o .:? "useLedgerAfterSlot" .!= DontUseLedgerPeers
 
 instance ToJSON NodeSetup where
   toJSON ns =
@@ -85,7 +87,7 @@ data RootConfig = RootConfig
     -- or domain name and a port number.
   , rootAdvertise    :: PeerAdvertise
     -- ^ 'advertise' configures whether the root should be advertised through
-    -- gossip.
+    -- peer sharing.
   } deriving (Eq, Show)
 
 instance FromJSON RootConfig where
@@ -121,6 +123,9 @@ data LocalRootPeersGroup = LocalRootPeersGroup
   { localRoots :: RootConfig
   , hotValency :: HotValency
   , warmValency :: WarmValency
+  , trustable   :: PeerTrustable
+    -- ^ 'trustable' configures whether the root should be trusted in fallback
+    -- state.
   } deriving (Eq, Show)
 
 -- | Does not use the 'FromJSON' instance of 'RootConfig', so that
@@ -134,6 +139,7 @@ instance FromJSON LocalRootPeersGroup where
                   <$> parseJSON (Object o)
                   <*> pure hv
                   <*> o .:? "warmValency" .!= WarmValency v
+                  <*> o .:? "trustable" .!= IsNotTrustable
 
 instance ToJSON LocalRootPeersGroup where
   toJSON lrpg =
@@ -142,6 +148,7 @@ instance ToJSON LocalRootPeersGroup where
       , "advertise" .= rootAdvertise (localRoots lrpg)
       , "hotValency" .= hotValency lrpg
       , "warmValency" .= warmValency lrpg
+      , "trustable" .= trustable lrpg
       ]
 
 newtype LocalRootPeersGroups = LocalRootPeersGroups
@@ -164,22 +171,32 @@ instance FromJSON PublicRootPeers where
 instance ToJSON PublicRootPeers where
   toJSON = toJSON . publicRoots
 
-data NetworkTopology = RealNodeTopology !LocalRootPeersGroups ![PublicRootPeers] !UseLedger
+data NetworkTopology = RealNodeTopology { ntLocalRootPeersGroups :: !LocalRootPeersGroups
+                                        , ntPublicRootPeers      :: ![PublicRootPeers]
+                                        , ntUseLedgerPeers       :: !UseLedgerPeers
+                                        , ntUseBootstrapPeers    :: !UseBootstrapPeers
+                                        }
   deriving (Eq, Show)
 
 instance FromJSON NetworkTopology where
   parseJSON = withObject "NetworkTopology" $ \o ->
-                RealNodeTopology <$> (o .: "localRoots"                                     )
-                                 <*> (o .: "publicRoots"                                    )
-                                 <*> (o .:? "useLedgerAfterSlot" .!= UseLedger DontUseLedger)
+                RealNodeTopology <$> (o .: "localRoots"                                  )
+                                 <*> (o .: "publicRoots"                                 )
+                                 <*> (o .:? "useLedgerAfterSlot" .!= DontUseLedgerPeers  )
+                                 <*> (o .:? "bootstrapPeers" .!= DontUseBootstrapPeers)
 
 instance ToJSON NetworkTopology where
   toJSON top =
     case top of
-      RealNodeTopology lrpg prp ul -> object [ "localRoots"         .= lrpg
-                                             , "publicRoots"        .= prp
-                                             , "useLedgerAfterSlot" .= ul
-                                             ]
+      RealNodeTopology { ntLocalRootPeersGroups
+                       , ntPublicRootPeers
+                       , ntUseLedgerPeers
+                       , ntUseBootstrapPeers
+                       } -> object [ "localRoots"         .= ntLocalRootPeersGroups
+                                   , "publicRoots"        .= ntPublicRootPeers
+                                   , "useLedgerAfterSlot" .= ntUseLedgerPeers
+                                   , "bootstrapPeers"     .= ntUseBootstrapPeers
+                                   ]
 
 --
 -- Legacy p2p topology file format
@@ -195,10 +212,12 @@ instance FromJSON (Legacy a) => FromJSON (Legacy [a]) where
 instance FromJSON (Legacy LocalRootPeersGroup) where
   parseJSON = withObject "LocalRootPeersGroup" $ \o -> do
                 hv@(HotValency v) <- o .: "hotValency"
+                wv <- o .:? "warmValency" .!= WarmValency v
                 fmap Legacy $ LocalRootPeersGroup
                   <$> o .: "localRoots"
                   <*> pure hv
-                  <*> pure (WarmValency v)
+                  <*> pure wv
+                  <*> o .: "trustable"
 
 instance FromJSON (Legacy LocalRootPeersGroups) where
   parseJSON = withObject "LocalRootPeersGroups" $ \o ->
@@ -213,9 +232,10 @@ instance FromJSON (Legacy PublicRootPeers) where
 instance FromJSON (Legacy NetworkTopology) where
   parseJSON = fmap Legacy
             . withObject "NetworkTopology" (\o ->
-                RealNodeTopology <$> fmap getLegacy (o .: "LocalRoots")
-                                 <*> fmap getLegacy (o .: "PublicRoots")
-                                 <*> (o .:? "useLedgerAfterSlot" .!= UseLedger DontUseLedger))
+              RealNodeTopology <$> fmap getLegacy (o .: "LocalRoots")
+                               <*> fmap getLegacy (o .: "PublicRoots")
+                               <*> (o .:? "useLedgerAfterSlot" .!= DontUseLedgerPeers)
+                               <*> pure DontUseBootstrapPeers)
 
 -- | Read the `NetworkTopology` configuration from the specified file.
 --
@@ -228,7 +248,12 @@ readTopologyFile tr nc = do
     Left e -> return . Left $ handler e
     Right bs ->
       let bs' = LBS.fromStrict bs in
-      first handlerJSON (eitherDecode bs')
+      (case eitherDecode bs' of
+        Left err -> Left (handlerJSON err)
+        Right t
+          | isValidTrustedPeerConfiguration t -> Right t
+          | otherwise                         -> Left handlerBootstrap
+      )
       `combine`
       first handlerJSON (eitherDecode bs')
 
@@ -256,6 +281,13 @@ readTopologyFile tr nc = do
     , "configuration flag. "
     , Text.pack err
     ]
+  handlerBootstrap :: Text
+  handlerBootstrap = mconcat
+    [ "You seem to have not configured any trustable peers. "
+    , "This is important in order for the node to make progress "
+    , "in bootstrap mode. Make sure you provide at least one bootstrap peer "
+    , "source. "
+    ]
 
 readTopologyFileOrError :: Tracer IO (StartupTrace blk)
                         -> NodeConfiguration -> IO NetworkTopology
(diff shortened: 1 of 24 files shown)


cardano-node/src/Cardano/Node/Configuration/TopologyP2P.hs
```diff
@@ -137,6 +142,7 @@ instance FromJSON LocalRootPeersGroup where
                   <$> parseJSON (Object o)
                   <*> pure hv
                   <*> o .:? "warmValency" .!= WarmValency v
+                  <*> o .: "peerTrustable"
```
bolt12, line 145: make this an optional field for backwards compatibility. Default to not trustable

cardano-node/src/Cardano/Node/Configuration/TopologyP2P.hs
```diff
@@ -216,9 +227,10 @@ instance FromJSON (Legacy PublicRootPeers) where
 instance FromJSON (Legacy NetworkTopology) where
   parseJSON = fmap Legacy
             . withObject "NetworkTopology" (\o ->
-                RealNodeTopology <$> fmap getLegacy (o .: "LocalRoots")
+              RealNodeTopology <$> fmap getLegacy (o .: "LocalRoots")
                                  <*> fmap getLegacy (o .: "PublicRoots")
-                                 <*> (o .:? "useLedgerAfterSlot" .!= UseLedger DontUseLedger))
+                                 <*> (o .:? "useLedgerPeers" .!= DontUseLedgerPeers)
+                                 <*> pure DontUseBootstrapPeers)
```
bolt12, line 233:
```suggestion
                               <*> fmap getLegacy (o .: "PublicRoots")
                               <*> (o .:? "useLedgerPeers" .!= DontUseLedgerPeers)
                               <*> pure DontUseBootstrapPeers)
```

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -122,6 +122,9 @@ import           Cardano.Node.TraceConstraints (TraceConstraints)
 import           Cardano.Tracing.Tracers
 import           Ouroboros.Network.PeerSelection.PeerSharing (PeerSharing (..))
 import           Ouroboros.Network.PeerSelection.State.LocalRootPeers (HotValency, WarmValency)
+import Ouroboros.Network.PeerSelection.LedgerPeers.Type (UseLedgerPeers)
+import Ouroboros.Network.PeerSelection.PeerTrustable (PeerTrustable)
+import Ouroboros.Network.PeerSelection.Bootstrap (UseBootstrapPeers)
```
bolt12, line 127:
```suggestion
import           Ouroboros.Network.PeerSelection.LedgerPeers.Type (UseLedgerPeers)
import           Ouroboros.Network.PeerSelection.PeerTrustable (PeerTrustable)
import           Ouroboros.Network.PeerSelection.Bootstrap (UseBootstrapPeers)
```

cardano-node/src/Cardano/Node/Tracing/Tracers/P2P.hs
```diff
@@ -802,10 +815,10 @@ instance LogFormatting PeerSelectionCounters where
         (fromIntegral hotBigLedgerPeers)
     , IntM
         "Net.PeerSelection.WarmLocalRoots"
-        (fromIntegral $ foldl' (\a (b, _) -> a + b) 0 localRoots)
+        (fromIntegral $ getHotValency $ foldl' (\a (b, _) -> a + b) 0 localRoots)
     , IntM
         "Net.PeerSelection.HotLocalRoots"
-        (fromIntegral $ foldl' (\a (_, b) -> a + b) 0 localRoots)
+        (fromIntegral $ getWarmValency $ foldl' (\a (_, b) -> a + b) 0 localRoots)
```
bolt12, line 821: These are swapped. Look at the string, good thing we have the newtypes

cardano-node/test/Test/Cardano/Node/Gen.hs
```diff
@@ -191,8 +195,27 @@ genPublicRootPeers =
...
+genDomainAccessPoint :: Gen DomainAccessPoint
+genDomainAccessPoint =
+    DomainAccessPoint <$> domains <*> Gen.integral Range.constantBounded
+  where
+    domains = Gen.element [ "test1"
+              , "test2"
+              , "test3"
+              , "test4"
+              , "test5"
+              ]
```
bolt12, line 213:
```suggestion
    domains = Gen.element 
                [ "test1"
                , "test2"
                , "test3"
                , "test4"
                , "test5"
                ]
```

bolt12 (commented):
Missing Changelog and check configuration file for any chance of being able to sync if in bootstrap mode.

cardano-node/src/Cardano/Node/Configuration/TopologyP2P.hs
```diff
@@ -167,22 +174,24 @@ instance FromJSON PublicRootPeers where
 instance ToJSON PublicRootPeers where
   toJSON = toJSON . publicRoots
 
-data NetworkTopology = RealNodeTopology !LocalRootPeersGroups ![PublicRootPeers] !UseLedger
+data NetworkTopology = RealNodeTopology !LocalRootPeersGroups ![PublicRootPeers] !UseLedgerPeers !UseBootstrapPeers
```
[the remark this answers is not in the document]
coot: :thinking: we should have called `UseBootstrapPeers` in `ouroboros-network` as `BootstrapPeersConfig`.

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
```diff
@@ -466,6 +467,16 @@ instance HasSeverityAnnotation (TracePeerSelection addr) where
 
       TraceDemoteBigLedgerPeersAsynchronous {} -> Warning
 
+      TraceUseBootstrapPeersChanged {} -> Info
+      TraceBootstrapPeersFlagChangedWhilstInSensitiveState -> Info
+
+      TraceLedgerStateJudgementChanged {} -> Info
+      TraceOnlyBootstrapPeers {}          -> Info
```
coot, line 474: I think this should be a `Notice`.  For end users, it doesn't matter as they are not reading log files, but for SPOs this should trigger a warning.

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
```diff
@@ -466,6 +467,16 @@ instance HasSeverityAnnotation (TracePeerSelection addr) where
...
+
+      TraceOutboundGovernorCriticalFailure {} -> Error
+
+      TraceDebugState {} -> Info
```
coot, line 478: That's for sure ought to be `Debug` level:
```suggestion
      TraceDebugState {} -> Debug
```

nix/nixos/cardano-node-service.nix
```diff
@@ -513,6 +516,16 @@ in {
         '';
       };
 
+      useBootstrapPeers = mkOption {
+        type = types.nullOr (types.either types.bool (types.listOf types.attrs));
+        default = false;
+        description = ''
+          If set, it will enable bootstrap peers.
+          To disable set this to false.
```
coot, line 524: :thinking: so we parse `false`, maybe it's enough to leave it unset - then no extra parsing is required.

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
```diff
@@ -1553,6 +1564,14 @@ instance ToJSON PeerSelectionTargets where
                  , "targetActiveBigLedgerPeers" .= nActiveBigLedgerPeers
                  ]
 
+instance ToJSON peerAddr => ToJSON (PublicRootPeers peerAddr) where
+  toJSON prp =
+    Aeson.object [ "kind" .= String "PublicRootPeers"
```
coot, line 1569: We often do the `kind` thing, but it would be more in the spirit of `json` to communicate this through the field name that includes the record.  But that would be a large change and I am not sure if it's worth the trouble.

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
bolt12: Maybe you are right 😄

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
karknu: Nope. In spite of the name these log messages are manually triggered.
Having them at `Debug` means that no SPO will ever have them enabled when we want the data for troobleshooting.

cardano-node/src/Cardano/Node/Configuration/TopologyP2P.hs
```diff
@@ -198,10 +207,12 @@ instance FromJSON (Legacy a) => FromJSON (Legacy [a]) where
 instance FromJSON (Legacy LocalRootPeersGroup) where
   parseJSON = withObject "LocalRootPeersGroup" $ \o -> do
                 hv@(HotValency v) <- o .: "hotValency"
+                wv <- o .:? "warmValency" .!= WarmValency v
                 fmap Legacy $ LocalRootPeersGroup
                   <$> o .: "localRoots"
                   <*> pure hv
-                  <*> pure (WarmValency v)
```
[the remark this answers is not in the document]
bolt12:
```
  [ "You are using legacy p2p topology file format."
  , "See https://github.com/intersectmbo/cardano-node/issues/4559"
  , "Note that the legacy p2p format will be removed in `1.37` release."
  ]
```

😬

cardano-node/src/Cardano/Node/Configuration/TopologyP2P.hs
```diff
@@ -259,6 +276,14 @@ readTopologyFile tr nc = do
...
     , Text.pack err
     ]
+  handlerBootstrap :: Text
+  handlerBootstrap = mconcat
+    [ "You seem to have not configured any trustable peer. "
+    , "This is important in order for the node to make progress "
+    , "in bootstrap mode. Make sure you provide at least one bootstrap peer "
+    , "source (by setting useBootstrapPeer topology file option) "
+    , "or mark a local root peer as trustable. "
+    ]
```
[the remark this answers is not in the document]
bolt12: > `bootstrap mode`: do we explain somewhere what is boostrap mode? Or maybe it would be better to say:
> 
> > This is important for the node to be able to make progress when syncing.

I have a pr for cardano-node-wiki where I explain what boot strap peers are: https://github.com/input-output-hk/cardano-node-wiki/pull/39/files#diff-743fb54f52af163e2ec1e256e645df88e52976e97c677f5aaa461821c4768f58R157-R189

cardano-node/src/Cardano/Node/Configuration/TopologyP2P.hs
```diff
@@ -167,22 +174,24 @@ instance FromJSON PublicRootPeers where
...
 
 instance FromJSON NetworkTopology where
   parseJSON = withObject "NetworkTopology" $ \o ->
-                RealNodeTopology <$> (o .: "localRoots"                                     )
-                                 <*> (o .: "publicRoots"                                    )
-                                 <*> (o .:? "useLedgerAfterSlot" .!= UseLedger DontUseLedger)
+                RealNodeTopology <$> (o .: "localRoots"                                  )
+                                 <*> (o .: "publicRoots"                                 )
+                                 <*> (o .:? "useLedgerAfterSlot" .!= DontUseLedgerPeers  )
+                                 <*> (o .:? "useBootstrapPeers" .!= DontUseBootstrapPeers)
```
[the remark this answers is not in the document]
bolt12: Yeah I think that makes sense

nix/nixos/cardano-node-service.nix
bolt12: I like having it so people looking at the source code can be clear what's it set to in default

cardano-node/src/Cardano/Node/Configuration/TopologyP2P.hs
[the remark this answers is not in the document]
bolt12: Should I remove it completely? @coot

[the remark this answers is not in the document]
coot: Yes, let's remove it :)

nix/nixos/cardano-node-service.nix
coot: What if we interpret an empty list as `DontUseBootstrapPeers`?

bolt12: We'd loose the ability of configuring a node to only use local roots as bootstrap peers.

coot: This conversation is resolved, but note that one can repeat a local roots in bootstrap peers.

cardano-node/src/Cardano/Node/Configuration/TopologyP2P.hs
```diff
@@ -256,6 +281,14 @@ readTopologyFile tr nc = do
     , "configuration flag. "
     , Text.pack err
     ]
+  handlerBootstrap :: Text
+  handlerBootstrap = mconcat
+    [ "You seem to have not configured any trustable peers. "
+    , "This is important in order for the node to make progress "
+    , "in bootstrap mode. Make sure you provide at least one bootstrap peer "
+    , "source (by setting 'bootstrapPeer' topology file option) "
```
coot, line 289: Maybe we don't need to mention how to enable bootstrap peers, since this warning is shown only if bootstrap peers are enabled.

nix/nixos/cardano-node-service.nix
bolt12: I don't think that's avoidable. What we decided was that if that's the case that's a misconfiguration and not really our fault, right?

cardano-node/src/Cardano/Node/Configuration/TopologyP2P.hs
```diff
@@ -264,3 +297,25 @@ readTopologyFileOrError tr nc =
...
+--
+-- Checking for chance of progress in bootstrap phase
+--
+
+-- | This function returns false if non-trustable peers are configured
+--
+hasTrustablePeers :: NetworkTopology -> Bool
+hasTrustablePeers (RealNodeTopology (LocalRootPeersGroups lprgs) _ _ ubp) =
+    case ubp of
+      DontUseBootstrapPeers   -> True
```
coot, line 310: I don't have a better name than `hasTrustablePeers`, but this name is a bit misleading if `DontUseBootstrapPeers` returns `True`.

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -840,21 +858,17 @@ producerAddressesNonP2P nt =
...
-            )
-            (groups lrpg)
-      , foldMap (Map.fromList . rootConfigToRelayAccessPoint . publicRoots) prp
-      )
-
-useLedgerAfterSlot
-  :: NetworkTopology
-  -> UseLedgerAfter
-useLedgerAfterSlot (RealNodeTopology _ _ (UseLedger ul)) = ul
+  -> ([(HotValency, WarmValency, Map RelayAccessPoint (PeerAdvertise, PeerTrustable))], Map RelayAccessPoint PeerAdvertise)
```
coot, line 861: I'd split the line since it's not very readable.

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -840,21 +858,17 @@ producerAddressesNonP2P nt =
...
+producerAddresses RealNodeTopology { ntLocalRootPeersGroups
+                                   , ntPublicRootPeers
+                                   } =
+  ( map (\lrp -> ( hotValency lrp
+                 , warmValency lrp
+                 , Map.fromList $ map (fmap (, trustable lrp))
```
coot, line 867: This is a bit ugly that we need to add this information this way. We cannot simply extend `RootConfig` because that type is used in various places.  A nicer way would be to provide `LocalRootConfig` type:
```hs
data LocalRootConfig = LocalRootConfig
  { localRootAccessPoints :: [RelayAccessPoint]
    -- ^ a list of relay access points, each of which is either an ip address
    -- or domain name and a port number.
  , localRootAdvertise    :: PeerAdvertise,
    -- ^ 'advertise' configures whether the root should be advertised through
    -- gossip.
  , localRootTrustable    :: Bool
    -- ^ ...
  } deriving (Eq, Show)

```

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
```diff
@@ -2438,11 +2478,50 @@ instance ToJSON addr
...
+                            <> "Encountered: " <> show invalid
+
+instance ToJSON LedgerStateJudgement where
+  toJSON YoungEnough = String "YoungEnough"
+  toJSON TooOld      = String "TooOld"
+
+instance FromJSON LedgerStateJudgement where
+  parseJSON (String "YoungEnough") = pure YoungEnough
+  parseJSON (String "TooOld")      = pure TooOld
+  parseJSON _                      = fail "Invalid JSON for LSJ"
```
coot, line 2505: such acronyms can be very frustrating in an error message.
```suggestion
  parseJSON _                      = fail "Invalid JSON for LedgerStateJudgement"
```

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
```diff
@@ -2438,11 +2478,50 @@ instance ToJSON addr
...
+  toJSON (UseLedgerPeers Always)             = Number 0
+  toJSON (UseLedgerPeers (After (SlotNo s))) = Number (fromIntegral s)
+
+instance ToJSON UseBootstrapPeers where
+    toJSON DontUseBootstrapPeers   = Null
+    toJSON (UseBootstrapPeers dps) = toJSON dps
+
+instance FromJSON UseBootstrapPeers where
+    parseJSON Null = pure DontUseBootstrapPeers
+    parseJSON v    = UseBootstrapPeers <$> parseJSON v
```
coot, line 2518: indentation

cardano-node/src/Cardano/Node/Configuration/TopologyP2P.hs
bolt12: Good point! I'll think of something different!

cardano-node/src/Cardano/Tracing/OrphanInstances/Network.hs
bolt12: where exactly?

coot: The file is indenting with 2 spaces not 4.

bolt12:
Merged on #481

coot:
@bolt12 I think you mean #5691.

bolt12:
Yes, my bad!

PR #5501, opened 2023-10-10
workbench: fix plan path logic

andreabedini:
The workbench shell can be in "cabal mode" or "nix mode".

In cabal mode, cardano-node is built locally with cabal, while in nix
mode cardano-node is provided pre-compiled by nix.

While preparing the manifest we check WB_MODE_CABAL to know whether we
should look for a build plan locally or in the nix store. This is to
make sure we don't pick a stray build plan left over from a previous
local build.


mgmeier (approved):
Thank you @andreabedini . LGTM!

PR #5509, opened 2023-10-17
Workbench | locli: various improvements

mgmeier:
# Description

* fix Plutus workload parse
* adjust analysis to solo node profiles
* decrease RAM req. for analyses on top of new tracing
* further small fixes

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [x] Self-reviewed the diff

### DIFF
--- a/bench/locli/src/Cardano/Analysis/API/Context.hs
+++ b/bench/locli/src/Cardano/Analysis/API/Context.hs
@@ -75,8 +75,8 @@ data PlutusParams
 instance FromJSON PlutusParams where
   parseJSON = withObject "PlutusParams" $ \v ->
     PlutusParams
-      <$> v .: "type"
-      <*> v .: "script"
+      <$> v .:? "type"    .!= ""
+      <*> v .:? "script"  .!= ""
 
 instance ToJSON PlutusParams where
   toJSON PlutusParams{..} =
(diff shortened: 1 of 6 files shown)


bench/locli/src/Cardano/Analysis/API/Context.hs
NadiaYvette, line 1: I ran into this too.

bench/locli/src/Cardano/Analysis/BlockProp.hs
NadiaYvette, line 1: This looks like golfing went wrong. Maybe the golfing could've been done as `take 1 . scanr1 min`, though golfing in general can throw people inconveniently when they need to actually get things done.

NadiaYvette (approved):
From what I can see, this helps with the Plutus config issue test runs are tripping over and fixes a few others, too.

bench/locli/src/Cardano/Analysis/BlockProp.hs
mgmeier: Indeed; the code implicitly assumed there is at least one peer fetch. However, the workbench has solo node profiles, where no peer fetch hapens.

bench/locli/src/Cardano/Analysis/API/Context.hs
mgmeier: The issue here was we process external JSON. This sometimes has `"key": {}` instead of `"key": null`, as `aeson` would do. So the derived `FromJSON` instance trips over this, as it expects the value to be parseable when it, being an empty object, should result in a `Nothing` value.

PR #5519, opened 2023-10-26
Include commits from source-repos in workbench manifest

andreabedini:
Currently `wb manifest` obtains a git commit hash of packages from chap by matching them by package-id.

This gives the wrong result if a package comes from a source-repository-package stanza. The package-id would match something on chap but the commit hash would be a different one.

This PR reworks the logic of `wb manifest` such that srp are correcntly identified and reported.

Unfortunately there are issues in both cabal-mode and nix mode:

In a workbench shell in cabal mode, many dependencies will be pre-installed, in that case we cannot tell where it came from. To simplify, we take the following approach:
- if it is from a source-repo use that
- otherwise see if it matches any package on chap
- otherwise return null
This means that in case a package exists on both hackage and chap, we will always report the commit from chap.

In nix mode; haskell.nix erases the location of source-repository-packages during the build so, nix-build-srps will again miss their hash.

### DIFF
--- a/nix/workbench/default.nix
+++ b/nix/workbench/default.nix
@@ -9,7 +9,7 @@ with lib;
 
 let
   # recover CHaP location from cardano's project
-  chapPackages = "${cardanoNodeProject.args.inputMap."https://input-output-hk.github.io/cardano-haskell-packages"}/foliage/packages.json";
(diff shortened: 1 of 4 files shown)


mgmeier (approved):
LGTM, thank you @andreabedini

PR #5535, opened 2023-11-07
workbench: various improvements + future proofing

mgmeier:
# Description

* Compiler version is now documented and reported from a deployed node cluster
* Constrain building `PlutusV1` and `PlutusV2`  to `plc 1.0`
* use profile RTS overrides for a cluster setup
* workbench's remote environment is now parametrizable
* shell script fixes for machines with 4 and less CPU cores
* set unlimited `ChainSyncIdleTimeout` for (upcoming) config option

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [x] Self-reviewed the diff


angerman (approved):
Looks sensible. Nothing that raised any obvious red flags.

NadiaYvette (approved):
The protocol parameter change is great! Good catch on the Plutus version in Loop.hs too.

nix/workbench/analyse/analyse.sh
NadiaYvette, line 1: It makes sense that a minimum number of kernel threads are needed. I think the -c option outputs 0 when no match is available, though if for some reason /proc/ isn't mounted where expected e.g. in some weirdly-configured container/chroot/etc., it could error as is handled in the script. Theoretically, this could golf the awk pipeline, though how few people could read it & understand what it does without fishing through the manuals might make it better to keep the awk pipeline:

```shell
# will fail if container/chroot doesn't mount /proc/ -- add 2>/dev/null || echo 0 to handle
num_threads=`grep -c processor /proc/cpuinfo`
num_threads=${num_threads/#%[0-5]/6}
```

I think there are two pieces that make it counterintuitive. First, things like ^ and $ are more typically embedded in the pattern for most pattern syntaxes. Second, the manuals don't document that # and % can be combined to do something like what ^pat$ does in more normal pattern syntaxes.

Icelandjack (approved):

nix/workbench/analyse/analyse.sh
mgmeier: 1. `/proc` mount should not be assumed - that's a valid case on MacOS e.g. I'll make the whole thing Mac compatible :) Thank you for pointing that out.
2. I tested, and the pattern unfortunately doesn't capture changing a 0-5 to a default 6. Will keep `awk` around for the time being, and yes, it's a little more expressive, too.

PR #5539, opened 2023-11-15
Add darwin.signingUtils

angerman:
Thanks to @KtorZ for noticing this.

on macOS:
```
$ nix build .#cardano-node -L
```
should produce `./result/bin/cardano-node`.

Prior to this change
```
$ codesign --verify ./result/bin/cardano-node 
./result/bin/cardano-node: invalid signature (code or signature have been modified)
```
after this change
```
$ codesign --verify ./result/bin/cardano-node -v
./result/bin/cardano-node: valid on disk
./result/bin/cardano-node: satisfies its Designated Requirement
```

### DIFF
--- a/nix/set-git-rev.nix
+++ b/nix/set-git-rev.nix
@@ -1,11 +1,19 @@
 { pkgs }:
 drv:
-pkgs.buildPackages.runCommand drv.name
+with pkgs;
+buildPackages.runCommand drv.name
 {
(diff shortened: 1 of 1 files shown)


yvan-sraka (approved):
I just tested it on my macOS M1 and confirm that it worked as expected :)

michalrus (approved):

Issue #5272, opened 2023-05-25
Optimize new tracing


jutaro:
## Performance enhancements for tracing, metrics, datapoints and forwarding.

Old tracing should soon be complemented and then superseded by new tracing. New tracing has a lot of enhancements:

- All tracers are always on and respect the configured severity
- All tracers can be configured individually based on hierarchical namespaces. This is possible down to individual messages.
- Traces are forwarded to a cardano-tracer process, which can handle the traces of many nodes
- When old tracing will be gone, tracing can be reconfigured in a running node
- A complete documentation of all trace messages, metrics and datapoints can be automatically generated.  Please see the generated tracing configuration reference at: [trace-messages](https://github.com/input-output-hk/cardano-node/blob/master/doc/new-tracing/tracers_doc_generated.md#trace-messages
)

All these enhancement should not burden cardano-node with additional performance penalties. This issue has been opened to measure the performance and profiling the running node in different scenarios, and then find optimizations for new tracing.

jutaro:
###Branch: jutaro/benchmark-new
###Commit: 5081370e513db717874606d3e009fa1f4513d1dd
###Log text lines emitted per host: 4120.6111111

<pre>
#+TITLE: 2023-05-24-10-28-50813_variance_8.1.0
#+SUBTITLE: filters: unfiltered, slot range: raw(0-402, I 402 total) filtered(0-402, I 402 total), filtered 1.0
#+DATE: 2023-05-24 10:34:30 UTC
#+VERSION: locli 1.29 (508137)
|---+---------------------------------------+---------+-------+----------+--------------------------------------------------------------|
| ! |                             Parameter | time, % | range | alloc, % |                                              source location |
|---+---------------------------------------+---------+-------+----------+--------------------------------------------------------------|
| # |                    decodeWithByteSpan |     3.5 |   1.8 |      5.9 |                src/Codec/CBOR/Decoding.hs:(1005,1)-(1009,29) |
| # |                                 trans |     2.2 |   0.8 |      1.9 |            src/Control/State/Transition/Extended.hs:416:1-36 |
| # |                              cpure_NP |     2.2 |   1.7 |      3.7 |                          src/Data/SOP/NP.hs:(256,1)-(258,29) |
| # |                                  fn_3 |     1.9 |   0.8 |      2.7 |                             src/Data/SOP/Classes.hs:147:1-55 |
[...]

PR #5541, opened 2023-11-17
Optimize new tracing further

jutaro:
This pull request addresses the remaining performance penalty observed after the optimization of the new tracing mechanism introduced in pull request #5272. The goal is to further optimize the code to minimize the performance impact associated with the new tracing features.

Changes Made
Implemented additional optimizations to mitigate the performance penalty against old tracing.
The optimization efforts aim to balance the introduction of new features with the need for improved performance.

### DIFF
--- a/trace-dispatcher/src/Cardano/Logging/Trace.hs
+++ b/trace-dispatcher/src/Cardano/Logging/Trace.hs
@@ -37,7 +37,7 @@ module Cardano.Logging.Trace (
   , withInnerNames
   ) where
 
-import           Control.Monad (join)
+import           Control.Monad (forM_, join)
 import           Control.Monad.IO.Unlift
 import qualified Control.Tracer as T
 import           Data.Maybe (isJust)
@@ -307,9 +307,7 @@ contramapMCond (Trace tr) mFunc =
     where
       rFunc arg = do
         condMes <- mFunc arg
-        case condMes of
-          Nothing -> pure ()
-          Just mes -> T.traceWith tr mes
+        forM_ condMes (T.traceWith tr)
 
 {-# INLINE contramapM' #-}
 contramapM' :: Monad m
--- a/trace-dispatcher/src/Cardano/Logging/Trace.hs
+++ b/trace-dispatcher/src/Cardano/Logging/Trace.hs
@@ -181,7 +181,6 @@ withSeverity (Trace tr) = Trace $
         (lc, Left e) -> (lc, Left e))
       tr
   where
-    {-# INLINE process #-}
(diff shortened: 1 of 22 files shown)


trace-dispatcher/src/Cardano/Logging/Trace.hs
```diff
@@ -248,6 +268,7 @@ withDetails (Trace tr) = Trace $
         (lc, Left e) -> (lc, Left e))
       tr
   where
+    {-# INLINE process #-}
```
mgmeier, line 271: Only top-level definitions can be marked `INLINE`. As it's ignored here, can we remove it?

trace-dispatcher/src/Cardano/Logging/Tracer/Composed.hs
```diff
@@ -87,24 +88,23 @@ mkCardanoTracer' trStdout trForward mbTrEkg tracerPrefix hook = do
...
-                        pure (metricsFormatter "" ekgTrace)
+                        pure (metricsFormatter ekgTrace)
 --                      >>= recordMetricsStatistics internalTr
                         >>= maybeSilent hasNoMetrics tracerPrefix True
                         >>= hook
 
     pure (messageTrace <> metricsTrace)
 
   where
+    {-# INLINE addContextAndFilter #-}
```
mgmeier, line 99: Same as above, `INLINE` works on top-level definitions only.

trace-dispatcher/src/Cardano/Logging/Types.hs
```diff
@@ -53,7 +53,7 @@ module Cardano.Logging.Types (
 
 
 import           Codec.Serialise (Serialise (..))
-import           Data.Aeson ((.=))
+-- import           Data.Aeson ((.=))
```
mgmeier, line 56: Please remove commented out line (+ the ones below).

trace-dispatcher/src/Cardano/Logging/Trace.hs
jutaro: Done

trace-dispatcher/src/Cardano/Logging/Tracer/Composed.hs
jutaro: Done

trace-dispatcher/src/Cardano/Logging/Types.hs
jutaro: Done

mgmeier (approved):
Looks great, thank you @jutaro

PR #5645, opened 2024-01-19
cardano-testnet: allow to pass genesis files

smelc:
# Description

We want to be able to tweak genesis files in `cardano-testnet`. This PR makes it possible to pass custom alonzo, conway, and shelley genesis files to `createTestnet`. A new function `createTestnetDefault` is introduced, that uses default genesis files for all 3.

# Context

This PR replaces https://github.com/IntersectMBO/cardano-node/pull/5643, whose design was deemed unsatisfying.

# Checklist

- [ ] New tests are added if needed and existing tests are updated.
- [X] Self-reviewed the diff


HOLYGHOST1111 (approved):

HOLYGHOST1111 (approved):

HOLYGHOST1111 (approved):

cardano-testnet/src/Testnet/Components/Configuration.hs
```diff
@@ -81,10 +82,12 @@ numSeededUTxOKeys = 3
 createSPOGenesisAndFiles
   :: (MonadTest m, MonadCatch m, MonadIO m, HasCallStack)
   => CardanoTestnetOptions
+  -> Maybe (ShelleyGenesis StandardCrypto) -- ^ The shelley genesis to use, otherwise a default is picked
```
Jimbo4350, line 85: My thoughts are we shouldn't use `Maybe` here. I want to force the caller to be explicit in what genesis they provide.

cardano-testnet/src/Testnet/Defaults.hs
```diff
@@ -412,13 +413,15 @@ defaultByronProtocolParamsJsonValue =
...
+                                               --   Some fields are overridden by the accompanying 'CardanoTestnetOptions'
   -> Api.ShelleyGenesis StandardCrypto
-defaultShelleyGenesis startTime testnetOptions =
+defaultShelleyGenesis startTime testnetOptions mBase =
   let testnetMagic = cardanoTestnetMagic testnetOptions
       slotLength = cardanoSlotLength testnetOptions
       epochLength = cardanoEpochLength testnetOptions
       maxLovelaceLovelaceSupply = cardanoMaxSupply testnetOptions
-  in Api.shelleyGenesisDefaults
+  in (fromMaybe Api.shelleyGenesisDefaults mBase)
```
Jimbo4350, line 424: So the problem here is if a user did provide a shelley genesis we would be overriding several fields.

cardano-testnet/src/Testnet/Start/Cardano.hs
```diff
@@ -122,8 +128,17 @@ startTimeOffsetSeconds = if OS.isWin32 then 90 else 15
 -- > │   └── node-spo{1,2,3}
 -- > └── utxo-keys
 -- >     └── utxo{1,2,3}.{addr,skey,vkey}
-cardanoTestnet :: CardanoTestnetOptions -> Conf -> H.Integration TestnetRuntime
-cardanoTestnet testnetOptions Conf {tempAbsPath} = do
+cardanoTestnet :: ()
+  => CardanoTestnetOptions
+  -> Conf
+  -> Maybe (ShelleyGenesis StandardCrypto) -- ^ The shelley genesis to use, otherwise a default is picked
+                                           --   Some fields are overridden by the accompanying 'CardanoTestnetOptions'
```
Jimbo4350, line 135: Same as above. I want callers to be explicit. We can leave a comment guiding them to the default genesis values.

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Cli/Alonzo/LeadershipSchedule.hs
```diff
@@ -0,0 +1,514 @@
+{-# LANGUAGE BlockArguments #-}
+{-# LANGUAGE DisambiguateRecordFields #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
```
Jimbo4350, line 6: Why was this added?

Jimbo4350 (changes requested):

Jimbo4350:
A viable option is exposing a `cardanoTestnetDefault` function that uses all of our defaults.

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Cli/Alonzo/LeadershipSchedule.hs
smelc: Copied this file to write a new test and erroneously `git add`ed it :shrug:

Removed :+1:

cardano-testnet/src/Testnet/Defaults.hs
smelc: Yes, but this is mitigated by the additional documentation I have added in all the call chain about this.

Also, discussing a similar API with @andreabedini, this overriding is the default expectation of callers (not doing it would be the surprise!).

So I'd keep it like this.

cardano-testnet/src/Testnet/Components/Configuration.hs
smelc: Done :+1:

cardano-testnet/src/Testnet/Start/Cardano.hs
smelc: Done :+1:

cardano-testnet/src/Testnet/Components/Configuration.hs
```diff
@@ -78,35 +79,32 @@ createConfigYaml (TmpAbsolutePath tempAbsPath') anyCardanoEra' = GHC.withFrozenC
...
 
+newtype NumPools = NumPools { unNumPools :: Int }
+
 createSPOGenesisAndFiles
   :: (MonadTest m, MonadCatch m, MonadIO m, HasCallStack)
-  => CardanoTestnetOptions
-  -> UTCTime -- ^ Start time
+  => NumPools
+  -> AnyCardanoEra -- ^ The era to use
+  -> ShelleyGenesis StandardCrypto -- ^ The shelley genesis to use.
```
smelc, line 88: I did this change to avoid mixing `CardanoTestnetOptions` and genesis files in the same signature. We should avoid having them together in a signature, because they contain duplicate data, e,g, the network magic is both in `CardanoTestnetOptions` and the genesis files.

So if we have them together, we are imposing on the caller to be careful to pass coherent data. This change avoids that, by using smaller types instead.

cardano-testnet/src/Testnet/Start/Cardano.hs
```diff
@@ -79,6 +90,19 @@ data ForkPoint
 startTimeOffsetSeconds :: DTC.NominalDiffTime
 startTimeOffsetSeconds = if OS.isWin32 then 90 else 15
 
+-- | Like 'cardanoTestnet', but using defaults for all configuration files.
+-- See 'cardanoTestnet' for additional documentation.
+cardanoTestnetDefault :: ()
+  => CardanoTestnetOptions
+  -> Conf
+  -> H.Integration TestnetRuntime
```
smelc, line 98: Here, we're not mixing `CardanoTestnetOptions` and genesis files, which is good (as I explained elsewhere). This is nice for the vast majority of callers, that don't need to be exposes to a complex API.

cardano-testnet/src/Testnet/Start/Cardano.hs
```diff
@@ -122,14 +146,38 @@ startTimeOffsetSeconds = if OS.isWin32 then 90 else 15
...
-  currentTime <- H.noteShowIO DTC.getCurrentTime
-  let tempAbsPath' = unTmpAbsPath tempAbsPath
+cardanoTestnet :: ()
+  => CardanoTestnetOptions -- ^ The options to use. Must be consistent with the genesis files.
+  -> Conf
+  -> UTCTime -- ^ The starting time. Must be the same as the one in the shelley genesis.
+  -> ShelleyGenesis StandardCrypto -- ^ The shelley genesis to use, for example 'Defaults.defaultShelleyGenesis'.
+                                   --   Some fields are overridden by the accompanying 'CardanoTestnetOptions'.
+  -> AlonzoGenesis -- ^ The alonzo genesis to use, for example 'Defaults.defaultAlonzoGenesis'.
+  -> ConwayGenesis StandardCrypto -- ^ The conway genesis to use, for example 'Defaults.defaultConwayGenesis'.
```
smelc, line 156: Here we mix `CardanoTestnetOptions` and genesis files. So callers must be careful to provide coherent data. If they don't, they will get caught by the new checks below. Ideally, in the future, we'll want to remove fields from `CardanoTestnetOptions`, so that the intersection with the data of genesis files becomes empty. This a bigger change though, so I chose this version, which is an improvement already, and minimizes disturbing users.

cardano-testnet/src/Testnet/Components/Configuration.hs
Jimbo4350: What about the `ConwayGenesis` and `AlonzoGenesis` parameters?

Jimbo4350: We should not expose the data constructor for `NumPools` and expose smart constructors instead: e.g `numPools :: CardanoTestnetOptions -> NumPoolNodes`

Jimbo4350 (approved):
LGTM! However you need to make the change regarding my comments about the smart constructors.

carbolymer (approved):

cardano-testnet/src/Testnet/Components/Configuration.hs
smelc: > What about the ConwayGenesis and AlonzoGenesis parameters?

@Jimbo4350> since `create-testnet-data` doesn't yet take alonzo and conway genesis as CLI parameters, they do not make sense here yet. Adding them is tracked by https://github.com/IntersectMBO/cardano-cli/issues/548

smelc: > We should not expose the data constructor for NumPools and expose smart constructors instead: e.g numPools :: CardanoTestnetOptions -> NumPoolNodes

Done :+1:

PR #5704, opened 2024-03-05
wb | bench storage

fmaste:
# Description

- Remove Cardano World QA profiles
- - QA class machines of Cardano World Nomad cluster are gone, support for it is removed along with any sensible defaults that were added to avoid deployment errors when both cluster (CW and the P&T exclusive one) were available.
- - Changes were made having in mind support for sporadic clusters to bench different hardware configurations.
- Add ssh utilities for perf Nomad client nodes
- - Commands were renamed to resolve the confusion it makes using Nomad "client" and Nomad "node" interchangeably. Nomad CLI uses `nomad nodes` to query the agents but we are trying to use "node" for a Cardano cluster node and "client" for a Nomad agent running as client (a client used to run Cardano nodes =) ).
- - Subcommands added/renamed: `wb nomad clients ready|machines|ssh`
- - `wb nomad clients ready`: Returns JSON array with "id", "name", "datacenter" and "ip" of all SRE-managed P&T exclusive Nomad clients available (status=ready).
- - `wb nomad clients machines`: Like `ready` but adds EC2 machines info for fingerprinting, to be used to ensure cloud runs are reproducible.
- - `wb nomad clients ssh (all|producers|CLIENT_NAME)`: self explanatory, `producers` is without the `explorer` node.
- Check available storage space before starting a cloud run
- - We defined the minimum we want available for the "value" profile.
- Replace idle-nomadperf with fast-nomadperf variants
- - A profile that tests the whole cycle.
- Better error reporting when incompatible genesis file format
- - With genesis files around 500 mega that have an incompatible format the node fails further in time and the workbench was assuming it started without errors, showing a misleading message later when the healthcheck failed.
- - The time needed for supervisord to consider a node as "running" now depends on the number of UTxO generated.
- Fix cloud runs that were not fetching the entrypoints' stdout and stderr
- - This only happened with cloud runs that had finished without problems.
- Add latency service and profile for the 52 perf nodes
- - A profile were every nodes makes multiple pings with different sizes to its producers.
- - The output format is not defined and there's no automatic parsing of the results.
- Use the regions defined in the profile with no assumptions.
- - Removed `validateNodeSpecs`.
- - Only differentiate between region "loopback" and the others. 
- Add support for "perf-ssd"
- - Using Nomad namespace and class abstractions we can support both types of machines.
- - Now the profile includes things that were part of the workbench, like the Nomad address, machine types, if we can download the logs using ssh, etc.
- - Made NOMAD_CLIENTS (pinning) work with any profile
- - Commands like `wb clients` now use the profile to build the Nomad server queries.
- Small changes

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [x] Self-reviewed the diff

# Note on CI
for you.


mgmeier (approved):
Excellent, @fmaste , thank you very much.

PR #5746, opened 2024-04-03
cardano-testnet: Avoid rewrite of sgMaxLovelaceSupply

smelc:
# Description

Avoid this [hacky rewrite](https://github.com/IntersectMBO/cardano-node/blob/master/cardano-testnet/src/Testnet/Components/Configuration.hs#L164) that was necessary for the [treasury growth test](https://github.com/IntersectMBO/cardano-node/blob/master/cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/LedgerEvents/TreasuryGrowth.hs) to pass (without the rewrite the treasury was negative).

# How to trust this PR

The treasury growth test now passes without the genesis' rewrite (try it with `PARALLEL_TESTNETS=1 DISABLE_RETRIES=1 cabal test cardano-testnet-test --test-options '-p "/Treasury Growth/"'`)

# Checklist

- [X] CI passes. See note on CI.
- [X] Self-reviewed the diff

### DIFF
--- a/cardano-testnet/src/Testnet/Components/Configuration.hs
+++ b/cardano-testnet/src/Testnet/Components/Configuration.hs
@@ -146,8 +146,8 @@ createSPOGenesisAndFiles (NumPools numPoolNodes) (NumDReps numDelReps) era shell
     , "--spec-conway",  inputGenesisConwayFp
     , "--testnet-magic", show testnetMagic
     , "--pools", show numPoolNodes
-    , "--total-supply",     show @Int 2_000_000_000_000
-    , "--delegated-supply", show @Int 1_000_000_000_000
+    , "--total-supply",     show @Int 2_000_000_000_000 -- 2 trillions
+    , "--delegated-supply", show @Int 1_000_000_000_000 -- 1 trillion
     , "--stake-delegators", show numStakeDelegators
     , "--utxo-keys", show numSeededUTxOKeys
     , "--drep-keys", show numDelReps
--- a/cardano-testnet/src/Testnet/Components/Configuration.hs
+++ b/cardano-testnet/src/Testnet/Components/Configuration.hs
@@ -166,12 +166,6 @@ createSPOGenesisAndFiles (NumPools numPoolNodes) (NumDReps numDelReps) era shell
 
   H.renameFile (tempAbsPath </> "byron-gen-command" </> "genesis.json") (genesisByronDir </> "genesis.json")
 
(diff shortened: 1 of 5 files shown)


cardano-testnet/src/Testnet/Components/Configuration.hs
```diff
@@ -137,8 +137,8 @@ createSPOGenesisAndFiles (NumPools numPoolNodes) era shelleyGenesis alonzoGenesi
     , "--spec-conway",  inputGenesisConwayFp
     , "--testnet-magic", show testnetMagic
     , "--pools", show numPoolNodes
-    , "--total-supply",     show @Int 2_000_000_000_000
-    , "--delegated-supply", show @Int 1_000_000_000_000
+    , "--total-supply",     show @Int 2_000_000_000_000 -- 2 trillions
```
Jimbo4350, line 140: So we are now overwriting the `inputGenesisShelleyFp`. We don't want to do this.

Jimbo4350 (changes requested):

cardano-testnet/src/Testnet/Components/Configuration.hs
Jimbo4350: It would be better to either:
- Disallow values that result in a negative treasury
- Add a check that will fail if the treasury in the node state is negative

smelc: @Jimbo4350> Please note that this was the case before this PR, I'm not changing this behavior here :slightly_smiling_face:

> Disallow values that result in a negative treasury

> Add a check that will fail if the treasury in the node state is negative

That was actually my first instinct, but alas, looking at the ledger's spec; it seems quite involved to catch cases that will cause treasury to be negative (at least with my level of knowledge) and also it seemed wrong to have this code here. I would favor reusing the ledger's core that does this (if any).

> We don't to hard code values unless its specifically contained in a default Haskell value

Agreed, but I would rather do that in a follow-up PR. This PR is an improvement over the current state, so it makes sense as a first step.

Jimbo4350: > Please note that this was the case before this PR, I'm not changing this behavior here 

Yep definitely wasn't blaming you. It was likely me that put those values there :clown_face:

> Agreed, but I would rather do that in a follow-up PR.

:+1:

Jimbo4350 (approved):

PR #5764, opened 2024-04-09
NodeVersion tracer for providing metrics

jutaro:
This adds a new trace message, which adds new metrics. Here is an example of what it provides:
```
cardano_build_info {
, version = "8.10.0"
, version_major="8"
, version_minor="10"
, version_patch="0"
, revision="fd7eb00d1697b0c6eb7f3e3d0ad6e94fcc5caa63"
, compiler_name="ghc"
, compiler_version="8.10"
, compiler_version_major="8"
, compiler_version_minor="10"
, architecture="x86_64"
, os_name="linux" 
} 1 
```
*NB* in the actual Prometheus output, this is 1 line only; newlines have been inserted here for readability only.

Additionally, it provided an alternative representation of the same data as trace messages for both machines and humans. Here are examples of how the data will be represented:

Machine Trace Message:
{"at":"2024-04-08T12:30:08.89384608Z","ns":"Version.NodeVersion","data":{"applicationName":"cardano-node","applicationVersion":"8.10.0","architecture":"x86_64","compilerName":"ghc","compilerVersion":"8.10","gitRevision":"f3fabe88bfc83f5e9c2a9d7af2293857cd8212c3","osName":"linux"},"sev":"Info","thread":"5","host":"deusXmachina"}

Human Trace Message:
[2024-04-08 13:46:26.5169Z][deusXmachina:Version.NodeVersion](Info,5) cardano-node 8.10.0 git rev f3fabe88bfc83f5e9c2a9d7af2293857cd8212c3 - linux-x86_64 - ghc-8.10

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -32,7 +32,7 @@ packages:
   trace-forward
 
 program-options
-  ghc-options: -Werror
+--  ghc-options: -Werror
 
 test-show-details: direct
 
--- a/cabal.project
+++ b/cabal.project
@@ -32,7 +32,7 @@ packages:
   trace-forward
 
 program-options
---  ghc-options: -Werror
+  ghc-options: -Werror
 
 test-show-details: direct
 
--- a/cardano-node/src/Cardano/Node/Run.hs
+++ b/cardano-node/src/Cardano/Node/Run.hs
@@ -288,8 +288,12 @@ handleNodeWithTracers cmdPc nc0 p networkMagic blockType runP = do
             >>= mapM_ (traceWith $ startupTracer tracers)
 
           traceWith (nodeVersionTracer tracers) getNodeVersion
+
+          blockForging <- snd (Api.protocolInfo runP)
           traceWith (startupTracer tracers)
-                    (BlockForgingUpdate NotEffective)
+                    (BlockForgingUpdate (if null blockForging
+                                          then DisabledBlockForging
+                                          else EnabledBlockForging))
 
           -- We ignore peer logging thread if it dies, but it will be killed
           -- when 'handleSimpleNode' terminates.
--- a/cardano-node/src/Cardano/Node/Run.hs
+++ b/cardano-node/src/Cardano/Node/Run.hs
@@ -8,8 +8,10 @@
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE PackageImports #-}
 {-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TupleSections #-}
+
+{-# LANGUAGE TypeApplications #-}
+
 
 {-# OPTIONS_GHC -Wno-unused-imports #-}
 
@@ -74,6 +76,8 @@ import           Paths_cardano_node (version)
 
 import qualified Cardano.Crypto.Init as Crypto
 
+import           Cardano.Node.Tracing.Tracers.NodeVersion (getNodeVersion)
+
 import           Cardano.Node.Configuration.Logging (LoggingLayer (..), createLoggingLayer,
                    nodeBasicInfo, shutdownLoggingLayer)
 import           Cardano.Node.Configuration.NodeAddress
@@ -102,12 +106,12 @@ import qualified Ouroboros.Network.Diffusion.P2P as P2P
 import           Ouroboros.Network.NodeToClient (LocalAddress (..), LocalSocket (..))
 import           Ouroboros.Network.NodeToNode (AcceptedConnectionsLimit (..), ConnectionId,
                    PeerSelectionTargets (..), RemoteAddress)
+import           Ouroboros.Network.PeerSelection.Bootstrap (UseBootstrapPeers (..))
+
 import           Ouroboros.Network.PeerSelection.RelayAccessPoint (RelayAccessPoint (..))
 import           Ouroboros.Network.Protocol.ChainSync.Codec
 import           Ouroboros.Network.Subscription (DnsSubscriptionTarget (..),
                    IPSubscriptionTarget (..))
-import           Ouroboros.Network.PeerSelection.Bootstrap
-                     (UseBootstrapPeers (..))
 
 import           Cardano.Node.Configuration.Socket (SocketOrSocketInfo (..),
                    gatherConfiguredSockets, getSocketOrSocketInfoAddr)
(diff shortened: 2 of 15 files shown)


trace-dispatcher/src/Cardano/Logging/Tracer/EKG.hs
```diff
@@ -84,3 +89,9 @@ ekgTracer storeOrServer = liftIO $ do
                         Right server -> creator2 name server
             let rgsMap' = Map.insert name gauge rgsMap
             pure (rgsMap', gauge)
+
+presentPrometheusM :: [(Text, Text)] -> Text
+presentPrometheusM labels = "{" <> foldr (\(k, v) acc -> k <> "=\"" <> v <>
+                                    (if k == fst (last labels)
+                                      then "\" " <> acc
+                                      else "\", " <> acc)) "" labels <> "} 1"
```
mgmeier, line 97: I find that a little hard to approach, and I think one can avoid checking being the last key in the list. Can we use something more expressive, like e.g.:
```
presentPrometheusM =
  label . map pair
  where
    label pairs = "{" <> intercalate "," pairs <> "} 1"
    pair (k, v) = k <> "=\"" <> v <> "\""
```

trace-dispatcher/src/Cardano/Logging/Tracer/EKG.hs
```diff
@@ -84,3 +89,9 @@ ekgTracer storeOrServer = liftIO $ do
                         Right server -> creator2 name server
             let rgsMap' = Map.insert name gauge rgsMap
             pure (rgsMap', gauge)
+
```
mgmeier, line 92: Could you provide in a comment what the implemented output format is? Like what to expect when presenting a `[("key1", "value1"), ("key2", "value2")]` for example?

cardano-node/src/Cardano/Node/Tracing/Tracers/NodeVersion.hs
```diff
@@ -0,0 +1,137 @@
...
+
+getNodeVersion :: NodeVersionTrace
+getNodeVersion =
+  let applicationName = "cardano-node"
+      applicationVersion = version
+      osName = pack os
+      architecture = pack arch
+      compilerName = pack System.Info.compilerName
+      compilerVersion = System.Info.compilerVersion
+
```
mgmeier, line 51: `fullCompilerVersion` appears first in `base-4.15-0.0`. To reliably get the patch version number, could we use conditionals like
```
#if MIN_VERSION_base(4,15,0)
[...]
```

For older versions, we could add (e.g. for GHC8.10.7) some sensible fallbacks based on the version bundled with a compiler, and/or make the patch version a `Maybe` type.
```
#elif MIN_VERSION_base(4,14,3) [... -> patch version 7]
#elif MIN_VERSION_base(4,14,1) [... -> patch version 2]
```

mgmeier: IMHO the version of `base` is the decisive factor - does it provide `fullCompilerVersion` or not? So making a choice based on that is independent of GHC version.

trace-dispatcher/src/Cardano/Logging/Tracer/EKG.hs
mgmeier: ❤️

cabal.project
NadiaYvette, line 1: If a warning is unavoidable, would it make more sense to use a `{-# OPTIONS -Wno-whatever-warning #-}` sort of pragma in the beginning of just the relevant file(s) vs. removing `-Werror` for the whole project?

cardano-node/src/Cardano/Node/Tracing/Tracers/NodeVersion.hs
jutaro: Done and tested

cabal.project
jutaro: I use it for me locally always, was an error to push this

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -280,12 +287,9 @@ handleNodeWithTracers cmdPc nc0 p networkMagic blockType runP = do
...
             >>= mapM_ (traceWith $ startupTracer tracers)
 
-          blockForging <- snd (Api.protocolInfo runP)
+          traceWith (nodeVersionTracer tracers) getNodeVersion
           traceWith (startupTracer tracers)
-                    (BlockForgingUpdate (if null blockForging
-                                          then EnabledBlockForging
-                                          else DisabledBlockForging))
-
+                    (BlockForgingUpdate NotEffective)
```
mgmeier, line 292: use equivalent logic from `TraceDispatcher{}` case

cardano-node/src/Cardano/Node/Tracing/Tracers/NodeVersion.hs
```diff
@@ -0,0 +1,153 @@
...
+    [("cardano_version_major", "Cardano node version information")
+    ,("cardano_version_minor", "Cardano node version information")
+    ,("cardano_version_patch", "Cardano node version information")
+    ,("haskell_compiler_major", "Cardano compiler version information")
+    ,("haskell_compiler_minor", "Cardano compiler version information")
+
+#if MIN_VERSION_base(4,15,0)
+    ,("haskell_compiler_patch", "Cardano compiler version information")
+#endif
+    ,("cardano_build_info", "Cardano node build info")
```
mgmeier, line 85: <3

cardano-node/src/Cardano/Tracing/Tracers.hs
```diff
@@ -672,6 +690,25 @@ sendEKGDirectDouble ekgDirect name val = do
...
+      Nothing -> do
+        label <- EKG.getLabel name (ekgServer ekgDirect)
+        Label.set label (presentPrometheusM labels)
+        pure $ Map.insert name label registeredMap
+  where
+    presentPrometheusM :: [(Text, Text)] -> Text
+    presentPrometheusM =
+      label . map pair
+      where
+        label pairs = "{" <> Text.intercalate "," pairs <> "} 1"
```
mgmeier, line 709: please double-check Prometheus page to not contain `... 1 1`

mgmeier (commented):
Additionally, please add full namespace to
```
BlockForgingUpdate b -> sendEKGDirectInt ekgDirect' "forging_enabled"
```

cardano-node/src/Cardano/Tracing/Tracers.hs
jutaro: You were right

mgmeier (approved):
Thanks @jutaro , great work, and very useful.

PR #5772, opened 2024-04-09
Remove orphan ToJSON instances that were moved to ledger

neilmayhew:
# Description

These instances were added to ledger in IntersectMBO/cardano-ledger#4250. This PR shouldn't be merged until node is updated to use the corresponding versions of the ledger packages.

The golden tests pass with the current instances. They can be used to validate the new instances and should then be discarded (in #5776) because they will no longer be testing code that's in node.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [x] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/cardano-node/cardano-node.cabal
+++ b/cardano-node/cardano-node.cabal
@@ -248,6 +248,8 @@ test-suite cardano-node-test
                       , cardano-crypto-wrapper
                       , cardano-api:{cardano-api, internal}
                       , cardano-ledger-core
+                      , cardano-ledger-allegra
+                      , cardano-ledger-alonzo
                       , cardano-node
                       , cardano-slotting
                       , directory
@@ -276,5 +278,6 @@ test-suite cardano-node-test
                         Test.Cardano.Node.POM
                         Test.Cardano.Tracing.NewTracing.Consistency
                         Test.Cardano.Tracing.OrphanInstances.HardFork
+                        Test.Cardano.Tracing.OrphanInstances.Shelley
 
   ghc-options:          -threaded -rtsopts "-with-rtsopts=-N -T"
--- a/cardano-node/cardano-node.cabal
+++ b/cardano-node/cardano-node.cabal
@@ -149,8 +149,8 @@ library
                       , cardano-crypto-class
                       , cardano-crypto-wrapper
                       , cardano-git-rev ^>=0.2.2
-                      , cardano-ledger-alonzo
-                      , cardano-ledger-allegra
+                      , cardano-ledger-alonzo >=1.7.1
+                      , cardano-ledger-allegra >=1.4.1
                       , cardano-ledger-api
                       , cardano-ledger-babbage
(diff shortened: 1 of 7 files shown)


neilmayhew:
cc @Jimbo4350

smelc:
@neilmayhew> since the tests were not launched on the original instances:

image

We don't really know if the instances from the ledger behave the same :upside_down_face: Maybe remove the commit removing the local instances until the time when we'll be able to merge this PR?

neilmayhew:
> @neilmayhew> since the tests were not launched on the original instances:
> 
> We don't really know if the instances from the ledger behave the same 🙃

We do, because I ran them locally on a detached head. :smile:

Also, my PR for ledger uses identical golden files (copied from this PR).

> Maybe remove the commit removing the local instances until the time when we'll be able to merge this PR?

That would be unfortunate, because the point of this PR was to reduce the work needed when the time comes to integrate with a newer ledger, by making it easier to find the instances that need to be removed.

I was assuming that whoever does the integration would do something similar to what I did above, by running the tests manually before and after the integration.

However, perhaps the least confusing thing would be to split this PR into three:

1. The golden tests, which would be merged right away.
2. Removing the instances, which would wait until integration.
3. Removing the golden tests, which would be merged right after the preceding PR (2).

If this is more acceptable to you, I'll create new PRs for (1) and (3) and update this one to become (2).

smelc:
> However, perhaps the least confusing thing would be to split this PR into three:
> 
> 1. The golden tests, which would be merged right away.
> 1. Removing the instances, which would wait until integration.
> 1. Removing the golden tests, which would be merged right after the preceding PR (2).

Yes I think this is the best way, to make sure nothing can go wrong during the integration. This procedures avoids the integrator  to have to understand in details what's happening, he will anyway be safe!

So I've approved https://github.com/IntersectMBO/cardano-node/pull/5775 and this one will serve as reference for the integration.

Thanks @neilmayhew :heart:

neilmayhew:
cc @Lucsanszky

cardano-node/cardano-node.cabal
```diff
@@ -149,8 +149,8 @@ library
                       , cardano-crypto-class
                       , cardano-crypto-wrapper
                       , cardano-git-rev ^>=0.2.2
-                      , cardano-ledger-alonzo
-                      , cardano-ledger-allegra
```
smelc, line 153: I think the bounds can be removed now, because here are the versions chosen on a recent version of `master`:

```shell
→ ghc-pkg list | grep ledger-alonzo
    cardano-ledger-alonzo-1.8.0.0
→ ghc-pkg list | grep ledger-allegra
    cardano-ledger-allegra-1.4.1.0
```

neilmayhew: Agreed. I'll close this PR now.

neilmayhew:
Superseded by recent integration

PR #5773, opened 2024-04-10
LogFormatting and MetaTrace instances for V2.FlavorImplSpecificTrace

fraser-iohk:
# Description

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
+++ b/cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
@@ -2090,10 +2090,42 @@ instance MetaTrace V1.BackingStoreValueHandleTrace where
     , Namespace [] ["Statted"]
     ]
 
-
 instance LogFormatting V2.FlavorImplSpecificTrace where
+  forMachine _dtal V2.FlavorImplSpecificTraceInMemory =
+    mconcat [ "kind" .= String "InMemory" ]
+  forMachine _dtal V2.FlavorImplSpecificTraceOnDisk =
+    mconcat [ "kind" .= String "OnDisk" ]
+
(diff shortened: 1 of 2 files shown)


cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
```diff
@@ -2090,10 +2090,42 @@ instance MetaTrace V1.BackingStoreValueHandleTrace where
...
+  namespaceFor V2.FlavorImplSpecificTraceInMemory =
+    Namespace [] ["InMemory"]
+  namespaceFor V2.FlavorImplSpecificTraceOnDisk =
+    Namespace [] ["OnDisk"]
+
+  severityFor (Namespace _ ["InMemory"]) _ = Just Info
+  severityFor (Namespace _ ["OnDisk"])   _ = Just Info
+  severityFor _                          _ = Nothing
+
+  -- suspicious
```
jasagredo, line 2114: You could have omitted this one. It defaults to public

jasagredo (approved):
They indeed look weird because they are empty but I think we can merge this

cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
fraser-iohk: I wanted to add explicit cases (and the "suspicious" comment) because I thought it'd be the kind of thing that's easy to forget about if / when other constructors get added -- I don't think it makes a big difference though, so I don't mind just using the default 🤷

Issue #5751, opened 2024-04-04
Missing metric for current forging mode on a blockproducer node


gitmachtl:
With the push to P2P also on the BlockProducer node, SPOs need to change there backup/failover infrastructure.

A test with node 8.9.1 showed, turning the blockproduction off on a node via a SIGHUP signal (and removed credential files), all last metrics of the blockproducer stays. So there is currently no way to detect the forging mode of a running node that way.

We would need an additional metric (accessable via prometheus interface or a cli query) with value 1 = forging and 0 = not forging, to be sure we know what the node is currently doing. And the backup/failover infra can handle the workmode accordingly.

Trying to get this data out of the logfiles is not a nice way, and i am sure we can add this metric to all the other available values.

coot:
We install a `SIGHUP` handler in `cardano-node` ([see](https://github.com/intersectmbo/cardano-node/blob/717b38ce896231a571202ee1524c6d88aebdc972/cardano-node/src/Cardano/Node/Run.hs#L690)) which contains the necessary information; this only requires adding an EKG counter in the node, which makes it quite easy to implement (no need to modify anywhere deeper in the stack, e.g. `ouroboros-consensus`).

gitmachtl:
@coot thx, please make sure that the EKG metric is also reported correctly as false(0) if the blockproducer node was started with the `--non-producing-node` option. that should be the default in an active/standby blockproducer backup infra. start it up with the credentials but in non producing mode. check if the state to promote it as an active producer is ok, if so, reload the settings via a SIGHUP signal and start forging.

coot:
I don't think I will be working on it, but I'll bring it to the attention of the core team.

mgmeier:
The metric `forging_enabled` (formerly `can_forge_blocks`) exists and should provide the required info.
However, it doesn't seem to observe the `non-producing-node` command line option yet.

PR #5780, opened 2024-04-12
'can_forge_blocks' metric

jutaro:
I have prepared a pull request to add a new Prometheus metric called 'can_forge_blocks'. This metric indicates whether a blockchain node has the capability to forge new blocks on the network.

The metric will be a boolean type, where a value of 1 signifies that the node can forge blocks, and 0 indicates otherwise.

This addition will provide valuable insights into the operational status and capabilities of our blockchain nodes, aiding in monitoring and decision-making processes.


gitmachtl:
so this is closing https://github.com/IntersectMBO/cardano-node/issues/5751

Issue #5599, opened 2023-12-21
DRep Activity


Jimbo4350:
The ledger marks DReps as inactive if they have not voted recently (determined by a parameter). This issue tests that, in that case, their stake is not considered to be part of the voting stake.

PR #5796, opened 2024-04-23
Add test for DRep Activity mechanism

palas:
# Description

This PR adds a test for DRep activity mechanism (which addresses [this issue](https://github.com/IntersectMBO/cardano-node/issues/5599)), that ensures that stake associated to DReps that are not actively participating in governance are not counted towards the minimum participation required for approving proposals.

We test it by:
- Creating a testnet with a single default DRep (which expires in 100 epochs)
- Have that DRep set the `drepActivity` param to `6`
- Register two new DReps (which are affected immediately by the new `drepActivity` param)
- Distribute the stake evenly among all the DReps
- Carry out some voting activity while they expire (this is to ensure the number of dormant epochs is low)
- Carry out a new proposal after the expiration time and vote `yes` only with the default DRep
- Check that only the proposal after the expiration passes due to inactivity

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [x] Self-reviewed the diff

### DIFF
--- a/cardano-testnet/src/Testnet/Components/DReps.hs
+++ b/cardano-testnet/src/Testnet/Components/DReps.hs
@@ -10,6 +10,7 @@ module Testnet.Components.DReps
   , signTx
   , submitTx
   , failToSubmitTx
+  , retrieveTransactionId
   ) where
 
 import           Cardano.Api (AnyCardanoEra (..), FileDirection (In), ShelleyBasedEra (..),
@@ -304,3 +305,24 @@ failToSubmitTx execConfig cEra signedTx = GHC.withFrozenCallStack $ do
   case exitCode of
     ExitSuccess -> H.failMessage GHC.callStack "Transaction submission was expected to fail but it succeeded"
     _ -> return ()
+
+-- | Retrieves the transaction ID (governance action ID) from a signed
+-- transaction file using @cardano-cli@.
+--
+-- This function takes the following parameters:
+--
+-- * 'execConfig': Specifies the CLI execution configuration.
+-- * 'signedTx': Signed transaction to be submitted, obtained using 'signTx'.
+--
+-- Returns the transaction ID (governance action ID) as a 'String'.
+retrieveTransactionId
+  :: (MonadTest m, MonadCatch m, MonadIO m)
+  => H.ExecConfig
+  -> File SignedTx In
+  -> m String
+retrieveTransactionId execConfig signedTxBody = do
+  txidOutput <- H.execCli' execConfig
+    [ "transaction", "txid"
+    , "--tx-file", unFile signedTxBody
+    ]
+  return $ mconcat $ lines txidOutput
--- a/cardano-testnet/src/Testnet/Components/DReps.hs
+++ b/cardano-testnet/src/Testnet/Components/DReps.hs
@@ -1,12 +1,10 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE ExistentialQuantification #-}
-{-# LANGUAGE InstanceSigs #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TypeApplications #-}
 
 module Testnet.Components.DReps
-  ( SomeKeyPair(..)
-  , VoteFile
+  ( VoteFile
   , generateDRepKeyPair
   , generateRegistrationCertificate
   , createCertificatePublicationTxBody
@@ -46,8 +44,8 @@ import           System.FilePath ((</>))
 import           Testnet.Components.Query (EpochStateView, findLargestUtxoForPaymentKey,
                    getCurrentEpochNo, getMinDRepDeposit, waitUntilEpoch)
 import qualified Testnet.Process.Run as H
-import           Testnet.Runtime (PaymentKeyInfo (paymentKeyInfoAddr, paymentKeyInfoPair),
-                   PaymentKeyPair (..), StakingKeyPair (StakingKeyPair, stakingSKey))
+import           Testnet.Runtime (KeyPair(..), PaymentKeyInfo (paymentKeyInfoAddr, paymentKeyInfoPair),
+                   PaymentKeyPair (..), StakingKeyPair (..), SomeKeyPair (..))
 import           Testnet.Start.Types (anyEraToString)
 
 import           Hedgehog (MonadTest, evalMaybe)
@@ -200,23 +198,6 @@ createVotingTxBody execConfig epochStateView sbe work prefix votes wallet = do
 -- Transaction signing
 
 data SignedTx
-
-class KeyPair a where
-  secretKey :: a -> FilePath
-
-instance KeyPair PaymentKeyPair where
-  secretKey :: PaymentKeyPair -> FilePath
-  secretKey = paymentSKey
-
-instance KeyPair StakingKeyPair where
-  secretKey :: StakingKeyPair -> FilePath
-  secretKey = stakingSKey
-
-data SomeKeyPair = forall a . KeyPair a => SomeKeyPair a
-
-instance KeyPair SomeKeyPair where
-  secretKey :: SomeKeyPair -> FilePath
-  secretKey (SomeKeyPair x) = secretKey x
 
 -- | Calls @cardano-cli@ to signs a transaction body using the specified key pairs.
 --
--- a/cardano-testnet/src/Testnet/Components/DReps.hs
+++ b/cardano-testnet/src/Testnet/Components/DReps.hs
@@ -1,20 +1,26 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE InstanceSigs #-}
 {-# LANGUAGE TypeApplications #-}
 
 module Testnet.Components.DReps
-  ( generateDRepKeyPair
+  ( SomeKeyPair(..)
+  , VoteFile
+  , generateDRepKeyPair
   , generateRegistrationCertificate
-  , createDRepRegistrationTxBody
+  , createCertificatePublicationTxBody
   , generateVoteFiles
   , createVotingTxBody
   , signTx
   , submitTx
   , failToSubmitTx
   , retrieveTransactionId
+  , registerDRep
+  , delegateToDRep
   ) where
 
-import           Cardano.Api (AnyCardanoEra (..), FileDirection (In), ShelleyBasedEra (..),
-                   renderTxIn)
+import           Cardano.Api (AnyCardanoEra (..), ConwayEra, EpochNo (EpochNo), FileDirection (In),
+                   ShelleyBasedEra (..), ToCardanoEra (toCardanoEra), renderTxIn)
 
 import           Cardano.CLI.Types.Common (File (..))
 
@@ -30,9 +36,11 @@ import           GHC.Stack (HasCallStack)
 import qualified GHC.Stack as GHC
 import           System.FilePath ((</>))
 
-import           Testnet.Components.Query (EpochStateView, findLargestUtxoForPaymentKey)
+import           Testnet.Components.Query (EpochStateView, findLargestUtxoForPaymentKey,
+                   getCurrentEpochNo, getMinDRepDeposit, waitUntilEpoch)
 import qualified Testnet.Process.Run as H
-import           Testnet.Runtime (PaymentKeyInfo (paymentKeyInfoAddr), PaymentKeyPair (..))
+import           Testnet.Runtime (PaymentKeyInfo (paymentKeyInfoAddr, paymentKeyInfoPair),
+                   PaymentKeyPair (..), StakingKeyPair (StakingKeyPair, stakingSKey))
 import           Testnet.Start.Types (anyEraToString)
 
 import           Hedgehog (MonadTest)
@@ -66,7 +74,7 @@ generateDRepKeyPair execConfig work prefix = do
 
 -- DRep registration certificate generation
 
-data DRepRegistrationCertificate
+data Certificate
 
 -- | Generates a registration certificate for a decentralized representative (DRep)
 -- using @cardano-cli@.
@@ -90,7 +98,7 @@ generateRegistrationCertificate
   -> String
   -> PaymentKeyPair
   -> Integer
-  -> m (File DRepRegistrationCertificate In)
+  -> m (File Certificate In)
 generateRegistrationCertificate execConfig work prefix drepKeyPair depositAmount = do
   let dRepRegistrationCertificate = File (work </> prefix <> ".regcert")
   void $ H.execCli' execConfig [ "conway", "governance", "drep", "registration-certificate"
@@ -104,48 +112,46 @@ generateRegistrationCertificate execConfig work prefix drepKeyPair depositAmount
 
 data TxBody
 
--- | Composes a decentralized representative (DRep) registration transaction body
---   (without signing) using @cardano-cli@.
+-- | Composes a certificate publication transaction body (without signing) using @cardano-cli@.
 --
 -- This function takes seven parameters:
 --
 -- * 'execConfig': Specifies the CLI execution configuration.
 -- * 'epochStateView': Current epoch state view for transaction building. It can be obtained
 --                     using the 'getEpochStateView' function.
--- * 'sbe': The Shelley-based era (e.g., 'ShelleyEra') in which the transaction will be constructed.
+-- * 'sbe': The Shelley-based era (e.g., 'ShelleyBasedEraShelley') in which the transaction will be constructed.
 -- * 'work': Base directory path where the transaction body file will be stored.
 -- * 'prefix': Prefix for the output transaction body file name. The extension will be @.txbody@.
--- * 'drepRegCert': The file name of the registration certificate for the DRep, obtained using
---                  'generateRegistrationCertificate'.
+-- * 'certificate': The file name of the certificate.
 -- * 'wallet': Payment key information associated with the transaction,
 --             as returned by 'cardanoTestnetDefault'.
 --
 -- Returns the generated @File TxBody In@ file path to the transaction body.
-createDRepRegistrationTxBody
+createCertificatePublicationTxBody
   :: (H.MonadAssertion m, MonadTest m, MonadCatch m, MonadIO m)
   => H.ExecConfig
   -> EpochStateView
   -> ShelleyBasedEra era
   -> FilePath
   -> String
-  -> File DRepRegistrationCertificate In
+  -> File Certificate In
   -> PaymentKeyInfo
   -> m (File TxBody In)
-createDRepRegistrationTxBody execConfig epochStateView sbe work prefix drepRegCert wallet = do
+createCertificatePublicationTxBody execConfig epochStateView sbe work prefix cert wallet = do
   let dRepRegistrationTxBody = File (work </> prefix <> ".txbody")
   walletLargestUTXO <- findLargestUtxoForPaymentKey epochStateView sbe wallet
   void $ H.execCli' execConfig
     [ "conway", "transaction", "build"
     , "--change-address", Text.unpack $ paymentKeyInfoAddr wallet
     , "--tx-in", Text.unpack $ renderTxIn walletLargestUTXO
-    , "--certificate-file", unFile drepRegCert
+    , "--certificate-file", unFile cert
     , "--witness-override", show @Int 2
     , "--out-file", unFile dRepRegistrationTxBody
     ]
   return dRepRegistrationTxBody
 
--- DRep vote file generation
-data DRepVoteFile
+-- Vote file generation
+data VoteFile
 
 -- | Generates decentralized representative (DRep) voting files (without signing)
 -- using @cardano-cli@.
@@ -163,7 +169,7 @@ data DRepVoteFile
 --               representing the DRep key pair and a 'String' representing the
 --               vote type (i.e: "yes", "no", or "abstain").
 --
--- Returns a list of generated @File DRepVoteFile In@ representing the paths to
+-- Returns a list of generated @File VoteFile In@ representing the paths to
 -- the generated voting files.
 generateVoteFiles :: (MonadTest m, MonadIO m, MonadCatch m)
   => H.ExecConfig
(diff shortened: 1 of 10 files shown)


palas:
Hm, it looks like the test is extemely flaky at the moment, I will look into that.

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/LedgerEvents/Gov/DRepActivity.hs
```diff
@@ -0,0 +1,301 @@
...
+  voteChangeProposal execConfig epochStateView sbe gov "vote2"
+                     secondGovernanceActionTxId secondGovernanceActionIndex secondPropVotes wallet1
+
+  (EpochNo epochAfterSecondProp) <- getCurrentEpochNo epochStateView sbe
+  H.note_ $ "Epoch after second prop: " <> show epochAfterSecondProp
+
+  void $ waitUntilEpoch (File configurationFile) (File socketPath) (EpochNo (epochAfterSecondProp + 5))
+  dRepActivityAfterSecondProp <- getDRepActivityValue execConfig
+
+  dRepActivityAfterSecondProp === 4 -- This is what we set in first prop, second prop should fail
```
Jimbo4350: Not for this PR, however it would be nice to pattern match on the expected error via: 
```
data TxValidationError era where
  ByronTxValidationError
    :: Consensus.ApplyTxErr Consensus.ByronBlock
    -> TxValidationError era

  ShelleyTxValidationError
    :: ShelleyBasedEra era
    -> Consensus.ApplyTxErr (Consensus.ShelleyBlock (ConsensusProtocol era) (ShelleyLedgerEra era))
    -> TxValidationError era
```

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/LedgerEvents/Gov/DRepActivity.hs
```diff
@@ -0,0 +1,301 @@
...
+
+  result2 <- checkDRepState sbe (File configurationFile) (File socketPath) execConfig
+                            (Just . map (drepExpiry . snd) . Map.toList)
+  H.note_ $ "DRep expiration dates: " <> show result2
```
Jimbo4350: Again here we should be asserting what the expected expiration epoch is supposed to be,

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/LedgerEvents/Gov/DRepActivity.hs
```diff
@@ -0,0 +1,301 @@
...
+
+  !propSubmittedResult <- findCondition (maybeExtractGovernanceActionIndex sbe (fromString governanceActionTxId))
+                                        (unFile configurationFile)
+                                        (unFile socketPath)
+                                        (EpochNo 30)
+
+  governanceActionIndex <- case propSubmittedResult of
+                             Left e ->
+                               H.failMessage callStack
+                                 $ "findCondition failed with: " <> displayError e
```
Jimbo4350, line 267: "makeActivityChangeProposal.findCondition"

Jimbo4350 (commented):
There is a lot going on in this PR. The simplest approach would have been potentially two properties:

1a. Modify Conway genesis with ridiculously low drep activity i.e 1 Epoch
2a. Test that the drep can no longer participate in voting (wait till 1 epoch has passed)

2a. Modify Conway genesis with slightly more reasonably drep activity i.e 2 Epochs
2b. Test that the drep can participate in voting before 2 Epochs have passed.

This should give us two smaller properties (code wise) that should run relatively quickly.

We could have then separately tested what happens if the activity is updated. It seems all three of these scenarios are in a single property.

cardano-testnet/src/Testnet/Components/DReps.hs
```diff
@@ -140,6 +143,89 @@ createDRepRegistrationTxBody execConfig epochStateView sbe work prefix drepRegCe
...
+-- This function takes the following parameters:
+--
+-- * 'execConfig': Specifies the CLI execution configuration.
+-- * 'work': Base directory path where the voting files and directories will be
+--           stored.
+-- * 'prefix': Name for the subfolder that will be created under 'work' to store
+--             the output voting files.
+-- * 'governanceActionTxId': Transaction ID string of the governance action.
+-- * 'governanceActionIndex': Index of the governance action.
+-- * 'allVotes': List of tuples where each tuple contains a 'PaymentKeyPair'
```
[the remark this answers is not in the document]
palas: Thanks

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/LedgerEvents/Gov/DRepActivity.hs
```diff
@@ -0,0 +1,301 @@
...
+  H.note_ $ "Sprocket: " <> show poolSprocket1
+  H.note_ $ "Abs path: " <> tempAbsBasePath'
+  H.note_ $ "Socketpath: " <> socketPath
+  H.note_ $ "Foldblocks config file: " <> configurationFile
+
+  gov <- H.createDirectoryIfMissing $ work </> "governance"
+
+  -- First proposal (set activity to something feasible in the test)
+  dRepActivityBeforeFirstProp <- getDRepActivityValue execConfig
+  dRepActivityBeforeFirstProp === 100 -- This is the default value
```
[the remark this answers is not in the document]
palas: Yes, I thought about doing that originally. But I think I would have had to modified the testnet creation code probably. Also, it is convenient to have a DRep that doesn't expire for the test, and this way we get the change of parameter tested for free

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/LedgerEvents/Gov/DRepActivity.hs
```diff
@@ -0,0 +1,301 @@
...
+  voteChangeProposal execConfig epochStateView sbe gov "vote1"
+                     firstGovernanceActionTxId firstGovernanceActionIndex firstPropVotes wallet0
+
+  (EpochNo epochAfterFirstProp) <- getCurrentEpochNo epochStateView sbe
+  H.note_ $ "Epoch after first prop: " <> show epochAfterFirstProp
+
+  void $ waitUntilEpoch (File configurationFile) (File socketPath) (EpochNo (epochAfterFirstProp + 2))
+  dRepActivityAfterFirstProp <- getDRepActivityValue execConfig
+
+  dRepActivityAfterFirstProp === 4 -- This is what we just set, should work
```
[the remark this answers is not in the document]
palas: You are right, I think that has been the issue in one of the executions, in fact.

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/LedgerEvents/Gov/DRepActivity.hs
```diff
@@ -0,0 +1,301 @@
...
+
+  result1 <- checkDRepState sbe (File configurationFile) (File socketPath) execConfig
+                            (Just . map (drepExpiry . snd) . Map.toList)
+  H.note_ $ "DRep expiration dates: " <> show result1
```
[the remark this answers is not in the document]
palas: It depends on when the epoch was registered and because it is not instantaneous, and because epochs are so fast in testnet, there is considerable fluctuation (of several epochs).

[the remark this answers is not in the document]
palas: It depends on which epoch the DRep was registered and because it takes time and epochs are so fast in testnet, there is considerable fluctuation (of several epochs).

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/LedgerEvents/Gov/DRepActivity.hs
palas: Huh, yes that would be much nicer. I didn't know you could do that, but it makes sense.

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/LedgerEvents/Gov/DRepActivity.hs
palas: Nice catch

palas:
> 1a. Modify Conway genesis with ridiculously low drep activity i.e 1 Epoch
> 2a. Test that the drep can no longer participate in voting (wait till 1 epoch has passed)
> 
> 2a. Modify Conway genesis with slightly more reasonably drep activity i.e 2 Epochs
> 2b. Test that the drep can participate in voting before 2 Epochs have passed.

Yes, dividing concerns probably would give us smaller tests. I am not sure that a DRep expiring prevents it from participating, I think it is just not counted when calculating minimum thresholds, that is what I was trying to test. That is a bit trickier because you still need a DRep that didn't expire to see if they can make the motion pass even though they are minority.

In any case, if we redesign, I would prefer to do it in a separate PR.

cardano-testnet/src/Testnet/Components/DReps.hs
```diff
@@ -100,50 +119,148 @@ generateRegistrationCertificate execConfig work prefix drepKeyPair depositAmount
...
+    [ "--witness-override", show @Int (length votes)
+    , "--out-file", unFile dRepVotingTxBody
+    ]
+  return dRepVotingTxBody
+
 -- Transaction signing
 
 data SignedTx
 
+class KeyPair a where
```
carbolymer, line 247: Nice idea. I think this should be moved to `Testnet.Runtime` to make it more accessible.

In general I think we could create a `KeyPair` type with a phantom variable which would represent keys and just use uninhabited types to mark the types. I mean, instead of:
```haskell
data PaymentKeyPair = PaymentKeyPair
  { paymentVKey :: FilePath
  , paymentSKey :: FilePath
  } deriving (Eq, Show)

data StakingKeyPair = StakingKeyPair
  { stakingVKey :: FilePath
  , stakingSKey :: FilePath
  } deriving (Eq, Show)
```
we could have
```haskell
data Staking
data Payment

data KeyPair (type :: Type) = KeyPair
  { vKey :: File type InOut
  , sKey :: File type InOut
  }
```
and just use `KeyPair Staking` and `KeyPair Payment`. But that's out of scope of this PR.

cardano-testnet/src/Testnet/Components/DReps.hs
```diff
@@ -218,3 +335,171 @@ failToSubmitTx execConfig cEra signedTx = GHC.withFrozenCallStack $ do
...
+    , "--tx-file", unFile signedTxBody
+    ]
+  return $ mconcat $ lines txidOutput
+
+-- | Register a Delegate Representative (DRep) using @cardano-cli@,
+-- generating a fresh key pair in the process.
+--
+-- This function takes the following parameters:
+--
+-- * 'execConfig': Specifies the CLI execution configuration.
```
carbolymer, line 365: I think this should go into each argument haddocks

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/LedgerEvents/Gov/DRepActivity.hs
```diff
@@ -0,0 +1,351 @@
...
+  -- Last proposal (set activity to something else again and it should pass, because of inactivity)
+  -- Because 2 out of 3 DReps were inactive, prop should pass
+  let lastTargetDRepActivity = secondTargetDRepActivity + fromIntegral numOfFillerProposals + 1
+  void $ activityChangeProposalTest execConfig epochStateView configurationFile socketPath ceo gov
+                                    "lastProposal" wallet0 [(1, "yes")] lastTargetDRepActivity
+                                    minEpochsToWaitIfChanging (Just lastTargetDRepActivity)
+                                    maxEpochsToWaitAfterProposal
+
+activityChangeProposalTest
+  :: (MonadTest m, MonadIO m, H.MonadAssertion m, MonadCatch m, Foldable t)
```
carbolymer, line 154: I feel that it may be easy to make a mistake when calling this function with so many arguments. Maybe it would be better to create a separate datatype for that purpose?

cardano-testnet/src/Testnet/Components/Query.hs
```diff
@@ -301,7 +302,12 @@ checkDRepState sbe configurationFile socketPath execConfig f = withFrozenCallSta
                   [ "checkDRepState: foldEpochState returned Nothing: "
                   , "This is probably an error related to foldEpochState." ]
       H.failure
-    Right (_, Just val) ->
+    Right (ConditionNotMet, Just _) -> do
```
carbolymer, line 305: Just out of curiosity, has this happened for you or is this just a safeguard?

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/LedgerEvents/Gov/DRepActivity.hs
```diff
@@ -0,0 +1,351 @@
...
+  H.writeFile proposalAnchorFile "dummy anchor data"
+
+  proposalAnchorDataHash <- H.execCli' execConfig
+    [ "conway", "governance"
+    , "hash", "anchor-data", "--file-text", proposalAnchorFile
+    ]
+
+  minDRepDeposit <- getMinDRepDeposit execConfig ceo
+
+  proposalFile <- H.note $ baseDir </> "sample-proposFal-anchor"
```
carbolymer, line 275:
```suggestion
  proposalFile <- H.note $ baseDir </> "sample-proposal-anchor"
```

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/LedgerEvents/Gov/DRepActivity.hs
```diff
@@ -0,0 +1,351 @@
...
+
+  let stakeVkeyFp = baseDir </> "stake.vkey"
+      stakeSKeyFp = baseDir </> "stake.skey"
+
+  _ <- P.cliStakeAddressKeyGen baseDir
+         $ P.KeyNames { P.verificationKeyFile = stakeVkeyFp
+                      , P.signingKeyFile = stakeSKeyFp
+                      }
+
+  proposalAnchorFile <- H.note $ baseDir </> "sample-proposFal-anchor"
```
carbolymer, line 265:
```suggestion
  proposalAnchorFile <- H.note $ baseDir </> "sample-proposal-anchor"
```

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/LedgerEvents/Gov/DRepActivity.hs
palas: Yes, I was tempted to do a monad that holds that info and refactor all the DRep tests and functions to use it. Maybe I can do that in a PR after these are merged

cardano-testnet/src/Testnet/Components/Query.hs
palas: Safeguard

cardano-testnet/src/Testnet/Components/DReps.hs
palas: 100%

cardano-testnet/src/Testnet/Components/DReps.hs
```diff
@@ -218,3 +254,139 @@ failToSubmitTx execConfig cEra signedTx = GHC.withFrozenCallStack $ do
...
+getLastPParamUpdateActionId :: (MonadTest m, MonadCatch m, MonadIO m)
+  => H.ExecConfig -- ^ Specifies the CLI execution configuration.
+  -> m (Maybe (String, Word32))
+getLastPParamUpdateActionId execConfig = do
+  govStateString <- H.execCli' execConfig
+    [ "conway", "query", "gov-state"
+    , "--volatile-tip"
+    ]
+
+  govStateJSON <- H.nothingFail (Aeson.decode (LBS.pack govStateString) :: Maybe Aeson.Value)
```
smelc, line 371:
```suggestion
  govStateJSON :: Aeson.Value <- H.execCliStdoutToJson execConfig
    [ "conway", "query", "gov-state"
    , "--volatile-tip"
    ]
```

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/LedgerEvents/Gov/DRepActivity.hs
```diff
@@ -0,0 +1,351 @@
...
+import           Hedgehog
+import qualified Hedgehog.Extras as H
+import qualified Hedgehog.Extras.Stock.IO.Network.Sprocket as IO
+
+-- | Execute me with:
+-- @DISABLE_RETRIES=1 cabal test cardano-testnet-test --test-options '-p "/DRep Activity/"'@
+hprop_check_drep_activity :: Property
+hprop_check_drep_activity = H.integrationWorkspace "test-activity" $ \tempAbsBasePath' -> do
+    -- Start a local test net
+  conf@Conf { tempAbsPath } <- mkConf tempAbsBasePath'
```
smelc, line 56:
```suggestion
  -- Start a local test net
  conf@Conf { tempAbsPath } <- mkConf tempAbsBasePath'
```

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/LedgerEvents/Gov/DRepActivity.hs
smelc: And you can probably also use a newly-introduced datatype to pass less arguments to `delegateToDRep`, for example these ones:

```hs
EpochStateView -- ^ Current epoch state view for transaction building. It can be obtained
FilePath -- ^ Path to the node configuration file as returned by 'cardanoTestnetDefault'.
FilePath -- ^ Path to the cardano-node unix socket file.
ShelleyBasedEra ConwayEra -- ^ The Shelley-based era (e.g., 'ConwayEra') in which the transaction will be constructed.
```

smelc (approved):

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/LedgerEvents/Gov/DRepActivity.hs
carbolymer, line 154:
```suggestion
  :: (HasCallStack, MonadTest m, MonadIO m, H.MonadAssertion m, MonadCatch m, Foldable t)
```

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/LedgerEvents/Gov/DRepActivity.hs
```diff
@@ -0,0 +1,351 @@
...
+    isSuccess epochAfterProp (EpochNo epochNo) (EpochInterval epochInterval) =
+      (epochAfterProp + minWait <= epochNo) &&
+      (case mExpected of
+        Nothing -> True
+        Just expected -> epochInterval == expected) &&
+      (epochNo <= epochAfterProp + maxWait)
+
+
+makeActivityChangeProposal
+  :: (H.MonadAssertion m, MonadTest m, MonadCatch m, MonadIO m)
```
carbolymer, line 235:
```suggestion
  :: (HasCallStack, H.MonadAssertion m, MonadTest m, MonadCatch m, MonadIO m)
```

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/LedgerEvents/Gov/DRepActivity.hs
```diff
@@ -0,0 +1,351 @@
...
+                             Left e ->
+                               H.failMessage callStack
+                                 $ "makeActivityChangeProposal failed waiting for gov action with: " <> displayError e
+                             Right Nothing ->
+                               H.failMessage callStack "Couldn't find proposal."
+                             Right (Just a) -> return a
+
+  return (governanceActionTxId, governanceActionIndex)
+
+voteChangeProposal :: (MonadTest m, MonadIO m, MonadCatch m, H.MonadAssertion m)
```
carbolymer, line 325:
```suggestion
voteChangeProposal :: (HasCallStack, MonadTest m, MonadIO m, MonadCatch m, H.MonadAssertion m)
```

carbolymer (approved):

Jimbo4350:
> > 1a. Modify Conway genesis with ridiculously low drep activity i.e 1 Epoch
> > 2a. Test that the drep can no longer participate in voting (wait till 1 epoch has passed)
> > 2a. Modify Conway genesis with slightly more reasonably drep activity i.e 2 Epochs
> > 2b. Test that the drep can participate in voting before 2 Epochs have passed.
> 
> Yes, dividing concerns probably would give us smaller tests. I am not sure that a DRep expiring prevents it from participating, I think it is just not counted when calculating minimum thresholds, that is what I was trying to test. That is a bit trickier because you still need a DRep that didn't expire to see if they can make the motion pass even though they are minority.
> 
> In any case, if we redesign, I would prefer to do it in a separate PR.

I think we should redesign.

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/LedgerEvents/Gov/DRepActivity.hs
palas: Yes, if we have the base config vars in a Monad, and I think that could easily include the `era` and the `EpochStateView`, then the remaining args are going to be really few. We can still also pass a single argument that is a record so that remaining params are labeled, definitely.

palas: Precisely, those are the ones I meant would be nice to hide with a monad. A configuration datatype with all the config info would be an alternative too, even though it is still one extra parameter which you have to include every time. Another thing I considered was returning a record of partially applied functions that include the config, but I discarded the idea because I thought it would be pretty ugly

palas:
> I think we should redesign.

Ok, I will check whether DReps expiring prevents them from participating and follow up on this

PR #5813, opened 2024-04-29
Remove consensus internal options from cli flags

jasagredo:
# Description

Some of the options presented as cli flags seem to be consensus internal or low-level options. This PR adds warnings when they are set via cli flags in preparation for future removal, they can still be configured in the JSON/YAML configuration.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [X] The version bounds in `.cabal` files are updated
- [x] Self-reviewed the diff

### DIFF
--- a/cardano-node/ChangeLog.md
+++ b/cardano-node/ChangeLog.md
@@ -3,6 +3,12 @@
 ## Next version
 
 - Use p2p network stack by default, warn when using the legacy network stack.
+- Deprecate some CLI flags corresponding to low-level consensus options. They are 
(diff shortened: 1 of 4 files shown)


mgmeier:
For Performance & Tracing: we do make use of the snapshot options only via the `--config` file. We don't require them as CLI options.

Jimbo4350 (approved):
LGTM! Updating the parser help with the configuration value needed to set in the config file would be useful.

CarlosLopezDeLara (approved):

PR #5823, opened 2024-05-03
Fix jq expression in CI

jasagredo:


jasagredo:
Prevents this error in CI which probably was corrupting the GHA cache (always writing to the same cache):
```
jq: error (at <stdin>:0): Cannot index array with string "style"
```

CarlosLopezDeLara (approved):

PR #5842, opened 2024-05-16
release-upload: generate checksums of artifacts

smelc:
# Context

* This is the cardano-node version of https://github.com/IntersectMBO/cardano-cli/pull/762
* Was asked by a user: https://github.com/IntersectMBO/cardano-cli/issues/755#issuecomment-2107020047
* Anyway a good practice to make it possible for users to verify the integrity of their downloads and a number of deployments tools require assets to have attached checksums.
* The change of version of `{download,upload}-artifact` is motivated by the fact that `v3` will be shutdown at the end of the year: https://github.com/actions/upload-artifact#actionsupload-artifact

# How to trust this PR

* Look at the assets generated when producing a release with this PR: https://github.com/IntersectMBO/cardano-node/releases/tag/untagged-0ccd023f738aead32ba8
* Observe it contains the checksums: 

image

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/.github/workflows/release-upload.yaml
+++ b/.github/workflows/release-upload.yaml
@@ -78,7 +78,7 @@ jobs:
               cp result/cardano-node-*-*.zip .
               ;;              
           esac
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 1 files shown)


carbolymer (approved):
makes senses 👍🏻

PR #5864, opened 2024-05-29
Churn counters fix

jutaro:
#### Description

This pull request addresses a fix for the churn counter in our current tracing implementation. With the new tracing infrastructure, it is necessary to declare new tracers in three separate locations. This redundancy is less than ideal and has been identified as an area for future improvement.

#### Key Changes

- **Churn Counter Adjustment**: Updated the churn counter logic to ensure proper functionality within the new tracing setup.
- **Tracer Declarations**: Added the necessary tracer declarations in all three required places to align with the current implementation requirements.

#### Context and Future Work

The current state of our tracing system is transitional. The new tracing functionality operates behind the legacy interface, which imposes certain restrictions and necessitates the multiple declarations. This approach, while functional, is not optimal. It is our goal to streamline this process once we can fully transition away from the old tracing infrastructure.

#### Impact

- Ensures the churn counter works correctly with the new tracing setup.
- Maintains compatibility with the existing tracing interface.

#### Next Steps

- **Refactoring**: Plan for a refactor to remove the old tracing infrastructure.
- **Streamlining Tracer Declarations**: Simplify tracer declarations to avoid redundancy and improve maintainability.

We appreciate any feedback on this fix and suggestions for improving our tracing system as we move forward.

#### Testing

- Verified that the churn counter operates correctly with the new tracer declarations.
- Conducted regression tests to ensure no adverse effects on existing tracing functionality.

### DIFF
--- a/cardano-node/src/Cardano/Node/Tracing/Consistency.hs
+++ b/cardano-node/src/Cardano/Node/Tracing/Consistency.hs
@@ -63,7 +63,7 @@ import           Ouroboros.Network.KeepAlive (TraceKeepAliveClient (..))
 import qualified Ouroboros.Network.NodeToClient as NtC
 import           Ouroboros.Network.NodeToNode (ErrorPolicyTrace (..), RemoteAddress, WithAddr (..))
 import qualified Ouroboros.Network.NodeToNode as NtN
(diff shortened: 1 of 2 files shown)


mgmeier (approved):
LGTM!

PR #5939, opened 2024-08-05
hls build fixes

NadiaYvette:
# Description

When starting nix develop independently of the makefile, hls fails to build, seemingly because of version skew between which versions of ghc and hls are compatible and the declarations in ./nix/pkgs.nix for them. This commit series goes about fix-ups to get nix develop to stop failing when started from a regular shell.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff


sgillespie (approved):
LGTM! 

My only concern is hacking flake.lock by hand. We should probably avoid this in the future. Since haskell.nix should be updated soon, I'm okay with it.

carbolymer (approved):

PR #5942, opened 2024-08-07
Update cardano-node after addition of the new BulkSync implementation

facundominguez:
This PR updates tracing and exposes the genesis configuration.

It requires https://github.com/IntersectMBO/ouroboros-consensus/pull/1179. `cabal.project` is made to point to a back-port of the BulkSync implementation on `ouroboros-network-0.16.1.1` and `ouroboros-consensus-0.20.0.0`.

### DIFF
--- a/cardano-node/src/Cardano/Node/Configuration/POM.hs
+++ b/cardano-node/src/Cardano/Node/Configuration/POM.hs
@@ -162,6 +162,9 @@ data NodeConfiguration
 
          -- Enable Peer Sharing
        , ncPeerSharing :: PeerSharing
+
+       -- Enable Genesis syncing protocol
+       , ncEnableGenesis :: Bool
        } deriving (Eq, Show)
 
 
@@ -225,6 +228,9 @@ data PartialNodeConfiguration
 
          -- Peer Sharing
        , pncPeerSharing :: !(Last PeerSharing)
+
+         -- Genesis syncing protocol
+       , pncEnableGenesis :: !(Last Bool)
        } deriving (Eq, Generic, Show)
 
 instance AdjustFilePaths PartialNodeConfiguration where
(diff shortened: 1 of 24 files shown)


cardano-testnet/src/Testnet/Defaults.hs
```diff
@@ -280,6 +280,7 @@ defaultYamlHardforkViaConfig era =
     , (proxyName (Proxy @TraceBlockFetchProtocolSerialised), Aeson.Bool False)
     , (proxyName (Proxy @TraceBlockFetchServer), Aeson.Bool False)
     , (proxyName (Proxy @TraceBlockchainTime), Aeson.Bool True)
+    , (proxyName (Proxy @TraceChainSyncJumping), Aeson.Bool False)
```
Jimbo4350, line 283: To keep with the ordering, this should be under `TraceChainSyncHeaderServer`.

cardano-node/src/Cardano/Node/Configuration/POM.hs
```diff
@@ -329,7 +332,8 @@ instance FromJSON PartialNodeConfiguration where
 
       -- Genesis syncing protocol
       -- DISABLED BY DEFAULT
-      pncEnableGenesis <- Last <$> v .:? "EnableGenesis" .!= Just False
+      pncEnableGenesis      <- Last <$> v .:? "EnableGenesis" .!= Just False
```
Jimbo4350, line 335: Why two configuration values? If the genesis config plans aren't present then we know to not enable it.

amesgen: Thanks, what would you prefer instead? Something like
```yaml
Genesis:
  Enable: True
  # optionally
  LowLevelOptions: 
    ....
```
?

amesgen:
This PR will likely be superseded by #6050 and the changes in https://github.com/IntersectMBO/cardano-node/tree/amesgen/genesis-10.2 (ie the last commit there basically is a cleaned-up version of this PR)

amesgen:
These changes were included in #6050

PR #5960, opened 2024-08-23
cardano-testnet: honor --max-supply value, by passing it to --create-testnet-data under the hood

smelc:
# Description

`cardano-testnet` has a `--max-supply` value which was being ignored. This PR fixes it, by passing the value to `--create-testnet-data`.

Fixes https://github.com/IntersectMBO/cardano-node/issues/5953

Related to this PR is https://github.com/IntersectMBO/cardano-cli/pull/874

# How to trust this PR:

* [ ] CI passes
* [ ] Ran `cardanot-testnet`'s tests locally using https://github.com/IntersectMBO/cardano-cli/pull/874 (showing that the combination of this PR and 874 is sound)

# Checklist

- [X] Self-reviewed the diff


cardano-testnet/src/Testnet/Start/Types.hs
```diff
@@ -59,7 +59,7 @@ cardanoDefaultTestnetOptions = CardanoTestnetOptions
   , cardanoSlotLength = 0.1
   , cardanoTestnetMagic = 42
   , cardanoActiveSlotsCoeff = 0.05
-  , cardanoMaxSupply = 100_020_000_000 -- 100 billions. This amount should be bigger than the 'byronTotalBalance' in Testnet.Start.Byron
+  , cardanoMaxSupply = 100_000_020_000_000 -- 100 000 billions Lovelace, so 100 millions ADA. This amount should be bigger than the 'byronTotalBalance' in Testnet.Start.Byron
```
smelc, line 62: This default (increased) amount was suggested by @CarlosLopezDeLara, who found the previous amount too low.

cardano-testnet/src/Testnet/Components/Configuration.hs
```diff
@@ -132,13 +132,14 @@ createSPOGenesisAndFiles
   :: (MonadTest m, MonadCatch m, MonadIO m, HasCallStack)
   => NumPools -- ^ The number of pools to make
```
Jimbo4350, line 133: Not for this PR but we need to make a hard decision about what overrides we allow otherwise this function signature (and others) will become unwieldly. I'd say lets just stick with what we have exposed for now and then return to this once `cardano-testnet` is in a better state.

cardano-testnet/src/Testnet/Components/Configuration.hs
```diff
@@ -181,8 +182,8 @@ createSPOGenesisAndFiles (NumPools numPoolNodes) (NumDReps numDelReps) sbe shell
     , "--spec-conway",  inputGenesisConwayFp
     , "--testnet-magic", show testnetMagic
     , "--pools", show numPoolNodes
-    , "--total-supply",     show @Int 2_000_000_000_000 -- 2 trillions
-    , "--delegated-supply", show @Int 1_000_000_000_000 -- 1 trillion
+    , "--total-supply",     show maxSupply
```
Jimbo4350, line 185: Nice :+1:

cardano-testnet/src/Testnet/Start/Types.hs
Jimbo4350: Hmmm. We need to figure out a way to eliminate this discrepancy, preferably at the type level. Perhaps we can have a type called `CommonGenesisConfiguration`.

Jimbo4350 (approved):
LGTM!

cardano-testnet/src/Testnet/Start/Types.hs
smelc: @Jimbo4350> I'm revisiting this type anyway for the work on per-node-configuration, so we'll be able to improve more soon.

That is why I'm leaving it like this now.

cardano-testnet/src/Testnet/Components/Configuration.hs
smelc: Yep this is going to happen :+1:

PR #5962, opened 2024-08-28
cardano-testnet-test: small improvements

smelc:
# Description

This PR removes initialization of fields in `cardano-testnet-test` that were useless because they were initializing said fields with the same value as the default values. It also does QoL improvements.

# Checklist

- [X] Self-reviewed the diff


carbolymer (approved):

coot:
Are we using P2P stack in tests? Note that the non-p2p stack will be removed at some point.

smelc:
@coot> I believe we don't use P2p in `cardano-testnet`, because here is what we have by default in the `configuration.yaml` used to start nodes:

```shell
→ cat /tmp/nix-shell.3H7AVw/cli-queries-test-23d51bd100cd79ee/configuration.yaml  | grep P2P
    "EnableP2P": false,
```

smelc:
Let me amend the PR then, to keep the flag then and honor it instead of doing nothing with it :slightly_smiling_face:

smelc:
Honoring P2P will affect other work I have stacked on this one, so I'll just keep the "small improvements" bits of this PR and merge. P2P honoring will come later.

coot:
Note that `non-p2p` will be removed.

PR #5973, opened 2024-09-04
Remove openssl support for snap-serer

erikd:
# Description

OpenSSL support was only need by `cardano-tracer` and even then is only used when `cardano-tracer` is compiled with the `rtview` flag is `True` (by default `False`).

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff

# Note on CI
for you.


cabal.project
```diff
@@ -48,7 +48,9 @@ package cryptonite
   flags: -support_rdrand
 
 package snap-server
-  flags: +openssl
+  -- OpenSSL support requires HsOpenSSL which will cause problems in the near future
+  -- https://github.com/haskell-cryptography/HsOpenSSL/issues/95
+  flags: -openssl
```
mgmeier, line 53: Unless `snap-server` remains somewhere in the node's dependency graph, this flag from now only affects `cardano-tracer`. That, however, will need to account for a non-default, non-SSL build of its dependency within the application code.

Merging this flag change would break `cardano-tracer` currently, please remove and postpone.

mgmeier (changes requested):

cabal.project
erikd: This currently builds with this change.

The `-openssl` flag on enables SSL for `cardano-tracer` when `cardano-tracer` is built with the `-rtview` flag which is not one be default.

mgmeier: Flags `-openssl` and `+rtview` result in a broken service (i.e., unable to start up). Handling this case needs to be adjusted in the application code before any merge.

mgmeier:
Closing this in favor of #5975 which will contain the same changes plus necessary application-side adjustements - removing the dependency on `HsOpenSSL`.

PR #5987, opened 2024-09-16
cardano-testnet: rename ShelleyTestnetOptions into GenesisOptions

smelc:
# Description

Doing a renaming agreed here: https://github.com/IntersectMBO/cardano-node/pull/5976#discussion_r1761109187

# Checklist

- [X] Self-reviewed the diff


carbolymer (approved):

Jimbo4350 (changes requested):
Requesting changes because I don't want this merged before 9.2 is released.

Jimbo4350 (approved):

Issue #5140, opened 2023-04-21
Adding Prometheus metric type comments


eMCeee89:
**Internal/External**
*External*

**Area**
*Other*

**Summary**
There are metric types missing in Prometheus metrics endpoint. It would be great to have them added for all the metrics. There is a problem if we want to scrape them using AWS CloudWatch agent because metrics without type are simply dropped and [a feature request for adding these types explicitly as a workaround](https://github.com/aws/amazon-cloudwatch-agent/issues/187) has never been implemented. It would be great to have this standard included, e.g. as a node_exporter has it. I am attaching a snippet of node_exporter metrics output, it is really about an extra line for every metric stating its type (help is optional but would be nice too):
```
# curl 127.0.0.1:9100/metrics
# HELP go_gc_duration_seconds A summary of the pause duration of garbage collection cycles.
# TYPE go_gc_duration_seconds summary
go_gc_duration_seconds{quantile="0"} 1.922e-05
go_gc_duration_seconds{quantile="0.25"} 2.694e-05
go_gc_duration_seconds{quantile="0.5"} 3.0461e-05
go_gc_duration_seconds{quantile="0.75"} 7.9441e-05
go_gc_duration_seconds{quantile="1"} 7.9441e-05
# HELP go_goroutines Number of goroutines that currently exist.
# TYPE go_goroutines gauge
go_goroutines 8
...
```

**Steps to reproduce**
1. Enable prometheus metrics endpoint
2. Go to prometheus metrics endpoint and observe missing TYPE comment lines for all the metrics

**Expected behavior**
I expect to have metric types defined when querying Prometheus metrics endpoint so these can be correctly scraped by AWS CloudWatch agent.

**System info (please complete the following information):**
- OS Name: Amazon Linux 2
- OS Version: amzn2-ami-ecs-kernel-5.10-hvm
- Node version: 1.35.7
- CLI version : 1.35.7 - linux-x86_64 - ghc-8.10, git rev 0000000000000000000000000000000000000000

jutaro:
[...]

PR #5997, opened 2024-09-26
cardano-tracer: OpenMetrics compliance for Prometheus; fix `forHuman` output in journald

mgmeier:
# Description

This PR is mainly a rewrite of rendering the Prometheus metric exposition in `cardano-tracer`:
* For the response, directly fold a `Builder` through the `Store` sample instead of manipulating `Text` values.
* Improved OpenMetrics compliance of Prometheus exposition; also addresses #5140 .
* Prometheus help annotations can be provided via the new optional config value `metricsHelp`.

Additionally, it:
* fixes forwarding `forHuman` output to journald.
* sets the period to invalidate a memoized routing table to 5s.
* adds JSON response to the EKG server (same as Prometheus).

It futher affects:
* `cardano-node` / `trace-dispatcher`: The `trace-documentation` CLI command has now an optional arg `--output-metric-help` for creating machine-readable JSON output of metrics docs.
* `workbench`: it adds a default (yet incomplete) metrics help dictionary to the cardano-tracer nix service config.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
  - [X] Code builds on Linux, MacOS and Windows for `ghc-8.10.7` and `ghc-9.2.7`
- [x] Self-reviewed the diff


cardano-tracer/src/Cardano/Tracer/Handlers/Metrics/Utils.hs
```diff
@@ -1,63 +1,42 @@
...
-  ( MetricName
-  , MetricValue
-  , MetricsList
-  , RouteDictionary(..)
-  , getListOfMetrics
-  , renderListOfConnectedNodes
-  , renderJson
-  , nodeNames
-  , computeRoutes
+  ( module Cardano.Tracer.Handlers.Metrics.Utils
```
jutaro, line 6: I have never seen this! What is the meaning of this?

cardano-tracer/src/Cardano/Tracer/Handlers/RTView/Utils.hs
```diff
@@ -1,23 +1,27 @@
 {-# LANGUAGE NamedFieldPuns #-}
 
 module Cardano.Tracer.Handlers.RTView.Utils
-  ( forAcceptedMetrics_
-  , forAcceptedMetricsUI_
-  , forConnected
-  , forConnected_
-  , forConnectedUI
-  , forConnectedUI_
+  ( module Cardano.Tracer.Handlers.RTView.Utils
```
jutaro, line 4: You like this crazy stuff

jutaro (approved):
LGTM

cardano-tracer/src/Cardano/Tracer/Handlers/Metrics/Utils.hs
Icelandjack: It re-exports: https://stackoverflow.com/questions/36044041/haskell-module-inside-module

mgmeier: This exports all symbols from that module. It's best to use the current module name here when explicit export lists are required, and you want to export *everything*.

cardano-tracer/src/Cardano/Tracer/Handlers/RTView/Utils.hs
mgmeier: Yes, I want to be explicit about: this module exports *every* definition and type it contains :)

Icelandjack (approved):
For MetricsHelp it suffices to use generalized newtype deriving but it doesn't hurt to be explicit. LGTM!

mgmeier:
> For MetricsHelp it suffices to use generalized newtype deriving but it doesn't hurt to be explicit. LGTM!

GND serializes as such: `{"Left":"testfile.json"}`. I'm trying to avoid any Haskell / aeson specific serializations. Instead, provide one parser allowing for two distinct JSON types without relying on serializing a constructor or tag.

PR #6001, opened 2024-10-01
cardano-testnet: don't specify delegated supply anymore. Use create-testnet-data default behavior

smelc:
# Description

This pops an earlier TODO and doesn't cause any runtime changes, because the behavior remains the same under the hood.

# Checklist

- [X] Self-reviewed the diff


carbolymer (approved):

Jimbo4350 (approved):

smelc:
Thanks @carbolymer :pray:

PR #6003, opened 2024-10-04
cardano-testnet: use new output of "transaction submit"

smelc:
# Description

Test this new additional output of `transaction submit`: https://github.com/IntersectMBO/cardano-cli/pull/925

# Checklist

- [X] New tests are added if needed and existing tests are updated.
- [X] Self-reviewed the diff

### DIFF
--- a/cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Cli/Transaction.hs
+++ b/cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Cli/Transaction.hs
@@ -12,7 +12,10 @@ import           Cardano.Api
 import qualified Cardano.Api.Ledger as L
 import           Cardano.Api.Shelley
 
+import           Cardano.CLI.Types.Common
+import           Cardano.Crypto.Hash.Class (hashToStringAsHex)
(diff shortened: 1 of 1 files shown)


cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Cli/Transaction.hs
```diff
@@ -115,11 +115,16 @@ hprop_transaction = integrationRetryWorkspace 0 "simple transaction build" $ \te
...
+  stdout <- execCli' execConfig
     [ anyEraToString cEra, "transaction", "submit"
     , "--tx-file", txbodySignedFp
     ]
 
+  txId <- if null stdout
+          then do
+            H.note_ "Expected output of \"transaction submit\" not to be empty"
+            H.failure
+          else pure $ last $ lines stdout
```
Jimbo4350, line 127: This is a poor way of obtaining the txid The txid can be calculated from the tx itself and you should query the mempool to confirm indeed if it was submitted or not.

Jimbo4350 (changes requested):

Jimbo4350 (approved):

PR #6047, opened 2024-12-04
Integrate ledger snapshot checksum

geo2a:
# Description

https://github.com/IntersectMBO/ouroboros-consensus/pull/1319, when merged, will checksum the ledger state snapshots to detect data corruption. This PR integrates this change into `cardano-node`.

In addition to the snapshot file itself, a `.checksum` file is written to dist, which contains the `CRC32` checksum of the snapshot data:
```bash
> ls mainnet/db/ledger
2223  2223.checksum

> cat mainnet/db/ledger/2223.checksum
8b578fb4
```

To preserve backwards compatibility, the absence of the checksum file does not block the node from starting up. Instead, a warning is issued when restoring from a snapshot that is missing the checksum file:

```bash
...
Listening on http://127.0.0.1:12798
[geo2a-la:cardano.node.ChainDB:Warning:5] [2024-12-04 13:09:30.72 UTC] Checksum file is missing for snapshot DiskSnapshot {dsNumber = 2223, dsSuffix = Nothing}
...
```

If the checksum differs from the file, there's a number of scenarios that all ultimately lead to the deletion of the snapshot and an attempt to restore from an older one:

- Data corruption of the snapshot file:
  ```
  [geo2a-la:cardano.node.ChainDB:Error:5] [2024-12-04 13:18:42.93 UTC] Invalid snapshot DiskSnapshot {dsNumber = 2223, dsSuffix = Nothing}InitFailureRead ReadSnapshotDataCorruption This is most likely an expected change in the serialization format, which currently requires a chain replay
  ```
  To model data corruption for testing purposes, one could use `dd`, i.e.:
  ```bash
  printf '\x0\x0' | dd of=mainnet/db/ledger/2223 bs=1 seek=100 count=3 conv=notrunc
  ```
- The `.checksum` file is invalid:
  ```
  [geo2a-la:cardano.node.ChainDB:Error:5] [2024-12-04 13:12:29.41 UTC] Invalid snapshot DiskSnapshot {dsNumber = 2223, dsSuffix = Nothing}InitFailureRead (ReadSnapshotInvalidChecksumFile 21127.checksum) This is most likely an expected change in the serialization format, which currently requires a chain replay
  ```
  that means violated the following condition: `length str == 8 && all isHexDigit str`

The feature is on by default and can be disabled by the following configuration option:
```yaml
DoDiskSnapshotChecksum: false
```

### DIFF
--- a/cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
+++ b/cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
@@ -1522,6 +1522,8 @@ instance ( StandardHash blk
              " This is most likely an expected change in the serialization format,"
           <> " which currently requires a chain replay"
         _ -> ""
+  forHuman (LedgerDB.SnapshotMissingChecksum snap) =
+      "Checksum file is missing for snapshot " <> showT snap
 
   forMachine dtals (LedgerDB.TookSnapshot snap pt enclosedTiming) =
     mconcat [ "kind" .= String "TookSnapshot"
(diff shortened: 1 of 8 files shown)


cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
```diff
@@ -1522,6 +1522,8 @@ instance ( StandardHash blk
              " This is most likely an expected change in the serialization format,"
```
jasagredo, line 1522: (can't comment there)
Probably this code above would benefit from some spaces in between:
```
"Invalid snapshot " <> showT snap <> showT failure <> context
```

geo2a: Agreed.

jasagredo (approved):

PR #6052, opened 2024-12-06
Enhancements and fixes related to logging, metrics, and testing

### LINKED ISSUE #5751: Missing metric for current forging mode on a blockproducer node

jutaro:
#### **Summary**
This pull request addresses multiple enhancements and fixes related to logging, metrics, and testing in the project.

---

#### **Changes**
1. **Let the Forging Mode metrics respect the command line config**  
   - Resolves [#5751](https://github.com/project/repo/issues/5751).  
   - Fixes a metric to track the **current forging mode** on a block producer node.  
   - Ensures visibility into the forging mode to improve monitoring and debugging.

2. **Enhanced Logging for Chain Events**  
   - Added the following fields to log messages for better traceability:  
     - **`tipBlockHash`**: Hash of the current tip block.  
     - **`tipBlockParentHash`**: Hash of the parent block of the tip.  
     - **`tipBlockIssuerVerificationKeyHash`**: Hash of the verification key of the tip block issuer.  
   - Affected log messages:  
     - `SwitchedToAFork`  
     - `AddedToCurrentChain`

3. **Consistency Testing for Mainnet Configuration**  
   - Added a **test case** to ensure **mainnet configuration consistency**.  
   - Validates the correctness of the mainnet configuration to prevent misconfigurations.

#### **Testing**
- Verified the new metric appears correctly in monitoring systems during runtime.  
- Confirmed the additional fields are included in the logs for the relevant events (`SwitchedToAFork`, `AddedToCurrentChain`).  
- Successfully ran the new mainnet configuration test case.

---
#### **Checklist**
- [x] Corrected metric for current forging mode.  
- [x] Updated log messages with additional fields.  
- [x] Implemented and verified the mainnet configuration consistency test.  
- [x] Updated relevant documentation if necessary.  

---

#### **Linked Issue**
Closes #5751

### DIFF
--- a/cardano-node/src/Cardano/Node/Run.hs
+++ b/cardano-node/src/Cardano/Node/Run.hs
@@ -232,11 +232,11 @@ handleNodeWithTracers cmdPc nc0 p networkMagic blockType runP = do
           startupInfo <- getStartupInfo nc p fp
           mapM_ (traceWith $ startupTracer tracers) startupInfo
           traceNodeStartupInfo (nodeStartupInfoTracer tracers) startupInfo
-
           -- sends initial BlockForgingUpdate
           blockForging <- snd (Api.protocolInfo runP)
+          let isNonProducing = ncStartAsNonProducingNode nc
           traceWith (startupTracer tracers)
-                    (BlockForgingUpdate (if null blockForging
+                    (BlockForgingUpdate (if isNonProducing || null blockForging
                                           then DisabledBlockForging
                                           else EnabledBlockForging))
 
@@ -278,10 +278,10 @@ handleNodeWithTracers cmdPc nc0 p networkMagic blockType runP = do
             >>= mapM_ (traceWith $ startupTracer tracers)
 
           traceWith (nodeVersionTracer tracers) getNodeVersion
-
+          let isNonProducing = ncStartAsNonProducingNode nc
           blockForging <- snd (Api.protocolInfo runP)
           traceWith (startupTracer tracers)
-                    (BlockForgingUpdate (if null blockForging
+                    (BlockForgingUpdate (if isNonProducing || null blockForging
                                           then DisabledBlockForging
                                           else EnabledBlockForging))
 
@@ -724,12 +724,13 @@ updateBlockForging startupTracer blockType nodeKernel nc = do
         Just Refl -> do
           -- TODO: check if runP' has changed
           blockForging <- snd (Api.protocolInfo runP')
+          let isNonProducing = ncStartAsNonProducingNode nc
           traceWith startupTracer
-                    (BlockForgingUpdate (bool EnabledBlockForging
-                                              DisabledBlockForging
-                                              (null blockForging)))
-
-          setBlockForging nodeKernel blockForging
+                    (BlockForgingUpdate (if isNonProducing || null blockForging
+                                          then DisabledBlockForging
+                                          else EnabledBlockForging))
+          unless (ncStartAsNonProducingNode nc) $
+            setBlockForging nodeKernel blockForging
         Nothing ->
           traceWith startupTracer
             $ BlockForgingBlockTypeMismatch
--- a/cardano-node/src/Cardano/Node/Run.hs
+++ b/cardano-node/src/Cardano/Node/Run.hs
@@ -729,7 +729,7 @@ updateBlockForging startupTracer blockType nodeKernel nc = do
                     (BlockForgingUpdate (if isNonProducing || null blockForging
                                           then DisabledBlockForging
                                           else EnabledBlockForging))
-          unless (ncStartAsNonProducingNode nc) $
+          unless isNonProducing $
             setBlockForging nodeKernel blockForging
         Nothing ->
           traceWith startupTracer
--- a/cardano-node/src/Cardano/Node/Run.hs
+++ b/cardano-node/src/Cardano/Node/Run.hs
@@ -729,8 +729,7 @@ updateBlockForging startupTracer blockType nodeKernel nc = do
                     (BlockForgingUpdate (if isNonProducing || null blockForging
                                           then DisabledBlockForging
                                           else EnabledBlockForging))
-          unless isNonProducing $
-            setBlockForging nodeKernel blockForging
+          setBlockForging nodeKernel blockForging
         Nothing ->
           traceWith startupTracer
             $ BlockForgingBlockTypeMismatch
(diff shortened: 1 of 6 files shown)


cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
```diff
@@ -493,8 +504,18 @@ instance ( LogFormatting (Header blk)
                | dtal == DDetailed ]
             ++ [ "events" .= toJSON (map (forMachine dtal) events)
                | not (null events) ]
+            ++ [ "tipBlockHash" .= tipBlockHash]
+            ++ [ "tipBlockParentHash" .= tipBlockParentHash]
+            ++ [ "tipBlockIssuerVerificationKeyHash" .= tipBlockIssuerVkHashText]
```
mgmeier, line 509: Can you use a 3-element list here (or add those fields to the non-conditional list above), avoiding spurious uses of `(++)`?

cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
```diff
@@ -506,6 +527,10 @@ instance ( LogFormatting (Header blk)
...
+
```
mgmeier, line 533: same as above

cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
```diff
@@ -2097,22 +2121,44 @@ data ChainInformation = ChainInformation
...
     }
+  where
+    tipIssuerVkHash :: BlockIssuerVerificationKeyHash
+    tipIssuerVkHash =
+      case AF.head frag of
+        Left AF.AnchorGenesis ->
+          NoBlockIssuer
+        Left (AF.Anchor _s _h _b) ->
+          NoBlockIssuer
+        Right blk -> getIssuerVerificationKeyHash blk
```
mgmeier, line 2161: Can the `Left` cases be combined into one? (I'm aware the code is copied from legacy tracing, but it seems redundant).

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -724,12 +724,13 @@ updateBlockForging startupTracer blockType nodeKernel nc = do
...
           traceWith startupTracer
-                    (BlockForgingUpdate (bool EnabledBlockForging
-                                              DisabledBlockForging
-                                              (null blockForging)))
-
-          setBlockForging nodeKernel blockForging
+                    (BlockForgingUpdate (if isNonProducing || null blockForging
+                                          then DisabledBlockForging
+                                          else EnabledBlockForging))
+          unless isNonProducing $
```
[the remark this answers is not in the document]
jutaro: Without unless

cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
jutaro: Ok

cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
jutaro: OK

cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
jutaro: Done

TerminadaDrep:
Would it be difficult to also enable logging of the reason for "SwitchedToAFork"?  IE: Longer chain vs Equal length but terminal block had lower VRF?
That would close [feature request 6014](<https://github.com/IntersectMBO/cardano-node/issues/6014>)

mgmeier:
@TerminadaDrep I've looked at the issue, and the data traced by `SwitchedToAFork` would be sufficient to give an underlying cause as to why a certain chain candidate won the Praos ordering over the previous one.
However, determining that cause given the old and new chain selection data needs to be implemented, and is outside the scope of this PR. We'll address it in a future one.

cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
```diff
@@ -680,7 +745,14 @@ instance MetaTrace  (ChainDB.TraceAddBlockEvent blk) where
         , ( "epoch"
           , "In which epoch is the tip of the current chain."
           )
+        , ( "forks"
+          , "counter for forks"
+          )
+        , ( "tipBlock"
+          , "Hash vallues for tipBlockHash, tipBlockParentHash and tipBlockIssuerVkHashText"
```
mgmeier, line 752: * typo `values`
* `tipBlockIssuerVkHashText` is not the field actual field name, it's the name you chose for the let binding. Should be `tipBlockIssuerVKeyHash` to match the trace msg.

cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
```diff
@@ -703,6 +775,9 @@ instance MetaTrace  (ChainDB.TraceAddBlockEvent blk) where
         , ( "epoch"
           , "In which epoch is the tip of the current chain."
           )
+        , ( "tipBlock"
+          , "Hash vallues for tipBlockHash, tipBlockParentHash and tipBlockIssuerVkHashText"
+          )
```
mgmeier, line 780: ^^ same as above

mgmeier (changes requested):

cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
jutaro: Fixed

cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
jutaro: Fixed

mgmeier (approved):
LGTM, thank you @jutaro

PR #6075, opened 2025-01-14
prepare for garnix

dermetfan:
# Description

- Add garnix cache to the flake nix config.
- Add `garnix.yaml` that builds only `.#checks`.
- Adjust the flake outputs to make garnix build everything that Hydra did.

Garnix does not build the `hydraJobs` output
so move them all into the `checks` output,
adhering to the flake output schema
because garnix does not build nested attrsets.

Note that this replaces the layout of the `checks` output
but it should still include all the checks that it had previously.
So no checks are removed, they just have a different name.
I hope that's ok and no checks are referred to by name somewhere.

# Checklist

- ~~New tests are added if needed and existing tests are updated.~~
- ~~Any changes are noted in the `CHANGELOG.md` for affected package~~
- ~~The version bounds in `.cabal` files are updated~~
- [x] CI passes
- [x] Self-reviewed the diff

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -2,8 +2,14 @@
   description = "Cardano Node";
 
   nixConfig = {
-    extra-substituters = [ "https://cache.iog.io" ];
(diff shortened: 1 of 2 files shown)


dermetfan:
We are not switching to garnix.

PR #6085, opened 2025-01-23
tx-generator: Core.hs: simplifications + better non-Byron support

smelc:
# Description

As part of my work on adapting `tx-generator` to upcoming changes to API (this is https://github.com/IntersectMBO/cardano-node/commits/smelc/adapt-to-api-removal-of-protocol-parameters-use-ledger-type-instead/), I:

1. found a place where the code could be simplified (this is the first commit)
2. found that Byron was not being supported in this place, and so removed the corresponding error case, by pushing the non-Byron constraints to callers.

Point 1. is good on its own while point 2. will make my life easier for the branch mentioned above, but it is optional (I can live with it).

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
+++ b/bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
@@ -162,19 +162,23 @@ cancelBenchmark = do
 getLocalConnectInfo :: ActionM LocalNodeConnectInfo
 getLocalConnectInfo = makeLocalConnectInfo <$> getEnvNetworkId <*> getEnvSocketPath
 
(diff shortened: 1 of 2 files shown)


bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
NadiaYvette, line 1: This new helper function `caseByronOrShelleyBasedEra` is wonderful, thanks for propagating its use onward!

NadiaYvette (approved):
Propagating the new helper function out to us is great & lifting the remote handling for Byron up a layer of calls is a good refactor too.

PR #6091, opened 2025-01-28
Fix block replay tracer in new tracing system

jasagredo:
# Description

Fixes the block replay tracer in the new tracing system. It incorrectly took the ReplayStart as the ReplayGoal thus started already at 100% of progress and never emited any more traces.

Also make the tracer a bit more useful by showing the current slot number.


jutaro (approved):
Looks good to me. Thanks Javier for spotting this!

PR #6104, opened 2025-02-05
Fix port display

Icelandjack:
Fixes the way the port is displayed in the node information (without an accidental layer of `Maybe`).


mgmeier:
Will be merged as part of another PR.

PR #6111, opened 2025-02-08
10.2.1 release preparation

neilmayhew:
Update CHaP index and `cardano-node` version

### DIFF
--- a/cardano-node/src/Cardano/Node/Run.hs
+++ b/cardano-node/src/Cardano/Node/Run.hs
@@ -939,10 +939,7 @@ mkP2PArguments NodeConfiguration {
         targetNumberOfEstablishedBigLedgerPeers = ncDeadlineTargetOfEstablishedBigLedgerPeers,
         targetNumberOfActiveBigLedgerPeers      = ncDeadlineTargetOfActiveBigLedgerPeers
     }
-    syncTargets = PeerSelectionTargets {
-      targetNumberOfRootPeers        = 0,
-      targetNumberOfKnownPeers       = 0,
-      targetNumberOfEstablishedPeers = 0,
+    syncTargets = deadlineTargets {
       targetNumberOfActivePeers               = ncSyncTargetOfActivePeers,
       targetNumberOfKnownBigLedgerPeers       = ncSyncTargetOfKnownBigLedgerPeers,
       targetNumberOfEstablishedBigLedgerPeers = ncSyncTargetOfEstablishedBigLedgerPeers,
--- a/cardano-tracer/src/Cardano/Tracer/Acceptors/Client.hs
+++ b/cardano-tracer/src/Cardano/Tracer/Acceptors/Client.hs
@@ -74,7 +74,7 @@ runAcceptorsClient tracerEnv tracerEnvRTView p (ekgConfig, tfConfig, dpfConfig) 
       , (runDataPointsAcceptorInit   tracerEnv dpfConfig errorHandler, 3)
       ]
  where
-  appInitiator protocolsWithNums =
+  appInitiator protocolsWithNums _ =
     OuroborosApplication
       [ MiniProtocol
          { miniProtocolNum    = MiniProtocolNum num
@@ -95,10 +95,12 @@ doConnectToForwarder
   -> LocalAddress
   -> Word32
   -> ProtocolTimeLimits (Handshake ForwardingVersion Term)
-  -> OuroborosApplication 'Mux.InitiatorMode
+  -> (    ForwardingVersionData
+       -> OuroborosApplication 'Mux.InitiatorMode
                           (MinimalInitiatorContext LocalAddress)
                           (ResponderContext LocalAddress)
                           LBS.ByteString IO () Void
+     )
   -> IO ()
 doConnectToForwarder snocket address netMagic timeLimits app = do
   done <- connectToNode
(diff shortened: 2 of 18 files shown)


cardano-tracer/test/Cardano/Tracer/Test/Forwarder.hs
```diff
@@ -190,8 +190,9 @@ doConnectToAcceptor TestSetup{..} snocket muxBearer address timeLimits (ekgConfi
 
   forwarderApp
     :: [(RunMiniProtocol 'Mux.InitiatorMode initCtx respCtx LBS.ByteString IO () Void, Word16)]
+    -> ForwardingVersionData
     -> OuroborosApplication 'Mux.InitiatorMode initCtx respCtx LBS.ByteString IO () Void
-  forwarderApp protocols =
+  forwarderApp protocols _ =
```
mgmeier, line 195: There's no motivation to change type signature of `forwarderApp`. Can that be reverted?

trace-dispatcher/src/Cardano/Logging/Forwarding.hs
```diff
@@ -307,8 +308,9 @@ doListenToAcceptor magic snocket makeBearer configureSocket address timeLimits
  where
   forwarderApp
     :: [(RunMiniProtocol 'Mux.ResponderMode initiatorCtx responderCtx LBS.ByteString IO Void (), Word16)]
+    -> ForwardingVersionData
     -> OuroborosApplication 'Mux.ResponderMode initiatorCtx responderCtx LBS.ByteString IO Void ()
-  forwarderApp protocols =
+  forwarderApp protocols _ =
```
mgmeier, line 313: As before, there's no motivation to change type signature of `forwarderApp` (twice). Can that be reverted?

mgmeier:
Maybe some context to making `ForwardingVersionData` an extra argument:
Its definition is `newtype ForwardingVersionData = ForwardingVersionData{ networkMagic :: NetworkMagic  }`. This means that no function who has the networkId in scope will ever need `ForwardingVersionData` as a seperate argument.

carbolymer (approved):
👍🏻 on chairman, submit-api and testnet

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -939,10 +939,7 @@ mkP2PArguments NodeConfiguration {
         targetNumberOfEstablishedBigLedgerPeers = ncDeadlineTargetOfEstablishedBigLedgerPeers,
         targetNumberOfActiveBigLedgerPeers      = ncDeadlineTargetOfActiveBigLedgerPeers
     }
-    syncTargets = PeerSelectionTargets {
-      targetNumberOfRootPeers        = 0,
-      targetNumberOfKnownPeers       = 0,
-      targetNumberOfEstablishedPeers = 0,
+    syncTargets = deadlineTargets {
       targetNumberOfActivePeers               = ncSyncTargetOfActivePeers,
```
coot, line 943: shoudn't this be `ncSyncTargetOfActiveBigledgerPeers`, as in the other fields?

crocodile-dentist: No, this is for allowing non-big ledger peers into our outbound connection set, eg. local root peers, but others as well if there is space, such as regular ledger peers or peers found via peer sharing. Turns out it is important for benchmark regression tests, but I didn't see a risk in allowing a node to connect to as long as we have our big ledger peers the consensus GSM should work as intended.

crocodile-dentist: Also, the ncSyncTargetOfActiveBigledgerPeers should be rather large - it is 30 by default now, while this target could be small to accomodate just the local root peers, for instance.
I've ran a Genesis sync test with 45 active peers in total w/o seeing any impact on performance, but there probably is some limit to it so this is another thing to tweak independently to avoid potential issues of having too many active peers.

neilmayhew:
@crocodile-dentist @coot:

The review feedback here is on changes that came from your PRs (#6107 #6114) that were targeting `release/10.2.x`. I merged those and cherry-picked them onto this branch because we wanted to release from `master` and it's not possible to rebase a `release` branch because of branch protections.

Feel free to push changes to this branch, or to create a new PR targeting `master` for your previous PRs. I'll then rebase this branch onto `master` after they're merged.

cardano-tracer/src/Cardano/Tracer/Acceptors/Client.hs
```diff
@@ -95,10 +95,12 @@ doConnectToForwarder
   -> LocalAddress
   -> Word32
   -> ProtocolTimeLimits (Handshake ForwardingVersion Term)
-  -> OuroborosApplication 'Mux.InitiatorMode
+  -> (    ForwardingVersionData
+       -> OuroborosApplication 'Mux.InitiatorMode
                           (MinimalInitiatorContext LocalAddress)
                           (ResponderContext LocalAddress)
                           LBS.ByteString IO () Void
+     )
```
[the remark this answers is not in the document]
coot: This is required by `ouroboros-network-framework`.  I changed it so it's more explicit what is available if, in the future, you'll need to customise the protocols based on negotiated version data.

mgmeier:
@coot Thanks for confirming this is indeed motivated by `simpleSingletonVersions`. As I pointed out, the forwarding application does not require the extra complexity of making behaviour dependent on negotiated version data. Moreover, the way version data definition is currently implemented, it's conceptually impossible to do so - which makes sense, at this has not been an option in the past.

I'd like the code to be very discriptive and reflect that situation, ideally also in its type signatures. I'll add a commit to make the `forwarderApp` a constant function.

mgmeier (approved):
Adressed my concerns in an additional commit.

PR #6119, opened 2025-02-19
Improve scripts/generate-release-changelog-links.hs

neilmayhew:
Add a CLI flag to write output to a file so that cabal output and progress messages aren't mixed in with it


aniketd:
Superseded by #6120

PR #6126, opened 2025-02-20
Fix a typo in a GitHub workflow name

neilmayhew:

PR #6127, opened 2025-02-20
fix `wait-for-hydra` for when there are too many check runs

dermetfan:
# Description

When there are more than 30 (the default pagination page size) check runs, the Hydra statuses are sometimes not found, blocking progress.

We can query only the status we need instead of searching though all statuses.

Also includes #6126, as requested, so we only have to merge one PR.

# Checklist

- ~~New tests are added if needed and existing tests are updated.~~
- ~~Any changes are noted in the `CHANGELOG.md` for affected package~~
- ~~The version bounds in `.cabal` files are updated~~
  - ~~Code is linted with `hlint`.  See `.github/workflows/check-hlint.yml` to get the `hlint` version~~
  - ~~Code is formatted with `stylish-haskell`.  See `.github/workflows/stylish-haskell.yml` to get the `stylish-haskell` version~~
  - ~~Code builds on Linux, MacOS and Windows for `ghc-8.10.7` and `ghc-9.2.7`~~
- [x] Self-reviewed the diff

### DIFF
--- a/.github/workflows/release-ghcr.yaml
+++ b/.github/workflows/release-ghcr.yaml
@@ -1,4 +1,4 @@
-name: Upload to ghci.io
+name: Upload to ghcr.io
 
 on:
   push:
--- a/.github/workflows/release-ghcr.yaml
+++ b/.github/workflows/release-ghcr.yaml
@@ -33,12 +33,12 @@ jobs:
             success)
               echo "ci/hydra-build:required succeeded"
               exit 0;;
-            failure)
-              echo "ci/hydra-build:required failed"
-              exit 1;;
-            *)
+            '')
               echo "ci/hydra-build:required pending. Waiting 30s..."
               sleep 30;;
+            *)
+              echo "ci/hydra-build:required terminated unsuccessfully"
+              exit 1;;
           esac
         done
 
--- a/.github/workflows/release-ghcr.yaml
+++ b/.github/workflows/release-ghcr.yaml
@@ -28,7 +28,7 @@ jobs:
     - name: Waiting for ci/hydra-build:required to complete
       run: |
         while [[ true ]]; do
-          conclusion=$(gh api repos/$GITHUB_REPOSITORY/commits/$GITHUB_SHA/check-runs --jq '.check_runs[] | select(.name == "ci/hydra-build:required") | .conclusion')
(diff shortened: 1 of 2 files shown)


johnalotoski (approved):

neilmayhew (approved):
LGTM, although I think it would be good to expand the cases for `"$conclusion"` as well, so that the default case produces an error.

.github/workflows/release-upload.yaml
```diff
@@ -25,20 +25,21 @@ jobs:
     name: "Wait for hydra check-runs"
     runs-on: ubuntu-latest
     steps:
-    - name: Get specific check run status
+    - name: Waiting for ci/hydra-build:required to complete
```
dermetfan, line 28: Updated to the same as in `release-ghcr.yaml`. Seems it was copied from an example and then forgotten to be updated.

.github/workflows/release-ghcr.yaml
```diff
@@ -28,17 +28,18 @@ jobs:
...
             success)
-              echo "ci/hydra-build:required succeeded"
+              echo "$check_name succeeded"
               exit 0;;
-            failure)
-              echo "ci/hydra-build:required failed"
-              exit 1;;
-            *)
-              echo "ci/hydra-build:required pending. Waiting 30s..."
+            '')
```
neilmayhew, line 37: I'm concerned that this won't catch the situation where there is no check to wait for, because the `jq` expression will return an empty string in that situation as well.

To handle this correctly would be a bit tricky, though. If the `jq` expression was `map(.conclusion)` you'd get either `[null]` (for in-progress) or `[]` (for no check). You'd then have to run another `jq` to check whether the array was empty (eg using `length`) and if not, to extract the actual conclusion.

I'm not sure, though, whether it's worth going to this much trouble. Also, it's possible the check doesn't exist because it hasn't started yet.

dermetfan: > Also, it's possible the check doesn't exist because it hasn't started yet.

That's right, and therefore I think there is nothing to do differently in the `[null]` vs the `[]` case.

If no check is ever delivered we just run into the regular timeout, but there's nothing we could have done to throw an error earlier, unless we also query the repo settings to see whether the check we expect is configured as required, but I guess that's out of scope for this script, and still would not catch misconfigurations in the remaining CI infra anyway.

neilmayhew (approved):
Great to have this fixed

PR #6138, opened 2025-03-04
cardano-testnet: harmonize node and genesis files API + allow (for example) to only pass a custom conway genesis

smelc:
# Description

* Work towards implementing option 3. from https://github.com/IntersectMBO/cardano-node/issues/6137: we support a mix of generated genesis files and genesis files provided by the user.
* Follow-up of problem reported by user: https://github.com/IntersectMBO/cardano-node/issues/6130#issuecomment-2692010489

Together with https://github.com/IntersectMBO/cardano-node/pull/6103 (6103 being stacked on this PR), this PR allows for the following:

* Having `cardano-testnet` usable as a standalone executable
* Allowing overriding 0, 1, or many genesis files. If a user doesn't specify a genesis file, `cardano-testnet` generates it. This is exemplified in https://github.com/IntersectMBO/cardano-node/pull/6103's test of the `cardano-testnet` executable, that specifies only the Alonzo and Conway genesis files, but leaves the Shelley genesis to be created automatically: https://github.com/IntersectMBO/cardano-node/blob/smelc/testnet-pass-config-files/cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/CommandLineExecutable.hs#L33
* This is done by passing (or not) custom genesis files. If a file is omitted, it's generated. There's no need for corresponding flags.

# Checklist

- [x] New tests are added if needed and existing tests are updated :left_right_arrow: Tested in https://github.com/IntersectMBO/cardano-node/pull/6103
- [x] Self-reviewed the diff


smelc:
@IntersectMBO/performance-and-tracing> this PR removes the creation of the genesis `spec` files. Given that we now have a complex mechanism for genesis files being passed by the user or generated on demand, maintaining the creation of `spec` files is too complex on our side.

carbolymer (approved):

PR #6148, opened 2025-03-14
cardano-testnet: allow to take node configuration file as input

smelc:
> [!WARNING]
>
> If this design is chosen (over https://github.com/IntersectMBO/cardano-node/pull/6103), then https://github.com/cardano-foundation/developer-portal/pull/1426 needs to be amended correspondingly

# Description

Fixes https://github.com/IntersectMBO/cardano-node/issues/6069

Fixes https://github.com/IntersectMBO/cardano-node/issues/6137

Variant of https://github.com/IntersectMBO/cardano-node/pull/6103 where passing the node configuration file and the genesis files are tied:

* Either you pass the node configuration file and all 3 genesis files,
* or you pass none of these; and `cardano-testnet` generates everything for you

# Follow-up issues to create

@Jimbo4350> please tick them if you agree :slightly_smiling_face:  I'll create them

* [ ] Have `pTestnetNodeOptions` return a list of `TestnetNodeOption` and `TestnetNodeOption` be `UserProvidedNodeOptions Filepath | SpoNodeOptions [Srting] | RelayNodeOptions [ String]`. In other words merge the types `TestnetNodeOptions` and `AutomaticNodeOption`. This will generalize the behavior so that the user can create multiple nodes, no matter whether they are using custom options or using default behavior (`Spo` or `Relay`). This will require a bit of generalization in [Cardano.hs](cardano-testnet/src/Testnet/Start/Cardano.hs) but not much. :left_right_arrow: shown here: https://github.com/IntersectMBO/cardano-node/pull/6155
* [ ] ~~Change `cardanoTestnet` to take `Filepath` for genesis files instead of Haskell values, maybe?~~ Decided against
* [ ] Check that the CLI flags changing genesis values are not used when passing a custom genesis file (TODOs are added in the code already about this).
* [ ] Reflect in the types the tying behavior between the node configuration file being passed by the user and the genesis files being passed by the user too. :left_right_arrow: shown here: https://github.com/IntersectMBO/cardano-node/pull/6155
* [X] Created already: https://github.com/IntersectMBO/cardano-node/issues/6122

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/cardano-testnet/src/Parsers/Run.hs
+++ b/cardano-testnet/src/Parsers/Run.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 
 module Parsers.Run
   ( commands
@@ -8,14 +9,27 @@ module Parsers.Run
   , opts
   ) where
 
-import           Cardano.CLI.Environment
+import           Cardano.Api.Ledger (StandardCrypto)
+import           Cardano.Api.Shelley (ShelleyGenesis)
 
+import           Cardano.CLI.Environment
+import           Cardano.Ledger.Alonzo.Genesis (AlonzoGenesis)
+import           Cardano.Ledger.Conway.Genesis (ConwayGenesis)
+import           Cardano.Node.Configuration.POM
+import           Cardano.Node.Types
+
+import qualified Data.Aeson as Aeson
 import           Data.Foldable
+import           Data.Monoid
+import           Data.Yaml (decodeFileThrow)
 import           Options.Applicative
 import qualified Options.Applicative as Opt
+import qualified System.Directory as System
+import           System.FilePath (takeDirectory, (</>))
 
 import           Testnet.Property.Run
 import           Testnet.Start.Cardano
+import           Testnet.Start.Types
 
 import           Parsers.Cardano
 import           Parsers.Help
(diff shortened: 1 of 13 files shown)


cardano-testnet/src/Parsers/Run.hs
```diff
@@ -52,5 +66,49 @@ runTestnetCmd = \case
...
+        Right nodeConfig -> pure nodeConfig
+    genesisFilepathToData :: Aeson.FromJSON a => GenesisFile -> IO (UserProvidedData a)
+    genesisFilepathToData (GenesisFile filepath) = do
+      exists <- System.doesFileExist filepath
+      if exists
+        then UserProvidedData <$> decodeFileThrow filepath
+        else
+          -- Here we forbid mixing defaulting of genesis files with providing a user node configuration file:
+          --
+          -- Because of this check, either the user passes a node configuration file AND he passes all the genesis files.
```
Jimbo4350, line 111: :+1:

cardano-testnet/src/Testnet/Start/Cardano.hs
```diff
@@ -244,6 +263,8 @@ cardanoTestnet
 
   -- Make Non P2P topology files
   forM_ (zip [1..] portNumbers) $ \(i, myPortNumber) -> do
+    -- TODO: if the user provided its own configuration file, and requested a P2P topology file,
```
Jimbo4350, line 266: Can you turn this into an issue so we can track it?

cardano-testnet/src/Testnet/Start/Cardano.hs
```diff
@@ -282,8 +303,9 @@ cardanoTestnet
         , "--database-path", nodeDataDir </> "db"
         ]
         <> spoNodeCliArgs
-        <> testnetNodeExtraCliArgs nodeOptions
+        <> maybe [] extraCliArgs nodeOptions
     pure $ eRuntime <&> \rt -> rt{poolKeys=mKeys}
+    -- TODO log the node's pid in a file. This is useful for killing the node later.
```
Jimbo4350, line 308: If this is deserving of an issue can you create it?

Jimbo4350 (approved):
LGTM :+1:

cardano-testnet/src/Testnet/Start/Cardano.hs
smelc: Follow-up issue created: https://github.com/IntersectMBO/cardano-node/issues/6149

cardano-testnet/src/Testnet/Start/Cardano.hs
smelc: Done here: https://github.com/IntersectMBO/cardano-node/issues/6150

PR #6156, opened 2025-03-18
New Tracing: PrometheusSimple backend, quality-of-life features and Node 10.3 package bumps

mgmeier:
# Description

This PR provides package version bumps for the new tracing system aimed at Node 10.3 release. Furthermore, it adds / changes the following features:

**cardano-node** / **trace-dispatcher** -- New version `2.9`
* New `PrometheusSimple` backend which runs a simple TCP server for direct exposition of metrics, without forwarding.
* New `maxReconnectDelay` config option in `TraceOptionForwarder`: Specifies maximum delay (seconds) between (re-)connection attempts of a forwarder (default: 60s).
* Fix: change semantics of several `Forge.*` metrics to counters - to match semantics of legacy tracing.
* Fix: remove defective `ForgingThreadStats`, keeping `ForgingStats`.
* Fix: correct `blockdelay_cdf*` metric
* Fix: correct `blockReplayProgress` metric
* Remove `ekg-wai` from dependencies.

**cardano-tracer** -- New version `0.3.2`
* When requesting forwarded metrics, ask for delta to previous request only. New config option `ekgRequestFull` defaults to `false`; set to `true` to revert this behavior.
* Add up-to-date symlink for each log rotation, pointing to the most recent file (`node.json` or `node.log`, depending on the log format).
* Fix: CLI option `--min-log-severity` is now observed correctly.
* Add `metricsNoSuffix` to configuration, which when set removes suffixes like `_int` from metrics names, making them more similar to those in the old system (optional boolean; default: `false`).
* Remove `metricsComp` metric name remapping from configuration.

**tx-generator** -- New version `2.15`
* Remove the dependency on `cardano-api`'s deprecated `ProtocolParameters` type.
* New cabal flag `withplutuslib` is added, enabling import and re-compilation of scripts from `plutus-scripts-bench` - default: false.

Last not least, this PR includes some optimizations to the new tracing system's forwarding, aimed at reducing CPU usage especially when under low / idle load.

_Known issues_: The new system forging metrics remain cumulative for all block producing credentials provided to the Node, over the runtime of the Node process. If more than 1 set of credentials is provided (for operating a so-called "dense pool"), or credentials are changed via SIGHUP, it's not reflected individually in the metrics.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [x] Self-reviewed the diff


bench/tx-generator/src/Cardano/Benchmarking/Tracer.hs
jutaro, line 1: maybe use maybe: maybe mempty forwardFormatter mbTrForward

bench/tx-generator/src/Cardano/Benchmarking/Tracer.hs
```diff
@@ -69,10 +69,10 @@ generatorTracer ::
 generatorTracer tracerName mbTrStdout mbTrForward = do
   forwardTrace <- case mbTrForward of
                         Nothing -> mempty
-                        Just trForward -> forwardFormatter Nothing trForward
+                        Just trForward -> forwardFormatter trForward
```
jutaro, line 72: Maybe maybe maybe: maybe mempty forwardFormatter mbTrForward

jutaro (commented):
I'm seeing that the traces in txgen are not filtered, but I find a hack for silencing, that I don't understand:
```
isPrefixSilent :: Text -> Bool
isPrefixSilent namespace =
  fromMaybe False $ do
    guard $ not $ Text.null namespace
    prefix <- find (any (namespace `Text.isPrefixOf`)) (Map.keys opts)
    conf <- prefix `Map.lookup` opts
    pure $ configSilent `elem` conf
  where
    TraceConfig{tcOptions = opts} = initialTraceConfig
```

If I would know the requirements, I can probably write a different version?

mgmeier:
> I'm seeing that the traces in txgen are not filtered, but I find a hack for silencing, that I don't understand:
> 
> ```
> isPrefixSilent :: Text -> Bool
> isPrefixSilent namespace =
>   fromMaybe False $ do
>     guard $ not $ Text.null namespace
>     prefix <- find (any (namespace `Text.isPrefixOf`)) (Map.keys opts)
>     conf <- prefix `Map.lookup` opts
>     pure $ configSilent `elem` conf
>   where
>     TraceConfig{tcOptions = opts} = initialTraceConfig
> ```
> 
> If I would know the requirements, I can probably write a different version?

That's not a hack, it's a shortcut. For the `tx-generator` with its low number of tracers, and top-level namespace entries only, it's easily possible to avoid constructing a tracer entirely when you already know that all of its messages would be filtered out (silenced).
So for that, null tracer (`pure mempty`) is created, instead of an actual one with filtering applied.

Do you think that that could be improved, or maybe it's unnecessary? It was implemented in the early days of the new tracing system, situation may have changed. I agree that if it stays like this, I should at least an explanatory comment.

bench/tx-generator/src/Cardano/Benchmarking/Tracer.hs
mgmeier: ✅

Icelandjack (approved):

PR #6181, opened 2025-04-16
Update cardano-node changelog

aniketd:
# Description

@jasagredo found the inconsistency that the change-log for `cardano-node` wasn't kept completely up-to-date and its name wasn't following the convention.

As to "why" this might be: in my under-qualified opinion, I think most likely due to an undocumented technical limitation in our automated operational infrastructure, or second-most likely due to an egregious oversight. If it is indeed the former, I expect this PR to trigger the CI action in question.

In this PR, we update the `cardano-node` change-log, add a link to it from the top-level change-log, and also update its reference in the respective `cardano-node.cabal` file.

> NOTE: The markdown link checker is expected to fail, since the link will only be valid once we merge this PR

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [x] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Changelogs for components can be found as follows:
 - [cardano-submit-api](https://github.com/IntersectMBO/cardano-node/blob/master/cardano-submit-api/CHANGELOG.md) 
 - [trace-forward](https://github.com/IntersectMBO/cardano-node/blob/master/trace-forward/CHANGELOG.md) 
 - [cardano-tracer](https://github.com/IntersectMBO/cardano-node/blob/master/cardano-tracer/CHANGELOG.md)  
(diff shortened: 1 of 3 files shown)


jasagredo (commented):
In general I think these changes are fine, but I don't know if they are exhaustive. Is it decided that we are keeping track of the changelog for cardano-node then in the file?

aniketd:
> Is it decided that we are keeping track of the changelog for cardano-node then in the file?

I made sense to me to keep things consistent across packages. Because there does not seem to be any other place (within `git` history) that we do this. This is in addition to the release notes on GitHub under "releases", of course, though. On Slack, I remember @disassembler mentioning that it makes sense to keep this change-log up-to-date from here on.

@jasagredo

PR #6192, opened 2025-04-24
renamed ForkPolicy configuration option as ResponderCoreAffinityPolicy

coot:


karknu (approved):
LGTM

cardano-node/src/Cardano/Node/Configuration/POM.hs
```diff
@@ -197,20 +197,22 @@ data NodeConfiguration
...
-       , ncForkPolicy :: NCForkPolicy
+       , ncResponderCoreAffinityPolicy :: ResponderCoreAffinityPolicy
        } deriving (Eq, Show)
 
--- | We expose the `Ouroboros.Network.Mux.ForkPolicy` as a `NodeConfiguration` field.
--- * `NoBindForkPolicy` corresponds to `Ouroboros.Network.Mux.noBindForkPolicy`
--- * `ResponderForkPolicy` corresponds to `Ouroboros.Network.Mux.responderForkPolicy`
+-- | We expose the `Mux.ForkPolicy` as a `NodeConfiguration` field.
+--
+-- * `NoResponderCoreAffinity` corresponds to `Mux.noBindForkPolicy`
```
Jimbo4350, line 205: What do these mean however? Can you link to documentation?

Jimbo4350 (approved):

cardano-node/src/Cardano/Node/Configuration/POM.hs
coot: It ought to be linked through haddocks, but there's something wrong with your haddocks pipeline: https://cardano-node.cardano.intersectmbo.org/cardano-node/Cardano-Node-Configuration-POM.html

coot: It will work if:
* in this PR we use full module names rather than qualification (which likely is is a bug in haddock)
* use a newer version of `cabal` (I tested it with `cabal-0.14.2.0`) (one needs to get rid of the `--local` switch which was removed from `cabal project-haddock` command). The `nix develop` exposes `cabal-install-3.10.3.0`.
* in your `github-page` action you might need to use a custom `cabal.project.local` file which enables documentation for dependencies. Locally I have:
```
documentation: False
package ouroboros-network-framework
  documentation: True
```
and it worked, but it's probably enough to change `documentation: True` and not include the per-package specific option (which would be annoying if one needed to do that for all dependencies); or maybe a custom `cabal.project.local` file is not needed at all.

PR #6195, opened 2025-04-24
Add snapshot-converter

johnalotoski:
# Description

* Adds the consensus snapshot-converter to pkgs and the release bins
* Fixes the ci artifact nixosTest for the larger tarball size
* Removes a local haskellNix windows override that has been upstreamed
* Bump iohkNix for LedgerDB cfg and make corresponding ci adjustments

# Checklist

- [X] Self-reviewed the diff


configuration/cardano/mainnet-config-bp.json
```diff
@@ -12,9 +12,15 @@
   "LastKnownBlockVersion-Alt": 0,
   "LastKnownBlockVersion-Major": 3,
   "LastKnownBlockVersion-Minor": 0,
+  "LedgerDB": {
+    "Backend": "V2InMemory",
+    "NumOfDiskSnapshots": 2,
+    "QueryBatchSize": 100,
+    "SnapshotInterval": 4320
```
jasagredo, line 19: For these three options, it is fine to make them explicit which would use something like `RequestedNumOfDiskSnapshots 2`, or leave them as undefined which will use `DefaultNumOfDiskSnapshots` which in the end will result in the same value.

jasagredo (approved):
This looks fine from my side.

disassembler (approved):

PR #6206, opened 2025-05-02
ghc: drop haskellNix ghc 8.10 variant support

johnalotoski:
# Description

Drops support for haskellNix ghc 8.10

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/nix/haskell.nix
+++ b/nix/haskell.nix
@@ -22,8 +22,8 @@ let
       src = ../.;
       name = "cardano-node";
       compiler-nix-name = lib.mkDefault "ghc96";
-      # extra-compilers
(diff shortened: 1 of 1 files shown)


mgmeier (approved):
Could we possibly also remove the `passthru.eventlogged` package (and the underlying `eventlogged` build config) from `haskell.nix`? Eventlog functionality is included by default with GHC9.4 and above; the `-l` switch in the cardano-node svc will keep working without disruption.

Approving anyway (and very happily).

cf. https://ghc.gitlab.haskell.org/ghc/doc/users_guide/phases.html#ghc-flag-eventlog
> Note that as of GHC 9.4 and later eventlog support is included in the RTS by default and the `-eventlog` is deprecated.

disassembler (approved):

PR #6209, opened 2025-05-06
scripts: replace heavy pandoc dependency with text-based markdown table in changelog link generator

mgmeier:
# Description

This small PR removes the heavy pandoc dependency from the changelog link generator script. The use case of that script is pretty narrow, and won't likely require an abstract representation of a table.

Instead, common functions from the `text` package are leveraged to render an identical output - significantly easing build time and dependency footprint of this script.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [x] Self-reviewed the diff


scripts/generate-release-changelog-links.hs
```diff
@@ -298,25 +291,33 @@ runGitHub :: GitHub.GitHubRW req res => GitHubAccessToken -> req -> res
...
-        [ mkCell [Pandoc.Str n]
-        , mkCell [Pandoc.Str (dispVer v)]
-        , mkCell (foldMap (\(fn, link) -> [Pandoc.Link mempty [Pandoc.Str fn] (link, fn)]) linkMaybe)
-        ]
-    mkCell t = Pandoc.Cell mempty Pandoc.AlignDefault 1 1 [Pandoc.Plain t]
+    mkHeader                        = ["Package", "Version", "Changelog"]
+    mkRow (PkgName n, v, linkMaybe) = [n        , dispVer v, dispLink linkMaybe]
+
+    -- example result: [CHANGELOG.md](https://github.com/IntersectMBO/cardano-base/blob/f11ddc7f/cardano-slotting/CHANGELOG.md "CHANGELOG.md")
+    dispLink (Just (file, link)) = mconcat ["[", file, "](", link, " \"", file, "\")"]
```
neilmayhew, line 305: I think using `format` would look better here, especially since the rest of the file uses it a lot. Perhaps:

```suggestion
    dispLink (Just (file, link)) = format $ "["%file%"]("%link%" \""%file%"\")"
```

However, if you prefer the `mconcat` version I think that's fine too.

neilmayhew (approved):
Nicely done! Clear, succinct code and good improvements overall.

I wasn't too bothered by the dependency weight of pandoc since this script needs to be run only once or twice during the entire release process and waiting an extra couple of minutes wasn't a problem. However, I still think this refactoring is worthwhile.

scripts/generate-release-changelog-links.hs
mgmeier: Great suggestion - adopted it.

PR #6215, opened 2025-05-14
Use Latest Haskell.nix

hamishmack:
Try using the latest haskell.nix

This also results in a new `nixpkgs` being used (required for windows cross compilation with this version of haskell.nix).

This means a new `actionlint` version is used and the GitHub actions have been updated for that.

### DIFF
--- a/.github/workflows/check-cabal-files.yml
+++ b/.github/workflows/check-cabal-files.yml
@@ -15,7 +15,7 @@ jobs:
       with:
         cabal-version: "3.10.2.0"
 
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
 
     - name: Cabal check
       run: ./scripts/ci/check-cabal-files.sh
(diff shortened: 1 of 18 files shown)


hamishmack:
Still to do:
* We need to remove the remaining dependencies on GHC <9.6.  Currently the version of hlint used builds with 9.2.8 (that requires 8.10.7 to bootstrap and is failing with the latest haskell.nix/nixpkgs).  Updating to the new version of hlint will need some updates.
*  It would be good if we could merge https://github.com/input-output-hk/cardano-automation/pull/8.
* This RP currently skips the custom emacs setup (the hasklig-mode is missing from the new nixpkgs).  We need to check if emacs in the workbench shell still does what is needed.
* For some reason GHC 9.6.7 seems to break the tests on macOS.  So I've pinned it to 9.6.6 hopefully that will fix the macOS failures (until we can work out why 9.6.7 fails).

mgmeier:
I've added a commit bumping the workbench emacs setup - it does seem to resolve any compatibility issues with up-to-date nixpkgs.

carbolymer (approved):

PR #6262, opened 2025-06-24
HOTFIX: Fix metric in ForgingStats tracer

jutaro:
This hotfix restores a single line that was erroneously overwritten in the commit
cardano-node: periodic ledger metrics tracer.

Specifically, the line added by Baldur in
cardano-node: Added missing number stat to the total.
was not carried over when simplifying the tracer type.


mgmeier (approved):
LGTM!

PR #6265, opened 2025-06-27
Removed stale comment, code formatting, bench config

coot:
# Description

- **Removed stale comment**
- **POM: code formatting**
- **workbench: enforce `PeerSharingDisabled` for benchmarking configs**

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [x] Self-reviewed the diff

### DIFF
--- a/cardano-node/src/Cardano/Node/Configuration/POM.hs
+++ b/cardano-node/src/Cardano/Node/Configuration/POM.hs
@@ -407,7 +407,6 @@ instance FromJSON PartialNodeConfiguration where
               Just True  -> Last $ Just EnabledP2PMode
 
       -- Peer Sharing
-      -- DISABLED BY DEFAULT
       pncPeerSharing <- Last <$> v .:? "PeerSharing"
 
       -- pncConsensusMode determines whether Genesis is enabled in the first place.
--- a/cardano-node/src/Cardano/Node/Configuration/POM.hs
+++ b/cardano-node/src/Cardano/Node/Configuration/POM.hs
@@ -672,9 +672,13 @@ defaultPartialNodeConfiguration =
             V2InMemory
             noDeprecatedOptions
     , pncProtocolIdleTimeout      = Last (Just Ouroboros.defaultProtocolIdleTimeout)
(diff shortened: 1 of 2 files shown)


mgmeier (approved):

cardano-node/src/Cardano/Node/Configuration/POM.hs
```diff
@@ -407,7 +407,6 @@ instance FromJSON PartialNodeConfiguration where
               Just True  -> Last $ Just EnabledP2PMode
 
       -- Peer Sharing
-      -- DISABLED BY DEFAULT
```
jasagredo, line 410: Instead of removing it, I would have written "ENABLED BY DEFAULT".

mgmeier: You could point to the module in `o-n` where the default value is defined.

jasagredo: That would indeed be even better as it cannot get out of sync again (or it would at least be clear it got out of sync)

coot: Check the commit I pushed.

jasagredo: That seems fine to guide the reader, but the haddocks themselves do not mention the default value. For example:
image
And it also doesn't have a link to the Source to go check if the default is defined somewhere around there.

coot: There's a `source` link to view the value where it's defined, [ref](https://ouroboros-network.cardano.intersectmbo.org/ouroboros-network/Ouroboros-Network-Diffusion-Configuration.html#v:defaultPeerSharing). If we include default values in haddocks, there's a chance they will become outdated; thus, it's not my preferred method.

mgmeier (approved):
LGTM

cardano-node/src/Cardano/Node/Configuration/POM.hs
jasagredo: Aha, good point. I agree

PR #6267, opened 2025-07-01
cardano-tracer: handle SIGINT same as cardano-node; ignore overflow in Forwarder backend lacking socket config

Icelandjack:
# Description

Two commits:
* *cardano-tracer*: Use same signal handler for SIGINT as SIGTERM.
* *trace-dispatcher*: Ignores overflow when there is no socket address specified.

# Checklist

  - [X] Code builds on Linux, MacOS and Windows for `ghc-9.6` and `ghc-9.12`
- [X] Self-reviewed the diff

### DIFF
--- a/cardano-tracer/src/Cardano/Tracer/Run.hs
+++ b/cardano-tracer/src/Cardano/Tracer/Run.hs
@@ -154,8 +154,8 @@ doRunCardanoTracer config rtViewStateDir tr protocolsBrake dpRequestors = do
     traceWith tr TracerShutdownInitiated
 #if RTVIEW
     backupAllHistory tracerEnv tracerEnvRTView
-#endif
     traceWith tr TracerShutdownHistBackup
+#endif
     applyBrake (teProtocolsBrake tracerEnv)
     traceWith tr TracerShutdownComplete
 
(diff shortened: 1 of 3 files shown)


cardano-node/cardano-node.cabal
```diff
@@ -189,7 +189,7 @@ library
                       , ouroboros-consensus ^>= 0.27
                       , ouroboros-consensus-cardano ^>= 0.25
                       , ouroboros-consensus-diffusion ^>= 0.23
-                      , ouroboros-consensus-protocol
+                      , ouroboros-consensus-protocol ^>= 0.12
```
mgmeier: Modifying `cardano-node`'s cabal file shouldn't be necessary.

cardano-tracer/src/Cardano/Tracer/Run.hs
```diff
@@ -154,8 +154,8 @@ doRunCardanoTracer config rtViewStateDir tr protocolsBrake dpRequestors = do
     traceWith tr TracerShutdownInitiated
 #if RTVIEW
     backupAllHistory tracerEnv tracerEnvRTView
-#endif
     traceWith tr TracerShutdownHistBackup
+#endif
```
mgmeier, line 158: ❤️

mgmeier (approved):

PR #6271, opened 2025-07-02
Change to allowList for release bins and add tx-generator

johnalotoski:
# Description

This PR adds `tx-generator` and removes several release binaries.
To do so, musl builds needed to be added and git hash stamping was applied to set commit version in the binary.
Workbench use of tx-generator was changed to a passthru noGitRev package version to avoid redundant building of the same source because of differing commit hash.

Release bin filtering was changed from a denyList to an acceptList  approach to avoid inclusion creep if the denyList isn't maintained.

# Checklist
- [X] Self-reviewed the diff

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -291,12 +291,27 @@
         exes;
 
       ciJobs = let
+        releaseBins = [
+          "bech32"
+          "cardano-cli"
+          "cardano-node"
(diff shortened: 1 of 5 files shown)


mgmeier (approved):
LGTM!

PR #6288, opened 2025-07-20
Do not retain ShelleyGenesis on benchmarks

jasagredo:
ShelleyGenesis contains stuffed UTxOs, and they are retained by this field.


jasagredo:
Close in favor of other changes @mgmeier is planning on doing on a separate PR

mgmeier:
This change was incorporated into #6295

PR #6294, opened 2025-07-30
New Tracing: Fix repeated "100%" trace events during chain replay

jutaro:
*Problem*

In rare corner cases, the node emits multiple identical trace events indicating 100% progress during chain replay. This results in noisy and redundant log output and can confuse operators or monitoring tools relying on progress metrics.

*Solution*

This PR introduces a minimal state-tracking mechanism (`ReplayBlockState`) to suppress duplicate progress trace messages, namely those indicating no significant advancement (e.g. repeated 100%).

*The fix:*

*  Stores the last slot number for which a progress event was emitted
*  Only emits a new event if the progress has increased by at least 0.01 percentage points
*  Uses `MVar` to manage replay state safely across events

*Additional Improvements*

*  Simplified and clarified `ReplayBlockStats`
*   Extracted human-readable and machine-readable progress computations (`progressForHuman`, `progressForMachine`)
*   Removed unnecessary state fields like `rpsDisplay` and `rpsLastProgress`

### DIFF
--- a/cardano-node/src/Cardano/Node/Tracing/Tracers/BlockReplayProgress.hs
+++ b/cardano-node/src/Cardano/Node/Tracing/Tracers/BlockReplayProgress.hs
@@ -9,7 +9,7 @@ module Cardano.Node.Tracing.Tracers.BlockReplayProgress
 import           Cardano.Api (textShow)
 
 import           Cardano.Logging
-import           Ouroboros.Consensus.Block (Point, SlotNo, realPointSlot)
+import           Ouroboros.Consensus.Block (SlotNo, realPointSlot)
 import qualified Ouroboros.Consensus.Storage.ChainDB as ChainDB
 import qualified Ouroboros.Consensus.Storage.LedgerDB as LedgerDB
 import           Ouroboros.Network.Block (pointSlot, unSlotNo)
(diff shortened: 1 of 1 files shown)


cardano-node/src/Cardano/Node/Tracing/Tracers/BlockReplayProgress.hs
```diff
@@ -71,28 +88,34 @@ instance MetaTrace ReplayBlockStats where
...
+      let (st', mbStats) = mbProduceBlockStats st msg
+      pure (st', fmap ((ctx,) . Right) mbStats)
+
+    mbProduceBlockStats :: ReplayBlockState -> ChainDB.TraceEvent blk -> (ReplayBlockState, Maybe ReplayBlockStats)
+    mbProduceBlockStats st@(ReplayBlockState (Just lastSlot))
+      (ChainDB.TraceLedgerDBEvent
+       (LedgerDB.LedgerReplayEvent
+        (LedgerDB.TraceReplayProgressEvent
+         (LedgerDB.ReplayedBlock curSlot [] _ (LedgerDB.ReplayGoal replayToSlot)))))
+            | progressFor (realPointSlot curSlot) replayToSlot - progressFor lastSlot replayToSlot >= 0.1 =
```
mgmeier: A threshold of `0.1` might mute meaningful progress information (especially with long chains like mainnet). Would a threshold of `0.01` make more sense - especially since you're  rounding to 2 decimal places (`round2`) above?

cardano-node/src/Cardano/Node/Tracing/Tracers/BlockReplayProgress.hs
```diff
@@ -71,28 +88,34 @@ instance MetaTrace ReplayBlockStats where
...
-replayBlockStats st@ReplayBlockStats {} _context _ = pure st
+    process :: MVar ReplayBlockState
+            -> (LoggingContext, Either TraceControl (ChainDB.TraceEvent blk))
+            -> IO (Maybe (LoggingContext, Either TraceControl ReplayBlockStats))
+    process _ (ctx, Left control) = pure (Just (ctx, Left control))
+    process var (ctx, Right msg) = modifyMVar var $ \st -> do
+      let (st', mbStats) = mbProduceBlockStats st msg
+      pure (st', fmap ((ctx,) . Right) mbStats)
+
+    mbProduceBlockStats :: ReplayBlockState -> ChainDB.TraceEvent blk -> (ReplayBlockState, Maybe ReplayBlockStats)
```
mgmeier: You could reduce code duplication and improve readability by changing the type to
```
mbProduceBlockStats :: ChainDB.TraceEvent blk -> ReplayBlockState -> (ReplayBlockState, Maybe ReplayBlockStats)
```

Like this, you need to match on the event only once, and then add a conditional guard like (pseudo-code):
```
arg: st@(ReplayBlockState rbs)
   ...
    < match >
       |  maybe True (\lastSlot -> progressFor etc.) rbs -> emit and update state
    _ -> (st, Nothing)
```

Like this, it becomes much better readable under which condition the trace is emitted, and also it's clear that there's only 1 code path so the trace emitted will be identical.

mgmeier (changes requested):

cardano-node/src/Cardano/Node/Tracing/Tracers/BlockReplayProgress.hs
```diff
@@ -71,28 +88,34 @@ instance MetaTrace ReplayBlockStats where
...
-      pure $ if (progress' == 0.0 && not rpsDisplay)
-                || ((progress' - rpsLastProgress) > 0.1)
-                then ReplayBlockStats True slotno endslot progress' progress'
-                else ReplayBlockStats False slotno endslot progress' rpsLastProgress
-replayBlockStats st@ReplayBlockStats {} _context _ = pure st
+    process :: MVar ReplayBlockState
+            -> (LoggingContext, Either TraceControl (ChainDB.TraceEvent blk))
+            -> IO (Maybe (LoggingContext, Either TraceControl ReplayBlockStats))
+    process _ (ctx, Left control) = pure (Just (ctx, Left control))
+    process var (ctx, Right msg) = modifyMVar var $ \st -> do
```
mgmeier, line 98: A *very* minor detail: As per the Haskell2010 report "A case expression is evaluated by pattern matching the expression e against the individual alternatives. The alternatives are tried sequentially, from top to bottom."

For block replay it's not much of an issue, but in general for the tracing system it's advisable to match on the `Right` case first (as it's already being done throughout the rest of the codebase).

mgmeier (approved):
LGTM, thank you @jutaro

PR #6334, opened 2025-09-26
Fill in missing pieces from Consensus for 10.6

jasagredo:
# Description

This PR fills some `undefined`s and provides some instances and fields coming from Consensus. Each commit is atomic.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -100,8 +100,8 @@ source-repository-package
   type: git
   location: https://github.com/IntersectMBO/ouroboros-consensus
   -- latest master
-  tag: 71b02607c8a39ed4d8c983b281b05452ed8c01ce
-  --sha256: sha256-/vnZnAPsEuqQMzG5NGHaWk9vyefBWMft7/rKQ+yyYTQ=
+  tag: ac1a8db76f4c7a38a9a6b962a40fa722d5bd55a6
+  --sha256: sha256-8MlAxCi1wXLc2p0csYTKZ4RW7+uqWvxOBs5IhISzwxk=
   subdir:
       ouroboros-consensus
       ouroboros-consensus-cardano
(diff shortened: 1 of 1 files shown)


cardano-node/src/Cardano/Tracing/OrphanInstances/Shelley.hs
```diff
@@ -1527,6 +1528,14 @@ instance
     -- TODO: fix
     Conway.ScriptIntegrityHashMismatch _ _ -> undefined
 
+instance ToObject (Praos.PraosTiebreakerView crypto) where
+  toObject v (Praos.PraosTiebreakerView sl issuer issueNo vrf) =
+    mconcat [ "slotNo" .= toObject v sl
+            , "issuer" .= textShow issuer
+            , "issueNo" .= textShow issueNo
+            , "vrf" .= textShow vrf
+            ]
```
amesgen, line 1537: This can be made a bit nicer (not using verbose `Show` instances) by adapting what is present on `main` for `PraosChainSElectView`:
https://github.com/IntersectMBO/cardano-node/blob/0f6b522cf720a298146875135cf91aa6043dbd74/cardano-node/src/Cardano/Tracing/OrphanInstances/Shelley.hs#L1330-L1346
(Also, I think it makes sense to use record field names to avoid future unexpected changes)

cardano-node/src/Cardano/Node/Tracing/Era/Shelley.hs
```diff
@@ -1477,6 +1478,14 @@ instance
     -- TODO: fix
     Conway.ScriptIntegrityHashMismatch _ _ -> undefined
 
+instance LogFormatting (Praos.PraosTiebreakerView crypto) where
+  forMachine _dtal (Praos.PraosTiebreakerView sl issuer issueNo vrf) =
+    mconcat [ "slotNo" .= condense sl
+            , "issuer" .= textShow issuer
+            , "issueNo" .= textShow issueNo
+            , "vrf" .= textShow vrf
+            ]
```
amesgen, line 1487: Same comment as for `ToObject (Praos.PraosTiebreakerView crypto)`

amesgen (commented):
The GSM part looks good; left comments regarding the tiebreaker view

fraser-iohk (commented):
KES agent tracing and shelley credentials changes look good 👍

PR #6346, opened 2025-10-21
Remove Integration Monad

Jimbo4350:
# Description

- Remove `Integration` monad from functions that we also want to use outside of property tests. 

- Introduce annotated exceptions (via `liftIOAnnotated`)  where the `Integration` monad has been removed in order to preserve debugging information.
```
┃ hprop_dump_config :: H.Property
    38 ┃ hprop_dump_config = integrationRetryWorkspace 2 "dump-config-files" $ \tmpDir -> H.runWithDefaultWatchdog_ $ do
       ┃ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
       ┃ │ ━━━ Exception (AnnotatedException SomeException) ━━━
       ┃ │ ! AnnotatedException !
       ┃ │ Underlying exception type: IOException
       ┃ │ 
       ┃ │ unsupported operation (Operation is not supported)
       ┃ │ 
       ┃ │ CallStack (from HasCallStack):
       ┃ │   liftIOCallStack, called at src/Testnet/Components/Configuration.hs:186:24 in cardano-testnet-10.0.1-inplace:Testnet.Components.Configuration
       ┃ │   createSPOGenesisAndFiles, called at src/Testnet/Start/Cardano.hs:133:8 in cardano-testnet-10.0.1-inplace:Testnet.Start.Cardano
       ┃ │   createTestnetEnv, called at test/cardano-testnet-test/Cardano/Testnet/Test/DumpConfig.hs:48:23 in cardano-testnet-10.0.1-inplace-cardano-testnet-test:Cardano.Testnet.Test.DumpConfig
       ┃ │   testCase, called at src/Hedgehog/Extras/Test/TestWatchdog.hs:166:66 in hedgehog-extras-0.8.0.0-26dc8622927009e7df9ba3764930f3c64dfd8a2e158af1686ffeb231b1a5219b:Hedgehog.Extras.Test.TestWatchdog
       ┃ │   a given constraint, called at src/Hedgehog/Extras/Test/TestWatchdog.hs:159:26 in hedgehog-extras-0.8.0.0-26dc8622927009e7df9ba3764930f3c64dfd8a2e158af1686ffeb231b1a5219b:Hedgehog.Extras.Test.TestWatchdog
       ┃ │   testCase, called at src/Hedgehog/Extras/Test/TestWatchdog.hs:144:11 in hedgehog-extras-0.8.0.0-26dc8622927009e7df9ba3764930f3c64dfd8a2e158af1686ffeb231b1a5219b:Hedgehog.Extras.Test.TestWatchdog
       ┃ │   runWithWatchdog, called at src/Hedgehog/Extras/Test/TestWatchdog.hs:159:26 in hedgehog-extras-0.8.0.0-26dc8622927009e7df9ba3764930f3c64dfd8a2e158af1686ffeb231b1a5219b:Hedgehog.Extras.Test.TestWatchdog
       ┃ │   runWithDefaultWatchdog, called at src/Hedgehog/Extras/Test/TestWatchdog.hs:166:36 in hedgehog-extras-0.8.0.0-26dc8622927009e7df9ba3764930f3c64dfd8a2e158af1686ffeb231b1a5219b:Hedgehog.Extras.Test.TestWatchdog
       ┃ │   runWithDefaultWatchdog_, called at test/cardano-testnet-test/Cardano/Testnet/Test/DumpConfig.hs:38:82 in cardano-testnet-10.0.1-inplace-cardano-testnet-test:Cardano.Testnet.Test.DumpConfig
```
# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/cardano-testnet/src/Testnet/Components/Configuration.hs
+++ b/cardano-testnet/src/Testnet/Components/Configuration.hs
@@ -165,9 +165,11 @@ createSPOGenesisAndFiles
   (TmpAbsolutePath tempAbsPath) =  do
   AnyShelleyBasedEra sbe <- pure cardanoNodeEra
 
-  let genesisShelleyDirAbs = takeDirectory inputGenesisShelleyFp
+  
+  let genesisShelleyDir = takeDirectory inputGenesisShelleyFp
+  
+  liftIOAnnotated $ System.createDirectoryIfMissing True genesisShelleyDir
 
-  genesisShelleyDir <- liftIOAnnotated $ System.createDirectoryIfMissing True genesisShelleyDirAbs >> pure genesisShelleyDirAbs
   let -- At least there should be a delegator per DRep
       -- otherwise some won't be representing anybody
       numStakeDelegators = max 3 (fromIntegral cardanoNumDReps) :: Int
--- a/cardano-testnet/src/Testnet/Process/RunIO.hs
+++ b/cardano-testnet/src/Testnet/Process/RunIO.hs
@@ -139,9 +139,6 @@ execFlexAny'
   -> m (ExitCode, String, String) -- ^ exit code, stdout, stderr
 execFlexAny' execConfig pkgBin envBin arguments = GHC.withFrozenCallStack $ do
   cp <- procFlex' execConfig pkgBin envBin arguments
-  --H.annotate . ("━━━━ command ━━━━\n" <>) $ case IO.cmdspec cp of
-  --  IO.ShellCommand cmd -> cmd
-  --  IO.RawCommand cmd args -> cmd <> " " <> L.unwords (argQuote <$> args)
   liftIOAnnotated $ IO.readCreateProcessWithExitCode cp ""
 
 
--- a/cardano-testnet/src/Testnet/Property/Assert.hs
+++ b/cardano-testnet/src/Testnet/Property/Assert.hs
@@ -100,7 +100,7 @@ assertExpectedSposInLedgerState output (NumPools numExpectedPools) execConfig = 
       let numPoolsInLedgerState = Set.size poolSet
       unless (numPoolsInLedgerState == numExpectedPools) $
         throwString $ unlines 
-          [ "Expected number of stake pooFvls not found in ledger state"
+          [ "Expected number of stake pools not found in ledger state"
           , "Expected: ", show numExpectedPools
           , "Actual: ", show numPoolsInLedgerState
           ]
--- a/cardano-testnet/src/Testnet/Runtime.hs
+++ b/cardano-testnet/src/Testnet/Runtime.hs
@@ -153,10 +153,11 @@ startNode tp node ipv4 port _testnetMagic nodeCmd = GHC.withFrozenCallStack $ do
 
     -- The port number if it is obtained using 'H.randomPort', it is firstly bound to and then closed. The closing
     -- and release in the operating system is done asynchronously and can be slow. Here we wait until the port
-
-    isClosed <- liftIOAnnotated $ Ping.waitForPortClosed 45 0.1 port
+    
+    let portWaitTimeout = 45
+    isClosed <- liftIOAnnotated $ Ping.waitForPortClosed portWaitTimeout 0.1 port
     unless isClosed $ 
-      throwString $ "Port is still in use after 30 seconds before starting node: " <> show port 
+      throwString $ "Port is still in use after " ++ show portWaitTimeout ++ " seconds before starting node: " <> show port 
 
     (Just stdIn, _, _, hProcess, _)
       <- firstExceptT ProcessRelatedFailure $ initiateProcess
--- a/cardano-testnet/src/Testnet/Start/Cardano.hs
+++ b/cardano-testnet/src/Testnet/Start/Cardano.hs
@@ -142,7 +142,7 @@ createTestnetEnv
 
     let producers = NodeId <$> filter (/= i) nodeIds
         topology = Defaults.defaultP2PTopology producers
-    liftIOAnnotated . LBS.writeFile  (nodeDataDir </> "topology.json") $ A.encodePretty topology
+    liftIOAnnotated . LBS.writeFile (nodeDataDir </> "topology.json") $ A.encodePretty topology
 
 -- | Starts a number of nodes, as configured by the value of the 'cardanoNodes'
 -- field in the 'CardanoTestnetOptions' argument. Regarding this field, you can either:


cardano-testnet/src/Testnet/Orphans.hs
```diff
@@ -0,0 +1,8 @@
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Testnet.Orphans () where
+
+import           RIO (RIO(..), liftIO)
```
carbolymer: Why not import `RIO.Orphans ()` here, and use `Testnet.Orphans` everywhere?

cardano-testnet/src/Testnet/Orphans.hs
```diff
@@ -0,0 +1,8 @@
...
+
+instance MonadFail (RIO env) where 
+  fail = liftIO . fail
```
carbolymer: Would it be possible to try to preserve the failure location? At least some additional information? In `IO` `fail` just raises an exception, so maybe throwing an exception here with `throwString` could be helpful in adding more location information? Maybe with addition of `checkpoint` from annotated-exceptions?

I tried capturing `HasCallStack` in `MonadFail` in the past but without success. Maybe using exceptions here would be more successful here - I haven't tried that I think.

cardano-testnet/src/Testnet/Process/RunIO.hs
```diff
@@ -0,0 +1,284 @@
...
+-- to launch via cabal exec.
+execFlex
+  :: String
+  -> String
+  -> [String]
+  -> RIO env String
+execFlex = execFlex' defaultExecConfig
+
+execFlex'
+  :: MonadIO m
```
carbolymer, line 111: Don't we need `HasCallStack` here?

cardano-testnet/src/Testnet/Process/RunIO.hs
```diff
@@ -0,0 +1,284 @@
...
+-- When running in a nix environment, the 'envBin' argument describes the environment variable
+-- that defines the binary to use to launch the process.
+--
+-- When running outside a nix environment, the `pkgBin` describes the name of the binary
+-- to launch via cabal exec.
+execFlex
+  :: String
+  -> String
+  -> [String]
+  -> RIO env String
```
carbolymer, line 107: Don't we need `HasCallStack` here?

cardano-testnet/src/Testnet/Process/RunIO.hs
```diff
@@ -0,0 +1,284 @@
...
+      <> env'
+    , execConfigCwd = Last $ Just tempBaseAbsPath
+    }
+
+
+execCli'
+  :: MonadIO m
+  => ExecConfig
+  -> [String]
+  -> m String
```
carbolymer, line 77: Don't we need `HasCallStack` here?

cardano-testnet/src/Testnet/Process/RunIO.hs
```diff
@@ -0,0 +1,284 @@
...
+  , execConfigCwd = mempty
+  }
+
+
+mkExecConfig :: ()
+  => MonadIO m
+  => FilePath
+  -> IO.Sprocket
+  -> Int -- ^ Network id
+  -> m ExecConfig
```
carbolymer, line 56: Don't we need `HasCallStack` here?

cardano-testnet/src/Testnet/Orphans.hs
Jimbo4350: Fair point, I'll investigate it.

cardano-testnet/src/Testnet/Process/RunIO.hs
```diff
@@ -0,0 +1,284 @@
...
+  -> [String]
+  -- ^ Arguments to the CLI command
+  -> RIO env CreateProcess
+  -- ^ Captured stdout
+procFlex = procFlex' defaultExecConfig
+
+
+liftIOAnnotated :: (HasCallStack, MonadIO m) => IO a -> m a 
+liftIOAnnotated action = GHC.withFrozenCallStack $
+  liftIO $ action `catch` (\(e :: SomeException) -> throwM $ exceptionWithCallStack e)
```
carbolymer: This is `catch` from `Prelude` right? So this should work with watchdog exceptions I think. Could you add a comment here that we're catching async exceptions here as well intentionally?

cardano-testnet/src/Testnet/Process/RunIO.hs
carbolymer, line 1: I'm not a fan of copying code downstream. Have you given a thought how we could reconcile your changes with hedgehog-extras, so possibly we could avoid the duplication?

My main issue with that is having to fix issues in both places, because we're using similar code in cardano-cli as well.

cardano-testnet/src/Testnet/Runtime.hs
```diff
@@ -149,14 +149,14 @@ startNode tp node ipv4 port _testnetMagic nodeCmd = GHC.withFrozenCallStack $ do
...
                              , "--host-addr", showIpv4Address ipv4
                              ]
-
-    nodeProcess <- newExceptT . fmap (first ExecutableRelatedFailure) . try $ procNode completeNodeCmd
+    nodeProcess <- newExceptT . fmap (first ExecutableRelatedFailure) . try $ runRIO () $ procNode completeNodeCmd
 
     -- The port number if it is obtained using 'H.randomPort', it is firstly bound to and then closed. The closing
     -- and release in the operating system is done asynchronously and can be slow. Here we wait until the port
-    -- is out of CLOSING state.
-    H.note_ $ "Waiting for port " <> show port <> " to be available before starting node"
```
carbolymer, line 158: This was a useful trace. I don't remember if we weren't experiencing hangups long time ago at this place. In general it would be nice to have a tracer available here so we could emit traces about the progress of the node startup.

cardano-testnet/src/Testnet/Runtime.hs
```diff
@@ -290,29 +289,28 @@ startLedgerNewEpochStateLogging testnetRuntime tmpWorkspace = withFrozenCallStac
...
+
+  let act = runExceptT $
+              foldEpochState
+                (configurationFile testnetRuntime)
+                (Api.File socketPath)
+                Api.QuickValidation
+                (EpochNo maxBound)
+                Nothing
+                (handler logFile diffFile)
+  void $ asyncRegister_ act
```
carbolymer: This part of the code has changed. Previously it was only logging if the output file did not exist already. Now it will mangle the output of multiple epoch state loggers if they get enabled at the same time.

This code should be inside `False` branch of the `case ` in line 299.

carbolymer (changes requested):

cardano-testnet/src/Testnet/Runtime.hs
```diff
@@ -362,3 +360,20 @@ instance (L.EraTxOut ledgerera, L.EraGov ledgerera, L.EraCertState ledgerera, L.
...
+
+-- | Runs an action in background, and registers its cancellation to 'MonadResource'.
+asyncRegister_ :: HasCallStack
+               => MonadResource m
+               => IO a -- ^ Action to run in background
+               -> m (ReleaseKey, H.Async a)
+asyncRegister_ act = GHC.withFrozenCallStack $ do 
+      allocate 
+        (do a <- H.async act
+            H.link a
```
carbolymer, line 378: Because of `link` here the comment is not true. The action will fork to the background, but `link` will cause the current thread to wait for the `act` to complete.

cardano-testnet/src/Testnet/Orphans.hs
lehins: Ironically, many years ago I tried to add this precise `MonadFail` instance to `rio` package, but Michael Snoyman stopped me, rightfully so.
Using `fail` in `IO` like monads is not a great idea, precisely because you will not get the source of the error and because it could be hidden by an invalid pattern match, eg:

```haskell
bar :: HasCallStack => RIO env ()
bar = do
  ...
  Just x <- readMaybe <$> readFile fileName
  ...

main = do
  ...
  runRIO env1 bar
  runRIO env2 bar
  ...
```

This will result in `HasCallStack` being useless and it will be impossible to tell which invocation of `bar` is at fault here.

That being said, GHC-9.14 finally has this problem solved. They added `HasCallStack` to `fail` :confetti_ball: 

https://ghc.gitlab.haskell.org/ghc/doc/libraries/base-4.22.0.0-inplace/Control-Monad.html#v:fail

So, if you switch to `throwString` here instead of `liftIO . fail`, then at least it will start behaving nicely with ghc-9.14 :smiley: 

With this addition, it might even make sense now to upstream such instance into `rio` package, but I haven't been involved with it for many years.

cardano-testnet/src/Testnet/Start/Cardano.hs
```diff
@@ -55,41 +59,51 @@ import qualified Data.Time.Clock as DTC
...
 testMinimumConfigurationRequirements options = withFrozenCallStack $ do
   when (cardanoNumPools options < 1) $ do
-    H.note_ "Need at least one SPO node to produce blocks, but got none."
-    H.failure
+    throwString "Need at least one SPO node to produce blocks, but got none."
+
+liftToIntegration :: RIO ResourceMap a -> H.Integration a 
+liftToIntegration  r = do 
+   rMap <- lift $ lift getInternalState 
+   liftIOAnnotated $ runRIO rMap r
```
carbolymer: Shouldn't that be `evalIO` instead? Or at least `evalIO . liftIOAnnotated`?
```suggestion
   H.evalIO $ runRIO rMap r
```
I think regular `liftIO` (which inside `liftIOAnnotated`) was losing some location information for me when reporting in the hedgehog. I don't know if that's still the case here.

cardano-testnet/src/Testnet/Start/Cardano.hs
```diff
@@ -271,62 +293,64 @@ cardanoTestnet
...
           Right (abstractTopology :: P2P.NetworkTopology NodeId) -> do
             topology <- mapM idToRemoteAddressP2P abstractTopology
-            H.lbsWriteFile topologyPath $ encode topology
+            liftIOAnnotated . LBS.writeFile topologyPath $ encode topology
           Left e ->
             -- There can be multiple reasons for why both decodings have failed.
             -- Here we assume, very optimistically, that the user has already
             -- instantiated it with a concrete topology file.
-            H.note_ $ "Could not decode topology file. This may be okay. Reason for decoding failure is:\n" ++ e
+            -- TODO: It is suspicious that this decoding can fail. Investigate further.
```
carbolymer: This will be irrelevant very soon, because non-p2p topology is gone and this double decoding will be removed in: https://github.com/IntersectMBO/cardano-node/pull/6331/files#diff-079ade366e18980b6232319946cc005904a8c865dadf467951da1d30f2069055R268

That's why it would be better to rebase this PR onto 10.6 integration PR. I expect you'll have a lot of conflicts.

The comment needs an update though.

cardano-testnet/src/Testnet/Start/Types.hs
```diff
@@ -285,18 +288,34 @@ data Conf = Conf
   , updateTimestamps :: UpdateTimestamps
   } deriving (Eq, Show)
 
--- | Create a 'Conf' from a temporary absolute path, with Genesis Hashes enabled
--- and updating time stamps disabled.
 -- Logs the argument in the test.
```
carbolymer: part of the haddock got left behind here

cardano-testnet/src/Testnet/Start/Types.hs
```diff
@@ -285,18 +288,34 @@ data Conf = Conf
...
+mkConfigAbs userOutputDir = do 
+  absUserOutputDir <-  makeAbsolute userOutputDir
+  dirExists <- doesDirectoryExist absUserOutputDir
+  let conf = mkConfig absUserOutputDir 
+  if dirExists then
+    -- Happens when the environment has previously been created by the user
+    return conf
+  else do
+    createDirectory absUserOutputDir
+    return conf
```
carbolymer: perhaps
```haskell
unless dirExists $
  createDirectory absUserOutputDir
pure conf
```
would be more readable?

cardano-testnet/src/Testnet/Start/Types.hs
```diff
@@ -41,6 +41,8 @@ module Testnet.Start.Types
   , NodeConfiguration
   , NodeConfigurationYaml
   , mkConf
+  , mkConfigAbs
+  , mkConfig
```
carbolymer: So we have `mkConf`, `mkConfig` and `mkConfigAbs` now. This is a bit confusing. I'm wondering if the naming could be improved here.

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/SanityCheck.hs
```diff
@@ -88,3 +95,35 @@ foldBlocksAccumulator _ _ allEvents _ _ =
   filterPoolReap :: LedgerEvent -> Bool
   filterPoolReap (PoolReap _) = True
   filterPoolReap _ = False
+
+
+hprop_asyncRegister_sanity_check :: Property
+hprop_asyncRegister_sanity_check =
```
carbolymer, line 101: You didn't include this test in `cardano-testnet-test.hs` so it's never executed.

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/SanityCheck.hs
```diff
@@ -88,3 +95,35 @@ foldBlocksAccumulator _ _ allEvents _ _ =
...
+hprop_asyncRegister_sanity_check = 
+    withTests 1 . property $ lift $ do 
+
+     beforeForkedThread <- getCurrentTime
+     (internalState,tId) <- runResourceT $ do 
+        s <- getInternalState
+        (_,asyncA) <- asyncRegister_ (threadDelay 10_000_000)
+        let tId = asyncThreadId asyncA
+        return (s,tId)
+     afterForkedThread <- getCurrentTime
```
carbolymer, line 110: FYI You can use https://hackage-content.haskell.org/package/hedgehog-extras-0.10.1.0/docs/Hedgehog-Extras-Test-TestWatchdog.html to verify if code paths were reached.

cardano-testnet/src/Testnet/Orphans.hs
carbolymer: >That being said, GHC-9.14 finally has this problem solved. They added HasCallStack to fail 🎊

shock

carbolymer:
[This failure is nice](https://github.com/IntersectMBO/cardano-node/pull/6346/checks?check_run_id=54375349183), has the exact line of the code which failed in the stacktrace:

screenshot-20251105_131530


Small thing: I'm guessing because we're not using `evalIO` in `liftToIntegration`, hedgehog is not highlighting the exact line in the test code. Can we try using `evalIO` there and see it improves things?

cardano-testnet/src/Testnet/Start/Cardano.hs
carbolymer: https://github.com/IntersectMBO/cardano-node/pull/6346#issuecomment-3490906731

cardano-testnet/src/Testnet/Process/RunIO.hs
Jimbo4350: I'll open a PR.

cardano-testnet/src/Testnet/Runtime.hs
Jimbo4350: Good catch :+1:

cardano-testnet/src/Testnet/Runtime.hs
Jimbo4350: > link will cause the current thread to wait for the act to complete.

This is not true. Have a look at `hprop_asyncRegister_sanity_check`.  We have a 10 second `threadDelay`  in the forked thread:
```
...
(_,asyncA) <- asyncRegister_ (threadDelay 10_000_000)
...
```

We get the time before and after and check if more than 5 seconds have passed. If the current thread waited for the 10 seconds to pass we would fail here:

```
     when (diff' >= 5) $
       throwString $ "Forked thread took too long: " <> show diff'
```

Jimbo4350: You can trigger the failure if you put a `threadDelay 10_000_000` in the `ResourceT` block i.e:

```
     ┏━━ test/cardano-testnet-test/Cardano/Testnet/Test/SanityCheck.hs ━━━
    100 ┃ hprop_asyncRegister_sanity_check :: Property
    101 ┃ hprop_asyncRegister_sanity_check = 
    102 ┃     withTests 1 . property $ lift $ do 
        ┃     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        ┃     │ ━━━ Exception (StringException) ━━━
        ┃     │ UnliftIO.Exception.throwString called with:
        ┃     │ 
        ┃     │ Forked thread took too long: 10.002805145s
        ┃     │ Called from:
        ┃     │   throwString (test/cardano-testnet-test/Cardano/Testnet/Test/SanityCheck.hs:118:8 in cardano-testnet-10.0.1-inplace-cardano-testnet-test:Cardano.Testnet.Test.SanityCheck)
    103 ┃ 
    104 ┃      beforeForkedThread <- getCurrentTime
    105 ┃      (internalState,tId) <- runResourceT $ do 
    106 ┃         s <- getInternalState
    107 ┃         (_,asyncA) <- asyncRegister_ (threadDelay 10_000_000)
    108 ┃         threadDelay 10_000_000 
    109 ┃         let tId = asyncThreadId asyncA
    110 ┃         return (s,tId)
    111 ┃      afterForkedThread <- getCurrentTime
    112 ┃      let diff' = diffUTCTime afterForkedThread beforeForkedThread
    113 ┃      -- The forked thread (asyncRegister_) may be some long running IO action.
    114 ┃      -- When the ResourceT block has finished this action should be cancelled. 
    115 ┃      -- Therefore we check to see that the thread is indeed cancelled when the 
    116 ┃      -- ResourceT block has finished.
    117 ┃      when (diff' >= 5) $
    118 ┃        throwString $ "Forked thread took too long: " <> show diff'
```

cardano-testnet/src/Testnet/Start/Cardano.hs
Jimbo4350: yes, good catch

cardano-testnet/src/Testnet/Start/Cardano.hs
Jimbo4350: I'm waiting for the 10.6 PR to get merged into master so I will only have to rebase once.

Jimbo4350:
> [This failure is nice](https://github.com/IntersectMBO/cardano-node/pull/6346/checks?check_run_id=54375349183), has the exact line of the code which failed in the stacktrace:
> screenshot-20251105_131530
> 
> Small thing: I'm guessing because we're not using `evalIO` in `liftToIntegration`, hedgehog is not highlighting the exact line in the test code. Can we try using `evalIO` there and see it improves things?

Using `evalIO` in `liftToIntegration` we get: 

```
       ┏━━ src/Testnet/Start/Cardano.hs ━━━
    96 ┃ liftToIntegration :: HasCallStack => RIO ResourceMap a -> H.Integration a 
    97 ┃ liftToIntegration  r = do 
    98 ┃    rMap <- lift $ lift getInternalState 
    99 ┃    H.evalIO $ runRIO rMap r
       ┃    ^^^^^^^^^^^^^^^^^^^^^^^^
       ┃    │ ━━━ Exception (IOException) ━━━
       ┃    │ unsupported operation (Operation is not supported)
    
       ┏━━ test/cardano-testnet-test/Cardano/Testnet/Test/DumpConfig.hs ━━━
    41 ┃ hprop_dump_config :: H.Property
    42 ┃ hprop_dump_config = integrationRetryWorkspace 2 "dump-config-files" $ \tmpDir -> H.runWithDefaultWatchdog_ $ do
       ┃ │ Workspace: /tmp/dump-config-files-0-test-0cf46f1eceefecce
       ┃ │ Workspace: /tmp/dump-config-files-1-test-f8bf352b3eb09851
       ┃ │ Workspace: /tmp/dump-config-files-2-test-54fa7671608a54bb
    43 ┃ 
    44 ┃   let testnetOptions = def { cardanoOutputDir = UserProvidedEnv tmpDir }
    45 ┃       genesisOptions = def { genesisEpochLength = 200 }
    46 ┃       byronGenesisFile = tmpDir </> "byron-genesis.json"
    47 ┃       shelleyGenesisFile = tmpDir </> "shelley-genesis.json"
    48 ┃ 
    49 ┃   -- Generate the sandbox
    50 ┃   conf <- mkConf tmpDir
       ┃   │ /tmp/dump-config-files-0-test-0cf46f1eceefecce
       ┃   │ /tmp/dump-config-files-1-test-f8bf352b3eb09851
       ┃   │ /tmp/dump-config-files-2-test-54fa7671608a54bb
    51 ┃   void $ liftToIntegration $ liftIO $ ioError unsupportedOperation
```

Using `liftIOAnnotated` we get:

```
 ┏━━ test/cardano-testnet-test/Cardano/Testnet/Test/DumpConfig.hs ━━━
    41 ┃ hprop_dump_config :: H.Property
    42 ┃ hprop_dump_config = integrationRetryWorkspace 2 "dump-config-files" $ \tmpDir -> H.runWithDefaultWatchdog_ $ do
       ┃ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
       ┃ │ ━━━ Exception (AnnotatedException SomeException) ━━━
       ┃ │ ! AnnotatedException !
       ┃ │ Underlying exception type: IOException
       ┃ │ 
       ┃ │ unsupported operation (Operation is not supported)
       ┃ │ 
       ┃ │ CallStack (from HasCallStack):
       ┃ │   liftIOAnnotated, called at src/Testnet/Start/Cardano.hs:99:4 in cardano-testnet-10.0.1-inplace:Testnet.Start.Cardano
       ┃ │   liftToIntegration, called at test/cardano-testnet-test/Cardano/Testnet/Test/DumpConfig.hs:51:10 in cardano-testnet-10.0.1-inplace-cardano-testnet-test:Cardano.Testnet.Test.DumpConfig
    ........
```

So we should change to `liftIOAnnotated` as we can tell where the exception was thrown from.

cardano-testnet/src/Testnet/Start/Types.hs
Jimbo4350: I removed `mkConf` from the exports and added a haddock.

cardano-testnet/src/Testnet/Process/RunIO.hs
```diff
@@ -0,0 +1,288 @@
...
+  contents <- liftIOAnnotated $ LBS.readFile planJsonFile
+
+  case eitherDecode contents of
+    Right plan -> case L.filter matching (plan & installPlan) of
+      (component:_) -> case component & binFile of
+        Just bin -> return $ addExeSuffix (T.unpack bin)
+        Nothing -> error $ "missing \"bin-file\" key in plan component: " <> show component <> " in the plan in: " <> planJsonFile
+      [] -> error $ "Cannot find \"component-name\" key with the value \"exe:" <> pkg <> "\" in the plan in: " <> planJsonFile
+    Left message -> error $ "Cannot decode plan in " <> planJsonFile <> ": " <> message
+  where matching :: Component -> Bool
```
carbolymer: This is an old buggy version of `binDist` which sometimes is unable to find an executable in `plan.json`. A fixed one is here: https://github.com/input-output-hk/hedgehog-extras/blob/c1130836a55f662b17d2e86daae04362d65655e1/src/Hedgehog/Extras/Test/Process.hs#L305

carbolymer (changes requested):

cardano-testnet/src/Testnet/Runtime.hs
carbolymer: You're right. Your version is fine.

carbolymer:
@Jimbo4350 

>So we should change to `liftIOAnnotated` as we can tell where the exception was thrown from.

Yes, seems that `evalIO` eats up the call stack, which is sad. However, all the annotations from `note_` are gone with `liftIOAnnotated` which is also a problem.

However we can have a cake and eat it too. See my commit `Fix liftToIntegration to report the exception location to Hedgehog`.

screenshot-20251107_121849

Small thing, that the exception type reported is `StringException` and `UnliftIO.Exception.throwString called with:` is a bit confusing. Additionally, if the caught exception carries a call stack, it will be rendered too leading to printing of 3 call stacks like here:

screenshot-20251107_124207

I guess it is fine, since this points to both location of the `error` there and `liftToIntegration`.

cardano-testnet/src/Testnet/Runtime.hs
```diff
@@ -362,3 +360,20 @@ instance (L.EraTxOut ledgerera, L.EraGov ledgerera, L.EraCertState ledgerera, L.
       , "rewardUpdate" .= nesRu
       , "currentStakeDistribution" .= nesPd
       ]
+
+
+-- | Runs an action in background, and registers its cancellation to 'MonadResource'.
+asyncRegister_ :: HasCallStack
```
carbolymer, line 371: I think this can be upstreamed to hedgehog-extras too. In the end I think there is hardly a benefit of using `evalM` there which requires `MonadTest` constraint: https://github.com/input-output-hk/hedgehog-extras/blob/c1130836a55f662b17d2e86daae04362d65655e1/src/Hedgehog/Extras/Test/Concurrent.hs#L97
I'm not sure what would have happened to make `async` throw right away (resource exhaustion perhaps, but that would signal larger problems anyway).

cardano-testnet/src/Testnet/Start/Cardano.hs
```diff
@@ -55,41 +62,53 @@ import qualified Data.Time.Clock as DTC
...
-    H.note_ "Need at least one SPO node to produce blocks, but got none."
-    H.failure
+    throwString "Need at least one SPO node to produce blocks, but got none."
+
+liftToIntegration :: (HasCallStack, MonadCatch m, MonadResource m, MonadTest m)
+                  => (forall n. (MonadCatch n, MonadResource n, MonadUnliftIO n, MonadFail n) => n a)
+                  -> m a
+liftToIntegration act =  do
+   internalState <- liftResourceT getInternalState
+   catch @_ @SomeException (liftIO $ runInternalState act internalState) (withFrozenCallStack $ failException . toException . stringException . displayException)
```
carbolymer: I have generalised `liftToIntegration` in the latest commit `Generalise liftToIntegration` so it's not tied to `RIO` or `Integration`. This way we could add it to `hedgehog-extras` together.

Small caveat: one has to foresee what constraints for `act` are needed upfront, otherwise GHC will not be happy.

We could use now a better (shorter preferably) name now.

cardano-testnet/src/Parsers/Run.hs
```diff
@@ -70,25 +70,25 @@ createEnvOptions CardanoTestnetCreateEnvOptions
...
-          createTestnetEnv
-            testnetOptions genesisOptions def
-            conf
-          cardanoTestnet testnetOptions conf
-      UserProvidedEnv nodeEnvPath ->
+        conf <- mkConfigAbs "testnet"
+        runRIO () $ createTestnetEnv
+                   testnetOptions genesisOptions def
+                   conf
+        withResourceMap (\rm -> void . runRIO rm $ cardanoTestnet testnetOptions conf)
```
carbolymer, line 87: What's the advantage over:
```suggestion
        void . runResourceT $ cardanoTestnet testnetOptions conf
```
?

cardano-testnet/src/Testnet/Start/Cardano.hs
carbolymer: The function type is a bit overkill, I think it could be just:
```haskell
liftToIntegration :: (HasCallStack, MonadCatch m, MonadResource m, MonadTest m)
                  => ResourceT IO a
                  -> m a
```
it'd be easier to understand I guess

cardano-testnet/src/Parsers/Run.hs
Jimbo4350: ~None. This was from a previous iteration of something I was trying.~
With my version of `liftToIntegration` it is needed to pass the resource map. However I haven't tried with your version of `liftToIntegration`.

cardano-testnet/src/Testnet/Property/Assert.hs
```diff
@@ -94,21 +92,21 @@ assertExpectedSposInLedgerState output (NumPools numExpectedPools) execConfig =
...
-                ]
+  ePoolSet <-  liftIOAnnotated (Aeson.eitherDecodeFileStrict' @(Set PoolId) output)
+  case ePoolSet of
+    Left err -> 
+      throwString $ "Failed to decode stake pools from ledger state: " <> err
+    Right poolSet -> do
+      let numPoolsInLedgerState = Set.size poolSet
+      unless (numPoolsInLedgerState == numExpectedPools) $
+        throwString $ unlines 
+          [ "Expected number of stake pooFvls not found in ledger state"
```
palas:
```suggestion
          [ "Expected number of stake pools not found in ledger state"
```

cardano-testnet/src/Testnet/Runtime.hs
```diff
@@ -149,14 +149,14 @@ startNode tp node ipv4 port _testnetMagic nodeCmd = GHC.withFrozenCallStack $ do
...
 
     -- The port number if it is obtained using 'H.randomPort', it is firstly bound to and then closed. The closing
     -- and release in the operating system is done asynchronously and can be slow. Here we wait until the port
-    -- is out of CLOSING state.
-    H.note_ $ "Waiting for port " <> show port <> " to be available before starting node"
-    H.assertM $ Ping.waitForPortClosed 30 0.1 port
+
+    isClosed <- liftIOAnnotated $ Ping.waitForPortClosed 45 0.1 port
+    unless isClosed $ 
+      throwString $ "Port is still in use after 30 seconds before starting node: " <> show port
```
palas:
```suggestion
    let portWaitTimeout = 45
    isClosed <- liftIOAnnotated $ Ping.waitForPortClosed portWaitTimeout 0.1 port
    unless isClosed $ 
      throwString $ "Port is still in use after " ++ show portWaitTimeout ++ " seconds before starting node: " <> show port 
```
The error message says 30 but the new wait is 45. I suggest parametrising

cardano-testnet/src/Testnet/Process/RunIO.hs
```diff
@@ -0,0 +1,299 @@
...
+  => ExecConfig
+  -> String -- ^ @pkgBin@: name of the binary to launch via 'cabal exec'
+  -> String -- ^ @envBin@: environment variable defining the binary to launch the process, when in Nix
+  -> [String]
+  -> m (ExitCode, String, String) -- ^ exit code, stdout, stderr
+execFlexAny' execConfig pkgBin envBin arguments = GHC.withFrozenCallStack $ do
+  cp <- procFlex' execConfig pkgBin envBin arguments
+  --H.annotate . ("━━━━ command ━━━━\n" <>) $ case IO.cmdspec cp of
+  --  IO.ShellCommand cmd -> cmd
+  --  IO.RawCommand cmd args -> cmd <> " " <> L.unwords (argQuote <$> args)
```
palas:
```suggestion
```
Did you mean to leave this commented out?

cardano-testnet/src/Testnet/Start/Cardano.hs
```diff
@@ -104,27 +121,28 @@ createTestnetEnv
...
   let nodeIds = fst <$> zip [1..] cardanoNodes
   forM_ nodeIds $ \i -> do
     let nodeDataDir = tmpAbsPath </> Defaults.defaultNodeDataDir i
-    H.evalIO $ IO.createDirectoryIfMissing True nodeDataDir
+    liftIOAnnotated $ IO.createDirectoryIfMissing True nodeDataDir
 
     let producers = NodeId <$> filter (/= i) nodeIds
         topology = Defaults.defaultP2PTopology producers
-    H.lbsWriteFile (nodeDataDir </> "topology.json") $ A.encodePretty topology
+    liftIOAnnotated . LBS.writeFile  (nodeDataDir </> "topology.json") $ A.encodePretty topology
```
palas:
```suggestion
    liftIOAnnotated . LBS.writeFile (nodeDataDir </> "topology.json") $ A.encodePretty topology
```
Double space

cardano-testnet/src/Testnet/Start/Cardano.hs
```diff
@@ -349,22 +366,20 @@ cardanoTestnet
 
   let tempBaseAbsPath = makeTmpBaseAbsPath $ TmpAbsolutePath tmpAbsPath
 
-  node1sprocket <- H.headM $ testnetSprockets runtime
+  node1sprocket <- case uncons $ testnetSprockets runtime of
+        Just (sprocket, _) -> pure sprocket
+        Nothing            -> throwString "No testnet sprocket available"
```
palas, line 371:
```suggestion
  node1sprocket <- case testnetSprockets runtime of
        sprocket:_ -> pure sprocket
        []         -> throwString "No testnet sprocket available"
```
Because it is a list we can just use pattern matching

cardano-testnet/src/Testnet/Components/Configuration.hs
```diff
@@ -155,11 +162,12 @@ createSPOGenesisAndFiles
...
   testnetOptions genesisOptions@GenesisOptions{genesisTestnetMagic}
   onChainParams
-  (TmpAbsolutePath tempAbsPath) = GHC.withFrozenCallStack $ do
+  (TmpAbsolutePath tempAbsPath) =  do
   AnyShelleyBasedEra sbe <- pure cardanoNodeEra
 
   let genesisShelleyDirAbs = takeDirectory inputGenesisShelleyFp
-  genesisShelleyDir <- H.createDirectoryIfMissing genesisShelleyDirAbs
+
+  genesisShelleyDir <- liftIOAnnotated $ System.createDirectoryIfMissing True genesisShelleyDirAbs >> pure genesisShelleyDirAbs
```
palas:
```suggestion
let genesisShelleyDir = takeDirectory inputGenesisShelleyFp
liftIOAnnotated $ System.createDirectoryIfMissing True genesisShelleyDir
```
If I understand this, it seems like a very complicated way of saying `let genesisShelleyDir = genesisShelleyDirAbs` but then again we really don't need two variable names for the same thing

palas (approved):
Nice! It looks good and I don't see any issues in the logic. I've written some style suggestions. And I think the changed files could use a formatter, because I've seen several spaces at the end of line, maybe `stylish-haskell` would be enough

cardano-testnet/src/Parsers/Run.hs
carbolymer: Changing of `liftToIntegration` may be worth exploring. I think you are not using the resource map actually anywhere, are you?

PR #6361, opened 2025-11-12
trace-forward | sink write and read optimizations

fmaste:
# Description

## Summary
Refactors `writeToSink` and `readFromSink` to simplify `STM` usage, improve performance by avoiding trace object copying, and remove dead code.

## Key Changes

- `writeToSink`: Removed the usage of two `TBQueue`s of different sizes. Previously, the queue was allowed to grow to max size, which implied expensive copying of trace objects.
- `readFromSink`: Simplified the logic into a single `atomically` block. It now uses a single length query with an implicit `retry`, replacing the previous conditional `check`.
- `STM` Boundaries: Clarified the distinction between `STM` and non-`STM` portions of the code, ensuring the `STM` transaction remains minimal.

## Cleanup

- Removed `connQueueSize` and `disconnQueueSize` in favor of `queueSize`.
- Removed the `wasUsed` flag as it is no longer required.
- Removed the `TVar` holding the `TBQueue` since the new logic removes the need for that specific synchronization protection.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/trace-dispatcher/src/Cardano/Logging/Types.hs
+++ b/trace-dispatcher/src/Cardano/Logging/Types.hs
@@ -487,50 +487,59 @@ instance AE.FromJSON Verbosity where
                                     <> "Unknown Verbosity: " <> show other
 
 data TraceOptionForwarder = TraceOptionForwarder {
-    tofConnQueueSize       :: Word
-  , tofDisconnQueueSize    :: Word
+    tofQueueSize           :: Word
   , tofVerbosity           :: Verbosity
   , tofMaxReconnectDelay   :: Word
 } deriving stock (Eq, Ord, Show, Generic)
 
--- A word regarding queue sizes:
--- In case of a missing forwarding service consumer, traces messages will be
--- buffered. This mitigates short forwarding interruptions, or delays at startup time.
+-- A word regarding queue size:
 --
--- The queue capacity should thus correlate to the expected log lines per second given
--- a particular tracing configuration - to avoid unnecessarily increasing memory footprint.
+-- In case of a missing forwarding service consumer, traces messages will be
+-- buffered. This mitigates short forwarding interruptions, or delays at startup
+-- time.
+--
+-- The queue capacity should thus correlate to the expected log lines per second
+-- given a particular tracing configuration - to avoid unnecessarily increasing
+-- memory footprint.
 --
 -- The default values here are chosen to accomodate verbose tracing output
--- (i.e., buffering 1min worth of trace data given ~32 messages per second). A config
--- that results in less than 5 msgs per second should also provide TraceOptionForwarder
--- queue size values considerably lower. The `disconnQueueSize` is the hard limit in that case.
+-- (i.e., buffering 1min worth of trace data given ~32 messages per second). A
+-- config that results in less than 5 msgs per second should also provide
+-- `TraceOptionForwarder` a queue size value considerably lower.
 --
--- The queue sizes tie in with the max number of trace objects cardano-tracer requests periodically,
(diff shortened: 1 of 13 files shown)


trace-dispatcher/src/Cardano/Logging/Types.hs
```diff
@@ -487,50 +487,59 @@ instance AE.FromJSON Verbosity where
...
+    parseJSON (AE.Object obj) = do
+      -- Field "queueSize" is the new field that replaces and unifies
+      -- both "connQueueSize" and "disconnQueueSize".
+      maybeQueueSize <- obj AE..:? "queueSize"
+      queueSize <- case maybeQueueSize of
+                     -- If the new field was provided we use it.
+                     (Just qs) -> return qs
+                     -- Else we look for the deprectaed fields.
+                     Nothing   -> do
+                       -- We keep the same default values.
```
mgmeier, line 523: While this works fine - wouldn't the code be a bit more expressive with alternative parsers, such as
`obj AE..: "queueSize" <|> [... parse legacy fieds...]`?
(first parser needs to fail for the second to evaluate)

trace-forward/src/Trace/Forward/Utils/TraceObject.hs
```diff
@@ -23,129 +32,96 @@ import qualified Trace.Forward.Protocol.TraceObject.Forwarder as Forwarder
...
+     if fromEnum n >= fromEnum queueLength
+     -- If the requested maximum number is more or equal length, return all.
+     then flushTBQueue queue -- Flush is non-blocking, can return empty.
+     -- If the requested maximum number is less than length, read `n` times.
+     else replicateM (fromIntegral n) (readTBQueue queue)
+  else
+   -- The queue is empty, return nothing or wait.
+   case blocking of
+     TokNonBlocking -> pure []
+     TokBlocking    -> retry
```
mgmeier, line 119: like above: super-optimize the blocking case

trace-forward/src/Trace/Forward/Utils/TraceObject.hs
```diff
@@ -23,129 +32,96 @@ import qualified Trace.Forward.Protocol.TraceObject.Forwarder as Forwarder
...
+                 TokNonBlocking -> NonBlockingReply res
+                 TokBlocking    -> BlockingReply $
+                                     -- Convert to a non-empty list.
+                                     case NE.nonEmpty res of
+                                       (Just ne) -> ne
+                                       Nothing -> if n > 0
+                                                  then error "impossible"
+                                                  -- I can't help returning zero
+                                                  -- items in a non-empty list.
+                                                  else error "Good luck!"
```
mgmeier: 3 small things here:
1. super-optimize the blocking case - it's the most common.
2. `NE.nonEmpty` boxes the value in a lazy `Maybe`. Can you use the same match on the list constructor `(:)` as before - and super-optimize on the non-error case (the old implementation didn't do that)?
3. The error distinction on `n` is somewhat confusing. Can you just use one error msg, that is a little more informative (e.g. "readFromSink: implementation error. STM txn must always block when queue is empty.")

trace-dispatcher/src/Cardano/Logging/Types.hs
fmaste: I found the one-liner alternative (pun intended) more difficult to read / error prone, but this is my personal opinion:  `(obj AE..: "queueSize" <|> (max <$> (obj AE..:? "connQueueSize" AE..!= 1024) <*> (obj AE..:? "disconnQueueSize" AE..!= 2048)))`

PR #6377, opened 2025-11-24
New tracing: new features and some maintenance

mgmeier:
# Description

This PR adds several quality-of-life features to the new tracing system, as well as improve robustness or code quality in several places. Broken down by component, the changes contained are:

cardano-tracer:
* Implement Prometheus HTTP service discovery (SD) under the URL `/targets`
* Add optional config field `"prometheusLabels": { "<labelname>": "<labelvalue>", ... }` for custom labels to be attached with Prometheus SD
* Use `TracerTrace.forMachine` directly instead of going via derived `TracerTrace.toJSON`; remove unused `TracerTrace` JSON instances
* Use proper 'camelCase' for machine-readable `TracerTrace`
* Proper tracing (vs. dumping to stdout) for `showProblemIfAny` and for forwarding connection interruptions
* Remove redundant `runInLoop` in favour of `trace-dispatcher`'s implementation
* Split up journal handler implementation into internal modules `Systemd` and `NoSystemd` (maintenance)

trace-forward:
* Provide `InitForwardingConfig` config record type for `initForwarding` and `initForwardingDelayed`

trace-dispatcher:
* `class LogFormatting`: remove redundant `forHumanFromMachine` and `forHumanOrMachine` (the system already does that inherently)
* Introduce type `Cardano.Logging.Types.TraceMessage.TraceMessage` with explicit codecs for JSON and CBOR
* Rework `PreFormatted` type and formatters to use `TraceMessage`; slightly optimize `humanFormatter'`
* Add CBOR formatting via `FormattedMessage.FormattedCBOR` constructor and a `cborFormatter'` function
* Replaced both `disconnectedQueueSize` and `connectedQueueSize` with `queueSize` in `TraceOptionForwarder` while keeping config parsing backwards compatible
* Add retry delay reset in `runInLoop` when the action runs sufficiently long
* Safely stop `standardTracer`'s stdout thread when there are no more producers (re-enabling the `cabal bench`)

Additionally, the PR:
* removes any use of `forHumanFromMachine` and `forHumanOrMachine` from the `cardano-node` codebase, replacing it with `forHuman` only where appropriate
* implements proper tracing (vs. dumping to stdout) for forwarding connection interruptions in `cardano-node` and `tx-generator`

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [x] Self-reviewed the diff

### DIFF
--- a/cardano-tracer/src/Cardano/Tracer/Handlers/ReForwarder.hs
+++ b/cardano-tracer/src/Cardano/Tracer/Handlers/ReForwarder.hs
@@ -26,6 +26,7 @@ import           Cardano.Tracer.MetaTrace
 import           Ouroboros.Network.Magic (NetworkMagic (..))
 import           Ouroboros.Network.NodeToClient (withIOManager)
 
+import           Control.Exception (SomeException (..))
 import           Control.Monad (when)
 import qualified Data.Text as Text
 
@@ -78,7 +79,8 @@ initReForwarder TracerConfig{networkMagic, hasForwarding}
         { initNetworkMagic          = NetworkMagic networkMagic
         , initEKGStore              = Nothing
         , initForwarderMode         = Log.Responder
-        , initOnForwardInterruption = Nothing         -- TODO:MKarg
+        , initOnForwardInterruption = Just $ \(SomeException e) ->
+            traceWith teTracer (TracerForwardingInterrupted initHowToConnect $ show e)
         , initOnQueueOverflow       = Nothing
         , ..
         }


cardano-tracer/src/Cardano/Tracer/Handlers/ReForwarder.hs
```diff
@@ -86,14 +70,24 @@ initReForwarder TracerConfig{networkMagic, hasForwarding}
             const $ return ()
 
   return (writesToSink', traceDP)
+  where
+    flattenNS = fmap (map (Text.intercalate "."))
 
+    initForwardingWith initHowToConnect =
```
jutaro, line 77: As far as I see, runInLoop now requires an explicit interruption handler; the re-forwarder builds InitForwardingWith with initOnForwardInterruption = Nothing, so connection drops or failures when re-forwarding will now be completely silent. In master the loop logged the broken connection and retry; after this change operators lose visibility into forwarding being down?

jutaro (changes requested):
A lot of changes, and I just have one question wit it. Otherwise it looks good

cardano-tracer/src/Cardano/Tracer/Handlers/ReForwarder.hs
mgmeier: You're completely right - it even says `, initOnForwardInterruption = Nothing         -- TODO:MKarg` in the reforwarder module 🤦 
Oversight on my part, will add the handler. Good catch!

PR #6387, opened 2025-12-03
Removed PeerT tracer

coot:
The information on block-fetch is exposed in other traces in
the `TraceFetchClientState` tracer:

* `SendFetchRequest` - when a block is requested
* `CompletedBlockFetch` - when a block is fetched, includes updated
  `PeerFetchInFlight` information, e.g. number of blocks in flight,
  bytes in flight, etc from that peer.

See also other messages in that tracer.


mgmeier (approved):
LGTM.

The metric `peersFromNodeKernel` will disappear as a consequence of this PR; however, I believe that information can be recovered by considering the `peerSelection_*`  bundle of metrics.

The complexity the `PeerT` tracer adds outweighs the little convenience it offers. I'm fine with removing it.

PR #6397, opened 2025-12-16
More consistent traces between immdb-server and cardano-node

ch1bo:
Cardano-node level changes where most changes are actually in https://github.com/IntersectMBO/ouroboros-consensus/pull/1812


bladyjoker (approved):
LGTM!

PR #6423, opened 2026-01-29
Leios prototype: monomorphic leios eb and separate traces

ch1bo:
Bumps the SRP and rewire metrics using the Leios-specific tracer


ch1bo:
Just opened a PR to keep you posted @bladyjoker and such we can link it up with the Leios project.

PR #6427, opened 2026-02-05
Update Dijkstra-era script purpose rendering to use Conway rendering logic

Savissy:
Motivation
Updated Dijkstra-era script purpose rendering to use Conway rendering logic instead of undefined, to prevent runtime failures during tracing output.
Description
Replace undefined in cardano-node/src/Cardano/Node/Tracing/Render.hs for: example Api.AlonzoEraOnwardsDijkstra -> undefined to Api.AlonzoEraOnwardsDijkstra -> renderConwayPlutusPurpose


Jimbo4350:
Superseded by #6475 which cherry-picks this fix directly onto master.

PR #6453, opened 2026-02-25
User can provide config for tx generator

palas:
# Description

Currently, the `tx-generator` config generation flags create the configuration from scratch, but what if the user has some preferences about how to create that configuration? Instead of having to patch the result every time, it would be better to be able to specify a partial configuration, and have `cardano-testnet` fill in the missing fields and overwrite those fields that are dependent of the generated testnet.

This PR, adds the new flag `--tx-generator-config` to specify a partial base configuration for `tx-generator`.

Additionally, this PR also removes the `tx-generator` related flags from the `create-env` subcommand, since they were not having any effect. In the future, it may be desirable to add the flags back with an appropriate implementation, but it is not obvious how because the socket paths are not currently generated by `create-env`, so the configuration would be incomplete, unless we also ask the user for that.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [x] Self-reviewed the diff

### DIFF
--- a/cardano-testnet/changelog.d/20260225_011141_palas_user_can_provide_config_for_tx_generator.md
+++ b/cardano-testnet/changelog.d/20260225_011141_palas_user_can_provide_config_for_tx_generator.md
@@ -1,0 +1,3 @@
+### Added
+
+- Added new flag `--tx-generator-config` to specify a partial base configuration for `tx-generator`.


palas:
After discussing further, we have decided that adding similar flags to `tx-generator` instead makes most sense and will make maintenance easier. So I am closing this PR in favour of that work

Issue #6455, opened 2026-02-25
cardano-testnet: Default to UpdateTimestamps in the UserProvidedEnv branch


palas:
## Problem

When using the two-step flow (`create-env` → `cardano --node-env`), genesis timestamps become stale between the two invocations. `create-env` writes start times as `now + 15s`, but by the time `cardano --node-env` runs, those times are in the past. The default for `updateTimestamps` is `DontUpdateTimestamps`, so unless the user explicitly passes `--update-time`, the testnet may fail to start.

In the direct `cardano` path (no `--node-env`), this is not a problem because `createTestnetEnv` and `cardanoTestnet` run back-to-back in the same process.

## Proposed fix

In `Parsers/Run.hs`, in the `UserProvidedEnv` branch of `runCardanoOptions`, pass `conf{updateTimestamps=UpdateTimestamps}` as the default. Consider adding a `--no-update-time` flag for opt-out.

This is a 1-line change. Updating already-fresh timestamps is harmless (just shifts them by ~15s).

## Risk: Very low

PR #6466, opened 2026-02-25
cardano-testnet: Default to `--update-time` and add `--no-update-time` flag

palas:
# Description

When using the two-step flow (`cardano-testnet create-env` → `cardano-testnet cardano --node-env`), genesis timestamps become stale between the two invocations. `create-env` writes start times as `now + 15s`, but by the time `cardano --node-env` runs, those times may be in the past. The default was to not update the timestamps, so unless the user explicitly passed `--update-time`, the testnet could fail to start, and this is something you would need to know by heart and remember, even though it is the most likely scenario.

This PR:

- Changes the default `--update-time`, so timestamps are refreshed automatically.
- Adds a `--no-update-time` flag to opt out of timestamp updates.
- Keeps `--update-time` for backward compatibility (it is now the default behavior).

Fixes: https://github.com/IntersectMBO/cardano-node/issues/6455

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [x] Self-reviewed the diff

### DIFF
--- a/cardano-testnet/changelog.d/20260226_005016_palas_default_to_update_timestamps.md
+++ b/cardano-testnet/changelog.d/20260226_005016_palas_default_to_update_timestamps.md
@@ -1,0 +1,9 @@
+### Added
+
+- Added `--preserve-timestamps` flag. When set, genesis file timestamps are kept as-is.
+
+### Changed
+
+- Timestamps in genesis files are now updated to the current date by default.
+- The `--update-time` flag is now internal (hidden, kept for backward compatibility).
+


cardano-testnet/src/Parsers/Cardano.hs
```diff
@@ -132,11 +132,14 @@ pMainnetParams = OA.flag' OnChainParamsMainnet
...
-  <> OA.help "Update the time stamps in genesis files to current date"
-  <> OA.showDefault
-  )
+pUpdateTimestamps =
+  -- Default to UpdateTimestamps, because when using the two-step flow
+  -- (cardano-testnet create-env → cardano-testnet cardano --node-env),
+  -- genesis timestamps can become stale.
+  -- See https://github.com/IntersectMBO/cardano-node/issues/6455
+  OA.flag' UpdateTimestamps (OA.long "update-time" <> OA.help "Update the time stamps in genesis files to current date (default, kept for backward compatibility)")
+  <|> OA.flag' DontUpdateTimestamps (OA.long "no-update-time" <> OA.help "Do not update the time stamps in genesis files to current date.")
```
Jimbo4350: This can be simplified with [value](https://hackage-content.haskell.org/package/optparse-applicative-0.19.0.0/docs/Options-Applicative.html#v:value),

cardano-testnet/src/Parsers/Cardano.hs
```diff
@@ -132,11 +132,14 @@ pMainnetParams = OA.flag' OnChainParamsMainnet
...
-  (  OA.long "update-time"
-  <> OA.help "Update the time stamps in genesis files to current date"
-  <> OA.showDefault
-  )
+pUpdateTimestamps =
+  -- Default to UpdateTimestamps, because when using the two-step flow
+  -- (cardano-testnet create-env → cardano-testnet cardano --node-env),
+  -- genesis timestamps can become stale.
+  -- See https://github.com/IntersectMBO/cardano-node/issues/6455
+  OA.flag' UpdateTimestamps (OA.long "update-time" <> OA.help "Update the time stamps in genesis files to current date (default, kept for backward compatibility)")
```
carbolymer:
```suggestion
  OA.flag' UpdateTimestamps (OA.long "update-time" <> OA.help "Update the time stamps in genesis files to current date (flag is deprecated) (default)")
```
It would be nice to remove this flag at some point, because it may be confusing.

palas: We can start by making it invisible at some point

cardano-testnet/src/Parsers/Cardano.hs
```diff
@@ -132,11 +132,14 @@ pMainnetParams = OA.flag' OnChainParamsMainnet
...
+  <|> OA.flag' DontUpdateTimestamps (OA.long "no-update-time" <> OA.help "Do not update the time stamps in genesis files to current date.")
+  <|> pure UpdateTimestamps
```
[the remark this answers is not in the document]
palas: The proposed solution wouldn't show correctly in the help. Also, I think the problem described falls more in the realm of the optparse-applicative implementation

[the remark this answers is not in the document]
palas: Same as for previous comment

cardano-testnet/src/Parsers/Cardano.hs
```diff
@@ -132,11 +132,14 @@ pMainnetParams = OA.flag' OnChainParamsMainnet
...
```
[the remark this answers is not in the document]
palas: It is all very self-descriptive, extracting functions would actually make it harder to follow

cardano-testnet/src/Parsers/Cardano.hs
```diff
@@ -132,11 +132,14 @@ pMainnetParams = OA.flag' OnChainParamsMainnet
   )
 
 pUpdateTimestamps :: Parser UpdateTimestamps
-pUpdateTimestamps = OA.flag DontUpdateTimestamps UpdateTimestamps
-  (  OA.long "update-time"
-  <> OA.help "Update the time stamps in genesis files to current date"
-  <> OA.showDefault
-  )
+pUpdateTimestamps =
```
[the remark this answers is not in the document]
palas: It is all very self-descriptive, extracting functions would actually make it harder to follow

cardano-testnet/src/Parsers/Cardano.hs
palas: @Jimbo4350, do you mean like this?

```
pUpdateTimestamps :: Parser UpdateTimestamps
pUpdateTimestamps =
  OA.flag UpdateTimestamps DontUpdateTimestamps
    (  OA.long "no-update-time"
    <> OA.help "Do not update the time stamps in genesis files to current date."
    )
```

That would be breaking, it would make `--update-time` not work anymore.
Do we want to do that anyway?

I cannot think of a way of use `value` while keeping the two options `--update-time` and `--no-update-time`, because `value` returns a `Mod` not a `Parser`

cardano-testnet/src/Parsers/Cardano.hs
Jimbo4350: We can do that now.

cardano-testnet/src/Parsers/Cardano.hs
Jimbo4350: ~Ok then use `internal`.~

Why not just make `update-time` optional? If you don't specify it then the time doesn't get updated.

palas: That is how it was before the PR, but you usually want to update the time, because otherwise the testnet won't work. And unless user is knowledgeable, they probably will not write the flag and then struggle to get the testnet working until they find out the hard way.

So defaulting to updating time is probably better user experience. Also, it makes the two workflows more similar: when you generate the configuration, modify it and then run the testnet, the time is probably outdated by then, so you need the extra flag that you wouldn't need if you generate and run in one command.

palas: Addressed by amending the existing commits

Issue #6457, opened 2026-02-25
cardano-testnet: Add `--params-file` and `--params-mainnet` flags to the direct `cardano` path


palas:
## Problem

The `--params-file` and `--params-mainnet` flags (for custom on-chain parameters) are only available via the `create-env` subcommand. The direct `cardano` subcommand hardcodes `DefaultParams` (via `def`) when calling `createTestnetEnv`, so there is no way to use custom on-chain params without the two-step flow.

This is a feature gap that makes the two-step flow strictly more capable than the direct path.

## Proposed fix

1. Add `pOnChainParams` to the `optsTestnet` parser in `Parsers/Cardano.hs`.
2. Thread the resulting `TestnetOnChainParams` through `CardanoTestnetCliOptions` (or a new field).
3. Pass it to `createTestnetEnv` in the `NoUserProvidedEnv` branch of `runCardanoOptions` in `Parsers/Run.hs`, instead of `def`.

~10 lines across `Parsers/Cardano.hs` and `Parsers/Run.hs`. Purely additive; existing behaviour unchanged when flags are absent.

## Depends on
None (independent)

## Risk: Very low

PR #6467, opened 2026-02-26
cardano-testnet: Add `--params-file` and `--params-mainnet` flags to `cardano` subcommand

palas:
# Description

The `--params-file` and `--params-mainnet` flags (for custom on-chain parameters) are only available via the `create-env` subcommand. The direct `cardano` subcommand hardcodes `DefaultParams` (via `def`) when calling `createTestnetEnv`, so there is no way to use custom on-chain params without the two-step flow.

This is a feature gap that made the two-step flow strictly more capable than the direct path.

This PR: Adds the `--params-file` and `--params-mainnet` flags to the `cardano` subcommand.

The flags are optional and if neither is provided the default is the same as before.

Fixes: https://github.com/IntersectMBO/cardano-node/issues/6457

# Checklist

- [x] New tests are added if needed and existing tests are updated.
- [x] CI passes. See note on CI.
  - [x] Code is linted with `hlint`.
  - [x] Code is formatted with `stylish-haskell`.
- [x] Self-reviewed the diff

### DIFF
--- a/cardano-testnet/changelog.d/20260226_013050_palas_add_params_perso_to_cardano_subcommand.md
+++ b/cardano-testnet/changelog.d/20260226_013050_palas_add_params_perso_to_cardano_subcommand.md
@@ -1,0 +1,4 @@
+### Added
+
+- Added `--params-file` and `--params-mainnet` flags to `cardano-testnet cardano` subcommand.
+


Jimbo4350:
I'm not sure we should add these flags to the direct `cardano` subcommand. The whole point of the two-step flow (`create-env` + `cardano --node-env`) is to be the customizable path — the direct `cardano` subcommand is meant to be the simple, opinionated "just spin up a testnet" experience.

If we mirror every `create-env` flag onto the direct path, we erode the reason the two-step flow exists in the first place. If someone needs custom on-chain params, they should use `create-env`.

palas:
> I'm not sure we should add these flags to the direct cardano subcommand. The whole point of the two-step flow (create-env + cardano --node-env) is to be the customizable path — the direct cardano subcommand is meant to be the simple, opinionated "just spin up a testnet" experience.
> 
> If we mirror every create-env flag onto the direct path, we erode the reason the two-step flow exists in the first place. If someone needs custom on-chain params, they should use create-env.

Yes, that is a perfectly valid option. What I was aiming for is for the two-step flow to only have the advantage of allowing you to edit the files in the middle of the process (which is almost the case already), other than that I was looking to make them both equivalent. Having that assumption should allow us to have a clearer logic in the implementation (have pretty much the same code for both, but split in two), and it should also make the story easier to explain

cardano-testnet/test/cardano-testnet-golden/files/golden/help/cardano.cli
```diff
@@ -11,6 +11,7 @@ Usage: cardano-testnet cardano [--num-pool-nodes COUNT]
   [--active-slots-coeff DOUBLE]
   [--node-env FILEPATH]
   [--update-time]
+  [--params-file FILEPATH | --params-mainnet]
```
carbolymer, line 14:
```suggestion
  [--params-blockfrost-file FILEPATH | --params-mainnet]
```
Blockfrost uses its own format for protocol params, we should be specific here. This is because we are also using different (cardano-ledger's) format of protocol parameters, so it may be confusing here a bit.

palas: Sounds good to me, but I will do it in a separate PR because this PR is just spreading an existing parameter for consistency, and your proposal is a breaking change, so I think it should be reviewed independently. https://github.com/IntersectMBO/cardano-node/issues/6488

7zkm7b8gw9-web (commented):
Ok

Issue #6249, opened 2025-06-12
gRPC: Server Dynamic reload mechanism


CarlosLopezDeLara:
Develop and document a dynamic reload mechanism for RPC server configurations. 

This feature must allow configuration updates at runtime without requiring a full cardano-node restart.

PR #6468, opened 2026-02-26
gRPC: Reload configuration on SIGHUP

carbolymer:
# Description

This PR adds a configuration reloading for gRPC on SIGHUP.

Closes: https://github.com/IntersectMBO/cardano-node/issues/6249

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/cardano-node/src/Cardano/Node/Run.hs
+++ b/cardano-node/src/Cardano/Node/Run.hs
@@ -43,8 +43,8 @@ import           Cardano.Node.Configuration.POM (NodeConfiguration (..),
                    defaultPartialNodeConfiguration, makeNodeConfiguration,
                    parseNodeConfigurationFP, getForkPolicy)
 import           Cardano.Node.Configuration.Socket (LocalSocketOrSocketInfo,
-                   SocketOrSocketInfo, SocketOrSocketInfo' (..),
-                   gatherConfiguredSockets, getSocketOrSocketInfoAddr)
+                   SocketOrSocketInfo, SocketOrSocketInfo' (..), gatherConfiguredSockets,
+                   getSocketOrSocketInfoAddr)
 import           Cardano.Node.Configuration.TopologyP2P
 import qualified Cardano.Node.Configuration.TopologyP2P as TopologyP2P
 import           Cardano.Node.Handlers.Shutdown
@@ -143,7 +143,9 @@ import           Control.Monad.Trans.Except.Extra (left, hushM)
 import           Control.Monad.Trans.Maybe (MaybeT(runMaybeT, MaybeT), hoistMaybe)
 import           "contra-tracer" Control.Tracer
 import           Data.Bits
+import           Data.Bifunctor (first)
 import           Data.Either (partitionEithers)
+import           Data.Functor.Identity (Identity (..))
 import           Data.IP (toSockAddr)
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
@@ -481,6 +483,7 @@ handleSimpleNode blockType runP tracers nc networkMagic onKernel = do
                                             (readTVar ledgerPeerSnapshotPathVar)
                                             (readTVar useLedgerVar)
                                             (const . pure $ ())
+    rpcConfigVar <- newTVarIO (ncRpcConfig nc)
 
     let nodeArgs = RunNodeArgs
           { rnGenesisConfig  = ncGenesisConfig nc
@@ -513,8 +516,8 @@ handleSimpleNode blockType runP tracers nc networkMagic onKernel = do
           , rnFeatureFlags = mempty -- TODO(10.7) forward this to CLI options?
           }
 #ifdef UNIX
-    -- initial `SIGHUP` handler, which only rereads the topology file but
-    -- doesn't update block forging.  The latter is only possible once
+    -- initial `SIGHUP` handler, which rereads the topology file and the RPC config from the main configuration file
+    -- but doesn't update block forging. The latter is only possible once
     -- consensus initialised (e.g. reapplied all blocks).
     _ <- Signals.installHandler
           Signals.sigHUP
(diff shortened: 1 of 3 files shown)


cardano-node/src/Cardano/Node/Tracing/Tracers/Startup.hs
```diff
@@ -576,6 +595,9 @@ ppStartupInfoTrace (NetworkConfig localRoots publicRoots useLedgerPeers peerSnap
 ppStartupInfoTrace (LedgerPeerSnapshotLoaded slotNo) =
     "Topology: Peer snapshot containing ledger peers recorded at " <> showT slotNo <> " loaded."
 
+ppStartupInfoTrace (RpcConfigUpdate config) = "Performing RPC configuration update: " <> config
+ppStartupInfoTrace (RpcConfigUpdateError err) = err
```
palas: Probably a good idea to add some context for the error too, like:
```
"Error while updating RPC configuration: " <> err
```

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -513,8 +516,8 @@ handleSimpleNode blockType runP tracers nc networkMagic onKernel = do
           , rnFeatureFlags = mempty -- TODO(10.7) forward this to CLI options?
           }
 #ifdef UNIX
-    -- initial `SIGHUP` handler, which only rereads the topology file but
-    -- doesn't update block forging.  The latter is only possible once
+    -- initial `SIGHUP` handler, which rereads the topology file or RPC config from the main configuration file
+    -- but doesn't update block forging.  The latter is only possible once
```
palas:
```suggestion
    -- but doesn't update block forging. The latter is only possible once
```

Double space, which was already there, but while we are at it...

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -513,8 +516,8 @@ handleSimpleNode blockType runP tracers nc networkMagic onKernel = do
           , rnFeatureFlags = mempty -- TODO(10.7) forward this to CLI options?
           }
 #ifdef UNIX
-    -- initial `SIGHUP` handler, which only rereads the topology file but
-    -- doesn't update block forging.  The latter is only possible once
+    -- initial `SIGHUP` handler, which rereads the topology file or RPC config from the main configuration file
```
palas:
```suggestion
    -- initial `SIGHUP` handler, which rereads the topology file and the RPC config from the main configuration file
```
It does both, so it should be "and"

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -569,14 +573,26 @@ handleSimpleNode blockType runP tracers nc networkMagic onKernel = do
...
               nc
       in
-      withAsync (runRpcServer (rpcTracer tracers) (ncRpcConfig nc, networkMagic))  $ \_ ->
+      let waitForRpcConfigChange oldConfig = atomically $ readTVar rpcConfigVar >>= \new -> check (new /= oldConfig)
+          rpcServerLoop = do
+            config@RpcConfig{isEnabled = Identity enabled} <- readTVarIO rpcConfigVar
+            if enabled
+              then
+                race_
+                  (runRpcServer (rpcTracer tracers) (config, networkMagic))
```
palas: I think it is worth highlighting (if I understood correctly) that if `runRpcServer`:
- Exits successfully but prematurely, it will restart immediately (if this happens at set-up, for example, it will enter a loop).
- Exits with an exception, I am guessing the exception will get propagated and kill the node.

Not sure whether that is desirable or not, maybe it is worth considering having a restart limit, or an exception catching mechanism here.

Also, it may be worth logging if the rpc server restarts for some reason other than config changed, (or if it crashes if you end up catching the exceptions)

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -43,7 +43,7 @@ import           Cardano.Node.Configuration.POM (NodeConfiguration (..),
                    defaultPartialNodeConfiguration, makeNodeConfiguration,
                    parseNodeConfigurationFP, getForkPolicy)
 import           Cardano.Node.Configuration.Socket (LocalSocketOrSocketInfo,
-                   SocketOrSocketInfo, SocketOrSocketInfo' (..),
+                   SocketConfig (..), SocketOrSocketInfo, SocketOrSocketInfo' (..),
```
palas:
```suggestion
                   SocketOrSocketInfo, SocketOrSocketInfo' (..),
```
It seems this `SocketConfig (..)` import is not used, not sure why the compiler doesn't complain

cardano-node/src/Cardano/Node/Run.hs
carbolymer: > Exits with an exception, I am guessing the exception will get propagated and kill the node.
This will not happen because the async of rpc server is never joined.

>it may be worth logging if the rpc server restarts for some reason other than config changed, (or if it crashes if you end up catching the exceptions)

Exception handler in cardano-rpc uses tracers to log the exception.

>Exits successfully but prematurely, it will restart immediately (if this happens at set-up, for example, it will enter a loop).

Good catch. I think we should disable gRPC in such case. To avoid looping.

palas (approved):
Thanks, it looks good now. The only thing is it would now be good now to log that the rpc server was disabled because it terminated, as part of `disableRpcServer` maybe

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -569,14 +573,33 @@ handleSimpleNode blockType runP tracers nc networkMagic onKernel = do
               (readTVar ledgerPeerSnapshotVar)
               nc
       in
-      withAsync (runRpcServer (rpcTracer tracers) (ncRpcConfig nc, networkMagic))  $ \_ ->
+      let waitForRpcConfigChange oldConfig = atomically $ readTVar rpcConfigVar >>= \new -> check (new /= oldConfig)
```
Jimbo4350: `rpcServerLoop` is non-trivial logic (config watching, crash recovery, disable/re-enable) buried in a nested `let` inside `handleSimpleNode`. Consider factoring it out as a standalone top-level function with its own type signature.

carbolymer:
hydra's green, success got not propagated to GH - CI stuck again: https://ci.iog.io/build/12243710/nixlog/2

PR #6473, opened 2026-03-04
bench: Integrate cardano-timeseries-io

Russoul:
# Description

Integrate Cardano Timeseries IO into cardano-tracer

### DIFF
--- a/cardano-tracer/bench/cardano-tracer-bench.hs
+++ b/cardano-tracer/bench/cardano-tracer-bench.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE OverloadedRecordDot #-}
 {-# LANGUAGE OverloadedStrings #-}
 
 import           Cardano.Logging hiding (LocalSocket)
@@ -61,7 +62,7 @@ main = do
   rtViewPageOpened <- newTVarIO False
 #endif
 
-  tracer <- mkTracerTracer $ SeverityF $ Just Warning
+  tracer <- mkTraceBundle $ SeverityF $ Just Warning
 
   let tracerEnv :: TracerConfig -> HandleRegistry -> TracerEnv
       tracerEnv config handleRegistry = TracerEnv
@@ -73,11 +74,12 @@ main = do
         , teCurrentDPLock         = currentDPLock
         , teDPRequestors          = dpRequestors
         , teProtocolsBrake        = protocolsBrake
-        , teTracer                = tracer
+        , teTracer                = tracer.assorted
         , teReforwardTraceObjects = \_-> pure ()
         , teRegistry              = handleRegistry
         , teStateDir              = Nothing
         , teMetricsHelp           = []
+        , teTimeseriesHandle      = Nothing
         }
 
       tracerEnvRTView :: TracerEnvRTView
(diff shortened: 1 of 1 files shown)


cardano-tracer/src/Cardano/Tracer/Handlers/Metrics/TimeseriesServer.hs
```diff
@@ -0,0 +1,154 @@
...
+    if maybe True (>= inputSanCfg.minimumPruningPeriodMillis) v then do
+      modifyConfig handle (\cfg -> Just cfg{pruningPeriodMillis = v})
+      send ok
+    else
+      send malformed
+  ["timeseries", "config", "retention"] | request.requestMethod == methodGet ->
+    expectEmptyQuery request send $ do
+      v <- (.retentionMillis) <$> readConfig handle
+      send $ responseLBS status200 contentHdrUtf8Text (encodeUtf8 (showT v))
+  ["timeseries", "config", "pruning"] | request.requestMethod == methodGet ->
```
mgmeier: Can you consolidate the `"config"` paths; example in pseudo-code:
```
  ["path"]
    | (== POST), Just bind <- optional -> do
    | (== GET) -> do
    | otherwise -> badMethod
```

cardano-tracer/src/Cardano/Tracer/Handlers/Metrics/TimeseriesServer.hs
```diff
@@ -0,0 +1,154 @@
...
+timeseriesApp :: InputSanitationConfig -> TimeseriesHandle -> Application
+timeseriesApp inputSanCfg handle request send = case request.pathInfo of
+  ["timeseries", "query"] | request.requestMethod == methodPost ->
+    expectEmptyQuery request send $ do
+      bs <- consumeRequestBodyStrict request
+      let query = decodeUtf8Lenient (BL.toStrict bs)
+      at <- getTimeMs
+      execute handle (fromIntegral at) query >>= \case
+        Left err -> send $
+          responseLBS status400 contentHdrUtf8Text (encodeUtf8 (asText err))
```
mgmeier: This is a server error (status code range 5xx) - probably use `500`?
If you want to make the distinction between `400` (e.g. malformed `query`, parse failed) and `500` (server-side evaluation error) - you'd need to discriminate `err` for the right code.

cardano-tracer/src/Cardano/Tracer/Handlers/Metrics/TimeseriesServer.hs
```diff
@@ -0,0 +1,154 @@
...
+notFound = responseLBS status404 [] ""
+
+expectEmptyQuery :: Request
+                -> (Response -> IO ResponseReceived)
+                -> IO ResponseReceived
+                -> IO ResponseReceived
+expectEmptyQuery request send kont =
+  case queryToQueryText request.queryString of
+    [] -> kont
+    _ -> send malformed
```
mgmeier: As per https://datatracker.ietf.org/doc/html/rfc3986, an unexpected or gratuitous query string does not invalidate an URI. Can you remove that constraint?

cardano-tracer/src/Cardano/Tracer/Handlers/Metrics/TimeseriesServer.hs
```diff
@@ -0,0 +1,154 @@
...
+    [(readTimeunit -> Just toMillis, Just (decimal -> Right (v, "")))] -> do
+      Just (Just (toMillis v))
+    [] -> Just Nothing
+    _ -> Nothing
+
+oneDurationItemQuery :: Request
+                     -> Maybe Word64
+oneDurationItemQuery request =
+  case queryToQueryText request.queryString of
+    [(readTimeunit -> Just toMillis, Just (decimal -> Right (v, "")))] -> Just (toMillis v)
```
mgmeier: Could you merge with the same expression in the prev. function into a small helper function?

In connection to the RFC linked above, please assume "open" query strings where you lookup the keys as per `readTimeUnit`, but ignore all others (i.e. not case-match on a list singleton).

cardano-tracer/cardano-tracer.cabal
```diff
@@ -123,6 +123,7 @@ library
                         Cardano.Tracer.Handlers.Metrics.Monitoring
```
mgmeier, line 123: Please bump version to `cardano-tracer-0.4.0` and complete changelog. Make timeseries the prominent new feature :) 
(The `NEXT` paragraph items are currently unreleased, and will also be part of `0.4.0`)

cardano-tracer/src/Cardano/Tracer/MetaTrace.hs
```diff
@@ -273,34 +293,56 @@ instance MetaTrace TracerTrace where
...
+          in configTracerTracer finalTracer $> finalTracer
+  where
+    configTracerTracer :: Trace IO TracerTrace -> IO ()
+    configTracerTracer tr = do
+      configReflection <- emptyConfigReflection
+      configureTracers configReflection initialTraceConfig [tr]
+     where
+       initialTraceConfig :: TraceConfig
+       initialTraceConfig =
+         TraceConfig
```
mgmeier: Please rebase on `master` and use the the change from there: `emptyTraceConfig` instead of `TraceConfig{..}`, omitting all `Nothing` fields.

cardano-tracer/src/Cardano/Tracer/MetaTrace.hs
```diff
@@ -273,34 +293,56 @@ instance MetaTrace TracerTrace where
...
  where
-   initialTraceConfig :: TraceConfig
-   initialTraceConfig =
+   cfg :: TraceConfig
+   cfg =
      TraceConfig
-     { tcForwarder         = Nothing
-     , tcNodeName          = Nothing
-     , tcResourceFrequency = Nothing
+     { tcForwarder              = Nothing
```
mgmeier: Same as previously: Please use `emptyTraceConfig` - reason is forward compatibility to the next release of `trace-dispatcher`.

PR #6504, opened 2026-03-30
Fix flaky MultiAssetReturnCollateral test

Jimbo4350:
## Summary
- Replace fixed 1-block wait + single query with `retryUntilJustM` polling (up to 10 blocks) for the multi-asset UTxO in `hprop_collateral_with_tokens`
- The mint transaction may not be confirmed within a single block, causing `findLargestMultiAssetUtxoWithAddress` to return `Nothing` and the test to fail spuriously
- This matches the retry pattern already used for the script address UTxO lookup on the next line

## Test plan
- [ ] CI passes the `Collateral With Multiassets` test on aarch64-linux (previously failing: check run [69208753482](https://github.com/IntersectMBO/cardano-node/pull/6502/checks?check_run_id=69208753482))

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Cli/Plutus/MultiAssetReturnCollateral.hs
+++ b/cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Cli/Plutus/MultiAssetReturnCollateral.hs
@@ -22,7 +22,7 @@ import           Testnet.Components.Configuration
 import           Testnet.Components.Query
 import           Testnet.Defaults
 import           Testnet.Process.Run (execCli', mkExecConfig)
-import           Testnet.Property.Util (integrationWorkspace)
+import           Testnet.Property.Util (integrationRetryWorkspace)
 import           Testnet.Types
 
 import           Hedgehog (Property)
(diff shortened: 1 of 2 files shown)


cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Cli/Plutus/MultiAssetReturnCollateral.hs
```diff
@@ -145,13 +145,9 @@ hprop_collateral_with_tokens = integrationWorkspace "collateral-with-tokens" $ \
   -- STEP 2: Attempt to spend from script with collateral containing tokens
   -- This will fail because collateral cannot contain non-ADA tokens
   
-  -- Wait for transactions to be processed and find UTxOs
-  _ <- waitForBlocks epochStateView 1
-  
-  -- Find the UTxO with tokens at wallet1 (for collateral)
-  txinCollateralWithTokensM <- 
+  -- Find the UTxO with tokens at wallet2 (for collateral)
+  (txinCollateralWithTokens, collateralTxOut) <- retryUntilJustM epochStateView (WaitForBlocks 10) $
```
carbolymer, line 149: >The mint transaction may not be confirmed within a single block, causing findLargestMultiAssetUtxoWithAddress to return Nothing and the test to fail spuriously

When this happens? Why it may not get confirmed within a single block?

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Cli/Plutus/MultiAssetReturnCollateral.hs
carbolymer, line 1: This test should use `integrationRetryWorkspace` instead of `integrationWorkspace`

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Cli/Plutus/MultiAssetReturnCollateral.hs
Jimbo4350: Hard to say exactly, but maybe a block gets forged before the tx has time to enter the mempool. We then wait for that block and query the node state, but it still reflects the pre-tx state. The tx only gets included in the next block, which updates the node state but by then the test has already failed.

carbolymer (approved):

PR #6511, opened 2026-04-07
add txsSyncDurationTotal counter

karknu:
# Description

Add a txsSyncDurationTotal counter for tracking the total time spent syncing the mempool.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [x] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
+++ b/cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
@@ -20,6 +20,7 @@ module Cardano.Node.Tracing.Tracers.Consensus
   , servedBlockLatest
   , ClientMetrics
   , txsMempoolTimeoutSoftCounterName
+  , txsSyncDurationTotalCounterName
   , impliesMempoolTimeoutSoft
   ) where
 
@@ -1087,6 +1088,9 @@ instance MetaTrace (TraceLocalTxSubmissionServerEvent blk) where
 txsMempoolTimeoutSoftCounterName :: Text.Text
 txsMempoolTimeoutSoftCounterName = "txsMempoolTimeoutSoft"
 
+txsSyncDurationTotalCounterName :: Text.Text
+txsSyncDurationTotalCounterName = "txsSyncDurationTotal"
+
 impliesMempoolTimeoutSoft :: TraceEventMempool blk -> Bool
 impliesMempoolTimeoutSoft = \case
   TraceMempoolRejectedTx _tx _txApplyErr details _mpSz ->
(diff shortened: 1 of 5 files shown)


cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
```diff
@@ -1184,7 +1188,9 @@ instance
     , IntM "mempoolBytes" (fromIntegral . unByteSize32 . msNumBytes $ mpSz)
     ]
   asMetrics (TraceMempoolSynced (FallingEdgeWith duration)) =
-    [ IntM "txsSyncDuration" (round $ 1000 * duration)
+    let durationMs = round (1000 * duration) :: Integer in
+    [ IntM "txsSyncDuration" durationMs
+    , CounterM txsSyncDurationTotalCounterName (Just (fromIntegral durationMs))
```
nfrisby, line 1193: For reference, I looked this up (for the nth time) just to check: `Nothing` is increment and `Just n` is `+n`

https://github.com/IntersectMBO/hermod-tracing/blob/0af5e41f81a6463a53b9ea109d85436b014de40e/trace-dispatcher/src/Cardano/Logging/Tracer/EKG.hs#L83-L84

nfrisby: EKG counters are integral. I think it's fine to round to milliseconds before contributing to the running total.

cardano-node/src/Cardano/Tracing/Metrics.hs
```diff
@@ -74,6 +78,14 @@ mapForgingStatsTxsProcessed fs f =
   atomicModifyIORef' (fsTxsProcessedNum fs) $
     \txCount -> join (,) $ f txCount
 
+mapForgingStatsTxsSyncDuration ::
```
nfrisby, line 81: Matches `mapForgingStatsTxsProcessed` just above ✅

nfrisby (commented):
LGTM

My only experience with the counters was from implementing this commit January https://github.com/IntersectMBO/cardano-node/commit/9afc1e7dd8b6b307e48b28bec1c22d10fac61dd3

So I primarily just compared to that. The only difference I'm seeing is that this counter would not be maintained when the Mempool tracer is turned off. I think that's probably fine. (But happened to not be something we wanted for the January commit.)

I hesitate to Approve, only because I don't own this repository. But if someone who owns it feels comfortable to Approve based on this Comment, feel free.

karknu:
> So I primarily just compared to that. The only difference I'm seeing is that this counter would not be maintained when the Mempool tracer is turned off. I think that's probably fine.

I implemented it like the `txsProcessedNum` counter. It and the other mempool counters/gauges have the same behavior where they depend on the tracer being enabled.

jutaro (approved):
No findings.

The new txsSyncDurationTotal metric looks consistent with the existing tracing/metrics pattern, and the naming/help-text updates match the behavior. Residual risk: I did not see tests for the new counter, so runtime/exporter behavior should still be checked manually.

PR #6516, opened 2026-04-08
Leios: Build against backward-compatible block decoder

ch1bo:
Bumps dependencies to use https://github.com/IntersectMBO/cardano-ledger/pull/5721


ch1bo:
Let's aim for merging #6537 and https://github.com/IntersectMBO/ouroboros-consensus/pull/1978 instead (and make sure that change is backwards compatible so we can keep using old recorded chains)

PR #6530, opened 2026-04-14
Fix flaky tests by using retryUntilJustM which should provide more reliable wait

carbolymer:
- Replaced `waitForBlocks 1` followed by immediate assertions with `retryUntilJustM` polling in seven test sites.
  The old pattern was flaky on aarch64-linux because `waitForBlocks` only guarantees the block number advanced, not that a submitted transaction was included in that block.
  Affected tests: Simple Script Mint, Simple Script CostCalculation, TxReferenceInputDatum (two sites), RegisterDeregisterStakeAddress (two sites), and MultiAssetReturnCollateral.

- Switched all remaining `integrationWorkspace` tests to `integrationRetryWorkspace 2` so that transient failures are retried automatically.
  Affected tests: FoldEpochState, Simple Script Mint, DRep Deposits, DRep Activity, Predefined Abstain DRep, PlutusV3 purposes, PlutusV2 two script certs, Collateral With Multiassets, Committee Add New, No Confidence, and Transaction Build Estimate.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/cardano-testnet/src/Testnet/Components/Query.hs
+++ b/cardano-testnet/src/Testnet/Components/Query.hs
@@ -27,6 +27,7 @@ module Testnet.Components.Query
   , waitUntilEpoch
   , waitForBlocks
   , retryUntilJustM
+  , retryUntilM
 
   , findAllUtxos
   , findUtxosWithAddress
@@ -182,6 +183,49 @@ retryUntilJustM esv timeout act = withFrozenCallStack $ do
         Nothing -> do
           H.threadDelay 300_000
           go startingValue
+
+    getCurrentValue = withFrozenCallStack $
+      case timeout of
+        WaitForEpochs _ -> unEpochNo <$> getCurrentEpochNo esv
+        WaitForSlots _ -> unSlotNo <$> getSlotNumber esv
+        WaitForBlocks _ -> unBlockNo <$> getBlockNumber esv
+
+    timeoutW64 =
+      case timeout of
+        WaitForEpochs (EpochInterval n) -> fromIntegral n
+        WaitForSlots n -> n
+        WaitForBlocks n -> n
+
+-- | Like 'retryUntilJustM' but takes a plain action and a predicate instead of
+-- an action returning 'Maybe'. On timeout, annotates the last value that failed
+-- the predicate. Intermediate attempts produce no annotations.
+retryUntilM
+  :: HasCallStack
+  => MonadIO m
+  => MonadTest m
+  => MonadAssertion m
+  => Show a
+  => EpochStateView
+  -> TestnetWaitPeriod -- ^ timeout
+  -> m a              -- ^ action to retry
+  -> (a -> Bool)      -- ^ predicate that must hold
+  -> m a
+retryUntilM esv timeout act predicate = withFrozenCallStack $ do
+  startingValue <- getCurrentValue
+  go startingValue
+  where
+    go startingValue = withFrozenCallStack $ do
+      result <- act
+      if predicate result
+        then pure result
+        else do
+          cv <- getCurrentValue
+          if timeoutW64 + startingValue < cv
+            then do
+              H.noteShow_ result
+              H.note_ $ "Predicate not satisfied after: " <> show timeout
+              H.failure
+            else H.threadDelay 300_000 >> go startingValue
 
     getCurrentValue = withFrozenCallStack $
       case timeout of
--- a/cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Cli/Plutus/MultiAssetReturnCollateral.hs
+++ b/cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Cli/Plutus/MultiAssetReturnCollateral.hs
@@ -4,16 +4,18 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
 
-module Cardano.Testnet.Test.Cli.Plutus.MultiAssetReturnCollateral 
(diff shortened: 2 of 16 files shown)


cardano-testnet/src/Testnet/Components/Query.hs
```diff
@@ -195,6 +196,49 @@ retryUntilJustM esv timeout act = withFrozenCallStack $ do
...
+      if predicate result
+        then pure result
+        else do
+          cv <- getCurrentValue
+          if timeoutW64 + startingValue < cv
+            then do
+              H.noteShow_ result
+              H.note_ $ "Predicate not satisfied after: " <> show timeout
+              H.failure
+            else H.threadDelay 300_000 >> go startingValue
```
palas, line 228:
```suggestion
retryUntilM esv timeout act predicate =
  retryUntilJustM esv timeout $ do
    result <- act
    pure $ if predicate result then Just result else Nothing
```
Would this be equivalent? If so it would be simpler to maintain.
....

Ah, we wouldn't get the `H.noteShow_ result`. That suggests this version is more general.

Maybe we can have an auxiliar version of `retryUntilM` without the notes, that returns an `Either`, for example, and then implement the actual `retryUntilM` and the `retryUntilJustM` in terms of it, by adding the note at the end in case of `Left`, and do the `failure`.

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Cli/Plutus/MultiAssetReturnCollateral.hs
```diff
@@ -144,14 +146,13 @@ hprop_collateral_with_tokens = integrationWorkspace "collateral-with-tokens" $ \
...
-  _ <- waitForBlocks epochStateView 1
-  
-  -- Find the UTxO with tokens at wallet1 (for collateral)
-  txinCollateralWithTokensM <- 
-    findLargestMultiAssetUtxoWithAddress epochStateView sbe $ T.pack maCollateralAddress
-  (txinCollateralWithTokens, collateralTxOut) <- H.evalMaybe txinCollateralWithTokensM
+  (txinCollateralWithTokens, collateralTxOut) <-
+    retryUntilJustM epochStateView (WaitForBlocks 5) $
+      -- Find the UTxO with tokens at wallet1 (for collateral)
+      findLargestMultiAssetUtxoWithAddress epochStateView sbe $ T.pack maCollateralAddress
```
palas: **[Important]** I think this is not actually waiting for the previous transaction, may be counterproductive. Because it is looking at an address that is created by the testnet, so it must exist before the transaction. It is defined like this:
```
let maCollateralAddress = T.unpack $ paymentKeyInfoAddr wallet2
```

palas (approved):
Nice improvement!

I wrote a refactoring suggestion (which could be done in a different PR).

And one potential issue that I think should be either addressed before merging or dismissed if I misunderstood, because if I am right it could actually be increasing flakiness in that particular test

cardano-testnet/src/Testnet/Components/Query.hs
carbolymer: Yep, that's the key difference. Adding show inside `retryUntilJustM` makes it dump an annotation every 300ms until failure which floods the logs. I've added this to prevent that. I like your idea, I'll do it in a follow up PR. We also have a bunch of retrying functions all doing the same, but slightly different. It would be good to unify them.

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Cli/Plutus/MultiAssetReturnCollateral.hs
carbolymer: But wallet2 is an output of a previous transaction:
```haskell
    , "--tx-out", collateralToBeValue
```
```haskell
      collateralToBeValue = mconcat [maCollateralAddress, "+", show @Int 5_000_000, "+", mintValue]

```

palas: But it can have several outputs (some from before), and the `findLargestMultiasset...` function finds any of them

carbolymer: fair, I've added filter by txId

cardano-testnet/src/Testnet/Components/Query.hs
Jimbo4350: I just spotted another problem. The `threadDelay` in `retryUntilJustM` and `retryUntilM` is hardcoded at 300ms regardless of which `TestnetWaitPeriod` constructor is used. This is problematic because slot time and epoch length are configurable i.e there's no guarantee that 300ms is a sensible polling interval for any of the three. For `WaitForEpochs` with typical configurations it means polling the node thousands of times per epoch, and even for `WaitForSlots` it could be too fast or too slow depending on the configured slot time.
                                                                                                                                                                                        
The delay should be abstracted into a function that computes a sensible polling interval from the actual slot time and epoch length in the genesis config, and both `retryUntilJustM` and `retryUntilM` should use it.

Jimbo4350 (approved):
LGT but the 300 ms issue needs to be addressed asap.

carbolymer:
Hydra green - status not updating for over a day https://ci.iog.io/jobset/IntersectMBO-cardano-node/pullrequest-6530

PR #6531, opened 2026-04-14
ReCon framework 1.1.0

Russoul:
* Support configurable timeunits in formulas via a CLI option.
* Support existential quantification over properties (both finite and infinite domain)
* Support Presburger arithmetic (over ℤ)
* Integrate Presburger arithmetic into the LTL
* An integration test that checks a few formulas against synthetic logs


mgmeier (approved):
LGTM - great work thank you @Russoul

PR #6535, opened 2026-04-15
cardano-testnet | Remove redundant `watchEpochStateUpdate`

carbolymer:
# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/cardano-testnet/src/Testnet/Components/Query.hs
+++ b/cardano-testnet/src/Testnet/Components/Query.hs
@@ -14,7 +14,7 @@ module Testnet.Components.Query
   , getEpochState
   , getSlotNumber
   , getBlockNumber
-  , watchEpochStateUpdate
+  , getEpochStateDetails
 
   , getMinDRepDeposit
   , getMinGovActionDeposit
@@ -60,10 +60,8 @@ import qualified Cardano.Ledger.State as L
 
 import           Prelude
 
-import           Control.Exception.Safe (MonadCatch)
 import           Control.Monad
 import           Control.Monad.Trans.Resource
-import           Control.Monad.Trans.State.Strict (put)
 import           Data.IORef
 import           Data.List (sortOn)
 import qualified Data.Map as Map
@@ -123,7 +121,7 @@ waitForEpochs
   -> EpochInterval  -- ^ Number of epochs to wait
   -> m EpochNo -- ^ The epoch number reached
 waitForEpochs epochStateView interval = withFrozenCallStack $ do
-  void $ watchEpochStateUpdate epochStateView interval $ \_ -> pure Nothing
+  void . retryUntilRightM epochStateView (WaitForEpochs interval) . pure $ Left ()
   getCurrentEpochNo epochStateView
 
 -- | Wait for the requested number of blocks
@@ -132,7 +130,6 @@ waitForBlocks
   => MonadIO m
   => MonadTest m
   => MonadAssertion m
-  => MonadCatch m
   => EpochStateView
   -> Word64 -- ^ Number of blocks to wait
   -> m BlockNo -- ^ The block number reached
@@ -140,12 +137,8 @@ waitForBlocks epochStateView numberOfBlocks = withFrozenCallStack $ do
   BlockNo startingBlockNumber <- getBlockNumber epochStateView
   H.note_ $ "Current block number: " <> show startingBlockNumber <> ". "
     <> "Waiting for " <> show numberOfBlocks <> " blocks"
-  H.noteShowM . H.nothingFailM . fmap (fmap BlockNo) $
-    watchEpochStateUpdate epochStateView (EpochInterval maxBound) $ \(_, _, BlockNo blockNumber) ->
-      pure $
-        if blockNumber >= startingBlockNumber + numberOfBlocks
-        then Just blockNumber
-        else Nothing
+  void . retryUntilRightM epochStateView (WaitForBlocks numberOfBlocks) . pure $ Left ()
+  getBlockNumber epochStateView
 
 data TestnetWaitPeriod
   = WaitForEpochs EpochInterval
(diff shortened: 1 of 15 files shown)


cardano-testnet/src/Testnet/Components/Query.hs
```diff
@@ -159,6 +153,34 @@ instance Show TestnetWaitPeriod where
     WaitForBlocks n -> "WaitForBlocks " <> show n
     WaitForSlots n -> "WaitForSlots " <> show n
 
+-- | Core retry loop. Repeats the action every 300ms until it returns 'Right'
+-- or the timeout is reached, in which case the last 'Left' is returned.
+retryByTimeout
```
palas: I think the name is misleading. To me, `retryByTimeout` suggests that it will retry once when the timeout hits. Maybe `retryUntilTimeout` instead?

`retryUntil` is better but there is already `retryUntilM` and they are both monadic... If what we want is making them easier to differentiate, I would focus on naming the difference, maybe `simpleRetryUntilM` or `eitherRetryUntilM` or something like that.

cardano-testnet/src/Testnet/Components/Query.hs
```diff
@@ -159,6 +153,34 @@ instance Show TestnetWaitPeriod where
...
+  => MonadIO m
+  => MonadTest m
+  => MonadAssertion m
+  => EpochStateView
+  -> TestnetWaitPeriod
+  -> m (Either e a)
+  -> m (Either e a)
+retryByTimeout esv timeout act = withFrozenCallStack $ do
+  startingValue <- getCurrentValue
+  go $ startingValue + timeoutW64
```
palas, line 168: Nice, I was thinking about this recalculation before but I didn't want to mention it

cardano-testnet/src/Testnet/Components/Query.hs
```diff
@@ -132,20 +130,16 @@ waitForBlocks
...
     <> "Waiting for " <> show numberOfBlocks <> " blocks"
-  H.noteShowM . H.nothingFailM . fmap (fmap BlockNo) $
-    watchEpochStateUpdate epochStateView (EpochInterval maxBound) $ \(_, _, BlockNo blockNumber) ->
-      pure $
-        if blockNumber >= startingBlockNumber + numberOfBlocks
-        then Just blockNumber
-        else Nothing
+  retryUntilM epochStateView (WaitForBlocks numberOfBlocks)
+    (getBlockNumber epochStateView)
+    (\(BlockNo bn) -> bn >= startingBlockNumber + numberOfBlocks)
```
palas: I think it is not very nice to have both the timeout and the action counting blocks, they may get out of sync too. I think it would be better to use `retryByTimeout` and rely on the timeout only, making the action always `Left =<< getBlockNumber epochStateView`, or `Left ()` and then getting the block number afterwards.

Same as in `waitForEpochs`

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Gov/GovActionTimeout.hs
```diff
@@ -136,8 +136,7 @@ hprop_check_gov_action_timeout = integrationRetryWorkspace 2 "gov-action-timeout
   void $ waitForEpochs epochStateView (EpochInterval $ unEpochInterval govActionLifetime + 1)
 
   -- Check proposal expired
-  mGovernanceActionTxIx <- watchEpochStateUpdate epochStateView (EpochInterval 2) $ \(anyNewEpochState, _, _) ->
-      return $ maybeExtractGovernanceActionIndex governanceActionTxId anyNewEpochState
-
-  mGovernanceActionTxIx H.=== Nothing
+  void $ retryUntilM epochStateView (WaitForEpochs $ EpochInterval 2)
+    (maybeExtractGovernanceActionIndex governanceActionTxId <$> getEpochState epochStateView)
+    isNothing
```
palas: Apparently, this is weakening the test, because it was originally testing that:
- The action is not available at the beginning of the call to `watchEpochStateUpdate`
- The action remained unavailable during two epochs

But I am not sure the later was intentional.

On the other hand, line 136 was clearly waiting for the proposal to expire. And the new usage of `retryUntilM` is allowing for that to happen 2 epochs later potentially, so that is not good.

We could either merge this with the line 136 and reduce the timeout to adjust it to the original expectation. Or if we think that verifying that "the action does not reappear for two extra epochs" is important, we would need to invert this retrying and make the timeout the success

palas (approved):
Very nice simplification 👍

There is a lot of small nuance in some changes, and probably most of it doesn't really make a difference in practice, so I think we can ignore it.

I pointed out a couple of small things that may be worth changing or double-checking

cardano-testnet/src/Testnet/Components/Query.hs
carbolymer: I think I'll go with retryUntilRightM

cardano-testnet/src/Testnet/Components/Query.hs
carbolymer: Good point.

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Gov/GovActionTimeout.hs
carbolymer: Good catch. 

>it was originally testing that (...) The action remained unavailable during two epochs

Looks like a double-check to me to deal with rollbacks. I don't expect rollbacks in one SPO testnets, but we can add a one epoch wait for a good measure, in case rollback happens on epoch boundary.

cardano-testnet/src/Testnet/EpochStateProcessing.hs
```diff
@@ -49,6 +50,32 @@ maybeExtractGovernanceActionIndex txid (AnyNewEpochState sbe newEpochState _) =
       | ti1 == L.extractHash ti2 = Just gai
     compareWithTxId _ x _ _ = x
 
+-- | Look up the @gasExpiresAfter@ epoch for the governance action submitted
```
palas, line 53: I was super confused until I found out that `gas` meant `GovActionState`. Maybe it would be good to say that, for example say "look up the expiration epoch in the governance action state, through the `gasExpiresAfter` lens", or something like that

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Gov/GovActionTimeout.hs
```diff
@@ -132,12 +126,31 @@ hprop_check_gov_action_timeout = integrationRetryWorkspace 2 "gov-action-timeout
...
+  -- rollback within the @k@-window can cross the removal boundary. Waiting
+  -- a full extra epoch past the boundary makes the removal @k@-deep stable
+  -- and eliminates the rollback race.
+  EpochNo currentE <- getCurrentEpochNo epochStateView
+  let timeoutEpochs
+        | expiryE >= currentE = fromIntegral (expiryE - currentE) + 3
+        | otherwise = 3
+  void $ retryUntilM epochStateView (WaitForEpochs $ EpochInterval timeoutEpochs)
+    (getCurrentEpochNo epochStateView)
+    (> EpochNo (expiryE + 1))
```
palas, line 150: Is it possible to use `waitForEpochs` here?

Also, I understand that: `timeoutEpochs` is calculated as `expiryE + 3` (essentially). But the predicate is computed as `> expiryE + 1`, so it is not consistent, right? Won't that `> expiryE + 1` trigger first always?

palas (approved):
Just one comment for clarification and one thing that I think is weird in the test gov action test. Otherwise, looking good 👍

cardano-testnet/src/Testnet/EpochStateProcessing.hs
carbolymer: I've renamed the function parameter

Jimbo4350 (approved):
Approved based on the work in: github.com/IntersectMBO/cardano-node/pull/6538

PR #6540, opened 2026-04-20
cardano-testnet | Test transaction autobalancing with withdrawal

carbolymer:
# Description

This test tries to reproduce https://github.com/IntersectMBO/cardano-cli/issues/965

The bug is not present anymore, so the test succeeds.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Cli/Transaction/WithdrawalReward.hs
+++ b/cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Cli/Transaction/WithdrawalReward.hs
@@ -1,0 +1,196 @@
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+
+-- | Smoke tests for @transaction build@ with a @--withdrawal@ flag,
+-- covering both key-witnessed and Plutus-script-witnessed withdrawals.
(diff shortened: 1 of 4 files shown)


palas (approved):
Everything looks correct. I've also checked and it fails if I removed the `withdrawal`. Curiously tests also seem to work without `--witness-override`, but not sure why

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Cli/Transaction/WithdrawalReward.hs
```diff
@@ -0,0 +1,196 @@
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+
+-- | Smoke tests for @transaction build@ with a @--withdrawal@ flag,
+-- covering both key-witnessed and Plutus-script-witnessed withdrawals.
+-- Loosely related to https://github.com/IntersectMBO/cardano-cli/issues/965.
```
Jimbo4350, line 8: Does this PR close 965?

Jimbo4350 (approved):

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Cli/Transaction/WithdrawalReward.hs
carbolymer: I think it proves that 965 is not reproducible now. So I think yes.

PR #6543, opened 2026-04-20
cardano-testnet: Add backwards-compatibility with some old cli/node versions

palas:
# Description

Makes `cardano-testnet` work seamlessly with older versions of `cardano-cli` and `cardano-node` (not all versions, just to some extent), without hard-coded version checks. The code probes the installed CLI's capabilities at runtime and adapts automatically — genesis files that the CLI doesn't produce are created via fallback, and the node configuration includes settings needed by older nodes.

**Motivation:** We aim to allow for `cardano-testnet` to run testnet with different versions of the `cardano-node`. This PR makes that easier, since we will only have to worry about running the correct node, and not about having the right configuration files generated.

**What changed:**

- **CLI capability detection** (`RunIO.hs`): `getCliHelpText` probes `cardano-cli <era> genesis create-testnet-data --help` to discover which `--spec-*` flags are supported.
- **Conditional spec flags** (`Configuration.hs`): `specFlag` takes a `(String -> Bool)` predicate and only passes flags the CLI advertises. Adding a future era is one extra `++ specFlag cliHas NewEra` line.
- **Byron genesis fallback** (`Configuration.hs`): When the CLI doesn't produce `byron-genesis.json` (older CLIs don't), it is created via the existing `createByronGenesis` from `Byron.hs`, with delegate keys copied into the expected pool-key directories.
- **Dijkstra genesis fallback** (`Configuration.hs`): When `dijkstra-genesis.json` is missing, the default is written directly so that the node configuration and hash computation stay uniform.
- **`Byron.hs` constraint relaxation**: `createByronGenesis` now requires only `MonadIO` (was `MonadTest, MonadCatch, MonadIO`) and uses `Testnet.Process.RunIO.execCli_`, so it can be called from both production code and tests.
- **`EnableP2P` in config** (`Defaults.hs`): Older nodes (≤ 9.x) require `EnableP2P: True` explicitly to parse P2P topology files; newer nodes already default to P2P so the flag is harmless.

**Tests with old versions of `cardano-node` and `cardano-cli` before this PR and after this PR:**

| Release | node   | cli       | Before | After |
|---------|--------|-----------|--------|-------|
| 9.0.0   | 9.0.0  | 9.0.0.0   | FAIL   | OK    |
| 9.1.0   | 9.1.0  | 9.2.1.0   | FAIL   | OK    |
| 9.1.1   | 9.1.1  | 9.2.1.0   | FAIL   | OK    |
| 9.2.0   | 9.2.0  | 9.4.1.0   | FAIL   | OK    |
| 9.2.1   | 9.2.1  | 9.4.1.0   | FAIL   | OK    |
| 10.1.1  | 10.1.1 | 10.1.0.0  | FAIL   | OK    |
| 10.1.2  | 10.1.2 | 10.1.1.0  | FAIL   | OK    |
| 10.1.3  | 10.1.3 | 10.1.1.0  | FAIL   | OK    |
| 10.1.4  | 10.1.4 | 10.1.1.0  | FAIL   | OK    |
| 10.2.1  | 10.2.1 | 10.4.0.0  | FAIL   | OK    |
| 10.3.1  | 10.3.1 | 10.7.0.0  | FAIL   | OK    |
| 10.4.1  | 10.4.1 | 10.8.0.0  | FAIL   | OK    |
| 10.5.1  | 10.5.1 | 10.11.0.0 | FAIL   | OK    |
| 10.5.2  | 10.5.2 | 10.11.0.0 | FAIL   | OK    |
| 10.5.3  | 10.5.3 | 10.11.0.0 | FAIL   | OK    |
| 10.5.4  | 10.5.4 | 10.11.0.0 | FAIL   | OK    |
| 10.6.2  | 10.6.2 | 10.15.0.0 | OK     | OK    |
| 10.6.3  | 10.6.3 | 10.15.0.0 | OK     | OK    |
| 10.6.4  | 10.6.4 | 10.15.0.0 | OK     | OK    |
| 10.7.1  | 10.7.1 | 10.16.0.0 | OK     | OK    |

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [x] Self-reviewed the diff

### DIFF
--- a/cardano-testnet/src/Testnet/Components/Configuration.hs
+++ b/cardano-testnet/src/Testnet/Components/Configuration.hs
@@ -262,7 +262,7 @@ createSPOGenesisAndFiles
      , inputGenesisDijkstraFp
      , tempAbsPath </> "byron.genesis.spec.json" -- Created by create-testnet-data
     ]
-    (\fp -> liftIOAnnotated $ whenM (System.doesFileExist fp) (System.removeFile fp))
+    $ \fp -> liftIOAnnotated $ whenM (System.doesFileExist fp) (System.removeFile fp)
 
   return genesisShelleyDir
   where
--- a/cardano-testnet/src/Testnet/Components/Configuration.hs
+++ b/cardano-testnet/src/Testnet/Components/Configuration.hs
@@ -53,13 +53,19 @@ import           System.FilePath.Posix (takeDirectory, (</>))
 import           Testnet.Blockfrost (blockfrostToGenesis)
 import qualified Testnet.Defaults as Defaults
 import           Testnet.Filepath
-import           Testnet.Process.RunIO (execCli_, liftIOAnnotated)
+import           Testnet.Process.RunIO (execCli_, getCliHelpText, liftIOAnnotated)
+import           Testnet.Start.Byron (ByronGenesisOptions (..), byronDefaultGenesisOptions,
+                   createByronGenesis)
 import           Testnet.Start.Types
 
 import qualified Hedgehog.Extras.Stock.OS as OS
 import qualified Hedgehog.Extras.Stock.Time as DTC
 
+import           Data.List (isInfixOf)
+import           Data.Maybe (fromMaybe)
+import           Data.Monoid.Extra (mwhen)
 import           RIO (MonadThrow, throwM)
+
 
 -- | Returns JSON encoded hashes of the era, as well as the hard fork configuration toggle.
 createConfigJson :: ()
@@ -72,9 +78,9 @@ createConfigJson :: ()
 createConfigJson (TmpAbsolutePath tempAbsPath) sbe = GHC.withFrozenCallStack $ do
   byronGenesisHash <- getByronGenesisHash $ tempAbsPath </> "byron-genesis.json"
   shelleyGenesisHash <- getHash ShelleyEra "ShelleyGenesisHash"
-  alonzoGenesisHash  <- getHash AlonzoEra  "AlonzoGenesisHash"
-  conwayGenesisHash  <- getHash ConwayEra  "ConwayGenesisHash"
-  dijkstraGenesisHash  <- getHash DijkstraEra  "DijkstraGenesisHash"
+  alonzoGenesisHash <- getHash AlonzoEra "AlonzoGenesisHash"
+  conwayGenesisHash <- getHash ConwayEra "ConwayGenesisHash"
+  dijkstraGenesisHash <- getHash DijkstraEra "DijkstraGenesisHash"
 
   pure $ mconcat
     [ byronGenesisHash
@@ -85,7 +91,7 @@ createConfigJson (TmpAbsolutePath tempAbsPath) sbe = GHC.withFrozenCallStack $ d
     , Defaults.defaultYamlHardforkViaConfig sbe
     ]
    where
-    getHash ::  MonadIO m => CardanoEra a -> Text.Text -> m (KeyMap Value)
+    getHash :: MonadIO m => CardanoEra a -> Text.Text -> m (KeyMap Value)
     getHash e = getShelleyGenesisHash (tempAbsPath </> Defaults.defaultGenesisFilepath e)
 
 createConfigJsonNoHash :: ()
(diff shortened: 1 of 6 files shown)


cardano-testnet/src/Testnet/Components/Configuration.hs
```diff
@@ -200,43 +206,86 @@ createSPOGenesisAndFiles
...
     , "--utxo-keys", show numSeededUTxOKeys]
     <> monoidForEraInEon @ConwayEraOnwards era (const ["--drep-keys", show cardanoNumDReps])
     <> [ "--start-time", DTC.formatIso8601 startTime
     , "--out-dir", tempAbsPath
     ]
 
+  -- Older cardano-cli versions do not generate byron-genesis.json from
+  -- create-testnet-data.  Create it via the legacy byron command when missing.
+  let byronGenesisPath = tempAbsPath </> Defaults.defaultGenesisFilepath ByronEra
+  unlessFileExists byronGenesisPath $ do
```
carbolymer, line 236: can you move this whole block into a separate function in where clause?

cardano-testnet/src/Testnet/Components/Configuration.hs
```diff
@@ -200,43 +206,86 @@ createSPOGenesisAndFiles
...
+  unlessFileExists dijkstraGenesisPath $
+    liftIOAnnotated $ LBS.writeFile dijkstraGenesisPath $ A.encodePretty dijkstraGenesis
+
   -- Remove the input files. We don't need them anymore, since create-testnet-data wrote new versions.
   forM_
     [  inputGenesisShelleyFp, inputGenesisAlonzoFp, inputGenesisConwayFp
+     , inputGenesisDijkstraFp
      , tempAbsPath </> "byron.genesis.spec.json" -- Created by create-testnet-data
     ]
     (\fp -> liftIOAnnotated $ whenM (System.doesFileExist fp) (System.removeFile fp))
```
carbolymer, line 265:
```suggestion
    $ \fp -> liftIOAnnotated $ whenM (System.doesFileExist fp) (System.removeFile fp)
```

cardano-testnet/src/Testnet/Process/RunIO.hs
```diff
@@ -93,6 +94,17 @@ execCli
   -> RIO env String
 execCli = GHC.withFrozenCallStack $ execFlex "cardano-cli" "CARDANO_CLI"
 
+-- | Best-effort query of the @cardano-cli@ help text for a subcommand.
+-- Returns 'Nothing' if the binary can't be found or the command fails.
+getCliHelpText :: MonadIO m => [String] -> m (Maybe String)
+getCliHelpText args = liftIO $ do
+  result <- try $ runRIO () $
+    execFlexAny' defaultExecConfig "cardano-cli" "CARDANO_CLI" (args ++ ["--help"])
+  let _ = result :: Either SomeException (ExitCode, String, String)
```
carbolymer, line 103: Why not move this annotation line above or below? I think also you'd want to use https://hackage.haskell.org/package/safe-exceptions-0.1.7.4/docs/Control-Exception-Safe.html#v:tryAny instead of `try` here.

cardano-testnet/src/Testnet/Components/Configuration.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/879c0f296

cardano-testnet/src/Testnet/Components/Configuration.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/c264d2f5b

cardano-testnet/src/Testnet/Process/RunIO.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/31c6e0316

Jimbo4350 (changes requested):
To clarify my earlier message. By "from this cardano-testnet onwards" I meant: make the current cardano-testnet work with the current cardano-cli release. That's the scope. I wasn't asking for a flag and backport mechanism or for us to collect past configurations. Since cardano-testnet has never been officially released there's no installed base to support, so we don't need a backwards compatibility path at all. We need forward compatibility with whatever cli is current.

palas:
Won't do, see: https://github.com/IntersectMBO/cardano-node/issues/6548

PR #6544, opened 2026-04-21
wb | genesis: leaner cache and Plutus cost models

fmaste:
# Workbench genesis

## Summary

- A "ripper" genesis backend: it splits a genesis into two independently-cached halves, the heavy dataset (keys, UTxO, delegators, DReps) and the per-era protocol parameters, then reassembles the final files with `cat`, `sed`, and `jq`.
- Restored non-default Plutus cost models. cardano-node now requires them through the Alonzo genesis `extraConfig` field, this PR adds support.

### Genesis creation

- Standardise on `cardano-cli latest genesis create-testnet-data`. The retired `create-staked` path and the separate byron genesis step are gone, byron now comes from `create-testnet-data` too.
- Small cache interface: `cache-hit`, `create-cache`, `derive-from-cache`, with one shared atomic commit (build in a temp dir, `mv` into place, so an interrupted run never leaves a half-written cache).
- Fix a cache-key leak: `create-testnet-data` bakes shelley `minUTxOValue` and conway `dRepDeposit` into dataset fields, so both are now part of the dataset cache key and injected into the specs.

#### Ripper backend

- **Separate caches.** Dataset and protocol parameters are cached independently, so profiles that share one half reuse it. Tweak a protocol parameter and the multi-hundred-MB dataset is not rebuilt, change pool or UTxO sizing and the protocol cache is untouched.
- **Plain tools.** Per-run assembly is `cat`, `sed`, and `jq`, with no `jq` pass over the large dataset and no pretty-printing.
- **Faithful.** The split and reassembly round-trip `create-testnet-data`'s output exactly, so the genesis the node sees is unchanged.

### Plutus cost models and profiles

- Restore arbitrary Plutus cost models through `genesis.alonzo.extraConfig`, the only path cardano-node still accepts for non-default parameter counts. The mainnet cost-model genesis fields were locked to fixed entry counts ([cardano-ledger#5241](https://github.com/IntersectMBO/cardano-ledger/pull/5241), [cardano-ledger#5379](https://github.com/IntersectMBO/cardano-ledger/pull/5379)), and `extraConfig.costModels` was added as the bypass for testnets and benchmarks ([cardano-ledger#5342](https://github.com/IntersectMBO/cardano-ledger/pull/5342), surfaced for the CLI in [cardano-cli#1352](https://github.com/IntersectMBO/cardano-cli/pull/1352)).
- Add Protocol Version 11 (preview and Dijkstra) support to `cardano-profile`.
- Re-derive `epoch-timeline.json` from a mainnet db-sync run and mainnet genesis, and restore the cost models that had been removed.

### Era as a first-class input

- New `--era-name` flag on `wb start` and `wb run allocate`, validated up front.
- The era is embedded in the run tag as a short code (`coay`, `bage`, `dira`, ...) and written straight to `meta.json` as `.meta.era`, instead of being reconstructed downstream.

### Simpler CLI and layout

- `wb run allocate` takes named flags only, and the profile name is read from `profile.json` (one source of truth, no positional arguments).
- On-disk genesis files use one naming convention everywhere, `genesis.<era>.json`, and the cache layout matches what `create-testnet-data` actually writes.
- Remove dead weight: pool `--relays` metadata, the genesis "spec" sub-command, and several unused key-shuffling helpers. `WB_BACKEND` becomes `WB_BACKEND_NAME`.
- The cloud backend no longer bundles the cache entry into the genesis tarball, which was already mounted.

### Compatibility

- Analysing older runs keeps working. `wb analyse` and the run helpers fall back to the previous genesis filenames and layout when the new ones are absent.

## Breaking changes

- `wb start` and `wb run allocate` now require `--era-name`.
- `wb run allocate` no longer accepts positional arguments; use the named flags.
- `WB_BACKEND` is renamed to `WB_BACKEND_NAME`.
- `--relays`, the genesis "spec" sub-command, and `create-staked` are removed.
- Non-default Plutus cost models must be supplied via `genesis.alonzo.extraConfig`.


mgmeier (approved):
LGTM - what an impressive chunk of work, thank you @fmaste

PR #6546, opened 2026-04-23
bench: cardano-recon-framework 1.1.1

Russoul:
* Support atoms that refer to property keys at arbitrary depth
* Replace crash-on-missing-key build flag with `--on-missing-key` CLI option


mgmeier (approved):
Great work @Russoul LGTM

PR #6549, opened 2026-04-24
cardano-testnet: declare supported cli/node versions

### LINKED ISSUE #6548: cardano-testnet: declare supported cli/node versions via README + --version output
## Summary

cardano-testnet currently has no user-facing declaration of which cardano-cli and cardano-node versions it supports. The only source of truth is `cardano-testnet.cabal` (lines 44–45) and users have no way to answer "which cli/node should I use with this cardano-testnet?" without reading cabal files.

## Prior art

- #5795 — explicitly titled "ADD COMPATIBILITY MATRIX", closed not-planned
- #5753 — docs.cardano.org matrix page 404'd
- #5903 — user asking what versions work with node 9.0.0

The existing matrix at https://docs.cardano.org/developer-resources/release-notes/comp-matrix/ covers cardano-node / db-sync / rosetta-java / graphql / wallet but **does not include cardano-cli or cardano-api**. The backing repo `cardano-foundation/docs-cardano-org` was archived 2025-07-22, so extending it is not a viable path.

## Proposal

Two one-time changes. Zero per-release maintenance.

### 1. README "Supported versions" section

Add to `cardano-testnet/README.md`:

> cardano-testnet is designed to run the cardano-cli and cardano-node that ship in the same release. Mixing binaries across releases is unsupported; use cli and node from the same release tarball as this cardano-testnet.
>
> The cardano-api and cardano-cli versions this build was compiled against are printed by `cardano-testnet --version`.

One rule covers every version pair. Leans on the existing major-version compat guarantee from consensus/network for same-release binaries.

### 2. Extend `cardano-testnet --version`

Report the cardano-api and cardano-cli versions this build was compiled against, sourced from the `VERSION_cardano_api` and `VERSION_cardano_cli` CPP macros cabal auto-generates based on the resolved dependency versions, so the cabal file remains the sole source of truth.

Target output:

```
cardano-testnet X.Y.Z - linux-x86_64 - ghc-9.6
git rev <sha>
built against cardano-api X.Y.Z.W
built against cardano-cli X.Y.Z.W
```

## Why not the alternatives

- **docs.cardano.org matrix update** — external dependency, repo archived, no update cadence we control.
- **Matrix-generating GHA** — higher upfront cost for value a single-sentence policy already captures.

## Acceptance

- [ ] `cardano-testnet/README.md` has a "Supported versions" section with the single-release policy statement.
- [ ] `cardano-testnet --version` output includes `built against cardano-api X.Y.Z.W` and `built against cardano-cli X.Y.Z.W`, sourced from cabal at build time.
- [ ] #5795 and #5903 are linked to the new README section and closed as resolved.
- [ ] PR #6543 either withdraws the backwards-compat mechanism or narrows to the forward-only fix against the current cli release.

palas:
# Description

Declares which `cardano-cli\` and `cardano-node` versions `cardano-testnet` supports,
via two zero-maintenance mechanisms:

1. **README "Supported versions" section** — states the single-release policy:
   use cli and node from the same release tarball as cardano-testnet.

2. **Extended `cardano-testnet version` output** — reports the `cardano-api` and
   `cardano-cli` versions the build was compiled against, using Cabal's
   auto-generated `VERSION_cardano_api` and `VERSION_cardano_cli` CPP macros.
   The cabal file remains the sole source of truth.

Also fixes the version banner which incorrectly said "cardano-node" instead of
"cardano-testnet".

Closes #6548

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [x] Self-reviewed the diff

### DIFF
--- a/cardano-testnet/src/Parsers/Version.hs
+++ b/cardano-testnet/src/Parsers/Version.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE TemplateHaskell #-}
 
 module Parsers.Version
@@ -33,6 +34,8 @@ runVersionOptions VersionOptions = do
     , " - ", os, "-", arch
     , " - ", compilerName, "-", showVersion compilerVersion
     , "\ngit rev ", T.unpack $(gitRev)
+    , "\nbuilt against cardano-api ", VERSION_cardano_api
+    , "\nbuilt against cardano-cli ", VERSION_cardano_cli
     ]
 
 cmdVersion :: Mod CommandFields VersionOptions
--- a/cardano-testnet/README.md
+++ b/cardano-testnet/README.md
@@ -1,1 +1,10 @@
 For information on how to use this package, see [Launching a Testnet](https://developers.cardano.org/docs/get-started/cardano-testnet)
+
+## Supported versions
+
+cardano-testnet is designed to run the cardano-cli and cardano-node that ship
+in the same release. Mixing binaries across releases is unsupported; use cli
+and node from the same release tarball as this cardano-testnet.
+
+The cardano-api and cardano-cli versions this build was compiled against are
+printed by `cardano-testnet version`.
--- a/cardano-testnet/changelog.d/20260424_015940_palas_tag_testnet_compatibility.md
+++ b/cardano-testnet/changelog.d/20260424_015940_palas_tag_testnet_compatibility.md
@@ -1,0 +1,5 @@
+### Added
+
+- Added "Supported versions" section to README declaring the single-release policy for cli/node compatibility.
+- `cardano-testnet version` now reports the cardano-api and cardano-cli versions it was built against.
+
--- a/cardano-testnet/test/cardano-testnet-golden/Cardano/Testnet/Test/Golden/Help.hs
+++ b/cardano-testnet/test/cardano-testnet-golden/Cardano/Testnet/Test/Golden/Help.hs
@@ -5,6 +5,7 @@
 module Cardano.Testnet.Test.Golden.Help
   ( golden_HelpAll
   , golden_HelpCmds
+  , golden_VersionCmd
   ) where
 
 import           Cardano.Testnet.Test.Golden.Util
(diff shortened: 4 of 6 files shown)


cardano-testnet/src/Parsers/Version.hs
```diff
@@ -29,10 +30,12 @@ optsVersion = pure VersionOptions
 runVersionOptions :: VersionOptions -> IO ()
 runVersionOptions VersionOptions = do
   IO.putStrLn $ mconcat
-    [ "cardano-node ", showVersion version
+    [ "cardano-testnet ", showVersion version
```
Jimbo4350: Why not render all the versions including cardano-node?

Jimbo4350 (commented):
We should implement a golden test for this.

cardano-testnet/src/Parsers/Version.hs
palas: Oh, I thought we were showing the `cardano-testnet` version and saying it was `cardano-node`'s

palas: Fixed in rebased commit

palas:
> We should implement a golden test for this.

@Jimbo4350, I've added a golden test in the latest commit. But, in retrospective, I am not sure it was a good idea. Because it needs to be changed in each bump. And I already deleted two lines to avoid the git rev (which would need to be changed in every commit), and the architecture (which would just never match), as well as the `cardano-node` version

Jimbo4350:
> > We should implement a golden test for this.
> 
> @Jimbo4350, I've added a golden test in the latest commit. But, in retrospective, I am not sure it was a good idea. Because it needs to be changed in each bump. And I already deleted two lines to avoid the git rev (which would need to be changed in every commit), and the architecture (which would just never match), as well as the `cardano-node` version

How do you suggest we police that the correct versions make it into the binary?

Jimbo4350 (approved):
LGTM

palas:
> How do you suggest we police that the correct versions make it into the binary?

I didn't know that was the goal. I hear `cardano-node` release is pretty tricky, so if this helps, then by all means. I guess that, in any case, it is good that it is obvious when the key dependencies have changed, that way it is harder that it happens by accident

PR #6550, opened 2026-04-24
cardano-testnet | Fix flaky height check in rpc test

carbolymer:
# Description

The chain height check was relying on an assumption that transaction goes into the next block, and sometimes the block nr check was returning stale info. 

This changes the implementation to use retry untilM, to reliably check for the height.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Rpc/Transaction.hs
+++ b/cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Rpc/Transaction.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE OverloadedLists #-}
 {-# LANGUAGE OverloadedStrings #-}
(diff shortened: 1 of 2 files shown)


cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Rpc/Transaction.hs
```diff
@@ -116,40 +120,30 @@ hprop_rpc_transaction = integrationRetryWorkspace 2 "rpc-tx" $ \tempAbsBasePath'
...
-  H.assertWith outputsAmounts $ elem (inject amount)
+    H.note_ $ "Ensure that there are 2 UTXOs in the address " <> show addrTxt1
+    utxosForAddress <- retryUntilM epochStateView (WaitForBlocks 10)
+      (do utxos <- H.evalIO $
+            Rpc.nonStreaming conn (Rpc.rpc @(Rpc.Protobuf UtxoRpc.QueryService "readUtxos")) def
+          flip filterM (utxos ^. U5c.items) $ \utxo -> do
+            utxoAddress <- deserialiseAddressBs addrInEra $ utxo ^. U5c.cardano . U5c.address
+            pure $ addr1 == utxoAddress
+      )
+      (\xs -> length xs == 2)
```
palas, line 142: This assumes the testnet generated address has a single UTxO, and it seems that is indeed the case. But I think it would be more robust to either use a freshly generated address or check the assumption. Also the last assertion ends up being `elem` instead of a `===`.

Also, now I realise this is really not from this PR, it was in the code already

palas (approved):
Nice 👍

Jimbo4350 (approved):

PR #6559, opened 2026-05-06
cardano-testnet: Add `--nodes` flag for per-node binary configuration

palas:
# Description

Adds a `--nodes` CLI flag to `cardano-testnet` that allows specifying the role (SPO or relay) and optionally a custom `cardano-node` binary for each node in the testnet. This enables running testnets with mixed node versions, which is needed for testing compatibility across node releases.

The new flag coexists with the existing `--num-pool-nodes` via `<|>` in the parser, so either can be used:

```
--nodes spo,spo:node-bin=/path/to/custom-node,relay,relay
--num-pool-nodes 3   # unchanged, still works
```

At creation time, custom binaries are validated by running `--version` and the result is recorded in a YAML env file (`node-data/nodeN/env`). At runtime, the env file is read to resolve the binary, falling back to the default resolution when no env file exists.

**Base PR:** #6563 (enforces SPOs come first and splits node list into SPO and relay)

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [x] Self-reviewed the diff

### DIFF
--- a/cardano-testnet/src/Parsers/Cardano.hs
+++ b/cardano-testnet/src/Parsers/Cardano.hs
@@ -3,6 +3,7 @@
 module Parsers.Cardano
   ( cmdCardano
   , cmdCreateEnv
+  , parseNodeSpecs
   ) where
 
 import           Cardano.Api (AnyShelleyBasedEra (..))
@@ -13,6 +14,7 @@ import           Cardano.Prelude (readMaybe)
 import           Prelude
 
 import           Control.Applicative (optional, (<|>))
+import           Control.Monad (unless)
 import           Data.Default.Class (def)
 import qualified Data.List as L
 import           Data.List.NonEmpty (NonEmpty ((:|)))
@@ -21,6 +23,10 @@ import           Data.Word (Word64)
 import           Options.Applicative (CommandFields, Mod, Parser)
 import qualified Options.Applicative as OA
 import           Options.Applicative.Types (readerAsk)
+import           Text.Parsec (char, many1, noneOf,
+                     sepBy1, string, try, (<?>), parse, eof, notFollowedBy)
+import qualified Text.Parsec as Parsec
+import qualified Text.Parsec.String as Parsec
 
 import           Testnet.Defaults (defaultEra)
 import           Testnet.Start.Cardano
@@ -55,7 +61,7 @@ pFromEnv = TestnetEnvOptions
 
 pCreationOptions :: Parser TestnetCreationOptions
 pCreationOptions = TestnetCreationOptions
-  <$> pTestnetNodeOptions
+  <$> pTestnetNodesWithOptions
   <*> pure (AnyShelleyBasedEra defaultEra)
   <*> pMaxLovelaceSupply
   <*> pNumDReps
@@ -105,27 +111,98 @@ pKesSource = OA.flag UseKesKeyFile UseKesSocket
   <>  OA.showDefault
   )
 
-pTestnetNodeOptions :: Parser TestnetNodeOptions
-pTestnetNodeOptions =
-  fmap (maybe cardanoDefaultTestnetNodeOptions mkPoolNodes) <$>
-    optional $ OA.option ensureAtLeastOne
-      (   OA.long "num-pool-nodes"
-      <>  OA.help "Number of pool nodes. Note this uses a default node configuration for all nodes."
-      <>  OA.metavar "COUNT"
-      )
+pTestnetNodesWithOptions :: Parser TestnetNodesWithOptions
+pTestnetNodesWithOptions =
+  pNodes <|> pNumPoolNodes <|> pure cardanoDefaultTestnetNodesWithOptions
   where
-    defaultSpoOption = NodeOptions []
-
-    mkPoolNodes num = TestnetNodeOptions
-      { optSpoNodes = defaultSpoOption :| L.replicate (num - 1) defaultSpoOption
-      , optRelayNodes = []
-      }
+    pNumPoolNodes :: Parser TestnetNodesWithOptions
+    pNumPoolNodes =
+      (\num -> TestnetNodesWithOptions { optSpoNodes = defaultSpoOption :| L.replicate (num - 1) defaultSpoOption, optRelayNodes = [] }) <$>
+        OA.option ensureAtLeastOne
+          (   OA.long "num-pool-nodes"
+          <>  OA.help "Number of pool nodes. Note this uses a default node configuration for all nodes."
+          <>  OA.metavar "COUNT"
+          )
+    defaultSpoOption = NodeWithOptions Nothing []
 
     ensureAtLeastOne :: OA.ReadM Int
     ensureAtLeastOne = readerAsk >>= \arg ->
       case readMaybe arg of
         Just n | n >= 1 -> pure n
         _ -> fail "Need at least one SPO node to produce blocks, but got none."
+
+    pNodes :: Parser TestnetNodesWithOptions
+    pNodes = OA.option readNodeSpecs
+      (   OA.long "nodes"
+      <>  OA.help "Comma-separated node specifications. SPO nodes must come before relay nodes. \
+                   \Each spec is a role (spo or relay) optionally followed by :node-bin=<path>. \
+                   \If the path contains commas, colons, double quotes, or backslashes, wrap it \
+                   \in double quotes and escape any literal double quotes as \\\" and backslashes \
+                   \as \\\\ within. To prevent bash from consuming the double quotes, enclose the \
+                   \whole argument in single quotes. \
+                   \Examples: --nodes spo,spo:node-bin=/path/to/bin,relay,relay | \
+                   \--nodes 'spo:node-bin=\"/path,with:commas\",relay'"
+      <>  OA.metavar "SPEC[,SPEC...]"
+      )
+
+    readNodeSpecs :: OA.ReadM TestnetNodesWithOptions
+    readNodeSpecs = readerAsk >>= either (fail . show) pure . parseNodeSpecs
+
+-- | Parse a @--nodes@ argument string into 'TestnetNodesWithOptions'.
+--
+-- SPO nodes are required to appear before relay nodes because:
+--
+-- 1. The testnet configuration assigns node directories by position (node-spo1,
+--    node-spo2, …, relay1, relay2, …). Allowing arbitrary ordering would require
+--    maintaining a separate mapping between pool indices and node positions, which
+--    much of the existing code does not expect.
+--
+-- 2. Silently reordering (e.g. turning @relay,spo@ into @spo,relay@) would
+--    violate the user's expectations about which node gets which configuration.
+--
+-- 3. Requiring SPOs first lets us represent the result directly as a 'NonEmpty'
+--    list of SPO nodes (guaranteeing at least one) plus a plain list of relays,
+--    so the type itself makes an invalid configuration unrepresentable.
+parseNodeSpecs :: String -> Either Parsec.ParseError TestnetNodesWithOptions
+parseNodeSpecs = parse (nodeSpecsParser <* eof) "Error parsing node specifications"
+  where
+    nodeSpecsParser :: Parsec.Parser TestnetNodesWithOptions
+    nodeSpecsParser = do
+      specs <- nodeSpec `sepBy1` char ','
+      let (spos, relays) = span (\(role, _) -> role == Spo) specs
+      unless (all (\(role, _) -> role == Relay) relays) $
+        fail "SPO nodes must come before relay nodes. Example: --nodes spo,spo,relay,relay"
+      case map snd spos of
+        [] -> fail "Need at least one SPO node to produce blocks."
+        (s:ss) -> pure $ TestnetNodesWithOptions
+          { optSpoNodes = s :| ss
+          , optRelayNodes = map snd relays
+          }
+
+    nodeSpec :: Parsec.Parser (NodeRole, NodeWithOptions)
+    nodeSpec = do
+      role <- nodeRole
+      bin <- optional $ char ':' *> nodeBinKV
+      pure (role, NodeWithOptions bin [])
+
+    nodeRole :: Parsec.Parser NodeRole
+    nodeRole =
+          Spo <$ try (string "spo" <* notFollowedBy (noneOf ",:\"\\"))
+      <|> Relay <$ try (string "relay" <* notFollowedBy (noneOf ",:\"\\"))
+      <?> "node role (\"spo\" or \"relay\")"
+
+    nodeBinKV :: Parsec.Parser FilePath
+    nodeBinKV = string "node-bin=" *> (quotedPath <|> unquotedPath) <?> "\"node-bin=<path>\", where <path> is the path to the node binary, optionally quoted if it contains special characters"
+
+    quotedPath :: Parsec.Parser FilePath
+    quotedPath = char '"' *> Parsec.many quotedChar <* char '"'
+      where
+        quotedChar = try (char '\\' *> (char '"' <|> char '\\')) <|> noneOf "\""
+
+    unquotedPath :: Parsec.Parser FilePath
+    unquotedPath = many1 (noneOf ",:\"\\")
+
+data NodeRole = Spo | Relay deriving Eq
 
 pOnChainParams :: Parser TestnetOnChainParams
 pOnChainParams = fmap (fromMaybe DefaultParams) <$> optional $
--- a/cardano-testnet/src/Testnet/Runtime.hs
+++ b/cardano-testnet/src/Testnet/Runtime.hs
@@ -57,7 +57,7 @@ import           Cardano.Node.Testnet.Paths (defaultSocketName)
 import qualified Testnet.Ping as Ping
 import           Testnet.Process.Run (ProcessError (..), initiateProcess)
 import           Testnet.Process.RunIO (execCli_, execKesAgentControl_, liftIOAnnotated,
-                   procKesAgent, procNode)
+                   procCustom, procKesAgent, procNode)
 import           Testnet.Types (TestnetKesAgent (..), TestnetNode (..),
                    TestnetRuntime (configurationFile), showIpv4Address, testnetSprockets)
 
@@ -121,11 +121,13 @@ startNode
   -- ^ Node port
   -> Int
   -- ^ Testnet magic
+  -> Maybe FilePath
+  -- ^ Optional custom node binary. 'Nothing' uses the default resolution.
   -> [String]
   -- ^ The command to execute to start the node.
   -- @--socket-path@, @--port@, and @--host-addr@ gets added automatically.
   -> ExceptT NodeStartFailure m TestnetNode
-startNode tp node ipv4 port _testnetMagic nodeCmd = GHC.withFrozenCallStack $ do
+startNode tp node ipv4 port _testnetMagic mNodeBin nodeCmd = GHC.withFrozenCallStack $ do
   let tempBaseAbsPath = makeTmpBaseAbsPath tp
       socketDir = makeSocketDir tp
       logDir = makeLogDir tp
@@ -156,7 +158,10 @@ startNode tp node ipv4 port _testnetMagic nodeCmd = GHC.withFrozenCallStack $ do
                              , "--port", show port
                              , "--host-addr", showIpv4Address ipv4
                              ]
-    nodeProcess <- newExceptT . fmap (first ExecutableRelatedFailure) . try $ runRIO () $ procNode completeNodeCmd
+    nodeProcess <- newExceptT . fmap (first ExecutableRelatedFailure) . try $ runRIO () $
+      case mNodeBin of
+        Nothing -> procNode completeNodeCmd
+        Just bin -> procCustom bin completeNodeCmd
 
     -- The port number if it is obtained using 'H.randomPort', it is firstly bound to and then closed. The closing
     -- and release in the operating system is done asynchronously and can be slow. Here we wait until the port
--- a/cardano-testnet/src/Testnet/Start/Cardano.hs
+++ b/cardano-testnet/src/Testnet/Start/Cardano.hs
@@ -17,9 +17,9 @@ module Testnet.Start.Cardano
   , TestnetCreationOptions(..)
   , TestnetRuntimeOptions(..)
   , TestnetEnvOptions(..)
-  , TestnetNodeOptions(..)
-  , NodeOptions(..)
-  , cardanoDefaultTestnetNodeOptions
+  , TestnetNodesWithOptions(..)
+  , NodeWithOptions(..)
+  , cardanoDefaultTestnetNodesWithOptions
 
   , TestnetRuntime (..)
 
@@ -28,7 +28,7 @@ module Testnet.Start.Cardano
   , createTestnetEnv
   , getDefaultAlonzoGenesis
   , getDefaultShelleyGenesis
-  , readNodeOptionsFromEnv
+  , readNodesWithOptionsFromEnv
   , retryOnAddressInUseError
 
   , liftToIntegration
@@ -47,11 +47,14 @@ import           Ouroboros.Network.PeerSelection.RelayAccessPoint (RelayAccessPo
 import           Prelude hiding (lines)
 
 import           Control.Concurrent (threadDelay)
-import           Control.Monad (forM, forM_, unless, when)
+import           Control.Monad (forM, forM_, guard, unless, when)
+import           Control.Monad.Trans.Maybe (runMaybeT)
+import           Control.Exception (IOException)
 import           Control.Monad.Catch
 import           Control.Monad.Trans.Resource (MonadResource, getInternalState)
 import           Data.Aeson
 import qualified Data.Aeson.Encode.Pretty as A
+import qualified Data.Yaml as Yaml
 import qualified Data.ByteString.Lazy as LBS
 import           Data.Default.Class ()
 import           Data.Either
@@ -67,12 +70,13 @@ import           Data.Time.Clock (NominalDiffTime)
 import qualified Data.Time.Clock as DTC
 import           GHC.Stack
 import qualified System.Directory as IO
+import qualified System.Process as Process
 import           System.FilePath ((</>))
 
 import           Testnet.Components.Configuration
 import qualified Testnet.Defaults as Defaults
-import           Cardano.Node.Testnet.Paths (defaultConfigFile, defaultPortFile,
-                   defaultUtxoAddrPath)
+import           Cardano.Node.Testnet.Paths (defaultConfigFile, defaultNodeEnvFile,
+                   defaultPortFile, defaultUtxoAddrPath)
 import           Testnet.Filepath
 import           Testnet.Handlers (interruptNodesOnSigINT)
 import           Testnet.Orphans ()
@@ -93,6 +97,7 @@ import           RIO.State (put)
 import           UnliftIO.Async
 import           UnliftIO.Exception (stringException)
 
+
 liftToIntegration :: HasCallStack => RIO ResourceMap a -> H.Integration a
 liftToIntegration r =  do
    rMap <- lift $ lift getInternalState
@@ -109,7 +114,7 @@ createTestnetEnv :: ()
 createTestnetEnv
   creationOptions@TestnetCreationOptions
     { creationEra=asbe
-    , creationNodes=TestnetNodeOptions{optSpoNodes, optRelayNodes}
+    , creationNodes=TestnetNodesWithOptions{optSpoNodes, optRelayNodes}
     }
   Conf
     { genesisHashesPolicy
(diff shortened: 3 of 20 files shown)


cardano-testnet/src/Testnet/Start/Cardano.hs
```diff
@@ -153,6 +156,16 @@ createTestnetEnv
     let topology = Defaults.defaultP2PTopology producers
     liftIOAnnotated . LBS.writeFile (nodeDataDir </> "topology.json") $ A.encodePretty topology
 
+    -- Write env file for nodes with custom binaries
+    case nodeBin nodeOption of
+      Nothing -> pure ()
+      Just bin -> do
```
```suggestion
    forM_ (nodeBin nodeOption) $ \bin -> do
```

cardano-testnet/src/Testnet/Runtime.hs
```diff
@@ -156,7 +158,10 @@ startNode tp node ipv4 port _testnetMagic nodeCmd = GHC.withFrozenCallStack $ do
                              , "--port", show port
                              , "--host-addr", showIpv4Address ipv4
                              ]
-    nodeProcess <- newExceptT . fmap (first ExecutableRelatedFailure) . try $ runRIO () $ procNode completeNodeCmd
+    nodeProcess <- newExceptT . fmap (first ExecutableRelatedFailure) . try $ runRIO () $
+      case mNodeBin of
+        Nothing -> procNode completeNodeCmd
+        Just bin -> pure (IO.proc bin completeNodeCmd){ IO.create_group = True }
```
carbolymer: This is too low-level at this stage. It is too magical and does not explain why we are creating group.
```haskell
Just bin -> pure (IO.proc bin completeNodeCmd){ IO.create_group = True }
```
I think this means that procFlex' should be refactored to support such usage.

I think this is also the right time to pay the technical debt from https://github.com/IntersectMBO/cardano-node/pull/6346 (see unresolved comments there):
1. Simplify (and rename) `liftToIntegration`
2. Upstream changes (especially duplicated functions in `Testnet.Process.RunIO` to hedgehog-extras.
3. I think `liftIOAnntated` when captures exception it still prints duplicated messages - you'd have to double check.

cardano-testnet/src/Testnet/Start/Cardano.hs
```diff
@@ -517,12 +530,49 @@ readNodeOptionsFromEnv envDir = do
...
+  exists <- IO.doesFileExist envFile
+  if not exists
+    then pure Nothing
+    else do
+      result <- Yaml.decodeFileEither envFile
+      case result of
+        Right NodeEnv{node_binary} -> pure (Just node_binary)
+        Left err -> throwString $ "Failed to parse node env file " <> envFile <> ": " <> show err
+
+getNodeVersion :: MonadIO m => FilePath -> m String
```
```suggestion
getNodeVersion :: HasCallStack => MonadIO m => FilePath -> m String
```

cardano-testnet/src/Testnet/Start/Cardano.hs
```diff
@@ -517,12 +530,49 @@ readNodeOptionsFromEnv envDir = do
...
+getNodeVersion bin = liftIO $ do
```
```suggestion
getNodeVersion bin = liftIOAnnotated $ do
```

cardano-testnet/src/Testnet/Start/Cardano.hs
```diff
@@ -517,12 +530,49 @@ readNodeOptionsFromEnv envDir = do
...
+readNodeBinFromEnvFile :: MonadIO m => FilePath -> m (Maybe FilePath)
+readNodeBinFromEnvFile envFile = liftIO $ do
+  exists <- IO.doesFileExist envFile
+  if not exists
+    then pure Nothing
+    else do
+      result <- Yaml.decodeFileEither envFile
+      case result of
+        Right NodeEnv{node_binary} -> pure (Just node_binary)
+        Left err -> throwString $ "Failed to parse node env file " <> envFile <> ": " <> show err
```
carbolymer: 1. Missing `HasCallStack`
2. Staircase is ugly
What do you think about:

```suggestion
readNodeBinFromEnvFile :: (HasCallStack, MonadIO m) => FilePath -> m (Maybe FilePath)
  readNodeBinFromEnvFile envFile = runMaybeT $ do
    guard =<< liftIOAnnotated (IO.doesFileExist envFile)
    NodeEnv{node_binary} <- either failParse pure =<< liftIOAnnotated (Yaml.decodeFileEither envFile)
    pure node_binary
    where
      failParse err = throwString $ "Failed to parse node env file " <> envFile <> ": " <> show err
```

cardano-testnet/src/Testnet/Start/Cardano.hs
```diff
@@ -517,12 +530,49 @@ readNodeOptionsFromEnv envDir = do
...
     parseNodeNum s = do
       rest <- stripPrefix "node" s
       readMaybe rest :: Maybe Int
+    readNodeOpt i = do
+      bin <- readNodeBinFromEnvFile (envDir </> defaultNodeEnvFile i)
+      pure $ NodeOption bin []
+
+data NodeEnv = NodeEnv
+  { node_binary :: FilePath
+  , node_version :: String
```
carbolymer: why snake case?

cardano-testnet/src/Parsers/Cardano.hs
```diff
@@ -107,26 +108,74 @@ pKesSource = OA.flag UseKesKeyFile UseKesSocket
...
+          case spos of
+            [] -> fail "Need at least one SPO node to produce blocks."
+            ((_,s):ss) -> pure $ TestnetNodeOptions
+              { optSpoNodes = s :| map snd ss
+              , optRelayNodes = map snd relays
+              }
+        Left err -> fail err
+
+    parseNodeSpec :: String -> Either String (String, NodeOption)
+    parseNodeSpec spec = case splitOnChar ':' spec of
```
carbolymer: this breaks on paths with `:`. Consider supporting escaping.

cardano-testnet/src/Testnet/Start/Types.hs
```diff
@@ -258,8 +258,9 @@ instance Default GenesisOptions where
     , genesisActiveSlotsCoeff = 0.05
     }
 
-newtype NodeOption = NodeOption
-  { nodeExtraCliArgs :: [String]
+data NodeOption = NodeOption
```
carbolymer: haddocks por favor

cardano-testnet/src/Testnet/Start/Cardano.hs
```diff
@@ -517,12 +530,49 @@ readNodeOptionsFromEnv envDir = do
...
+    then pure Nothing
+    else do
+      result <- Yaml.decodeFileEither envFile
+      case result of
+        Right NodeEnv{node_binary} -> pure (Just node_binary)
+        Left err -> throwString $ "Failed to parse node env file " <> envFile <> ": " <> show err
+
+getNodeVersion :: MonadIO m => FilePath -> m String
+getNodeVersion bin = liftIO $ do
+  output <- Process.readProcess bin ["--version"] ""
```
carbolymer, line 574: If the binary does not exist it will just throw an error `does not exist (No such file or directory)`.
It would be worth catching and rethrowing with more error information.

cardano-testnet/src/Testnet/Start/Types.hs
```diff
@@ -279,10 +280,10 @@ instance Default (UserProvidedData a) where
 
 cardanoDefaultTestnetNodeOptions :: TestnetNodeOptions
 cardanoDefaultTestnetNodeOptions = TestnetNodeOptions
-  { optSpoNodes = NodeOption [] :| []
-  , optRelayNodes = [ NodeOption []
-                    , NodeOption []
-                    ]
+  { optSpoNodes = NodeOption Nothing [] :| []
+  , optRelayNodes = [ NodeOption Nothing []
+                 , NodeOption Nothing []
```
carbolymer: I think formatting is broken here

cardano-testnet/src/Parsers/Cardano.hs
```diff
@@ -107,26 +108,74 @@ pKesSource = OA.flag UseKesKeyFile UseKesSocket
...
+      (   OA.long "nodes"
+      <>  OA.help "Comma-separated node specifications. SPO nodes must come before relay nodes. \
+                   \Each spec is a role (spo or relay) optionally followed by key=value pairs \
+                   \separated by colons. \
+                   \Example: --nodes spo,spo:node-bin=/path/to/bin,relay,relay"
+      <>  OA.metavar "SPEC[,SPEC...]"
+      )
+
+    readNodeSpecs :: OA.ReadM TestnetNodeOptions
+    readNodeSpecs = readerAsk >>= \arg ->
```
carbolymer: this could be simpler with `parsec`, it also ignores `:` in paths:
```haskell
  import           Cardano.Api.Parser.Text (Parser, runParser, char, many1, noneOf,
                     notFollowedBy, alphaNum, sepBy1, string, try, (<?>))

  readNodeSpecs :: OA.ReadM TestnetNodeOptions
  readNodeSpecs = readerAsk >>= either fail pure . runParser nodeSpecsParser . Text.pack

  nodeSpecsParser :: Parser TestnetNodeOptions
  nodeSpecsParser = do
    specs <- nodeSpec `sepBy1` char ','
    let (spos, relays) = span (\(role, _) -> role == Spo) specs
    unless (all (\(role, _) -> role == Relay) relays) $
      fail "SPO nodes must come before relay nodes. Example: --nodes spo,spo,relay,relay"
    case map snd spos of
      [] -> fail "Need at least one SPO node to produce blocks."
      (s:ss) -> pure $ TestnetNodeOptions
        { optSpoNodes = s :| ss
        , optRelayNodes = map snd relays
        }

  nodeSpec :: Parser (NodeRole, NodeOption)
  nodeSpec = do
    role <- nodeRole
    bin <- optional $ char ':' *> nodeBinKV
    pure (role, NodeOption bin [])

  nodeRole :: Parser NodeRole
  nodeRole =
        Spo <$ try (string "spo" <* notFollowedBy alphaNum)
    <|> Relay <$ try (string "relay" <* notFollowedBy alphaNum)
    <?> "node role (spo or relay)"

  nodeBinKV :: Parser FilePath
  nodeBinKV = string "node-bin=" *> many1 (noneOf ",") <?> "node-bin=<path>"

  data NodeRole = Spo | Relay deriving Eq
```

carbolymer (dismissed):
Exception handling is most important here

cardano-testnet/src/Testnet/Start/Cardano.hs
palas: Applied here: https://github.com/IntersectMBO/cardano-node/pull/6559/changes/68aa58b46007a5cd9c896a18df086781ce627e50

cardano-testnet/src/Testnet/Start/Cardano.hs
palas: Applied here: https://github.com/IntersectMBO/cardano-node/pull/6559/changes/68aa58b46007a5cd9c896a18df086781ce627e50

cardano-testnet/src/Testnet/Start/Cardano.hs
palas: Applied here: https://github.com/IntersectMBO/cardano-node/pull/6559/changes/68aa58b46007a5cd9c896a18df086781ce627e50

cardano-testnet/src/Testnet/Start/Cardano.hs
palas: Applied here: https://github.com/IntersectMBO/cardano-node/pull/6559/changes/68aa58b46007a5cd9c896a18df086781ce627e50

cardano-testnet/src/Testnet/Start/Cardano.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/6742d240ca11bad4b5b73edaf1873ee185ac9a10

cardano-testnet/src/Testnet/Start/Cardano.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/3d06d0ec1df3bfc85849422ccdb0432b94e01445

cardano-testnet/src/Parsers/Cardano.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/24c416f16f77b5cb658ddb5902fdd36aec8b38fd

cardano-testnet/src/Parsers/Cardano.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/24c416f16f77b5cb658ddb5902fdd36aec8b38fd

cardano-testnet/src/Testnet/Start/Types.hs
palas: Addressed in previous PR already

cardano-testnet/src/Testnet/Start/Types.hs
palas: Addressed in previous PR already

cardano-testnet/src/Testnet/Runtime.hs
palas: Makes sense. I think what is relevant to this PR is the moving of the low-level code. So I've addressed that by adding `procCustom` in: https://github.com/IntersectMBO/cardano-node/commit/ffe73e82ce497143f6f242eb2550bae22a215695

Now regarding the rest of issues:
- I've investigated the duplicate messages and it seems `liftIOAnnotated` is fine, it rethrows, but `liftToIntegration` does produce duplicate messages. So I've created an issue to fix that and rename `liftToIntegration`: https://github.com/IntersectMBO/cardano-node/issues/6567.
- I have seen the duplication with respect to `hedgehog-extras` and I've created an issue: https://github.com/input-output-hk/hedgehog-extras/issues/111. For upstreaming the `MonadIO` versions of the files, adding `HasCallStack`, and exporting `binDist`, as well as adding `procCustom`. Once that is relealsed we can replace the calls from `cardano-testnet` to use those.

cardano-testnet/test/cardano-testnet-golden/files/golden/help/create-env.cli
```diff
@@ -11,6 +12,17 @@ Usage: cardano-testnet create-env [--num-pool-nodes COUNT]
   Create a sandbox for Cardano testnet
 
 Available options:
+  --nodes SPEC[,SPEC...]   Comma-separated node specifications. SPO nodes must
```
Jimbo4350, line 15: My question would be is do we expect more options to be specified beyond what is added today? If so then we may be better off specifying a file.

cardano-testnet/src/Testnet/Runtime.hs
palas: Please, let me know if I missed anything

Jimbo4350: @carbolymer can you create issues regarding the tech debt from #6346?

cardano-testnet/test/cardano-testnet-golden/files/golden/help/create-env.cli
palas: It is hard to tell, I imagine there must be things you could potentially want to specify per node, but then there is also the possibility of creating the env and modifying before executing. The functionality that I was thinking about that seems would be convenient would be to have multipliers, like `3x:spo,4x:relay:node-bin=<...>`.

cardano-testnet/src/Parsers/Cardano.hs
```diff
@@ -105,28 +110,85 @@ pKesSource = OA.flag UseKesKeyFile UseKesSocket
...
     ensureAtLeastOne = readerAsk >>= \arg ->
       case readMaybe arg of
         Just n | n >= 1 -> pure n
         _ -> fail "Need at least one SPO node to produce blocks, but got none."
 
+    pNodes :: Parser TestnetNodesWithOptions
+    pNodes = OA.option readNodeSpecs
+      (   OA.long "nodes"
+      <>  OA.help "Comma-separated node specifications. SPO nodes must come before relay nodes. \
+                   \Each spec is a role (spo or relay) optionally followed by :node-bin=<path>. \
```
Jimbo4350, line 138: This needs to exist in documentation somewhere it can't only be in the help text.

cardano-testnet/src/Parsers/Cardano.hs
```diff
@@ -105,28 +110,85 @@ pKesSource = OA.flag UseKesKeyFile UseKesSocket
...
+-- | Parse a @--nodes@ argument string into 'TestnetNodesWithOptions'.
+parseNodeSpecs :: String -> Either Parsec.ParseError TestnetNodesWithOptions
+parseNodeSpecs = parse (nodeSpecsParser <* eof) "Error parsing node specifications"
+  where
+    nodeSpecsParser :: Parsec.Parsec String () TestnetNodesWithOptions
+    nodeSpecsParser = do
+      specs <- nodeSpec `sepBy1` char ','
+      let (spos, relays) = span (\(role, _) -> role == Spo) specs
+      unless (all (\(role, _) -> role == Relay) relays) $
+        fail "SPO nodes must come before relay nodes. Example: --nodes spo,spo,relay,relay"
```
Jimbo4350, line 174: Why? Does the order actually matter or is it a matter of the presence of SPOs? Can you also separate the parsing and validation logic here?

cardano-testnet/src/Parsers/Cardano.hs
```diff
@@ -105,28 +110,85 @@ pKesSource = OA.flag UseKesKeyFile UseKesSocket
...
+      )
+
+    readNodeSpecs :: OA.ReadM TestnetNodesWithOptions
+    readNodeSpecs = readerAsk >>= either (fail . show) pure . parseNodeSpecs
+
+-- | Parse a @--nodes@ argument string into 'TestnetNodesWithOptions'.
+parseNodeSpecs :: String -> Either Parsec.ParseError TestnetNodesWithOptions
+parseNodeSpecs = parse (nodeSpecsParser <* eof) "Error parsing node specifications"
+  where
+    nodeSpecsParser :: Parsec.Parsec String () TestnetNodesWithOptions
```
Jimbo4350: Can you write a basic property test for this that confirms we get the expected number of SPOs and relays?

cardano-testnet/src/Testnet/Start/Cardano.hs
```diff
@@ -517,12 +530,48 @@ readNodeOptionsFromEnv envDir = do
...
     [] -> throwString "No SPO node directories found in environment"
   where
     parseNodeNum s = do
       rest <- stripPrefix "node" s
       readMaybe rest :: Maybe Int
+    readNodeOpt i = do
+      bin <- readNodeBinFromEnvFile (envDir </> defaultNodeEnvFile i)
+      pure $ NodeWithOptions bin []
+
+data NodeEnv = NodeEnv
```
Jimbo4350, line 548: Missing haddocks

cardano-testnet/src/Testnet/Start/Types.hs
```diff
@@ -259,15 +259,16 @@ instance Default GenesisOptions where
     }
 
 -- | Configuration specific to each node
-newtype NodeOptions = NodeOptions
-  { nodeExtraCliArgs :: [String] -- ^ Extra CLI arguments passed to @cardano-node run@
+data NodeWithOptions = NodeWithOptions
+  { nodeBin :: Maybe FilePath -- ^ Path to the @cardano-node@ binary to use for running this node. 'Nothing' uses the default resolution mechanism.
```
Jimbo4350: Can you expand on this a little. What's a potential purpose of specifying the node binary and roughly what is the default resolution mechanism?

Jimbo4350 (commented):
LGTM; suggested some minor changes. Will have another look after changes/discussion.

cardano-testnet/src/Testnet/Runtime.hs
carbolymer: - https://github.com/IntersectMBO/cardano-node/issues/6569
  - https://github.com/IntersectMBO/cardano-node/issues/6568
  - https://github.com/input-output-hk/hedgehog-extras/issues/112

cardano-testnet/src/Testnet/Start/Cardano.hs
```diff
@@ -517,12 +530,48 @@ readNodeOptionsFromEnv envDir = do
   when (null spoFlags) $
     throwString "No SPO node directories found in environment"
```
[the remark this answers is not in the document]
palas: There are two versions of `throwString`,  there is one that works on in `MonadIO`, and I guess that is the one used here, because the code does compile: https://hoogle.haskell.org/?hoogle=throwString&scope=set%3Astackage

cardano-testnet/src/Testnet/Start/Cardano.hs
```diff
@@ -517,12 +530,48 @@ readNodeOptionsFromEnv envDir = do
...
+instance ToJSON NodeEnv where
+  toJSON NodeEnv{nodeBinary, nodeVersion} =
+    object [ "node_binary" .= nodeBinary
+           , "node_version" .= nodeVersion
+           ]
+
+readNodeBinFromEnvFile :: (HasCallStack, MonadIO m) => FilePath -> m (Maybe FilePath)
+readNodeBinFromEnvFile envFile = runMaybeT $ do
+  guard =<< liftIOAnnotated (IO.doesFileExist envFile)
+  NodeEnv{nodeBinary} <- either failParse pure =<< liftIOAnnotated (Yaml.decodeFileEither envFile)
```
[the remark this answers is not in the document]
palas: There are two versions of throwString, there is one that works on in MonadIO, and I guess that is the one used here, because the code does compile: https://hoogle.haskell.org/?hoogle=throwString&scope=set%3Astackage

cardano-testnet/src/Parsers/Cardano.hs
palas: We have several options, but I thought this was the least problematic one. For example, we could:
- **Allow random orderings**: but that would affect a lot of code that assumes SPOs go before relays, this wasn't a problem before because you couldn't specify, also we would need to store a mapping between pools and nodes, which currently are just folders numbered with consecutive numbers starting in 1.
- **Sort them before generating the config**: but I think that would be misleading. For example, if someone writes `relay,relay,spo,relay`, we could silently do `node1=spo, node2=relay, node3=relay, node4=relay`, but I think that would break the users expectations.
- **Complain when spos are not the first nodes**: this forces the user to reorder and avoids the false expectation. Also, it allows us to not have to modify the rest of the code. And it has another very good advantage, we can easily enforce that there is at least one spo by just using a pair (spo=`NEL NodeWithOptions`, relay=`[NodeWithOptions]`). If we don't ensure that SPO nodes go first, this becomes tricker.

So I thought the last one (complaining) was the best approach.

About parsing versus validation: I can do that, but I think it is a bad idea, because:

- As it is, it is impossible to "forget" to validate, and that is very good. In fact, I already forgot validating in one place last week and copilot caught it.
- The types already enforce what is being validated, so we would need to introduce a new intermediate type that doesn't enforce it for it to be returned by the parser.

Also it goes against the idea of ["parse, don't validate"](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/).

cardano-testnet/src/Parsers/Cardano.hs
```diff
@@ -105,28 +110,85 @@ pKesSource = OA.flag UseKesKeyFile UseKesSocket
...
+
+    quotedPath :: Parsec.Parsec String () FilePath
+    quotedPath = char '"' *> Parsec.many quotedChar <* char '"'
+      where
+        quotedChar = try (char '\\' *> (char '"' <|> char '\\')) <|> noneOf "\""
+
+    unquotedPath :: Parsec.Parsec String () FilePath
+    unquotedPath = many1 (noneOf ",:\"\\")
+
+data NodeRole = Spo | Relay deriving Eq
```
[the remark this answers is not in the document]
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/375d6a980fd8332479e783f96e8f463f86131326

cardano-testnet/src/Parsers/Cardano.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/375d6a980fd8332479e783f96e8f463f86131326

cardano-testnet/src/Testnet/Start/Types.hs
```diff
@@ -31,9 +31,9 @@ module Testnet.Start.Types
   , UpdateTimestamps(..)
   , TestnetOnChainParams(..)
   , mainnetParamsRequest
-  , TestnetNodeOptions(..)
-  , NodeOptions(..)
-  , cardanoDefaultTestnetNodeOptions
+  , TestnetNodesWithOptions(..)
+  , NodeWithOptions(..)
+  , cardanoDefaultTestnetNodesWithOptions
```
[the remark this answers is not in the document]
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/9ff2ba95b04b6f5c97eef7ee4da7b36b7f2315bc

cardano-testnet/src/Testnet/Start/Cardano.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/18b1af1850d475838c95983c8b92c6c8b7e0d569

cardano-testnet/src/Testnet/Start/Types.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/050a2bc5a8859f0228e3e12bdf454b42914e7cc5

cardano-testnet/src/Parsers/Cardano.hs
palas: Created issue: https://github.com/IntersectMBO/cardano-node/issues/6571

cardano-testnet/src/Parsers/Cardano.hs
carbolymer:
```suggestion
    nodeSpecsParser :: Parsec.Parser TestnetNodesWithOptions
```
but you'd need to import `Text.Parsec.String`

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Parser.hs
carbolymer, line 1: could we revert the order: put the tests first, then utility functions at the bottom?

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Parser.hs
```diff
@@ -0,0 +1,190 @@
...
+      H.failure
+    Right result -> do
+      nodeBin (NEL.head (optSpoNodes result)) H.=== Just "/usr/bin/cardano-node"
+      length (optRelayNodes result) H.=== 1
+
+-- | Quoted paths with commas, colons, backslashes, and quotes parse correctly.
+unit_quoted_paths :: H.Property
+unit_quoted_paths = H.withTests 1 . H.property $ do
+  case parseNodeSpecs "spo:node-bin=\"/path,with,commas\"" of
+    Left err -> do { H.footnote $ show err; H.failure }
```
carbolymer: What is the advantage of this over `H.leftFail`? Footnotes in addition are shown far away from a failure point.

carbolymer (commented):
Looks good to me. I'm leaving approval to @Jimbo4350

cardano-testnet/src/Parsers/Cardano.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/aa330e53f746bfc36c9abdd24e0bd4b422bfc52e

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Parser.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/2a2205708062e3d58a0656fdd0ab57653154559a

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Parser.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/c012c8e6a55b6d092c8a0637ad23ec830315b62f

cardano-testnet/src/Parsers/Cardano.hs
Jimbo4350: Regarding the ordering, fair enough but your explanation should be documented.

Jimbo4350 (approved):
LGTM, one minor comment

cardano-testnet/src/Parsers/Cardano.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/db7bdf56ee3bcce57575d687d54935511a37aec7

palas:
Backed up in: https://github.com/IntersectMBO/cardano-node/tree/backup/testnet-specify-node-bin-per-node

PR #6563, opened 2026-05-06
cardano-testnet: Enforce SPOs come first and split node list into SPO and relay

palas:
# Description

The `NodeOption` type was refactored from a sum type (`SpoNodeOptions | RelayNodeOptions`) into a record, and a new `TestnetNodeOptions` container enforces at the type level that there is at least one SPO (`NonEmpty`) and that SPOs come before relays (by structure). The parser also validates this ordering.

`readNodeOptionsFromEnv` now validates that node directories are consecutively numbered and that SPOs come before relays, matching the assumptions made by `cardanoTestnet`.

**Follow-up PR:** #6559 (adds `--nodes` flag for per-node binary configuration on top of this)

I would recommend looking at the diff with white-space disabled.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [x] Self-reviewed the diff


cardano-testnet/src/Testnet/Start/Cardano.hs
```diff
@@ -511,19 +497,32 @@ retryOnAddressInUseError act = withFrozenCallStack $ go maximumTimeout retryTime
...
 
 -- | Read node options from an existing testnet environment directory.
--- Scans @node-data/@ for node directories and checks @pools-keys/@ to
--- classify each node as SPO or relay.
-readNodeOptionsFromEnv :: MonadIO m => FilePath -> m (NonEmpty NodeOption)
+-- Scans @node-data/@ for node directories numbered @node1, node2, ...@
+-- and checks @pools-keys/@ to classify each as SPO or relay.
+-- Validates that nodes are consecutively numbered starting from 1,
+-- and that all SPO nodes come before relay nodes.
+readNodeOptionsFromEnv :: MonadIO m => FilePath -> m TestnetNodeOptions
```
```suggestion
readNodeOptionsFromEnv :: HasCallStack => MonadIO m => FilePath -> m TestnetNodeOptions
```

cardano-testnet/src/Testnet/Start/Types.hs
```diff
@@ -259,12 +258,14 @@ instance Default GenesisOptions where
     , genesisActiveSlotsCoeff = 0.05
     }
 
--- | Whether a node should be an SPO or just a relay.
--- The '@String' arguments will be appended to the default options when starting the node.
-data NodeOption
-  = SpoNodeOptions [String]
-  | RelayNodeOptions [String]
-  deriving (Eq, Show)
+newtype NodeOption = NodeOption
```
carbolymer: It has multiple options so it should be plural, no?
```suggestion
newtype NodeOptions = NodeOptions
```

Also haddocks on both types would be nice.

cardano-testnet/src/Testnet/Start/Types.hs
```diff
@@ -259,12 +258,14 @@ instance Default GenesisOptions where
...
+  { nodeExtraCliArgs :: [String]
+  } deriving (Eq, Show)
```
carbolymer, line 264: I guess it could be useful.
```suggestion
  } deriving (Eq, Show, IsList)
```

carbolymer (approved):

cardano-testnet/src/Testnet/Start/Types.hs
palas: I don't think it makes sense because we are adding the `nodeBin` field in the next PR

palas: But I didn't know `IsList` existed, so it is interesting

cardano-testnet/src/Testnet/Start/Types.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/pull/6563/changes/86e23ba3b3b5cdeeab5e460ca4ba702db34ad4eb

cardano-testnet/src/Testnet/Start/Cardano.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/pull/6563/changes/86e23ba3b3b5cdeeab5e460ca4ba702db34ad4eb

cardano-testnet/src/Testnet/Start/Types.hs
Jimbo4350, line 261: A better name for this may be `NodeWithArgs`.

palas: I don't think so, because we'll be adding more things. Next PR will be adding the binary path soon, so it needs to be more generic. I just split it in two PRs for easier reviewing

Jimbo4350: A binary points to a (soon to be) running node and the `[String]` is the node arguments (if any). `NodeOption` doesn't meant much.

PR #6580, opened 2026-05-26
cardano-node: remove iohk-monitoring tracing backend

jutaro:
## Summary

Removes the legacy `iohk-monitoring`-based tracing infrastructure from
`cardano-node`, leaving `trace-dispatcher` as the sole tracing backend.

`trace-dispatcher` has been the production default since node 8.x. The old
`iohk-monitoring` path was retained during a transition period but is no longer
used or tested in production deployments. Removing it eliminates ~11 000 lines
of dead code and 9 transitive build dependencies.

## What changed

**Deleted (~8 100 lines, 15 modules):**
- All of `Cardano.Tracing.*` — `Config`, `HasIssuer`, `Metrics`, `Render`,
  `Shutdown`, `Startup`, `Tracers`, and all `OrphanInstances.*` modules
  (Byron, Common, Consensus, HardFork, Network, Shelley)
- `Cardano.Node.Configuration.Logging` — the iohk-monitoring init layer
- `Cardano.Node.Tracing.Compat` — compatibility shim

**Modified:**
- `Cardano.Node.Run` — simplified to use only `initTraceDispatcher`; all
  legacy `createLoggingLayer` / `setupTrace` / `Cardano.BM.*` imports gone
- `Cardano.Node.TraceConstraints` — removes `HasIssuer`, `ToObject`, and
  related iohk-monitoring constraints
- `cardano-node.cabal` — drops `iohk-monitoring`, `lobemo-backend-*`,
  `lobemo-scribe-systemd`, `ekg-wai`, `tracer-transformers`, `scientific`
- Nix service config — removes legacy logging-layer options
- Benchmark profiles — `"tracing_backend"` updated to `"trace-dispatcher"`

## Breaking change

Any node config still using `"UseTraceDispatcher": false` or legacy
logging config keys will need to be migrated to `trace-dispatcher`
configuration before upgrading. See:

1. [Cardano Developer Portal: New tracing Quick Start](https://developers.cardano.org/docs/get-started/infrastructure/node/new-tracing-system/new-tracing-system/#node-side-configuration-of-new-tracing)
2. [trace-dispatcher documentation: configuration](https://github.com/IntersectMBO/hermod-tracing/blob/master/trace-dispatcher/doc/trace-dispatcher.md#configuration)
3. [trace-dispatcher: sample config](https://github.com/IntersectMBO/hermod-tracing/blob/master/trace-dispatcher/doc/config.yaml)
4. [cardano-node: mainnet sample config](https://github.com/IntersectMBO/cardano-node/blob/master/configuration/cardano/mainnet-config.yaml)

# Config options removed with iohk-monitoring

The following node configuration keys have no effect after the removal of the
`iohk-monitoring` tracing backend. All tracing is now handled exclusively by
`trace-dispatcher`.

## Parsed by the node but silently ignored

These three keys are still accepted by the JSON parser to avoid hard failures
on existing configs, but have no effect:

| Key | Previous effect |
|---|---|
| `TurnOnLogging` | Master on/off switch for the iohk-monitoring logging layer |
| `TurnOnLogMetrics` | Enabled/disabled node metrics collection via iohk-monitoring |
| `UseTraceDispatcher` | Chose between legacy (`false`) and trace-dispatcher (`true`) backend |

## Parsed by `Cardano.Tracing.Config` (module deleted)

| Key | Previous effect |
|---|---|
| `TracingVerbosity` | Rendering verbosity: `MinimalVerbosity` / `NormalVerbosity` / `MaximalVerbosity` |
| `TraceAcceptPolicy` | Per-tracer on/off toggle (bool) |
| `TraceBackingStore` | " |
| `TraceBlockFetchClient` | " |
| `TraceBlockFetchDecisions` | " |
| `TraceBlockFetchProtocol` | " |
| `TraceBlockFetchProtocolSerialised` | " |
| `TraceBlockFetchServer` | " |
| `TraceBlockchainTime` | " |
| `TraceChainDB` / `TraceChainDb` | " |
| `TraceChainSyncClient` | " |
| `TraceChainSyncBlockServer` | " |
| `TraceChainSyncHeaderServer` | " |
| `TraceChainSyncProtocol` | " |
| `TraceChurnMode` | " |
| `TraceConnectionManager` | " |
| `TraceConnectionManagerCounters` | " |
| `TraceConnectionManagerTransitions` | " |
| `TraceCsj` | " |
| `TraceDevotedBlockFetch` | " |
| `TraceDiffusionInitialization` | " |
| `TraceDNS` / `TraceDnsResolver` / `TraceDnsSubscription` | " |
| `TraceErrorPolicy` / `TraceLocalErrorPolicy` | " |
| `TraceForge` / `TraceForgeStateInfo` | " |
| `TraceGDD` | " |
| `TraceGsm` | " |
| `TraceHandshake` / `TraceLocalHandshake` | " |
| `TraceInboundGovernor` | " |
| `TraceInboundGovernorCounters` | " |
| `TraceInboundGovernorTransitions` | " |
| `TraceIpSubscription` | " |
| `TraceKeepAliveClient` / `TraceKeepAliveProtocol` | " |
| `TraceKesAgent` | " |
| `TraceLedgerPeers` | " |
| `TraceLocalChainSyncProtocol` | " |
| `TraceLocalConnectionManager` | " |
| `TraceLocalMux` / `TraceLocalMuxBearer` / `TraceLocalMuxChannel` | " |
| `TraceLocalRootPeers` | " |
| `TraceLocalServer` | " |
| `TraceLocalStateQueryProtocol` | " |
| `TraceLocalTxMonitorProtocol` | " |
| `TraceLocalTxSubmissionProtocol` / `TraceLocalTxSubmissionServer` | " |
| `TraceMempool` | " |
| `TraceMux` / `TraceMuxBearer` / `TraceMuxChannel` | " |
| `TracePeerSelection` / `TracePeerSelectionActions` / `TracePeerSelectionCounters` | " |
| `TracePeerSharingProtocol` | " |
| `TracePublicRootPeers` | " |
| `TraceSanityCheckIssue` | " |
| `TraceServer` | " |
| `TraceTxCounters` | " |
| `TraceTxInbound` / `TraceTxOutbound` | " |
| `TraceTxLogic` | " |
| `TraceTxSubmissionProtocol` / `TraceTxSubmission2Protocol` | " |

## Read directly by the `iohk-monitoring` library

These keys were consumed by `Cardano.BM.Configuration` — the node never parsed
them itself. They are now entirely ignored:

| Key | Previous effect |
|---|---|
| `minSeverity` | Global severity filter (`Debug` / `Info` / `Notice` / `Warning` / `Error` / `Critical`) |
| `rotation` | Log file rotation (`rpLogLimitBytes`, `rpKeepFilesNum`, `rpMaxAgeHours`) |
| `setupBackends` | Which backends to initialise (`KatipBK`, `EKGViewBK`, etc.) |
| `defaultBackends` | Default backend routing for namespaces |
| `setupScribes` | Log output targets (file, stdout, journald, trace-forwarder) |
| `defaultScribes` | Default scribe routing |
| `hasEKG` | Port for EKG web metrics interface |
| `hasPrometheus` | Host/port for Prometheus metrics endpoint |
| `options.mapBackends` | Per-namespace backend routing overrides |
| `options.mapSubtrace` | Per-namespace subtrace behaviour overrides |
| `options.mapSeverity` | Per-namespace severity filter overrides |

## Replacements in trace-dispatcher

The `trace-dispatcher` backend has its own equivalents for severity filtering
and output routing, configured under `TraceOptions` in the trace-dispatcher
config file. Refer to the trace-dispatcher documentation for the current config
format.


jutaro:
The change in [Byron.hs:122] is:

Before:
( "txid" .= txId tx )
which relied on instance ToJSON (TxId (GenTx ByronBlock)) from the now-deleted Cardano.Tracing.OrphanInstances.Byron, which was:

instance ToJSON (TxId (GenTx ByronBlock)) where
  toJSON = String . Text.take 8 . renderTxId
After:

( "txid" .= (Text.take 8 . renderTxId $ txId tx) )
The output is identical — both produce an 8-character hex string. The change just inlines the behavior that used to live in the deleted orphan instance, because that instance no longer exists.

cardano-node/src/Cardano/Node/Tracing/Documentation.hs
```diff
@@ -265,14 +269,6 @@ docTracersFirstPhase condConfigFileName = do
     configureTracers configReflection trConfig  [shutdownTr]
     shutdownTrDoc <- documentTracer (shutdownTr :: Logging.Trace IO ShutdownTrace)
 
-    chainDBTr <- mkCardanoTracer'
```
mgmeier, line 268: Why was this removed?

cardano-node/src/Cardano/Node/Tracing/Documentation.hs
```diff
@@ -412,13 +408,6 @@ docTracersFirstPhase condConfigFileName = do
     consensusStartupErrorTrDoc <- documentTracer (consensusStartupErrorTr ::
       Logging.Trace IO ConsensusStartupException)
 
-    consensusGddTr <- mkCardanoTracer
```
mgmeier, line 415: Why was this removed?

cardano-node/src/Cardano/Node/Tracing/Tracers.hs
```diff
@@ -58,12 +61,19 @@ import qualified Ouroboros.Network.Diffusion as Diffusion
...
 import           Network.Mux.Trace (TraceLabelPeer (..))
 import qualified Network.Mux.Trace as Mux
 import           Network.Mux.Tracing ()
 
+instance ToJSON NodeToNodeVersionData where
+  toJSON = String . Text.pack . show
+
+instance ToJSON NodeToClientVersionData where
+  toJSON = String . Text.pack . show
+
```
mgmeier, line 71: Isn't there a `ToJSON` for both of the above in `cardano-diffusion/orphan-instances/Cardano/Network/OrphanInstances.hs`?
Using `Show` here seems to lose structure.

cardano-node/src/Cardano/Node/Tracing/Tracers.hs
```diff
@@ -316,11 +326,6 @@ mkConsensusTracers configReflection trBase trForward mbTrEKG _trDataPoint trConf
                 ["Consensus", "Startup"]
     configureTracers configReflection trConfig [consensusStartupErrorTr]
 
-    !consensusGddTr <- mkCardanoTracer
```
mgmeier, line 319: Every `TraceGDDEvent` is silently discarded. This makes Genesis Density Disconnection events completely invisible

cardano-node/src/Cardano/Node/Configuration/POM.hs
```diff
@@ -352,17 +341,9 @@ instance FromJSON PartialNodeConfiguration where
...
-      pncTraceConfig     <-  if pncLoggingSwitch'
-                             then do
-                               partialTraceSelection <- parseJSON $ Object v
-                               if useTraceDispatcher
-                               then return $ Last $ Just $ PartialTraceDispatcher partialTraceSelection
-                               else return $ Last $ Just $ PartialTracingOnLegacy partialTraceSelection
-                             else return $ Last $ Just PartialTracingOff
+      _turnOnLogging :: Maybe Bool <- v .:? "TurnOnLogging"
+      _turnOnLogMetrics :: Maybe Bool <- v .:? "TurnOnLogMetrics"
+      _useTraceDispatcher :: Maybe Bool <- v .:? "UseTraceDispatcher"
```
mgmeier: These can go, right?

cardano-node/src/Cardano/Node/Tracing/Tracers/Startup.hs
```diff
@@ -129,23 +129,21 @@ getStartupInfo nc (SomeConsensusProtocol whichP pForInfo) fp = do
...
                    ]
 
-instance ToJSON blkVersion => ToJSON (ConsensusNetworkVersionTuple NodeToNodeVersion blkVersion) where
+instance Show blkVersion => ToJSON (ConsensusNetworkVersionTuple NodeToNodeVersion blkVersion) where
     toJSON (ConsensusNetworkVersionTuple nodeToClientVersion blockVersion) =
-      Aeson.object [ "nodeToNodeVersion" .= nodeToClientVersion
-                   , "blockVersion" .= blockVersion
+      Aeson.object [ "nodeToNodeVersion" .= String (pack $ show nodeToClientVersion)
+                   , "blockVersion" .= String (pack $ show blockVersion)
                    ]
```
mgmeier, line 143: Doesn't that change the type from `Aeson.Number` to `Aeson.String`? There should be serializations in `cardano-diffusion/orphan-instances/Cardano/Network/OrphanInstances.hs` for both of the above.

cardano-node/src/Cardano/Node/Tracing/Tracers/Startup.hs
```diff
@@ -233,9 +231,9 @@ instance ( Show (BlockNodeToNodeVersion blk)
                , "message" .= String msg ]
   forMachine _dtal (NetworkConfig localRoots publicRoots useLedgerPeers peerSnapshotFileMaybe) =
       mconcat [ "kind" .= String "NetworkConfig"
-               , "localRoots" .= toJSON localRoots
-               , "publicRoots" .= toJSON publicRoots
-               , "useLedgerAfter" .= useLedgerPeers
+               , "localRoots" .= String (showT localRoots)
+               , "publicRoots" .= String (showT publicRoots)
+               , "useLedgerAfter" .= String (showT useLedgerPeers)
```
mgmeier: Again, I would expect there to be a `ToJSON` in network; using `Show` here seems to lose log structure.

cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
```diff
@@ -91,21 +90,13 @@ withAddedToCurrentChainEmptyLimited tr = do
...
           , LedgerSupportsProtocol blk
-          , InspectLedger blk
-          , HasIssuer blk
-          , LogFormatting (ReasonForSwitch (TiebreakerView (BlockProtocol blk)))
-
           ) => LogFormatting (ChainDB.TraceEvent blk) where
   forHuman ChainDB.TraceLastShutdownUnclean        =
     "ChainDB is not clean. Validating all immutable chunks"
-  forHuman (ChainDB.TraceAddBlockEvent v)          = forHuman v
+  forHuman ChainDB.TraceAddBlockEvent{}           = "ChainDB add block event"
```
mgmeier: There's no dispatch happening anymore.
This seems critical.

cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
```diff
@@ -131,8 +122,8 @@ instance (  LogFormatting (Header blk)
                  RisingEdge -> "risingEdge" .= True
                  FallingEdgeWith pt -> "fallingEdge" .= forMachine dtal pt
              ]
-  forMachine details (ChainDB.TraceAddBlockEvent v) =
-    forMachine details v
+  forMachine _details ChainDB.TraceAddBlockEvent{} =
+    mconcat [ "kind" .= String "AddBlockEvent" ]
```
mgmeier: There's no dispatch happening anymore.
This seems critical.

cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
```diff
@@ -159,7 +150,7 @@ instance (  LogFormatting (Header blk)
 
   asMetrics ChainDB.TraceLastShutdownUnclean         = []
   asMetrics (ChainDB.TraceChainSelStarvationEvent _) = []
-  asMetrics (ChainDB.TraceAddBlockEvent v)           = asMetrics v
+  asMetrics ChainDB.TraceAddBlockEvent{}            = []
```
mgmeier: There's no dispatch happening anymore.
This seems critical.

cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
```diff
@@ -616,8 +600,7 @@ instance ( LogFormatting (Header blk)
             ++ [ "events" .= toJSON (map (forMachine DDetailed) events)
                | not (null events) ]
             ++ [ "tipBlockHash" .= tipBlockHash
-               , "tipBlockParentHash" .= tipBlockParentHash
-               , "tipBlockIssuerVKeyHash" .= tipBlockIssuerVkHashText]
```
mgmeier, line 620: The `"tipBlockIssuerVKeyHash"` key is missing after that change.

cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
```diff
@@ -650,8 +627,7 @@ instance ( LogFormatting (Header blk)
...
```
mgmeier, line 654: The `"tipBlockIssuerVKeyHash"` key is missing after that change.

cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
```diff
@@ -705,37 +681,25 @@ instance ( LogFormatting (Header blk)
...
-            BlockIssuerVerificationKeyHash bs ->
-              Text.decodeLatin1 (B16.encode bs)
     in  [ DoubleM "density" (fromRational density)
         , IntM    "slotNum" (fromIntegral slots)
         , IntM    "blockNum" (fromIntegral blocks)
         , IntM    "slotInEpoch" (fromIntegral slotInEpoch)
         , IntM    "epoch" (fromIntegral (unEpochNo epoch))
         , PrometheusM "tipBlock" [("hash",tipBlockHash)
-                                 ,("parent_hash",tipBlockParentHash)
-                                 ,("issuer_verification_key_hash", tipBlockIssuerVkHashText)]
```
mgmeier, line 738: The `issuer_verification_key_hash` metric is missing after the change.

cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
```diff
@@ -1182,21 +1146,24 @@ instance MetaTrace (ChainDB.TraceGCEvent blk) where
...
     forHuman ChainDB.InitialChainSelected{} =
         "Initial chain selected"
     forHuman ChainDB.StartedInitChainSelection {} =
         "Started initial chain selection"
 
-    forMachine dtal (ChainDB.InitChainSelValidation v) = forMachine dtal v
+    forMachine _dtal ChainDB.InitChainSelValidation{} =
+      mconcat
+        [ "kind" .= String "Follower.InitChainSelValidation"
+        ]
```
mgmeier: Dispatch is missing.
This seems critical.

cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
```diff
@@ -1182,21 +1146,24 @@ instance MetaTrace (ChainDB.TraceGCEvent blk) where
 -- -- TraceInitChainSelEvent
 -- --------------------------------------------------------------------------------
 
-instance (ConvertRawHash blk, LedgerSupportsProtocol blk)
+instance LedgerSupportsProtocol blk
   => LogFormatting (ChainDB.TraceInitChainSelEvent blk) where
-    forHuman (ChainDB.InitChainSelValidation v) = forHuman v
+    forHuman ChainDB.InitChainSelValidation{} = "Initial chain selection validation"
```
mgmeier: Dispatch is missing.
This seems critical.

cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
```diff
@@ -1182,21 +1146,24 @@ instance MetaTrace (ChainDB.TraceGCEvent blk) where
...
+      mconcat
+        [ "kind" .= String "Follower.InitChainSelValidation"
+        ]
     forMachine _dtal ChainDB.InitialChainSelected =
       mconcat ["kind" .= String "Follower.InitialChainSelected"]
     forMachine _dtal ChainDB.StartedInitChainSelection =
       mconcat ["kind" .= String "Follower.StartedInitChainSelection"]
 
-    asMetrics (ChainDB.InitChainSelValidation v) = asMetrics v
+    asMetrics ChainDB.InitChainSelValidation{} = []
```
mgmeier: Dispatch is missing.
This seems critical.

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
```diff
@@ -85,6 +84,10 @@ import           Data.Time (NominalDiffTime)
 import           Data.Word (Word32, Word64)
 import           Network.TypedProtocol.Core
 
+enclosingValue :: ToJSON a => Enclosing' a -> Value
+enclosingValue RisingEdge = object [ "edge" .= String "Starting" ]
+enclosingValue (FallingEdgeWith a) = object [ "edge" .= toJSON a ]
+
```
mgmeier, line 91: Defined elsewhere as `enclosingObject`.
Isn't there a way to keep this generic, like with the `EnclosedTime` class instance before?

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
```diff
@@ -231,7 +234,7 @@ instance (ConvertRawHash blk, LedgerSupportsProtocol blk)
     TraceDownloadedHeader h ->
       mconcat
         [ "kind" .= String "DownloadedHeader"
-        , tipToObject (tipFromHeader h)
+        , "header" .= String (showT $ headerPoint h)
```
mgmeier: Using the `Show` instance seems suspicious - wasn't it an Object before`

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
```diff
@@ -658,15 +661,13 @@ instance MetaTrace (TraceDecisionEvent peer (Header blk)) where
...
       => LogFormatting (TraceDecisionEvent peer (Header blk)) where
   forHuman = Text.pack . show
 
-  forMachine dtal (PeersFetch xs) =
+  forMachine _dtal (PeersFetch xs) =
     mconcat [ "kind" .= String "PeerFetch"
-            , "decisions" .= if dtal >= DMaximum
-                               then toJSON (Verbose <$> xs)
-                               else toJSON xs
+            , "decisions" .= List.foldl' (\acc x -> forMachine DDetailed x : acc) [] xs
```
mgmeier: Are you sure this produces the same output?
There's already a mismatch in detail levels.

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
```diff
@@ -792,7 +790,7 @@ instance (HasHeader header, ConvertRawHash header) =>
               , "block" .= String
                 (case pt of
                   GenesisPoint -> "Genesis"
-                  BlockPoint _ h -> renderHeaderHash (Proxy @header) h)
+                  BlockPoint slot _ -> "slot-" <> showT slot)
```
mgmeier: This is completely wrong.
The output must look like
```
...
"ns":"BlockFetch.Client.CompletedBlockFetch","data":{"block":"00c854f75889b72e14e59e7ed891a467f1b21e28378b092efc786c63b769f67f", ...}

File: cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs [hunk 21]
```diff
@@ -769,12 +770,9 @@ instance (HasHeader header, ConvertRawHash header) =>
     forMachine dtal (BlockFetch.SendFetchRequest af gsv) =
       mconcat $
               [ "kind" .= String "SendFetchRequest"
-              , "head" .= String (renderChainHash
-                                  (renderHeaderHash (Proxy @header))
-                                  (AF.headHash af))
```
mgmeier, line 774: The `head` key is missing after the change.

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
```diff
@@ -769,12 +770,9 @@ instance (HasHeader header, ConvertRawHash header) =>
...
               , "length" .= toJSON (fragmentLength' af)]
               ++
-              [ "deltaq" .= toJSON gsv | dtal >= DDetailed ]
+              [ "deltaq" .= String (Text.pack $ show gsv) | dtal >= DDetailed ]
```
mgmeier: The field has a `ToJSON` in network, IIRC. It's an Object, not a String.

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
```diff
@@ -1141,7 +1139,7 @@ instance
   forMachine dtal (TraceMempoolManuallyRemovedTxs txs0 txs1 mpSz) =
     mconcat
       [ "kind" .= String "TraceMempoolManuallyRemovedTxs"
-      , "txsRemoved" .= txs0
+      , "txsRemoved" .= map (String . Text.pack . show) (toList txs0)
```
mgmeier: This is suspicious. It might result in the same output, it may not.
This needs scrutiny.

mgmeier (changes requested):

cardano-node/src/Cardano/Node/Tracing/Documentation.hs
jutaro: Fixed. chainDBTr was removed accidentally because HasIssuer — which the ChainDB tracer depends on — was incorrectly deleted as part of the iohk-monitoring cleanup. HasIssuer is needed code (it computes the block issuer VKey hash for ChainDB events). It has been moved to Cardano.Node.Tracing.Tracers.HasIssuer and chainDBTr has been restored.

cardano-node/src/Cardano/Node/Tracing/Documentation.hs
jutaro: Fixed. consensusGddTr was also removed accidentally (see next comment). The tracer block has been restored in both Documentation.hs and Tracers.hs

cardano-node/src/Cardano/Node/Tracing/Tracers.hs
jutaro: I searched cardano-diffusion for ToJSON NodeToNodeVersionData and ToJSON NodeToClientVersionData but found no existing instances. If you can point to the exact location in cardano-diffusion/orphan-instances/Cardano/Network/OrphanInstances.hs, I'll replace the local orphan instances with an import instead. For now the Show-based instances keep the build working; happy to switch once the location is confirmed.

cardano-node/src/Cardano/Node/Tracing/Tracers.hs
jutaro: Fixed. The consensusGddTr tracer has been restored with mkCardanoTracer and properly wired up via traceWith consensusGddTr. GDD events are fully visible again.

cardano-node/src/Cardano/Node/Configuration/POM.hs
jutaro: Agreed. All three (_turnOnLogging, _turnOnLogMetrics, _useTraceDispatcher) have been removed.

cardano-node/src/Cardano/Node/Tracing/Tracers/Startup.hs
jutaro: Fixed. The ConsensusNetworkVersionTuple instances have been reverted to use ToJSON blkVersion => constraints with direct .= nodeToClientVersion / .= blockVersion, preserving the original JSON structure.

cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
jutaro: Fixed. The root cause was that Cardano.Tracing.HasIssuer was incorrectly deleted — it is pure trace-dispatcher functionality (computing the block issuer VKey hash), not iohk-monitoring code. It has been moved to Cardano.Node.Tracing.Tracers.HasIssuer. With the HasIssuer blk constraint restored, dispatch has been fully restored for forHuman, forMachine, and asMetrics on TraceAddBlockEvent.

cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
jutaro: Fixed. tipBlockIssuerVerificationKeyHash has been restored to the ChainInformation record, HasIssuer blk added back to chainInformation, and the tipBlockIssuerVkHashText let-binding and "tipBlockIssuerVKeyHash" field restored in both DDetailed branches of AddedToCurrentChain and SwitchedToAFork.

cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
jutaro: Fixed. The PrometheusM "tipBlock" calls now include ("issuer_verification_key_hash", tipBlockIssuerVkHashText) for AddedToCurrentChain and ("issuer_VKey_hash", tipBlockIssuerVkHashText) for SwitchedToAFork, matching master.

cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
jutaro: Fixed

cardano-node/src/Cardano/Node/Tracing/Tracers/ChainDB.hs
jutaro: Fixed

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
jutaro: The local enclosingValue was introduced because enclosingObject is not re-exported from Cardano.Logging. The semantics are identical.

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
jutaro: Fixed. TraceDownloadedHeader now uses tipToObject (tipFromHeader h) as in master, restoring the structured Object output.

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
jutaro: This matches master: the outer match uses the passed _dtal but the inner forMachine DDetailed x calls are hard-coded to DDetailed in master too. No behaviour change here.

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
jutaro: Fixed. CompletedBlockFetch now renders the block hash using renderHeaderHash (Proxy @header) h as in master. The ConvertRawHash header constraint has been restored on the instance.

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
jutaro: Fixed. The "head" .= String (renderChainHash (renderHeaderHash (Proxy @header)) (AF.headHash af)) field has been restored in SendFetchRequest.

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
jutaro: Fixed. deltaq now uses toJSON gsv as in master. If there is a specific ToJSON instance in cardano-diffusion / network that should be in scope here, please point to it — but the existing instance should produce the correct Object output.

cardano-node/src/Cardano/Node/Tracing/Tracers/Startup.hs
mgmeier: I can't see any change to the `Startup.hs` module in your follow-up commit.

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
mgmeier: I do not see that change in your follow-up commit.

cardano-node/src/Cardano/Node/Tracing/Tracers.hs
mgmeier: https://github.com/IntersectMBO/ouroboros-network/blob/d842a238a9ac329a4bb676dd95e46ee25a7a1a94/cardano-diffusion/orphan-instances/Cardano/Network/OrphanInstances.hs#L102-L115

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
mgmeier: So - what's the motivation for it...? Can we not leave this location completely unchanged?

cardano-node/src/Cardano/Node/Tracing/StateRep.hs
```diff
@@ -49,6 +50,24 @@ deriving instance ToJSON ChunkNo
...
+
+instance FromJSON NPV.NodeToClientVersion where
+  parseJSON = parseBoundedEnum "NodeToClientVersion"
+
+parseBoundedEnum :: (Bounded a, Enum a, Show a) => String -> Value -> Parser a
+parseBoundedEnum name = withText name $ \value ->
+  case filter ((== value) . Text.pack . show) [minBound .. maxBound] of
+    [version] -> pure version
+    _ -> fail $ "Invalid " <> name <> ": " <> Text.unpack value
+
```
mgmeier, line 52: Please use https://github.com/IntersectMBO/ouroboros-network/blob/d842a238a9ac329a4bb676dd95e46ee25a7a1a94/cardano-diffusion/orphan-instances/Cardano/Network/OrphanInstances.hs#L65-L100

cardano-node/src/Cardano/Node/Configuration/TopologyP2P.hs
```diff
@@ -48,6 +56,48 @@ instance AdjustFilePaths CardanoNetworkTopology where
...
+      <*> v .:? "bootstrapPeers" .!= DontUseBootstrapPeers
+
+peerAdvertiseFromBool :: Bool -> PeerAdvertise
+peerAdvertiseFromBool True = DoAdvertisePeer
+peerAdvertiseFromBool False = DoNotAdvertisePeer
+
+peerTrustableFromBool :: Bool -> PeerTrustable
+peerTrustableFromBool True = IsTrustable
+peerTrustableFromBool False = IsNotTrustable
+
```
mgmeier, line 51: Please have a look again at `cardano-diffusion`. There seems to be some redudancy.

mgmeier (changes requested):

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
mgmeier: I think the original ToJSON might have been based on some call to `renderTxId`...?
I don't think it was `Show` based at all.

cardano-node/src/Cardano/Node/Tracing/Tracers/Startup.hs
jutaro: The change is in commit d4b85596a. ConsensusNetworkVersionTuple now carries ToJSON blkVersion => and the fields use .= nodeToClientVersion / .= nodeToNodeVersion directly — the structured Number output comes from Cardano.Network.OrphanInstances, brought into scope via the empty import added to Startup.hs in that commit.

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
jutaro: The current code matches master exactly — the outer match uses the passed dtal, and the inner forMachine DDetailed x calls are hard-coded to DDetailed in master too. This was an accidental reformatting during the cleanup; the output is identical. Happy to revert to the exact master whitespace/structure if that helps review.

cardano-node/src/Cardano/Node/Configuration/TopologyP2P.hs
jutaro: Done in d4b85596a. Removed the local FromJSON/ToJSON instances for PeerTrustable, UseBootstrapPeers, and NetworkTopology UseBootstrapPeers PeerTrustable. Cardano.Network.OrphanInstances () is imported to bring those in scope.

cardano-node/src/Cardano/Node/Tracing/StateRep.hs
jutaro: Done in d4b85596a. The local ToJSON/FromJSON instances for NodeToNodeVersion and NodeToClientVersion have been removed. Cardano.Network.OrphanInstances () is imported, which provides the instances at L65–100.

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
jutaro: Fixed in latest PR

cardano-node/src/Cardano/Node/Tracing/Tracers/Startup.hs
jutaro: You're correct. Master uses .= toJSON localRoots, .= toJSON publicRoots, and .= useLedgerPeers directly — all three types have ToJSON instances in ouroboros-network/cardano-diffusion (HotValency, WarmValency, RelayAccessPoint, CardanoLocalRootConfig, PeerAdvertise, UseLedgerPeers). Our branch introduced showT unnecessarily. Reverted lines 235–237 to the master formulation.

mgmeier (approved):
Voila, finally, the great moment is upon us.
LGTM!

cardano-testnet/src/Testnet/Defaults.hs
```diff
@@ -263,71 +257,6 @@ defaultYamlHardforkViaConfig sbe =
                 , ("TestDijkstraHardForkAtEpoch", Aeson.Number 0)
                 ]
                 )
-  -- | Various tracers we can turn on or off
```
carbolymer, line 284: How does one configure tracing now? What are the defaults?

For example,  I was using snippet below to enable `PrometheusSimple` endpoint - how do I do it now?

>The trace-dispatcher backend has its own equivalents for severity filtering
and output routing, configured under TraceOptions in the trace-dispatcher
config file. Refer to the trace-dispatcher documentation for the current config
format.

Few more words how to configure tracing now w/ links would be greatly appreciated.

mgmeier: Thanks @carbolymer 
1. I think that particular snippet was removed without proper motivation. Reverted in the feedback commit below. Please quickly check that it restores your use case.
2. The material describing new tracing config is probably too extensive for the PR. However, it's all linked now above in the PR description.

carbolymer: Thanks, that should be enough.

carbolymer (approved):

cardano-node/cardano-node.cabal
```diff
@@ -164,38 +148,30 @@ library
...
                       , directory
                       , dns
-                      , ekg-wai
                       , ekg-core
                       , filepath
                       , generic-data
                       , hashable
                       , hostname
                       , io-classes:{io-classes,strict-stm,si-timers} ^>= 1.8
-                      , iohk-monitoring ^>= 0.2
```
Jimbo4350, line 174: 🥳 🚀

Jimbo4350 (approved):
LGTM! Sayonara `iohk-monitoring`!

PR #6584, opened 2026-05-28
cardano-recon: replace positional FILE/FILES args with named --formulas/--traces (v1.2.1)

Russoul:
## Summary

- Replace the positional `FILE` argument in `cardano-recon` with a named `--formulas FILE` option.
- Replace the positional `FILES` argument with a repeatable `--traces FILE` option (pass once per file).
- Improve the `--timeunit` help message to clearly describe how the value is used.
- Bump `cardano-recon-framework` version to 1.2.1 and add changelog entry.


jutaro (commented):
LGTM

mgmeier (approved):
LGTM

PR #6586, opened 2026-06-01
Move remaining path functions from `Defaults` module in `cardano-testnet` to `Cardano.Node.Testnet.Paths` in `cardano-node`

### LINKED ISSUE #6532: Consider moving some defaults from `Testnet.Defaults` to `cardano-node`
Look for other defaults in `Testnet.Defaults` and check if it makes sense to move them to `Cardano.Node.Testnet.Paths` in `cardano-node`.

_Originally posted by @Jimbo4350 in https://github.com/IntersectMBO/cardano-node/pull/6518#pullrequestreview-4105785311_

palas:
# Description

Closes #6532.

Move all path and filename definition functions from `Testnet.Defaults` (in `cardano-testnet`) to
`Cardano.Node.Testnet.Paths` (in `cardano-node`), so that consumers of generated testnet
configurations can reference canonical paths without depending on `cardano-testnet`.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [x] Self-reviewed the diff


cardano-node/src/Cardano/Node/Testnet/Paths.hs
```diff
@@ -4,66 +4,255 @@
 -- testnet configurations depend on this module so that directory
 -- layout changes are kept in sync at compile time.
 module Cardano.Node.Testnet.Paths
```
carbolymer, line 6: Suggestion: use singular nouns for modules
```suggestion
module Cardano.Node.Testnet.Path
```

cardano-testnet/src/Testnet/Defaults.hs
```diff
@@ -96,10 +117,19 @@ import           Data.Time (UTCTime)
 import           Data.Word (Word64)
 import           Lens.Micro
 import           Numeric.Natural
-import           System.FilePath ((</>))
-
 import           Test.Cardano.Ledger.Core.Rational
-import           Cardano.Node.Testnet.Paths (defaultNodeName, defaultNodeDataDir, defaultUtxoSKeyPath,
+import           Cardano.Node.Testnet.Paths (defaultCommitteeKeysDir, defaultCommitteeName,
```
carbolymer: I guess explicit imports don't make sense for this one? The functions are quite distinguishable imo.

carbolymer (approved):

cardano-testnet/src/Testnet/Defaults.hs
palas: Ok, they do all start with "default"

cardano-node/src/Cardano/Node/Testnet/Paths.hs
palas: I agree with the principle of having the types and the modules that have them singular, like in `Address`.

But, in this case, I am not sure it is a good idea, because we are providing many paths statically. And I think calling it `Path` would kind of convey that we are providing a `Path` datatype or something like that, or functions to handle paths in general.

To me, it is like an app would typically have a `File` menu (in singular), but it still haves the `Settings` option (in plural), `Setting` would sound weird.

Also the original module was called `Defaults` (in plural).

cardano-testnet/src/Testnet/Defaults.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/4b991a8da4091f59821d32968b3daa4324d058ba

PR #6590, opened 2026-06-03
Fix node config reading in gRPC config reload on SIGHUP

### LINKED ISSUE #6589: [BUG] - SIGHUP RPC config reload always fails with "Missing YAML config file"
**Internal/External**

**Area**

**Summary**
On `SIGHUP`, the node reloads its RPC configuration from the node config file.
This reload always fails with:

```
Error while updating RPC configuration: Missing YAML config file
```

logged at **Error** severity. The reload is non-fatal (it is silently skipped),
but every `SIGHUP` now emits a spurious `Error`-level trace.

This applies to the new tracing system.

**Steps to reproduce**
1. Start a node (new tracing enabled).
2. Send `SIGHUP` to the node process: `kill -HUP <pid>`.
3. Observe the node log.

Observed log line:

```
[<host>:Startup.RpcConfigUpdateError](Error,166) Error while updating RPC configuration: Missing YAML config file
```

**Expected behavior**
A `SIGHUP` RPC config reload from an otherwise valid config file should succeed
(or, if RPC is disabled, be a no-op) and not emit an `Error`-level trace.

- OS Name: Fedora 43
- Node version: 11.0.1

**Additional context**
The `SIGHUP` handler calls `updateRpcConfiguration`, passing the node config
file path (`ncConfigFile nc`):

`cardano-node/src/Cardano/Node/Run.hs`
```haskell
updateRpcConfiguration startupTracer (ncConfigFile nc) rpcConfigVar
```

`updateRpcConfiguration` re-parses the file and runs the full
`makeNodeConfiguration` validation:

`cardano-node/src/Cardano/Node/Run.hs`
```haskell
updateRpcConfiguration tracer configFilePath rpcConfigVar = do
  result <- fmap (join . first Exception.displayException)
              . try @Exception.SomeException
              . fmap makeNodeConfiguration
              . parseNodeConfigurationFP
              $ Just configFilePath
  case result of
    Left err ->
      -- reload failure, we don't do anything this time
      traceWith tracer (RpcConfigUpdateError $ pack err)
    ...
```

`parseNodeConfigurationFP` only decodes the file *contents*:

`cardano-node/src/Cardano/Node/Configuration/POM.hs`
```haskell
parseNodeConfigurationFP (Just (ConfigYamlFilePath fp)) = do
    nc <- decodeFileThrow fp
    pure $ adjustFilePaths (takeDirectory fp </>) nc
```

The `FromJSON PartialNodeConfiguration` instance never reads `pncConfigFile` —
the config file path is normally injected from the `--config` **CLI argument**
by the option parser, not stored inside the file. So after decoding,
`pncConfigFile == mempty`.

`makeNodeConfiguration` then requires `pncConfigFile` and fails:

`cardano-node/src/Cardano/Node/Configuration/POM.hs`
```haskell
makeNodeConfiguration pnc = do
  configFile <- lastToEither "Missing YAML config file" $ pncConfigFile pnc
  ...
```

→ `Left "Missing YAML config file"` → traced as `RpcConfigUpdateError` at
severity `Error`.

Because the config file path is structurally never present in the parsed file,
this validation can never pass on the reload path — the failure is
deterministic for any node.

carbolymer:
Fix gRPC reload issue, when the reload was failing due to missing node configuration file path. The fix propagates original CLI config to SIGHUP handler.

Closes: https://github.com/IntersectMBO/cardano-node/issues/6589

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/cardano-node/src/Cardano/Node/Run.hs
+++ b/cardano-node/src/Cardano/Node/Run.hs
@@ -265,7 +265,7 @@ handleNodeWithTracers cmdPc nc p@(SomeConsensusProtocol blockType runP) = do
                                   then DisabledBlockForging
                                   else EnabledBlockForging))
 
-  handleSimpleNode blockType runP tracers nc networkMagic
+  handleSimpleNode blockType runP tracers nc cmdPc networkMagic
     (\nk -> do
         setNodeKernel nodeKernelData nk
         traceWith (nodeStateTracer tracers) NodeKernelOnline)
@@ -305,13 +305,16 @@ handleSimpleNode
   -> Api.ProtocolInfoArgs blk
   -> Tracers RemoteAddress LocalAddress blk IO
   -> NodeConfiguration
+  -> PartialNodeConfiguration
+  -- ^ Original CLI configuration, used for SIGHUP config reload so CLI
+  -- overrides are preserved when re-reading the YAML file.
   -> NetworkMagic
   -> (NodeKernel IO RemoteAddress LocalConnectionId blk -> IO ())
   -- ^ Called on the 'NodeKernel' after creating it, but before the network
   -- layer is initialised.  This implies this function must not block,
   -- otherwise the node won't actually start.
   -> IO ()
-handleSimpleNode blockType runP tracers nc networkMagic onKernel = do
+handleSimpleNode blockType runP tracers nc cmdPc networkMagic onKernel = do
   logStartupWarnings
 
   logDeprecatedLedgerDBOptions
@@ -439,7 +442,7 @@ handleSimpleNode blockType runP tracers nc networkMagic onKernel = do
               (readTVar ledgerPeerSnapshotPathVar)
               (readTVar useLedgerVar)
               (writeTVar ledgerPeerSnapshotVar)
-            updateRpcConfiguration (startupTracer tracers) (ncConfigFile nc) rpcConfigVar
+            updateRpcConfiguration (startupTracer tracers) cmdPc rpcConfigVar
             traceWith (startupTracer tracers) (BlockForgingUpdate NotEffective)
           )
           Nothing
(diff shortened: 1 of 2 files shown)


cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -881,21 +885,18 @@ rpcServerLoop startupTracer rpcTracer rpcConfigVar networkMagic = go
...
 #ifdef UNIX
--- | Reload RPC configuration from the configuration file
-updateRpcConfiguration :: Tracer IO (StartupTrace blk) -- ^ tracer tracing the configuration reload
-                       -> ConfigYamlFilePath -- ^ node configuration file, to reload configuration from
-                       -> StrictTVar IO RpcConfig -- ^ TVar storing RPC configuration
+-- | Reload RPC configuration by re-reading the YAML config file and merging
+-- with CLI overrides, exactly like startup does via 'buildNodeConfiguration'.
+updateRpcConfiguration :: Tracer IO (StartupTrace blk)
+                       -> PartialNodeConfiguration
+                       -> StrictTVar IO RpcConfig
```
palas: This change lost the haddocks over the parameters

cardano-node/test/Test/Cardano/Node/POM.hs
```diff
@@ -303,6 +306,57 @@ eExpectedConfig = do
...
 
+-- A socket config with a node socket path, needed for RPC-enabled tests
+-- because makeRpcConfig validates that a node socket exists when RPC is enabled.
+testSocketConfigWithPath :: Last SocketConfig
+testSocketConfigWithPath =
+  Last . Just $ SocketConfig mempty mempty mempty (Last . Just $ File "node.socket")
+
+-- | CLI --grpc-enable + YAML silent -> RPC stays enabled.
+-- This is the exact regression scenario from issue #6589: a node started with
+-- --grpc-enable must not have RPC silently disabled on SIGHUP.
```
palas: This is technically not a regression test. But all these tests are very good documentation, and I don't think it is worth doing a proper regression test because it would be unnecessarily complex

palas (approved):
Nice 👍

cardano-node/src/Cardano/Node/Run.hs
```diff
@@ -881,21 +885,18 @@ rpcServerLoop startupTracer rpcTracer rpcConfigVar networkMagic = go
...
-              . fmap makeNodeConfiguration
-              . parseNodeConfigurationFP
-              $ Just configFilePath
+updateRpcConfiguration tracer cmdPc rpcConfigVar = do
+  result <- try @Exception.SomeException $ buildNodeConfiguration cmdPc
   case result of
     Left err ->
       -- reload failure, we don't do anything this time
-      traceWith tracer (RpcConfigUpdateError $ pack err)
+      traceWith tracer (RpcConfigUpdateError $ pack (Exception.displayException err))
```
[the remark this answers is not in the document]
carbolymer: not an issue

cardano-node/src/Cardano/Node/Run.hs
carbolymer: restored

Jimbo4350 (approved):

PR #6575, opened 2026-05-21
Bump cardano-api 11.3, cardano-cli 11.1, cardano-rpc 11.0, plutus 1.65 and CHaP

carbolymer:
## Description

Bump dependencies and CHaP index in cardano-node:

- **cardano-api** `^>= 11.0` to `^>= 11.3`
- **cardano-cli** `^>= 11.0` to `^>= 11.1`
- **cardano-rpc** `^>= 10.2` to `^>= 11.0`
- **plutus-ledger-api**, **plutus-tx**, **plutus-tx-plugin** `^>= 1.63` to `^>= 1.65` (in plutus-scripts-bench)
- **CHaP index-state** bumped to `2026-05-27`
- **CHaP flake input** updated

### Migration fixes

- **TxReferenceInputDatum test**: rewritten to use experimental API (`Exp.makeTransactionBodyAutoBalance`, `Exp.makeUnsignedTx`, `Exp.signTx`, `Exp.setTxSupplementalDatums`)
- **SearchUtxos test**: updated Test 3 to expect `GrpcInvalidArgument` for invalid address predicates (cardano-rpc 11.0 now rejects malformed addresses instead of returning empty)
- **cardano-submit-api**: replaced deprecated `getTxBody` with `Tx` pattern synonym
- **DRep.hs**, **Transaction.hs**: removed `Certificate` from `Cardano.Api` hiding list (no longer exported)

### Changelogs

- https://github.com/IntersectMBO/cardano-api/blob/master/cardano-api/CHANGELOG.md
- https://github.com/IntersectMBO/cardano-cli/blob/master/cardano-cli/CHANGELOG.md
- https://github.com/IntersectMBO/cardano-api/blob/master/cardano-rpc/CHANGELOG.md

## Checklist

- [x] Commit sequence broadly makes sense and commits have useful messages
- [x] New tests are added if needed and existing tests are updated
- [x] Any changes are noted in the `CHANGELOG.md` for affected package
- [x] The version bounds in `.cabal` files are updated
- [x] CI passes

PR #6601, opened 2026-06-17
Fix flaky ProposeNewConstitution test

### LINKED ISSUE #6593: Fix flaky ProposeNewConstitution test with per-test genesis override and two-round voting
## Summary

The `hprop_ledger_events_propose_new_constitution` test is flaky because CLI `query proposals` reads from the pulsing snapshot, which lags the live ledger state by one epoch.
Combined with a 0% `dvtUpdateToConstitution` threshold, any vote triggers immediate ratification, creating a tiny window where vote assertions fail.

## Root cause

Two facts interact to produce the race:

1. **Pulsing snapshot lag** - `query proposals` reads from the DRep pulsing snapshot, which is set once at the epoch boundary and not modified during pulsing. New proposals and votes only appear in this snapshot after the next epoch transition.
2. **Zero ratification threshold** - `dvtUpdateToConstitution` defaults to `0 %! 2` in `Testnet.Defaults`, meaning the proposal is ratified as soon as pulsing completes, regardless of votes.

On master, the test submits all 9 votes in a single transaction, then checks for ratification via `foldEpochState`, and only afterwards queries CLI proposals.
By this point the proposal may already have been ratified and removed from gov-state, causing `length proposalsArray === 1` to fail on an empty array.
The vote count assertions (`length proposalsDRepVotes === numVotes`) similarly fail when the snapshot has not yet refreshed or the proposal is gone.

## Current workaround (PR #6575)

PR #6575 reworked the test ordering - moving the CLI proposal-structure query before voting (race-free) and adding `unless` guards around the post-vote CLI assertions.
Reviewers (@Jimbo4350, @palas) correctly flagged the `unless` guards as weakening the test - they silently skip the vote-count checks whenever the snapshot is stale or the proposal is already ratified.

### A. Override Conway genesis per-test

Follow the pattern from the `NoConfidence` test: split `createAndRunTestnet` into `createTestnetEnv` + `H.rewriteJsonFile` + `cardanoTestnet`.
Override two parameters:

- `dvtUpdateToConstitution = 51 %! 100` - a 51% threshold, ensuring a minority of yes-votes cannot trigger ratification.
- `ucppGovActionLifetime = EpochInterval 4` - 80s lifetime at `genesisEpochLength = 200`, giving ample time for snapshot refresh. The master default of `EpochInterval 1` (20s) is too tight.

No committee setup is needed.
With `ucppCommitteeMinSize = 0` (the default), `committeeAccepted` returns `True` vacuously.

### B. Stage votes in two rounds

- **Round 1:** Submit 3 yes votes out of 9 DReps = 33.3% of active stake. This is below the 51% threshold, so the proposal stays in gov-state indefinitely.
- **Round 2:** Submit +1 yes, +3 no, +2 abstain. The new ratio is 4 yes out of 7 participating (abstain DReps are excluded from the denominator) = 57.1%, which exceeds 51% and triggers ratification.

### C. Deterministic assertions - no `unless` guards

- **After round 1:** Use `retryUntilJustM` until the pulsing snapshot shows the proposal with exactly 3 DRep votes. Assert vote counts unconditionally. The proposal cannot be ratified at 33.3% < 51%, so it will always be present.
- **After round 2:** Use `retryUntilJustM` until the snapshot shows the full vote counts (4 yes, 3 no, 2 abstain). Assert unconditionally, then verify ratification via the existing `foldEpochState` check.

### D. Technical notes

- Vote ratio is stake-weighted: non-voting DReps count in the denominator, abstain DReps are excluded.
- The pulsing snapshot's proposals map is set once at the epoch boundary and not modified mid-epoch - only the vote tallies are updated as pulsing progresses.
- The `retryUntilJustM` guard condition must match the expected postcondition exactly (proposal present AND vote count matches), otherwise the retry exits early and the assertion fails.

## Acceptance criteria

- [ ] The test runs without `unless` guards or other conditional skips around assertions.
- [ ] CLI `query proposals` vote counts are checked unconditionally in both rounds.
- [ ] The test passes reliably in CI (no flaky failures over 10+ runs).
- [ ] No changes to `Testnet.Defaults` - all overrides are local to this test.

carbolymer:
## Summary

- Raise `dvtUpdateToConstitution` to 51% (from the default 0%) via Conway genesis override, so a minority of yes-votes cannot trigger instant ratification.
- Split voting into two rounds: round 1 (3 yes = 33%, below threshold) keeps the proposal alive for deterministic ledger and CLI assertions; round 2 (+1 yes, +3 no, +2 abstain = 4/7 = 57.1%) crosses the threshold and triggers ratification.
- Replace `waitForGovActionVotes` + immediate assertions with `retryUntilJustM` polling that exits only when the exact expected vote count appears in both ledger state and CLI output.
- Extend `govActionLifetime` to 4 epochs and ratification timeout to epoch 20 so the proposal survives the pulsing snapshot refresh cycle.
- Move `query proposals` assertion before voting so the proposal structure is verified while it has zero votes (deterministic).

## Root cause

The default `dvtUpdateToConstitution = 0%` means any single DRep yes-vote immediately satisfies the ratification threshold.
Combined with the pulsing snapshot lag (the snapshot only picks up new votes at epoch boundaries), the window where the proposal exists with votes but is not yet ratified is extremely narrow.
The old test submitted all 9 votes at once and then asserted on the transient "proposal with votes" state, which was a race condition.

## Approach

Structure the test so that the states we are interested in are longer-lived (addressing [review feedback](https://github.com/IntersectMBO/cardano-node/pull/6575#discussion_r2093277083)):
1. Override Conway genesis to set a meaningful ratification threshold
2. Vote in rounds so the below-threshold state is stable for assertions
3. Use exact-count polling guards so assertions never fire on stale data

## Test plan

- [x] Test passes consistently with `DISABLE_RETRIES=1` (4/4 runs, ~75-96s each)
- [ ] CI green

### DIFF
--- a/cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Gov/ProposeNewConstitution.hs
+++ b/cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Gov/ProposeNewConstitution.hs
@@ -13,8 +13,10 @@ import           Cardano.Api.Experimental (Some (..), obtainCommonConstraints)
 import           Cardano.Api.Ledger (EpochInterval (..))
 
 import qualified Cardano.Crypto.Hash as L
+import qualified Cardano.Ledger.Conway.Genesis as L
 import qualified Cardano.Ledger.Conway.Governance as L
 import qualified Cardano.Ledger.Conway.Governance as Ledger
+import qualified Cardano.Ledger.Conway.PParams as L
 import qualified Cardano.Ledger.Hashes as L
 import qualified Cardano.Ledger.Shelley.LedgerState as L
 import           Cardano.Testnet
@@ -28,6 +30,7 @@ import qualified Data.Aeson.Lens as Aeson
 import           Data.Default.Class
 import           Data.Maybe
 import           Data.Maybe.Strict
+import           Data.Ratio ((%))
 import qualified Data.Text as Text
 import           Data.Text.Encoding (decodeUtf8)
 import qualified Data.Vector as Vector
@@ -48,6 +51,7 @@ import           Testnet.Process.Cli.Transaction
 import           Testnet.Process.Run (addEnvVarsToConfig, execCli', mkExecConfig)
 import           Testnet.Process.RunIO (liftIOAnnotated)
 import           Testnet.Property.Util (integrationRetryWorkspace)
+import           Testnet.Start.Cardano (liftToIntegration)
 import           Testnet.Start.Types
 import           Testnet.Types
 
@@ -66,12 +70,21 @@ hprop_ledger_events_propose_new_constitution = integrationRetryWorkspace 2 "prop
 
   work <- H.createDirectoryIfMissing $ tempAbsPath' </> "work"
 
-  -- Generate model for votes
-  let allVotes :: [(String, Int)]
-      allVotes = zip (concatMap (uncurry replicate) [(4, "yes"), (3, "no"), (2, "abstain")]) [1..]
-  annotateShow allVotes
+  -- Generate model for votes in two rounds.
+  -- The Conway DRep voting ratio is: yes / (yes + no + non-voters). Abstainers are excluded
+  -- from the denominator entirely. Non-voters count against (as implicit no).
+  -- Round 1: 3 yes, 6 non-voting. Ratio = 3/9 = 33.3%, below 51% threshold - proposal stays alive.
+  -- Round 2: +1 yes, +3 no, +2 abstain. Ratio = (3+1)/(9-2) = 57.1% (abstain excluded), triggers ratification.
+  let round1Votes :: [(String, Int)]
+      round1Votes = zip (replicate 3 "yes") [1..]
+  annotateShow round1Votes
 
-  let numVotes :: Int
+  let round2Votes :: [(String, Int)]
+      round2Votes = zip (concatMap (uncurry replicate) [(1, "yes"), (3, "no"), (2, "abstain")]) [4..]
+  annotateShow round2Votes
+
+  let allVotes = round1Votes ++ round2Votes
+      numVotes :: Int
       numVotes = length allVotes
   annotateShow numVotes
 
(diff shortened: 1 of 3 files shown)


cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Gov/ProposeNewConstitution.hs
```diff
@@ -66,12 +70,21 @@ hprop_ledger_events_propose_new_constitution = integrationRetryWorkspace 2 "prop
...
-  let allVotes :: [(String, Int)]
-      allVotes = zip (concatMap (uncurry replicate) [(4, "yes"), (3, "no"), (2, "abstain")]) [1..]
-  annotateShow allVotes
-
-  let numVotes :: Int
+  -- Generate model for votes in two rounds.
+  -- The Conway DRep voting ratio is: yes / (yes + no + non-voters). Abstainers are excluded
+  -- from the denominator entirely. Non-voters count against (as implicit no).
+  -- Round 1: 3 yes, 6 non-voting. Ratio = 3/9 = 33.3%, below 51% threshold - proposal stays alive.
+  -- Round 2: +1 yes, +3 no, +2 abstain. Ratio = 4/(4+3) = 57.1% (abstain excluded), triggers ratification.
```
palas:
```suggestion
  -- Round 2: +1 yes, +3 no, +2 abstain. Ratio = (3+1)/(9-2) = 57.1% (abstain excluded), triggers ratification.
```
It took me way too long to understand that we were aggregating votes of both rounds, and I think the "4+3" was part of the cause of the confusion, because in round 1 we are not adding yes and no votes, so it doesn't make sense to do it in round 2 either, what we are really doing is removing the abstentions from the total of 9, or counting the non-abstentions assuming everyone voted

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Gov/ProposeNewConstitution.hs
```diff
@@ -334,9 +328,112 @@ hprop_ledger_events_propose_new_constitution = integrationRetryWorkspace 2 "prop
...
+     voteTxFp <- signTx execConfig cEra gov "round1-signed-vote-tx" voteTxBodyFp signingKeys
+     submitTx execConfig cEra voteTxFp
+
+     waitForGovActionVotes epochStateView (EpochInterval 1)
+
+     -- Verify votes in ledger state
+     govState <- getGovState epochStateView ceo
+     govActionState <- H.headM $ govState ^. L.cgsProposalsL . L.pPropsL . to toList
+     let votes = govActionState ^. L.gasDRepVotesL . to toList
+     length (filter ((== L.VoteYes) . snd) votes) === 3
```
palas, line 348: Why not check the other types of vote too

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Gov/ProposeNewConstitution.hs
```diff
@@ -334,9 +328,112 @@ hprop_ledger_events_propose_new_constitution = integrationRetryWorkspace 2 "prop
   proposalsTag <- H.evalMaybe $ proposalsProcedure ^? Aeson.key "govAction" . Aeson.key "tag" . Aeson._String
   proposalsTag === "NewConstitution"
 
-  -- Check the stake pool votes are empty
-  proposalsStakePoolVotes <- H.evalMaybe $ proposal ^? Aeson.key "stakePoolVotes" . Aeson._Object
-  proposalsStakePoolVotes === mempty
+  -- Round 1: submit 3 yes votes. Ratio = 3 yes / (3 yes + 6 non-voting) = 33.3%, below 51% threshold.
+  -- The proposal cannot be ratified, so it persists across epoch boundaries.
+  do let drepVotes = [(defaultDRepKeyPair idx, vote) | (vote, idx) <- round1Votes]
```
palas, line 333: I think this `do` is redundant, is it for reducing the scope of the inner variables? Same with the `do` for the second round

palas (approved):
Looks good, just a couple of really minor nits

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Gov/ProposeNewConstitution.hs
carbolymer: Yes, it just reduces the scope and makes the test more readable.

Jimbo4350 (approved):

Issue #1245, opened 2020-06-12
[QA] - Metadata in transactions


dorin100:

dorin100:
https://hydra.iohk.io/build/3050757/download/1/delegation_design_spec.pdf

dorin100:
https://github.com/input-output-hk/cardano-node/issues/1038

dorin100:
https://jira.iohk.io/browse/CAD-1138

PR #6602, opened 2026-06-17
Update tx-generator to not use deprecated cardano-api API

### LINKED ISSUE #6592: Migrate tx-generator to the experimental Cardano API (Conway and later)
## Background

Raised by @Jimbo4350 in review of PR #6575.

The experimental Cardano API (`Cardano.Api.Experimental`) provides `makeUnsignedTx` and `signTx` as the supported transaction construction path, replacing direct ledger-level construction.
It only supports Conway and Dijkstra eras.

Mainnet is on Conway (protocol version 9+) and all production benchmark profiles already target Conway exclusively.
No active profile sets an era below Conway.
The pre-Conway code paths are dead in production.

This targets the master branch version of the code.

## Scope

Paths are relative to `bench/tx-generator/`.

### Core transaction building

- [ ] **`src/Cardano/TxGenerator/Tx.hs`** - rewrite `genTx` from manual `Ledger.mkBasicTx` / `WitVKey` / `ShelleyTx` construction to `Exp.makeUnsignedTx` + `Exp.makeKeyWitness` + `Exp.signTx`.

### Era dispatch - narrow to Conway + Dijkstra

- [ ] **`src/Cardano/Benchmarking/Script/Core.hs:71-81`** - `withEra` dispatches on every era from Shelley to Dijkstra. Remove pre-Conway branches.
- [ ] **`src/Cardano/TxGenerator/Fund.hs:80-88`** - `getFundWitness` pattern-matches on every era in `InAnyCardanoEra`. Narrow to Conway and Dijkstra.
- [ ] **`src/Cardano/TxGenerator/Utils.hs:27-35`** - `liftAnyEra` pattern-matches on every era in `InAnyCardanoEra`. Narrow to Conway and Dijkstra.
- [ ] **`nix/nixos/tx-generator-service.nix:120-128`** - the `era` option enum lists shelley through conway. Narrow to conway and dijkstra.

### Selftest era migration

- [ ] **`src/Cardano/Benchmarking/Script/Selftest.hs:103`** - change `AnyCardanoEra AllegraEra` to `AnyCardanoEra ConwayEra`.

The era binding propagates consistently to all `AddFund`, `Submit`, and `createChange` actions - no witness mismatch.
The change is structurally safe: no golden files, no asserted transaction sizes, no hardcoded expected values exist.

However, the protocol parameter conversion takes a different code path for Conway (via `toConwayPParams` → `toBabbagePParams` → `toAlonzoCommonPParams`) and requires fields not needed for Allegra (`executionUnitPrices`, `maxTxExecutionUnits`, `utxoCostPerByte`, `costModels`, etc.).
The current `data/protocol-parameters.json` has `protocolVersion: {major: 6}` (Shelley/Allegra vintage) - semantically wrong for Conway.

- [ ] **Update selftest protocol parameters file** to `data/protocol-parameters-v9.json` (or v10), which already exists in the data directory with Conway-appropriate values (PlutusV1/V2/V3 cost models, `utxoCostPerByte: 538`, version 9).

- [ ] **`test/Bench.hs:21`** - uses `runSelftest` for criterion benchmarks. Only checks `Right`/`Left`, no size assertions. Verify it passes after the era + parameter file change.

Unaffected areas:
- SizedMetadata costs differ between eras (Allegra: 39 bytes first entry, Conway: 42 bytes) but are not exercised by the selftest (`TxMetadataNone` throughout).
- Plutus script handling is not exercised (selftest uses `PayToAddr` exclusively).
- Genesis configuration is not used (selftest loads protocol parameters directly from JSON).
- Transaction CBOR size will change (Conway body has additional fields) but is unasserted.

### Other files

- [ ] **`src/Cardano/TxGenerator/PureExample.hs`** - hardcodes `BabbageEra` throughout. Rewrite to Conway using the experimental API.
- [ ] **`src/Cardano/Benchmarking/GeneratorTx/SizedMetadata.hs:48-95`** - per-era cost properties from Shelley to Dijkstra. Keep for completeness.

## Definition of done

- [ ] tx-generator builds with no warnings on the experimental API.
- [ ] `selftest` CLI passes on Conway with the v9/v10 protocol parameters.
- [ ] `test/Bench.hs` criterion benchmark passes.
- [ ] No remaining sub-Conway transaction construction paths in `bench/tx-generator/`.
- [ ] Per-era dispatchers (`withEra`, `getFundWitness`, `liftAnyEra`) narrowed to Conway + Dijkstra.
- [ ] Nix era enum updated.

carbolymer:
## Description

Removes all use of the deprecated cardano-api API from tx-generator.

The original plan was to migrate onto the experimental API and narrow era support to Conway/Dijkstra.
That plan changed: tx-generator now builds every transaction, for every era Shelley through Dijkstra, on a single track via `Cardano.Api.Compatible.Tx.createCompatibleTx`, with per-era dispatch retained through `withEra`/`shelleyBasedEraConstraints`.
The experimental API is still used, but only for witness vocabulary types (`AnyWitness`, `PlutusScriptWitness`, `AnyPlutusScript`, `SignedTx`), not for a separate transaction-building path.

- Build real Plutus spending witnesses: fund witnesses are typed `AnyWitness (LedgerEra era)`, `makePlutusContext` constructs the witness via the new `Cardano.TxGenerator.Setup.Plutus` helpers (`toAnyPlutusScript`, `mkPlutusSpendingWitness`), and `genTx` consumes each fund's stored witness.
- Plutus V1/V2 workloads are supported through the compatible API from Alonzo/Babbage onwards respectively (gated by `scriptLanguageSupportedInEra`); this depends on IntersectMBO/cardano-api#1282.
- Metadata sizing (`SizedMetadata`) and genesis funding are on the same `createCompatibleTx` track, generic over `IsShelleyBasedEra`.
- Fix the fee estimate double-counting key witnesses (the preview tx is already signed, so no dummy witnesses need to be added).
- Adapt the Ogmios submission path to the experimental `SignedTx`.
- Dijkstra is routed through the same path but is currently blocked by known runtime stubs in cardano-api (`shelleyBasedEraConstraints` has no Dijkstra arm yet); the Dijkstra metadata test cases are included as commented-out TODOs, disabled until that lands.
- workbench: pass the run's supervisord config explicitly to every `supervisorctl` call, so runs work on hosts that have their own `/etc/supervisord.conf`.
- Drop now-unused build-deps and the `-Wno-deprecations` pragmas across the library, `calibrate-script` and `tx-generator-apitest`, now that nothing deprecated remains.

Validated end to end with `make ci-test-hydra-autonix`: no `MissingScriptWitnessesUTXOW` rejections, script-spending blocks adopted (2.4 kB blocks passing the analysis filters).

> [!NOTE]
> `cabal.project` temporarily pins `source-repository-package`s for cardano-api (IntersectMBO/cardano-api#1282 head) and cardano-cli (IntersectMBO/cardano-cli#1408 head), plus a `cardano-data <1.3.1` constraint and a `-Wno-name-shadowing` override for `cardano-rpc` (Cardano.Api's new `blockNo` re-export shadows a local binding there).
> All of these are to be dropped once the pinned dependencies are released to CHaP.

## Checklist

- [ ] New tests are added if needed and existing tests are updated. These may include:
  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] CI passes
  - [ ] Code is linted with `hlint`
  - [ ] Code is formatted with `stylish-haskell`
- [ ] Self-reviewed the diff

### DIFF
--- a/bench/tx-generator/app/calibrate-script.hs
+++ b/bench/tx-generator/app/calibrate-script.hs
@@ -9,26 +9,18 @@
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeOperators #-}
 
 {-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
 
-import           Cardano.Api hiding (eraProtVerHigh)
-import           Cardano.Api.Experimental (AnyWitness (..), IsEra (useEra),
-                   PlutusScriptDatum (..), PlutusScriptPurpose (..), SignedTx (..),
-                   eraProtVerHigh, makeKeyWitness, makeUnsignedTx,
-                   obtainCommonConstraints, signTx, toPlutusSLanguage)
-import           Cardano.Api.Experimental.AnyScriptWitness (AnyPlutusScriptWitness (..),
-                   createPlutusSpendingScriptWitness)
-import qualified Cardano.Api.Experimental as Exp (PlutusScriptInEra (..),
-                   PlutusScriptOrReferenceInput (..), PlutusScriptWitness (..),
-                   evaluateTransactionFee)
-import           Cardano.Api.Experimental.Plutus (AnyPlutusScript (..),
-                   plutusScriptInEraSLanguage)
+import           Cardano.Api
+import           Cardano.Api.Experimental (IsEra (useEra), SignedTx (..), makeKeyWitness,
+                   makeUnsignedTx, obtainCommonConstraints, signTx)
+import qualified Cardano.Api.Experimental as Exp (evaluateTransactionFee)
+import           Cardano.Api.Experimental.Plutus (AnyPlutusScript (..))
 import qualified Cardano.Api.Experimental.Tx as Exp
 
 import           Cardano.Ledger.Core (mkCoinTxOut)
-import qualified Cardano.Ledger.Plutus.Language as L (PlutusLanguage, Plutus (..), PlutusBinary (..),
-                   SLanguage (..), decodePlutusRunnable)
 
 import           Cardano.Benchmarking.Compiler (keyBenchmarkInputs)
 import           Cardano.Benchmarking.GeneratorTx.SizedMetadata (mkMetadata)
@@ -42,7 +34,6 @@ import           Cardano.TxGenerator.Types
 import           Cardano.TxGenerator.Utils (keyAddress, mkTxIn)
 
 import           Control.Exception
-import           Data.Typeable (Typeable)
 import           Data.Aeson (decodeFileStrict')
 import qualified Data.ByteString.Lazy.Char8 as BSL (writeFile)
 
@@ -176,17 +167,7 @@ runPlutus strategy budgetType protoParamFile plutusDef@PlutusOn{..} scales
   = do
     protocolParameters <- readProtocolParametersOrDie protoParamFile
     (script, resolvedTo) <- either (error . show) pure =<< readPlutusScript plutusScript
-    let era = useEra @ConwayEra
-    anyPlutusScript <- obtainCommonConstraints era $
-      case script of
-        ScriptInAnyLang _lang (PlutusScript version (PlutusScriptSerialised sbs)) -> do
-          let slang = toPlutusSLanguage version
-              decode :: forall l. (L.PlutusLanguage l, Typeable l) => L.SLanguage l -> IO (AnyPlutusScript (ShelleyLedgerEra ConwayEra))
-              decode _ = case L.decodePlutusRunnable @l (eraProtVerHigh era) (L.Plutus (L.PlutusBinary sbs)) of
-                Left err -> throwIO $ userError $ "script decode failed: " ++ show err
-                Right runnable -> pure $ AnyPlutusScript (Exp.PlutusScriptInEra runnable)
-          obtainLangConstraints slang $ decode slang
-        _ -> throwIO $ userError "expected a Plutus script"
+    anyPlutusScript <- either (throwIO . userError . show) pure $ toAnyPlutusScript (shelleyBasedEra @ConwayEra) script
 
     let
       redeemerDef   = Right plutusDef
@@ -515,23 +496,25 @@ writeResultsJSON jsonName summaries = do
 --   uses it to augment the budget summary with txn size and fee.
 --   * If anything fails to evaluate, the summary is returned unchanged.
 --   * This function is currently monomorphic in the ledger era and will resolve era parameters to Conway.
-approximateTxProperties :: AnyPlutusScript (ShelleyLedgerEra ConwayEra) -> LedgerProtocolParameters ConwayEra -> (PlutusBudgetSummary, ScriptRedeemer) -> IO PlutusBudgetSummary
-approximateTxProperties (AnyPlutusScript ps) ledgerProtocolParameters (summary, redeemer) = do
+approximateTxProperties
+  :: forall era
+   . era ~ ConwayEra
+  => AnyPlutusScript (ShelleyLedgerEra era)
+  -> LedgerProtocolParameters era
+  -> (PlutusBudgetSummary, ScriptRedeemer)
+  -> IO PlutusBudgetSummary
+approximateTxProperties anyPlutusScript ledgerProtocolParameters (summary, redeemer) = do
   putStrLn $ "--> approximating txn size and fee for: " ++ messageId summary
   let
-    era = useEra @ConwayEra
+    era = useEra @era
     pparams = unLedgerProtocolParameters ledgerProtocolParameters
-    slang = plutusScriptInEraSLanguage ps
     dummyDatum = unsafeHashableScriptData $ ScriptDataNumber 0
-    datum = mkSpendingScriptDatum slang dummyDatum
-    anyWitness =
-      let witness = Exp.PlutusScriptWitness slang (Exp.PScript ps) datum redeemer (budgetUsedPerTxInput summary)
-      in AnyPlutusScriptWitness (AnyPlutusSpendingScriptWitness (createPlutusSpendingScriptWitness slang witness))
+    anyWitness = mkPlutusSpendingWitness anyPlutusScript dummyDatum redeemer (budgetUsedPerTxInput summary)
   obtainCommonConstraints era (do
     -- Corresponds to the metadata inserted in benchmarking workloads, which is why it's needed for the estimate.
     -- default value taken from: `add_tx_size` in nix/nixos/tx-generator-service.nix
     dummyMetadata <- either (\e -> throwIO $ userError $ "approximateTxProperties: mkMetadata: " ++ e) pure
-                       $ mkMetadata @ConwayEra 100
+                       $ mkMetadata @era 100
     let
       expMetadata = case dummyMetadata of
         TxMetadataNone      -> mempty
@@ -539,8 +522,8 @@ approximateTxProperties (AnyPlutusScript ps) ledgerProtocolParameters (summary, 
 
       -- just placeholders
       dummyTxIn ix = mkTxIn $ "900fc5da77a0747da53f7675cbb7d149d46779346dea2f879ab811ccc72a2162#" <> textShow @Int ix
-      dummyTxOut :: Exp.TxOut (ShelleyLedgerEra ConwayEra)
-      dummyTxOut = Exp.TxOut $ mkCoinTxOut (toShelleyAddr $ keyAddress @ConwayEra (Testnet (NetworkMagic 42)) keyBenchmarkInputs) (Coin 1_000_000)
+      dummyTxOut :: Exp.TxOut (ShelleyLedgerEra era)
+      dummyTxOut = Exp.TxOut $ mkCoinTxOut (toShelleyAddr $ keyAddress @era (Testnet (NetworkMagic 42)) keyBenchmarkInputs) (Coin 1_000_000)
 
       -- build a dummy tx akin to what we'd get in the tx-generator's benchmarking workload;
       -- it just needs to be sufficient to get our approximations.
@@ -558,7 +541,7 @@ approximateTxProperties (AnyPlutusScript ps) ledgerProtocolParameters (summary, 
 
     let
       witVKey = makeKeyWitness era unsignedTx (WitnessPaymentKey keyBenchmarkInputs)
-      dummyTx :: SignedTx ConwayEra
+      dummyTx :: SignedTx era
       dummyTx = signTx era [] [witVKey] unsignedTx
 
     pure summary
@@ -569,20 +552,4 @@ approximateTxProperties (AnyPlutusScript ps) ledgerProtocolParameters (summary, 
       putStrLn $ "approximation failed: " ++ show e
               ++ "\n--> using unmodified summary"
       pure summary
-
--- TODO: remove these vendored functions and import from
--- Cardano.Api.Experimental when the published cardano-api version includes them.
-
-obtainLangConstraints :: L.SLanguage lang -> ((L.PlutusLanguage lang, Typeable lang) => a) -> a
-obtainLangConstraints L.SPlutusV1 f = f
-obtainLangConstraints L.SPlutusV2 f = f
-obtainLangConstraints L.SPlutusV3 f = f
-obtainLangConstraints L.SPlutusV4 f = f
-
-mkSpendingScriptDatum :: L.SLanguage lang -> HashableScriptData -> PlutusScriptDatum lang 'SpendingScript
-mkSpendingScriptDatum = \case
-  L.SPlutusV1 -> SpendingScriptDatum
-  L.SPlutusV2 -> SpendingScriptDatum
-  L.SPlutusV3 -> SpendingScriptDatum . Just
-  L.SPlutusV4 -> SpendingScriptDatum . Just
 
--- a/bench/tx-generator/app/calibrate-script.hs
+++ b/bench/tx-generator/app/calibrate-script.hs
@@ -1,8 +1,10 @@
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE DataKinds #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE NumericUnderscores #-}
+{-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
@@ -10,27 +12,44 @@
 
 {-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
 
-import           Cardano.Api
+import           Cardano.Api hiding (eraProtVerHigh)
+import           Cardano.Api.Experimental (AnyWitness (..), IsEra (useEra),
+                   PlutusScriptDatum (..), PlutusScriptPurpose (..), SignedTx (..),
+                   eraProtVerHigh, makeKeyWitness, makeUnsignedTx,
+                   obtainCommonConstraints, signTx, toPlutusSLanguage)
+import           Cardano.Api.Experimental.AnyScriptWitness (AnyPlutusScriptWitness (..),
+                   createPlutusSpendingScriptWitness)
+import qualified Cardano.Api.Experimental as Exp (PlutusScriptInEra (..),
+                   PlutusScriptOrReferenceInput (..), PlutusScriptWitness (..),
+                   evaluateTransactionFee)
+import           Cardano.Api.Experimental.Plutus (AnyPlutusScript (..),
+                   plutusScriptInEraSLanguage)
+import qualified Cardano.Api.Experimental.Tx as Exp
+
+import           Cardano.Ledger.Core (mkCoinTxOut)
+import qualified Cardano.Ledger.Plutus.Language as L (PlutusLanguage, Plutus (..), PlutusBinary (..),
+                   SLanguage (..), decodePlutusRunnable)
 
 import           Cardano.Benchmarking.Compiler (keyBenchmarkInputs)
 import           Cardano.Benchmarking.GeneratorTx.SizedMetadata (mkMetadata)
 import           Cardano.TxGenerator.Calibrate.Utils
 import           Cardano.TxGenerator.PlutusContext
-import           Cardano.TxGenerator.ProtocolParameters ( ProtocolParameters (..), convertToLedgerProtocolParameters, 
-                   toLedgerPParams)
+import           Cardano.TxGenerator.ProtocolParameters (ProtocolParameters (..),
+                   convertToLedgerProtocolParameters)
 import           Cardano.TxGenerator.Setup.Plutus
 import           Cardano.TxGenerator.Tx (txSizeInBytes)
 import           Cardano.TxGenerator.Types
 import           Cardano.TxGenerator.Utils (keyAddress, mkTxIn)
 
 import           Control.Exception
+import           Data.Typeable (Typeable)
 import           Data.Aeson (decodeFileStrict')
 import qualified Data.ByteString.Lazy.Char8 as BSL (writeFile)
+
 import           Data.Char
 import           Data.Function (on, (&))
 import           Data.List (nub, sort, transpose)
 import           Data.List.Extra (split)
-
 import           Data.Map.Strict as Map (Map, empty, fromList, union)
 import           Data.Maybe
 import qualified Data.Text as T
@@ -157,6 +176,17 @@ runPlutus strategy budgetType protoParamFile plutusDef@PlutusOn{..} scales
   = do
     protocolParameters <- readProtocolParametersOrDie protoParamFile
     (script, resolvedTo) <- either (error . show) pure =<< readPlutusScript plutusScript
+    let era = useEra @ConwayEra
+    anyPlutusScript <- obtainCommonConstraints era $
+      case script of
+        ScriptInAnyLang _lang (PlutusScript version (PlutusScriptSerialised sbs)) -> do
+          let slang = toPlutusSLanguage version
+              decode :: forall l. (L.PlutusLanguage l, Typeable l) => L.SLanguage l -> IO (AnyPlutusScript (ShelleyLedgerEra ConwayEra))
+              decode _ = case L.decodePlutusRunnable @l (eraProtVerHigh era) (L.Plutus (L.PlutusBinary sbs)) of
+                Left err -> throwIO $ userError $ "script decode failed: " ++ show err
+                Right runnable -> pure $ AnyPlutusScript (Exp.PlutusScriptInEra runnable)
+          obtainLangConstraints slang $ decode slang
+        _ -> throwIO $ userError "expected a Plutus script"
 
     let
       redeemerDef   = Right plutusDef
@@ -189,7 +219,10 @@ runPlutus strategy budgetType protoParamFile plutusDef@PlutusOn{..} scales
       jsonName  = "summaries_" ++ scriptName <.> "json"
       csvName   = "scaling_" ++ scriptName <.> "csv"
 
-    summariesWithApprox <- mapM (approximateTxProperties script protocolParameters) summaries
+    ledgerProtocolParameters <-
+          either (error . docToString . prettyError) pure
+            $ convertToLedgerProtocolParameters shelleyBasedEra protocolParameters
+    summariesWithApprox <- mapM (approximateTxProperties anyPlutusScript ledgerProtocolParameters) summaries
     writeResultsJSON jsonName summariesWithApprox
     writeResultsCSV csvName summariesWithApprox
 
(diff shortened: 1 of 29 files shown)


carbolymer:
Experimental API migration guide: https://gist.github.com/carbolymer/1967d68a8d4ed4b80cbbf50ee8698bf8

bench/tx-generator/src/Cardano/Benchmarking/Script/Selftest.hs
```diff
@@ -100,7 +100,7 @@ testScript protocolFile submitMode =
           , teDescription = fromString "Genesis Initial UTxO Signing Key"
           , teRawCBOR = "X \vl1~\182\201v(\152\250A\202\157h0\ETX\248h\153\171\SI/m\186\242D\228\NAK\182(&\162"
           }
-    era = AnyCardanoEra AllegraEra
+    era = AnyCardanoEra ConwayEra
```
carbolymer, line 103: selftest is a smoke test for tx-generator so I presume we don't have to use Allegra here, and conway is fine?

bench/tx-generator/app/calibrate-script.hs
```diff
@@ -485,71 +488,86 @@ writeResultsJSON jsonName summaries = do
...
+  let
+    era = either (\deprecated -> error $ "unsupported era: " ++ show deprecated) id
+               $ sbeToEra shelleyBasedEra
+  evaluate
+    ( obtainCommonConstraints era $
+        let
+          ledgerPParams =
+            either (error . docToString . prettyError) id
+              $ toLedgerPParams shelleyBasedEra protocolParameters
+
```
Jimbo4350: We are creating two `PParams era` values here where one would suffice.

bench/tx-generator/app/calibrate-script.hs
```diff
@@ -485,71 +488,86 @@ writeResultsJSON jsonName summaries = do
 approximateTxProperties :: ScriptInAnyLang -> ProtocolParameters -> (PlutusBudgetSummary, ScriptRedeemer) -> IO PlutusBudgetSummary
```
Jimbo4350: Full migration away from the "old" api would remove the use of `ProtocolParameter`s here.

bench/tx-generator/app/calibrate-script.hs
```diff
@@ -485,71 +488,86 @@ writeResultsJSON jsonName summaries = do
...
+          ledgerPParams =
+            either (error . docToString . prettyError) id
+              $ toLedgerPParams shelleyBasedEra protocolParameters
+
+          ledgerProtocolParameters :: LedgerProtocolParameters ConwayEra
+          ledgerProtocolParameters =
+            either (error . docToString . prettyError) id
+              $ convertToLedgerProtocolParameters shelleyBasedEra protocolParameters
+
+          witness :: Witness WitCtxTxIn ConwayEra
```
Jimbo4350: This is still using the old api's `Witness` type.

Jimbo4350 (changes requested):

bench/tx-generator/app/calibrate-script.hs
carbolymer: fixed

bench/tx-generator/app/calibrate-script.hs
carbolymer: removed usage of `ProtocolParameters`.

bench/tx-generator/app/calibrate-script.hs
carbolymer: removed the usage

bench/tx-generator/src/Cardano/Benchmarking/GeneratorTx/SubmissionClient.hs
```diff
@@ -189,7 +191,8 @@ txSubmissionClient tr bmtr initialTxSource endOfProtocolCallback =
   fromGenTxId (Block.GenTxIdAlonzo  (Mempool.ShelleyTxId i)) = fromShelleyTxId i
   fromGenTxId (Block.GenTxIdBabbage (Mempool.ShelleyTxId i)) = fromShelleyTxId i
   fromGenTxId (Block.GenTxIdConway  (Mempool.ShelleyTxId i)) = fromShelleyTxId i
-  fromGenTxId _ = error "TODO: fix incomplete match"
+  fromGenTxId (Block.GenTxIdDijkstra (Mempool.ShelleyTxId i)) = fromShelleyTxId i
+  fromGenTxId (Block.GenTxIdByron _) = error "TODO: fix incomplete match"
```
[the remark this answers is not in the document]
carbolymer: this is fine

bench/tx-generator/src/Cardano/Benchmarking/GeneratorTx/SizedMetadata.hs
```diff
@@ -27,67 +30,31 @@ maxBSSize = 64
...
-    && prop_mapCostsBabbage
-    && prop_bsCostsShelley
-    && prop_bsCostsAllegra
-    && prop_bsCostsMary
-    && prop_bsCostsAlonzo
-    && prop_bsCostsBabbage
+  =    prop_mapCostsConway
+    && prop_mapCostsDijkstra
     && prop_bsCostsConway
+    && prop_bsCostsDijkstra
```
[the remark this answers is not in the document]
carbolymer: this is dead code anyway

bench/tx-generator/src/Cardano/TxGenerator/Tx.hs
```diff
@@ -147,45 +150,42 @@ sourceTransactionPreview txGenerator inputFunds valueSplitter toStore =
...
+        expMetadata = case metadata of
+          TxMetadataNone -> mempty
+          TxMetadataInEra _ m -> m
+        txBodyContent =
+          Exp.defaultTxBodyContent
+            & Exp.setTxIns expInputs
+            & Exp.setTxInsCollateral collateralIns
+            & Exp.setTxOuts expOutputs
+            & Exp.setTxFee fee
+            & Exp.setTxMetadata expMetadata
```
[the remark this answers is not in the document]
carbolymer: protocol params are unused but we can wire them anyway

bench/tx-generator/app/calibrate-script.hs
```diff
@@ -189,7 +218,10 @@ runPlutus strategy budgetType protoParamFile plutusDef@PlutusOn{..} scales
       jsonName  = "summaries_" ++ scriptName <.> "json"
       csvName   = "scaling_" ++ scriptName <.> "csv"
 
-    summariesWithApprox <- mapM (approximateTxProperties script protocolParameters) summaries
+    let ledgerProtocolParameters =
+          either (error . docToString . prettyError) id
```
carbolymer:
```suggestion
    ledgerProtocolParameters <-
          either (error . docToString . prettyError) pure
```

bench/tx-generator/src/Cardano/TxGenerator/Fund.hs
```diff
@@ -76,15 +77,10 @@ getFundCoin (Fund (InAnyCardanoEra _ a)) = case _fundVal a of
...
 -- `getFundWitness` otherwise wraps `_fundWitness`.
-getFundWitness :: forall era. IsShelleyBasedEra era => Fund -> Witness WitCtxTxIn era
-getFundWitness fund = case (cardanoEra @era, fund) of
-  (ByronEra   , Fund (InAnyCardanoEra ByronEra   a)) -> _fundWitness a
-  (ShelleyEra , Fund (InAnyCardanoEra ShelleyEra a)) -> _fundWitness a
-  (AllegraEra , Fund (InAnyCardanoEra AllegraEra a)) -> _fundWitness a
-  (MaryEra    , Fund (InAnyCardanoEra MaryEra    a)) -> _fundWitness a
-  (AlonzoEra  , Fund (InAnyCardanoEra AlonzoEra  a)) -> _fundWitness a
-  (BabbageEra , Fund (InAnyCardanoEra BabbageEra a)) -> _fundWitness a
+getFundWitness :: forall era. IsEra era => Fund -> Witness WitCtxTxIn era
```
Jimbo4350: Another type that needs to be replaced: `Witness`

bench/tx-generator/src/Cardano/TxGenerator/Genesis.hs
```diff
@@ -99,39 +102,42 @@ genesisExpenditure ::
...
-mkGenesisTransaction :: forall era .
-     IsShelleyBasedEra era
-  => SigningKey GenesisUTxOKey
+mkGenesisTransaction ::
+     Era era
+  -> SigningKey GenesisUTxOKey
   -> SlotNo
   -> L.Coin
   -> [TxIn]
   -> [TxOut CtxTx era]
```
Jimbo4350: `TxOut CtxTx era` needs to be replaced with experimental's `TxOut era` type.

Jimbo4350 (changes requested):
Some more leftover types.

bench/tx-generator/src/Cardano/TxGenerator/Fund.hs
carbolymer: replaced

bench/tx-generator/src/Cardano/TxGenerator/Genesis.hs
carbolymer: replaced

bench/tx-generator/app/calibrate-script.hs
```diff
@@ -157,6 +175,17 @@ runPlutus strategy budgetType protoParamFile plutusDef@PlutusOn{..} scales
   = do
     protocolParameters <- readProtocolParametersOrDie protoParamFile
     (script, resolvedTo) <- either (error . show) pure =<< readPlutusScript plutusScript
+    let era = useEra @ConwayEra
+    anyPlutusScript <- obtainCommonConstraints era $
+      case script of
+        ScriptInAnyLang _lang (PlutusScript version (PlutusScriptSerialised sbs)) -> do
```
Jimbo4350: This is still an old api type and we should be using `AnyScript era`. However there is no helper function to make this easy so I added it here: https://github.com/IntersectMBO/cardano-api/pull/1245

bench/tx-generator/app/calibrate-script.hs
```diff
@@ -481,75 +513,76 @@ writeResultsJSON jsonName summaries = do
...
+    dummyMetadata <- either (\e -> throwIO $ userError $ "approximateTxProperties: mkMetadata: " ++ e) pure
+                       $ mkMetadata @ConwayEra 100
+    let
+      expMetadata = case dummyMetadata of
+        TxMetadataNone      -> mempty
+        TxMetadataInEra _ m -> m
+
+      -- just placeholders
+      dummyTxIn ix = mkTxIn $ "900fc5da77a0747da53f7675cbb7d149d46779346dea2f879ab811ccc72a2162#" <> textShow @Int ix
+      dummyTxOut   = TxOut (keyAddress (Testnet (NetworkMagic 42)) keyBenchmarkInputs) (lovelaceToTxOutValue shelleyBasedEra 1_000_000) TxOutDatumNone ReferenceScriptNone
```
Jimbo4350: Still using the old api's `TxOut` here.

bench/tx-generator/app/calibrate-script.hs
```diff
@@ -481,75 +513,76 @@ writeResultsJSON jsonName summaries = do
...
+          & Exp.setTxFee (Coin 1_000_000)
+          & Exp.setTxMetadata expMetadata
+          & Exp.setTxProtocolParams pparams
+
+    unsignedTx <- either (\err -> throwIO $ userError $ "approximateTxProperties: " ++ show err) pure
+                    $ makeUnsignedTx era txBodyContent
+
+    let
+      witVKey = makeKeyWitness era unsignedTx (WitnessPaymentKey keyBenchmarkInputs)
+      dummyTx :: Tx ConwayEra
```
Jimbo4350: This is the old api's `Tx` type.

bench/tx-generator/src/Cardano/Benchmarking/GeneratorTx/SizedMetadata.hs
```diff
@@ -107,31 +73,32 @@ listMetadata :: [TxMetadataValue] -> TxMetadata
...
-dummyTxSize :: forall era . IsShelleyBasedEra era => AsType era -> Maybe TxMetadata -> Int
-dummyTxSize _p m = (dummyTxSizeInEra @era) $ metadataInEra m
+  era = useEra @era
+  expMetadata = case metadata of
+    TxMetadataNone -> mempty
+    TxMetadataInEra _ m -> m
+  txBodyContent = Exp.defaultTxBodyContent
+    & Exp.setTxIns [(mkTxIn "dbaff4e270cfb55612d9e2ac4658a27c79da4a5271c6f90853042d1403733810#0", AnyKeyWitnessPlaceholder)]
+    & Exp.setTxMetadata expMetadata
+  dummyTx :: Tx era
```
Jimbo4350: Old api's `Tx` type.

bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
```diff
@@ -504,19 +499,8 @@ makePlutusContext ScriptSpec{..} = do
...
-          Nothing -> error $ "runPlutusBenchmark: " ++ show version ++ " not supported in era: " ++ show (cardanoEra @era)
-          Just scriptLang -> PlutusScriptWitness
-                              scriptLang
-                              version
-                              (PScript script')               -- TODO: add capability for reference inputs from Babbage era onwards
-                              (ScriptDatumForTxIn $ Just scriptData)
-                              scriptRedeemer
-                              executionUnits
-      in return (ScriptWitness ScriptWitnessForSpending scriptWitness, script, getScriptData scriptData, scriptFee)
+    ScriptInAnyLang _ PlutusScript{} ->
```
Jimbo4350: Old api's `ScriptInAnyLang` type here.

bench/tx-generator/src/Cardano/TxGenerator/Tx.hs
```diff
@@ -147,45 +150,43 @@ sourceTransactionPreview txGenerator inputFunds valueSplitter toStore =
...
+            & Exp.setTxProtocolParams pparams
+    unsignedTx <- first (\err -> TxGenError $ "genTx: " ++ show err) $ makeUnsignedTx era txBodyContent
+    let witVKeys = [makeKeyWitness era unsignedTx (WitnessPaymentKey key) | key <- allKeys]
+    let tx = case signTx era [] witVKeys unsignedTx of
+                SignedTx signedLedgerTx -> ShelleyTx shelleyBasedEra signedLedgerTx
```
Jimbo4350: Old api's `Tx` type used here.

bench/tx-generator/src/Cardano/TxGenerator/UTxO.hs
```diff
@@ -25,59 +26,52 @@ makeToUTxOList fkts values
...
+          Nothing -> error "mkUtxOScript: scriptLanguageSupportedInEra==Nothing"
+          Just{} -> makeShelleyAddressInEra
+                         (shelleyBasedEra @era)
+                         networkId
+                         (PaymentCredentialByScript $ hashScript script')
+                         NoStakeAddress
+    mkTxOut v = case forEraMaybeEon (cardanoEra @era) of
+      Nothing -> error "mkUtxOScript: scriptDataSupportedInEra==Nothing"
+      Just tag -> TxOut
+                    plutusScriptAddr
```
Jimbo4350: Old api's `TxOut` type used here.

Jimbo4350 (changes requested):
There are still a few sites that use the old api's types. If you want to merge what you have and deal with them in a follow up PR that's fine but let me know.

bench/tx-generator/app/calibrate-script.hs
Jimbo4350: ^ This can stay as is for now since 1245 needs to be merged and released.

bench/tx-generator/src/Cardano/TxGenerator/Genesis.hs
```diff
@@ -99,39 +102,42 @@ genesisExpenditure ::
...
 genesisExpenditure networkId inputKey addr value fee ttl outputKey
-  = second (\tx -> (tx, Fund $ InAnyCardanoEra cardanoEra $ fund tx)) eTx
+  = obtainCommonConstraints era $
+      second (\tx -> (tx, Fund $ InAnyCardanoEra cardanoEra $ fund tx))
+        $ mkGenesisTransaction era (castKey inputKey) ttl fee [pseudoTxIn]
+            [Exp.TxOut $ toShelleyTxOutAny shelleyBasedEra txout]
  where
-  eTx         = mkGenesisTransaction (castKey inputKey) ttl fee [pseudoTxIn] [txout]
+  era = useEra @era
   txout       = TxOut addr value TxOutDatumNone ReferenceScriptNone
```
Jimbo4350: Old `TxOut type` used here.

bench/tx-generator/src/Cardano/TxGenerator/UTxO.hs
```diff
@@ -25,59 +26,52 @@ makeToUTxOList fkts values
...
-  mkNewFund :: L.Coin -> TxIx -> TxId -> Fund
-  mkNewFund val txIx txId = Fund $ InAnyCardanoEra (cardanoEra @era) $ FundInEra {
-      _fundTxIn = TxIn txId txIx
-    , _fundWitness = KeyWitness KeyWitnessForSpending
-    , _fundVal = lovelaceToTxOutValue (shelleyBasedEra @era ) val
-    , _fundSigningKey = Just key
-    }
+mkUTxOVariant networkId key value = obtainCommonConstraints (useEra @era) $
+  let
+    mkTxOut v = TxOut (keyAddress @era networkId key) (lovelaceToTxOutValue (shelleyBasedEra @era) v) TxOutDatumNone ReferenceScriptNone
```
Jimbo4350: Old `TxOut` type used here.

bench/tx-generator/app/calibrate-script.hs
carbolymer: Agreed, deferring until #1245 is merged and released. `ScriptInAnyLang` comes from `readPlutusScript`, so the whole script-loading path (`Setup.Plutus`, `PlutusContext`, `Core`) will move to `AnyScript` in one follow-up.

bench/tx-generator/app/calibrate-script.hs
carbolymer: fixed - built directly as experimental `TxOut` via ledger's `mkCoinTxOut`

bench/tx-generator/app/calibrate-script.hs
carbolymer: fixed - `dummyTx` stays `SignedTx` now

bench/tx-generator/src/Cardano/Benchmarking/GeneratorTx/SizedMetadata.hs
carbolymer: fixed - kept as `SignedTx`, sized via `serialiseToRawBytes`

bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
carbolymer: same as calibrate-script:181 - comes from `readPlutusScript`; the whole loading path moves to `AnyScript` in a follow-up once cardano-api#1245 is released

bench/tx-generator/src/Cardano/TxGenerator/Tx.hs
carbolymer: fixed - `genTx` returns `SignedTx` and the pipeline carries it end to end; the single remaining `ShelleyTx` wrap is in `mkTxInModeCardano`, since `TxInMode` has no constructor for the experimental type

bench/tx-generator/src/Cardano/TxGenerator/UTxO.hs
carbolymer: fixed - ledger `TxOut` via `mkCoinTxOut` + `datumTxOutL`; there's no experimental datum helper yet, this is the same recipe cardano-api's own `FromJSON (TxOut era)` uses internally

bench/tx-generator/src/Cardano/TxGenerator/Genesis.hs
carbolymer: fixed

bench/tx-generator/src/Cardano/TxGenerator/UTxO.hs
carbolymer: fixed - `mkCoinTxOut` + experimental `TxOut`

carbolymer:
Pushed the rest of the migration: all `Tx`/`TxOut` sites now use the experimental API. The old API remains in exactly two places: the `TxInMode`/`submitTxToNodeLocal` submission edge (no experimental submission path exists yet) and `ScriptInAnyLang` script loading, which waits for IntersectMBO/cardano-api#1245 - follow-up PR once that's released.

Jimbo4350 (approved):

bench/tx-generator/src/Cardano/TxGenerator/Tx.hs
```diff
@@ -139,53 +142,49 @@ sourceTransactionPreview ::
...
-    & setTxFee fee
-    & setTxValidityLowerBound TxValidityNoLowerBound
-    & setTxValidityUpperBound (defaultTxValidityUpperBound sbe)
-    & setTxMetadata metadata
-    & setTxProtocolParams (BuildTxWith (Just ledgerParameters))
+genTx era (LedgerProtocolParameters pparams) (collateralIns, collFunds) fee metadata inFunds outputs =
+  obtainCommonConstraints era $ do
+    let allKeys = mapMaybe getFundKey $ inFunds ++ collFunds
+        expInputs = map (\f -> (getFundTxIn f, AnyKeyWitnessPlaceholder)) inFunds
+        expMetadata = case metadata of
```
mgmeier: The witness needs to taken from the fund as well (`getFundWitness f`) - the placeholder won't work if we submit Plutus script txns.
The placeholder always leads to a `MissingScriptWitnessesUTXOW` error from ledger for all Plutus workloads.

bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
```diff
@@ -504,19 +500,8 @@ makePlutusContext ScriptSpec{..} = do
...
+      return (AnyKeyWitnessPlaceholder, script, getScriptData scriptData, scriptFee)
```
mgmeier: A correct script witness needs to be constructed here by means of the new API, carrying all the values same as before.
This will then be used in a `Fund`, so it goes hand-in-hand with my other comment. The placeholder is a bug breaking Plutus workloads.

bench/tx-generator/src/Cardano/Benchmarking/Script/Selftest.hs
mgmeier: Yes, Conway is fine.

mgmeier (changes requested):

mgmeier:
There's still this in the status quo: `Cardano.TxGenerator.Fund.FundInEra`'s `_fundWitness` field (and `getFundWitness`'s return type, and `Cardano.TxGenerator.UTxO.mkUTxOScript`'s witness parameter) are typed `AnyWitness era`, where `era` is cardano-api's API-level tag. A genuine Plutus script witness (AnyPlutusScript/AnyPlutusScriptWitness) requires era to be an actual cardano-ledger-core `Era` instance. The API-level `ConwayEra` tag doesn't have one.

 So addressing this will have major ripples throughout the code base.

mgmeier:
> pre-Conway eras are no longer exercised by the benchmarking workbench

That is only partially true. The workbench still supports benchmarking profiles pre-Conway. We do not exercise them, because they're not relevant to Cardano Mainnet currently. Secondly, `tx-generator` has use cases outside workbench, and constraining one does not imply constraining the other.

`tx-generator` aims to be a general-purpose submission service, runnable for any testnet, for any eras. Tying the tool to the new API will effectively remove that capability and attach it closely to the state of Mainnet.

I'm aware the reduction of era coverage is a direct constraint imposed by the new, experimental API. This leads to some general questions:
* The existing API is named "deprecated" in the PR. What's its remaining lifecycle?
* Is there a replacement in the new API, or besides it, for anything pre-Conway, or will that just disappear from `cardano-api`?
* The new API is named "experimental" - does the tx-generator really need to buy in *now*?
* Is there motive and means for the tx-generator to remain a more general rather than a quite era-specific tool?

mgmeier:
In lack of a suitable test case, I've pushed a (TEMP) commit to validate a fix:
* `make ci-test-hydra-autonix` - let it finish (should be <10mins)
* Inspect `run/current/node-*/stdout` - you'll see the error under the trace namespace `Mempool.RejectedTx`.
* If they've disappeared - `run/current/analysis/chain.txt` should show (mostly) blocks of 2-3kiB in size. This means, the script txns have been adopted.

bench/tx-generator/src/Cardano/TxGenerator/Tx.hs
carbolymer: Fixed.
`genTx` now builds its inputs as `(getFundTxIn f, getFundWitness f)`, so each fund supplies its stored witness.
Key-locked funds still yield `AnyKeyWitnessPlaceholder`, which is correct for them: their signatures come from `signTx`'s vkey witnesses, built from the funds' signing keys.
Script-locked funds yield their real script witness (see the other thread for where it is built).
`getFundWitness` returns `AnyWitness (LedgerEra era)` now, matching what `setTxIns` expects.

bench/tx-generator/src/Cardano/Benchmarking/Script/Core.hs
carbolymer: Fixed.
`makePlutusContext` now builds the real spending witness: `toAnyPlutusScript` decodes the loaded script into the ledger representation, and `mkPlutusSpendingWitness` assembles the witness from script, datum, redeemer and execution units.
Both helpers live in `Cardano.TxGenerator.Setup.Plutus` and are shared with calibrate-script, which had the only previously working copy of this logic.
The witness flows into the `Fund` via `mkUTxOScript`, so it goes hand in hand with the `genTx` change exactly as you said.

carbolymer:
@mgmeier Fixed, and validated with your repro. Mechanics are in the two inline threads; summary below.

Validation (`make ci-test-hydra-autonix`, run `2026-08-05-20-28-00000-1101-ci-test-hydra-coay-sup`):
no `MissingScriptWitnessesUTXOW` on either node, and the script-spending blocks in `chain.txt` are 2472 B each and pass the analysis filters.
I dropped your `(TEMP)` trace commit.
Add the regression test whenever it suits you.

Two things I hit during validation. Backlog material, not this PR:
1. In one run an early fork orphaned the genesis fund import tx. The generator kept chaining txs off the dead UTxO, so the whole run failed with `ConwayMempoolFailure "All inputs are spent"` while logging "Done." at every step. The fund model has no confirmation wait and no rollback recovery.
2. Every workbench `supervisorctl` call except the `supervisord` launch relies on the client's default config search. A host with its own `/etc/supervisord.conf` either breaks the run or silently talks to the wrong server. I have added a commit to this PR passing `-c` at all 12 call sites.

carbolymer:
Thanks for the thorough review and for pushing on these questions - they deserved proper answers, so I did some homework before replying.
The questions are fair, and your conservative position on a tool this many things depend on is the right instinct.

TL;DR: the deprecated API is going away and pre-Conway tx building will not get a full replacement.
tx-generator would stay general across "current era + next", which keeps hard-fork comparisons on one tool.
For older eras there is a workable path via `Cardano.Api.Compatible.Tx`: value workloads need only small additions to the existing API, while Plutus support would mean building new witness machinery - a separate, larger effort.
Both are on the table; details below.

**1. Lifecycle of the deprecated API?**
It is being removed piece by piece, and pre-Conway tx building will not get a like-for-like replacement.
cardano-cli already made this move; its interface targets the current era plus the upcoming one.
I realise that is not the answer a general-purpose tool hopes for, which is why the next point matters more.

**2. Replacement for pre-Conway?**
Partly - and I checked its coverage against what tx-generator actually uses, so this is specific rather than hopeful.
`Cardano.Api.Compatible.Tx` builds simple transactions Shelley through Conway: inputs and outputs (including multi-asset and datums), explicit fee, certificates, governance actions.
It returns the ordinary `Tx era`, so signing and submission plumbing work unchanged.
Missing for tx-generator's needs: metadata (the size-padded workloads require it) and TTL.
Both are small additions to one function, and since we maintain cardano-api, we can make them on request.
Plutus is the larger piece: Compatible has no spending witnesses, no collateral and no script integrity hash, so old-era Plutus means building that machinery.
That is a medium-to-large effort whichever way we approach it: extend Compatible upstream, or assemble the transactions directly against the ledger API inside tx-generator and lift the code upstream once proven (Compatible is written in the same ledger-lens style, so it would migrate cleanly).
On use cases: you mentioned the workbench still supports pre-Conway profiles, and the definitions in `cardano-profile` bear that out.
There is a family explicitly grouped as pre-Voltaire Plutus (`plutus-nomadperf`, `plutus-secp-*-nomadperf`, `ci-bench-plutus*`), all Babbage-shaped genesis with V1/V2 workloads, and `ERA` is a free parameter, so any of them can run as a genuine Babbage or Alonzo cluster.
Under the default `ERA=conway` they hard-fork to Conway at epoch 0, and this PR handles them as-is - our validation run was exactly that shape.
The case that needs the larger tier is an explicit `ERA=babbage` or `ERA=alonzo` run.
Do those still happen in practice?
That is the fact that decides whether the Plutus tier is needed or stays hypothetical.

**3. Does tx-generator need to buy in now?**
A fair challenge, and honestly the naming does not help us here: "experimental" suggests a side experiment, when in practice it is the next version of the API and where development happens.
Since the deprecated API is being removed, the migration would happen at some point regardless; the timing question is real, though.
The argument for now rather than later: feedback from a demanding consumer still influences the design - this PR alone surfaced several gaps that we then fixed in cardano-api, which benefits both sides.

**4. Can tx-generator remain a general tool?**
I think yes, in the sense you care about.
The proposal: it tracks the same rolling window as the rest of the toolchain - current mainnet era plus the next one, moving forward at each hard fork - so era N vs N+1 comparisons stay on one tool.
You are right that workbench profiles and tx-generator have separate scopes, and constraining one should not silently constrain the other; the tiers below are meant to keep that promise concrete rather than aspirational.
One observation about the status quo, offered as context rather than a counterargument: a single run has always been pinned to one era.
Funds are typed to the era they were created in, so a fund made in era N cannot be spent by a transaction built for era N+1 - Note [Era transitions] in `Fund.hs` records exactly this limitation.

The offer, summarised:

| Tier | What it enables | Size | Status |
|---|---|---|---|
| Nothing extra | Conway + Dijkstra, all workloads | - | this PR |
| Compatible + metadata/TTL | pre-Conway value and metadata workloads | small | on request |
| Old-era Plutus | pre-Voltaire Plutus profiles run with `ERA=babbage`/older | medium to large | depends on whether such runs still happen |

You said earlier that if this keeps the tool general without most of the maintenance burden, that would be optimal - I think the small tier gets us exactly there, and the Plutus tier is the "let's have a look together" case you offered.
The small tier we can simply do.
For the Plutus tier, the deciding question is the one above: are the pre-Voltaire Plutus profiles still exercised below Conway?

PR #6609, opened 2026-06-23
tx-generator: generic remote submission transport (with an Ogmios backend)

palas:
# Description

Adds a **generic remote submission transport** to `tx-generator`: point it at a submission endpoint and every transaction is submitted there as the endpoint's native calls, instead of via the local socket / Node-to-Node protocols. The transport is **backend-agnostic**; the first (and currently only) backend is **Ogmios** (JSON-RPC 2.0 `submitTransaction` over a WebSocket).

It is a **functional submission transport, not a benchmarking one** — useful for exercising a node through an endpoint end-to-end, not for measuring throughput.

Following review, the submission path is intentionally **not Ogmios-specific**: it is a generic interface with Ogmios as one implementation. Adding another backend later is a localised change (a new endpoint type + a transport builder + a dispatch arm), with no changes to the submission loop or its error handling.

## What it adds

- **New config keys `submissionEndpointType` + `submissionEndpointURI`** in the high-level JSON config, which **must be set together** — e.g. `"submissionEndpointType": "Ogmios"`, `"submissionEndpointURI": "ws://127.0.0.1:1337"`. When set, every phase — genesis import, UTxO splitting, and the benchmark phase — submits through the endpoint.
- **A backend-agnostic transport interface** (new module `Cardano.Benchmarking.Script.Submission`), with Ogmios as one backend (`Cardano.Benchmarking.Script.Ogmios`).
- **README + CHANGELOG** updates and a package version bump to **2.17** (adds `websockets` + `network-uri` deps).
- **`test-ogmios.sh`**: an integration test that spins up `cardano-testnet` + Ogmios and submits through it.
- **Bug fix (independent of the transport):** runs that never start the benchmark machinery (`debugMode: true`, or low-level scripts with no `Benchmark` phase) used to crash at shutdown with *"AsyncBenchmarkControl absent"*. They now exit cleanly — `runScript` returns `Maybe AsyncBenchmarkControl` instead of fabricating a no-op control.

## Implementation nuances

- **Backend-agnostic by construction.** `submitLoop` drives a `SubmitTransport` (one `submitOne` action per backend) and owns the rejection policy, the sent/failed tally, tracing, and the exit-code logic. A backend's rejection type stays its own and is only rendered for tracing (a `Pretty` constraint), so no transport detail leaks into the loop. `SubmitMode`'s constructor is now `SubmitToEndpoint SubmissionEndpointType`.
- **Requires `debugMode: true`.** A submission endpoint ignores `tps`/`targetNodes` and produces no benchmark metrics, so the compiler **rejects an endpoint config without `debugMode: true`** at compile time (fail fast, before any node interaction). `submissionEndpointType` and `submissionEndpointURI` must be set together, or the compiler rejects the config. Low-level `json` scripts are unaffected.
- **Only submission goes through the endpoint.** Protocol-parameter / era queries and protocol startup still use the local node socket + config, so local node access is still required.
- **One request in flight at the same time (Ogmios backend).** Strictly one tx per round trip over a single connection, so throughput is round-trip-bound. Ogmios pipelining (correlated by JSON-RPC `id`) is deliberately *not* used yet; it is noted in the module docs as a potential next step.
- **Rejections fail the run, by stream shape.** A rejected tx makes the process exit non-zero. Setup phases (chained txs) abort at the first rejection; the benchmark phase (independent txs) finishes the stream then fails on the tally.

## Testing

`test-ogmios.sh` is a self-contained integration test (nix-resolved Ogmios + `cardano-testnet`). It is **not wired into CI** (it needs nix). The Ogmios URL parser (`parseOgmiosUrl`) is exported to keep it unit-testable.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [x] Self-reviewed the diff


Jimbo4350 (changes requested):
This should not be Ogmios specific, it needs to be generic and at most a small section on how to use with Ogmios.

bench/tx-generator/src/Cardano/Benchmarking/Script/Ogmios.hs
```diff
@@ -0,0 +1,241 @@
...
+
+-- | Submit a single transaction and await its response on the same connection.
+-- A transaction rejection is returned as 'Left'; a protocol-level fault throws
+-- 'OgmiosProtocolError' (caught by 'withOgmiosTransport').
+ogmiosSubmitOne
+  :: IsShelleyBasedEra era
+  => WS.Connection -> IORef Int -> Tx era -> IO (Either OgmiosRejection ())
+ogmiosSubmitOne conn reqIdRef tx = do
+  reqId <- atomicModifyIORef' reqIdRef $ \n -> (n + 1, n)
+  WS.sendTextData conn $ Aeson.encode (mkSubmitRequest tx reqId)
```
carbolymer: if the process stalls on the other end, even receiving data might get stuck. worth wrapping in `timeout`.

bench/tx-generator/src/Cardano/Benchmarking/Script/Ogmios.hs
```diff
@@ -0,0 +1,241 @@
...
+-- mirrored request id alongside the result.
+parseOgmiosResponse :: LBS.ByteString -> Either String (Maybe Int, OgmiosResult)
+parseOgmiosResponse bs =
+  case Aeson.eitherDecode bs of
+    Left err -> Left err
+    Right val -> Aeson.parseEither parseResponse val
+ where
+  parseResponse = Aeson.withObject "OgmiosResponse" $ \obj -> do
+    respId <- obj .:? "id"
+    mResult <- obj .:? "result"
```
carbolymer, line 254: can `result` be null? if it can be, it will be treated as error.

carbolymer (commented):
-

bench/tx-generator/src/Cardano/Benchmarking/Script/Ogmios.hs
```diff
@@ -0,0 +1,241 @@
...
+        return $ OgmiosSuccess txId
+      Nothing -> do
+        errVal <- obj .: "error"
+        Aeson.withObject "error" (\errObj -> do
+          code <- errObj .: "code"
+          msg  <- errObj .: "message"
+          dat  <- errObj .:? "data"
+          return $ OgmiosError code msg (fromMaybe Null dat)
+          ) errVal
+    return (respId, result)
```
carbolymer, line 256: Staircase pattern: nested `case mResult` → `withObject "result"` → `withObject "transaction"`, then `Nothing` → `withObject "error"` - four levels of rightward drift.

Consider extracting `parseSuccess` and `parseError` helpers, e.g.:

```haskell
parseResponse = Aeson.withObject "OgmiosResponse" $ \obj -> do
    respId <- obj .:? "id"
    mResult <- obj .:? "result"
    result <- maybe (parseError obj) parseSuccess mResult
    return (respId, result)

parseSuccess :: Value -> Aeson.Parser OgmiosResult
parseSuccess = Aeson.withObject "result" $ \r -> do
    txObj <- r .: "transaction"
    txId <- Aeson.withObject "transaction" (.: "id") txObj
    return $ OgmiosSuccess txId

parseError :: Aeson.Object -> Aeson.Parser OgmiosResult
parseError obj = do
    errVal <- obj .: "error"
    Aeson.withObject "error" (\errObj ->
      OgmiosError <$> errObj .: "code" <*> errObj .: "message"
                  <*> (fromMaybe Null <$> errObj .:? "data")
      ) errVal
```

bench/tx-generator/src/Cardano/Benchmarking/Script/Submission.hs
```diff
@@ -0,0 +1,125 @@
...
+-- of mutually independent transactions, so submit it to the end and let the
+-- final tally decide.
+onRejectionFor :: Generator -> OnRejection
+onRejectionFor generator = case generator of
+  NtoM {}        -> ContinueOnRejection
+  Take _ g       -> onRejectionFor g
+  Cycle g        -> onRejectionFor g
+  Sequence gs
+    | any ((ContinueOnRejection ==) . onRejectionFor) gs -> ContinueOnRejection
+  _              -> AbortOnRejection
```
carbolymer: `onRejectionFor` is decided once for the whole stream, but `Sequence` can mix chained and independent sub-generators. If a low-level JSON script uses `Sequence [Split ..., NtoM ...]`, `NtoM` makes the whole sequence `ContinueOnRejection` - so a rejected `Split` tx doesn't abort the loop, and the rest of the chained split txs (plus the `NtoM` txs that depend on their outputs) are submitted anyway, all doomed.

The high-level compiler emits each phase as a separate `Submit` action, so this doesn't bite there. But the logic is still wrong for mixed `Sequence` generators in raw scripts.

One option: apply the policy per sub-generator instead of per stream - abort when the current sub-generator says abort, continue when it says continue. Alternatively, since `Sequence` with mixed dependency chains is an edge case, a simpler fix would be to just use `AbortOnRejection` for `Sequence` unconditionally (abort is always safe, just conservative).

bench/tx-generator/src/Cardano/Benchmarking/Script/Ogmios.hs
```diff
@@ -0,0 +1,241 @@
...
+  | OgmiosError Int Text Value
+  deriving (Eq, Show)
+
+-- | Parse a JSON-RPC 2.0 response to @submitTransaction@, returning the
+-- mirrored request id alongside the result.
+parseOgmiosResponse :: LBS.ByteString -> Either String (Maybe Int, OgmiosResult)
+parseOgmiosResponse bs =
+  case Aeson.eitherDecode bs of
+    Left err -> Left err
+    Right val -> Aeson.parseEither parseResponse val
```
carbolymer:
```suggestion
parseOgmiosResponse = fmap (Aeson.parseEither parseResponse) . Aeson.eitherDecode
```

bench/tx-generator/test-ogmios.sh
carbolymer, line 1: This looks like a good use case for `process-compose`, rather than writing own shell glue for handling processes.

bench/tx-generator/src/Cardano/Benchmarking/Compiler.hs
```diff
@@ -195,16 +198,27 @@ benchmarkingPhase wallet collateralWallet = do
...
   outputs <- askNixOption _nix_outputs_per_tx
   txParams <- askNixOption txGenTxParams
+  endpoint <- resolveSubmissionEndpoint
   doneWallet <- newWallet "done_wallet"
+  -- A submission endpoint is a functional submission transport, not a
+  -- benchmarking one: it ignores tps and targetNodes and produces no
+  -- submission metrics, so a config that asks for a real benchmark must fail
+  -- fast here instead of running unpaced and unmeasured.
+  submitMode <- case (endpoint, debugMode) of
+    (Just (eType, uri), True)  -> pure $ SubmitToEndpoint eType uri
```
carbolymer: I think `uri` should be already parsed at this stage instead of being `String`

bench/tx-generator/src/Cardano/Benchmarking/Script/Types.hs
```diff
@@ -185,6 +187,7 @@ data SubmitMode where
   DumpToFile  :: !FilePath -> SubmitMode
   DiscardTX   :: SubmitMode
   NodeToNode  :: NonEmpty NodeIPv4Address -> SubmitMode --deprecated
+  SubmitToEndpoint :: !SubmissionEndpointType -> !String -> SubmitMode
```
carbolymer: I am not convinced if this is the most convenient representation here. I'd just add separate `SubmitModes` for each backend i.e. keep ogmios here instead in `SubmissionEndpointType`.

bench/tx-generator/src/Cardano/Benchmarking/Script/Submission.hs
carbolymer, line 1: I like that this module is abstracted away from the supported backends. 👍🏻

carbolymer (commented):
LGTM but I'm leaving approval for perf team

bench/tx-generator/src/Cardano/Benchmarking/Script/Ogmios.hs
```diff
@@ -101,102 +94,79 @@ defaultOgmiosPort = 1337
...
-  _              -> AbortOnRejection
-
-submitAllOgmios :: forall era. IsShelleyBasedEra era
-  => OnRejection -> String -> TxStream IO era -> ActionM ()
-submitAllOgmios onRejection url txStream =
+-- | A transaction rejection reported by Ogmios, carried as the error type of
+-- the 'SubmitTransport' and rendered for tracing via 'Pretty'. Deliberately
+-- not exported: the submission loop only needs to render it, so no other
+-- module should depend on its shape.
+data OgmiosRejection = OgmiosRejection !Int !Text !Value
```
Jimbo4350: Haddocks for this type is missing. What is `Int`? What is `Text`?

Jimbo4350:
On [`Submission.hs` line 11](https://github.com/IntersectMBO/cardano-node/blob/1c2ad4bcdca481a0716fb0d17d25f6671d0d7b5b/bench/tx-generator/src/Cardano/Benchmarking/Script/Submission.hs#L11):

Nit: "the script monad" — did you mean `ActionM`? It's imported two lines below and used directly in the signature, so naming/linking it (`'ActionM'`) would be more precise than the informal phrase

bench/tx-generator/src/Cardano/Benchmarking/Script/Types.hs
Jimbo4350: I'm good with either. `String` needs a haddock though and the `Type` in `SubmissionEndpointType` is redundant.

Jimbo4350 (commented):
Closer. I'll have another look when you've reconciled @carbolymer 's comments (where you see fit).

bench/tx-generator/src/Cardano/Benchmarking/Script/Ogmios.hs
palas: Not for `SubmitTransactionResponse`, it is required and it must be an object with exactly one `transaction` sub-object in it, according to this: https://ogmios.dev/api/#operation-subscribe-/?SubmitTransaction
But I didn't know `.:?` also returned `Nothing` on `null`

bench/tx-generator/src/Cardano/Benchmarking/Script/Ogmios.hs
palas: Do you mean even submitting? Yes, apparently it can, good call

bench/tx-generator/src/Cardano/Benchmarking/Script/Submission.hs
palas: Hm, for the low-level, it is tricky to know what to do. Continuing is more conservative. But I am not even sure we should abort on split, because maybe nothing depends on it. Maybe we should just abort always on low level, or something like that

bench/tx-generator/src/Cardano/Benchmarking/Script/Types.hs
palas: > I'd just add separate SubmitModes for each backend i.e. keep ogmios here instead in SubmissionEndpointType.

That is what I did originally, and then I changed it so that it wasn't so ogmios specific. In retrospective, I kind of like the current approach, and I think it will probably make adding other endpoints easier in the future. It may be that when we implement gRPC we want to do it differently, but well...

palas: > I'm good with either. String needs a haddock though and the Type in SubmissionEndpointType is redundant.

I think removing `Type` here would be misleading, because `SubmissionEndpoint` suggests that it is a URL, and it is actually something like "Ogmios"/"gRPC". It doesn't represent a type in the Haskell sense.

bench/tx-generator/src/Cardano/Benchmarking/Script/Ogmios.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-node/pull/6609/changes/7716d4b7a954eddcaf06c4e8682453397b7c11df

bench/tx-generator/src/Cardano/Benchmarking/Script/Ogmios.hs
palas: Applied here: https://github.com/IntersectMBO/cardano-node/pull/6609/changes/cf9b22943d1fb35def8a8fafacdea7d9b926703c

bench/tx-generator/src/Cardano/Benchmarking/Script/Ogmios.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-node/pull/6609/changes/6ab928516fe8ec51ee4bcd8bb794a308b784b654

bench/tx-generator/src/Cardano/Benchmarking/Script/Submission.hs
palas: I've changed it so that `Sequence` always aborts in: https://github.com/IntersectMBO/cardano-node/pull/6609/changes/0a2d98abf0cb4d23d25d513ed4e4dbee8c7f23be

bench/tx-generator/src/Cardano/Benchmarking/Compiler.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-node/pull/6609/changes/cf2c1a6357b22c0c1822e533009a67eedc1bc190

bench/tx-generator/test-ogmios.sh
palas: I am not sure which one you mean. Do you have a link? But in any case, this script is a bit ad-hoc/disposable, it doesn't even go in the CI, at least for now

bench/tx-generator/src/Cardano/Benchmarking/Script/Types.hs
palas: In the end, I have renamed it to `SubmissionEndpointProtocol` here: https://github.com/IntersectMBO/cardano-node/pull/6609/changes/8c635b0bb1ff5f0632fc60f84be37c89b559a20d

bench/tx-generator/src/Cardano/Benchmarking/Script/Ogmios.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-node/pull/6609/changes/8c635b0bb1ff5f0632fc60f84be37c89b559a20d

Jimbo4350 (approved):
Nice work 👍

mgmeier (changes requested):
The test script sits at the package root, unreferenced by the cabal file, CI, or any nix file; unprecedented in this project. This means, weak conformance guarantees, and likely bit-rot.

It also defaults to `ref=testnet-tx-gen-tests` on the Ogmios side, which does not look stable.

I wonder if it could be a more stable test when seen from and implemented on the Ogmios side. @carbolymer has already pointed at using a nicer abstraction than plain shell script.

I'm hesitant to keep it with the `tx-generator`.

mgmeier:
Another feature request I'd have for this PR: Make the new mode of operation configurable via nix.
Basically, register the new options with `nix/nixos/tx-generator-service.nix`.

mgmeier:
My last RFC concerns the semantic expansion of `debugMode`. It cannot be treated according to its original scope any longer, as it now doubles as a confirmation for the new modus operandi (MO). While I understand the motivation to draw a hard line between the Ogmios submission mode and an actual benchmark - a perfectly valid distinction - I don't think piggy-backing on that flag is the best choice.

The `submissionEndpointURI` is a well-formed absolute URI, required together with its protocol — that's already a deliberate, hard-to-do-by-accident act. A boolean twin field (whether it's `debugMode` or a new dedicated one) adds ceremony without adding much real protection against a config error - IMHO. Furthermore, it only works with the `JsonHL` (high-level) command, where as low-level values can't rely on it. Last but not least, it currently depends on the ordering of steps in `Compiler.hs`, and the fact that a `RWST` monad with a lazy compiler error is used there. Potential refactors may break this unknowingly.

My suggestions:
* Drop the semantic expansion of `debugMode`. The added safety is not that large.
* If the "opt-in" into the new submission MO should be explicit - we can think of something else.
  * One possibility would be allowing an empty list for `_nix_targetNodes`. Backwards-compatible, but Ogmios, or any other `SubmissionEndpointProtocol` could require this to be empty, and a benchmark to be non-empty.
  * Introduce a boolean twin field, such as `_nix_submissionEndpointConfirmed :: Bool`. Not my preferred option, but would still be more explicit than `debugMode` - and being explicit is what you were going for.
  * Whatever happens - trace evidence of what the service does :)
  * Something else???

mgmeier (approved):
Great work @palas - thank you for adding this feature to `tx-generator`.

PR #6611, opened 2026-06-29
wb | workbench cleanup, static Make targets, advance genesis

fmaste:
Removes dead UTxO-HD/RTS knobs and their profiles, replaces the generated Make targets with explicit static rules, and advances the epoch-300 profiles to Voltaire while keeping their 64k block size.

- **wb | remove lmdb support with its profiles**: drop the LMDB UTxO-HD backend (`P.lmdb`, `Node.utxo_lmdb`, the `V1LMDB` ledger config) and its 4 profiles (`ci-bench-lmdb`, `value-volt-lmdb-nomadperf`, `value-volt-lmdb-cgmem-nomadperf`, `plutus-volt-lmdb-nomadperf`).
- **wb | remove gc parallel and load balance support and its profiles**: drop the `-qg1`/`-qb1` RTS knobs (`rtsGcParallel`/`rtsGcLoadBalance`) and their 4 `rtsqg1` profiles.
- **wb | remove RTS profiling from profile definitions, shell parameter only** — drop `rtsHeapProf` and the 3 `6-dense-*-rtsprof` profiles; the `6-dense-*-rtsprof` targets now inject `-hT -l` via `WB_PROFILING` at the Make level (a non-profiled build), instead of baking RTS profiling into the profile.
- **wb | rewrite Make shell targets as explicit static rules, drop jq**: the per-profile shell targets were generated by a `proftgt` macro built out of `eval`/`call`/`foreach` (meta-programming that produced the targets dynamically: hard to read and impossible to grep). Replace it with one explicit, literal, greppable static-pattern-rule stanza per variant (`-nix`, `-prof`, `-auto`, `-autostay`, `-nomadexec`, cloud, ...), each flag emitted as an explicit `ARG_*_YES/NO` fragment. Also move the workbench Make code under `nix/workbench/` (`lib.mk`, `profiles.mk`) and drop `jq` plus the `nix build && cat result` dance from `list-profiles`/`show-profile`/`ps`.
- **wb | move all profiles using genesis epoch 300 to voltaire keeping 64k block size**: switch every `genesisVariant300` profile to the new `genesisVariantVoltaire64k` (epoch 537 + the `blocksize64k` overlay), advancing the genesis while preserving the 64k `maxBlockBodySize` those profiles already had


mgmeier (approved):
LGTM

PR #6440, opened 2026-02-13
bench: PlutusV3 and PV11 scripts, profiles, and calibration tooling

fmaste:
The PR extends the Plutus benchmarking library with new scripts targeting Plutus V3 and Protocol Version 11, wires them into the benchmarking toolchain end-to-end, and adds supporting profile and reporting infrastructure.

On the library side, `plutus-scripts-bench` gains PlutusV3 variants of the existing ECDSA and Schnorr signature loop scripts, a V3 `CustomCall` variant, and two new PV11 scripts: `MultiScalarMulG1`, a BLS12-381 multi-scalar multiplication benchmark based on CIP-0133, and `ExpModInteger`, a fixed-iteration loop exercising the new `expModInteger` builtin. Alongside the additions, `mkPlutusBenchScript` was refactored to accept raw serialised UPLC bytes and a `PlutusScriptVersion` rather than a pre-wrapped `ScriptInAnyLang`, which eliminates repetitive boilerplate that had accumulated across every script module. Unused GHC extensions and redundant imports were removed throughout.

The tx-generator integration includes updated `scripts-fallback/` entries and redeemer JSON files for all new and updated scripts, a PV11 protocol parameters file for use with `calibrate-script`, and removal of the now-stale `scripts-legacy/` entries for the 2023-era ECDSA and Schnorr scripts.

On the profile and workbench side, the PV11 preview genesis overlay is updated with van Rossem cost models synced from mainnet. The following new benchmark profiles are added:

- `plutuscall-voltv11-ripemd`, `plutuscall-voltv11-schnorrv3`, `plutuscall-voltv11-mscalmul`, `plutuscall-voltv11-expmod` — PV11 plutuscall profiles for the new and updated scripts
- `ci-bench-v11-plutus`, `ci-bench-v11-plutusv3` — miniature CI profiles for PV11 Plutus workloads
- `plutusv3-expmod-nomadperf` — profile for the nomad cluster under PV11

Finally, `locli` gains a `PlutusBudgetSummary` module that outputs a Plutus workload calibration CSV as part of comparison reports.

PR #6615, opened 2026-07-10
bench: add ExpModInteger benchmark script and plutuscall-voltv11-expmod profile

Russoul:
## Summary
- Port a fixed-iteration `expModInteger` benchmark loop into `plutus-scripts-bench`, decoding a flat `(n, a, m)` from the redeemer (base `a=3`, prime modulus `m=2^61-1`), with `e` incrementing by 1 per iteration rather than doubling, to keep per-call cost roughly constant across the run and avoid any compiler hoisting of a loop-invariant computation.
- Each iteration force-evaluates `expModInteger a e m` (`let !_ = ...`) rather than checking it against a running expected result — this is a benchmark script, not a correctness test, so there's nothing to carry forward between iterations. Includes the same "redeemer counter must be >= 1000000" sanity check the other benchmark scripts have.
- Wire the script into `getAllScripts`/the cabal `other-modules` list, and add a `data/ExpModInteger.redeemer.json` + `scripts-fallback/ExpModInteger.plutus` for `tx-generator`/`calibrate-script` resolution.
- Add a calibrated `plutuscall-voltv11-expmod` profile (`LimitTxPerBlock_4`, fee from `calibrate-script run ExpModInteger txperblock_4`) and register it via `wb_profiles.mk`.

Calibration and the profile were verified against real local cluster runs (`make plutuscall-voltv11-expmod-auto`), confirming genuine `ExpModInteger` transaction submissions with the expected redeemer values. The script and its calibrated `txFee` were iterated on a few times (inlining helpers, flattening the redeemer, dropping the correctness check), each time re-verified with `calibrate-script` and a real cluster run.

Note: the redeemer's loop-counter placeholder must not equal `1000000` — `scriptDataModifyNumber`'s "already found a number to change" check is a before/after equality comparison, so a placeholder that already equals the reset target is a false negative and the reset falls through to the next number instead. Documented inline where this matters.


mgmeier:
The changes from this PR have been cherry-picked into #6440 and will be merged as part of that PR.

PR #6617, opened 2026-07-13
Remove local trace-resources package

jutaro:
Removes the locally-vendored trace-resources/ package directory and its local-package entry in cabal.project.
trace-resources is already published to CHaP, so cabal/nix now resolve it from there instead of building it locally — same as any other non-local dependency.
Drops now-stale references to the deleted directory from CODEOWNERS, CHANGELOG.md, CONTRIBUTING.md, .github/workflows/stylish-haskell.yml, and the schema-gen README.
No dependency renames, no new nix inputs, no behavior change.


mgmeier (approved):
Execllent, thank you @jutaro

PR #6620, opened 2026-07-16
fix(node): handle SIGTERM across startup phases

paolino:
# Description

Handle `SIGTERM` correctly while `cardano-node` is loading its configuration.

A `SIGTERM` received during startup was previously delivered as `ExitSuccess`. Configuration parsing could catch and wrap that exception, producing a generic parse error and making the process exit with status 1 instead of shutting down successfully.

This change introduces a phase-aware signal path:

- During startup, `SIGTERM` throws a dedicated asynchronous `SigTermException`.
- The top-level exception handler converts it to `ExitSuccess`.
- Immediately before diffusion starts, the signal handler switches back to `ExitSuccess`, which the diffusion layer already recognizes as an expected runtime shutdown.

This prevents configuration parsers from treating startup termination as a parsing failure while preserving the existing runtime shutdown behavior.

## Regression tests

The deterministic `cardano-node-test` properties cover the signal path without relying on process-start timing:

- `SigTermException` is classified as `SomeAsyncException`, so ordinary configuration-parser exception handling cannot wrap it.
- The top-level exception handler converts startup `SigTermException` to `ExitSuccess`.
- Startup and runtime phases select their intended exception types.
- On Unix, the real POSIX handler is installed before the test raises `SIGTERM`; the top-level handler must then exit successfully.

The installed-handler property passed 100 tests locally. The previous process-level test was removed because its fixed startup delay raced with handler installation on Linux and macOS CI.

## Antithesis reproduction

Antithesis originally exposed this intermittent failure in run `75fa032ed4fd4931ff63d020d9f48208-56-17`.

Failing property:

```text
container: p4, exit code: 1
```

Relevant run logs:

```text
vtime 5404.813501  p4 stderr: OtherParseException ExitSuccess
vtime 5404.813501  p4 stderr: cardano-node: Generic parse exception:
vtime 5404.813501  p4 stderr: ExitSuccess
vtime 5405.110873  containers_meta: p4 died, exit code 1
```

The patched image was subsequently exercised in three-hour Antithesis run `ce9c380a77c6c17f77edae0f35146a08-56-17`:

- 65 properties passed, 0 failed
- `OtherParseException`: 0 events
- `Generic parse exception`: 0 events

The clean three-hour run is broader fault-injection regression coverage, not proof of the fix by itself: the failure is intermittent and unpatched runs also completed without reproducing it. The deterministic properties above provide red/green coverage of async classification, top-level conversion, phase selection, and actual POSIX signal delivery.

The Antithesis reports are restricted to authorized tenant users, so the stable run IDs and reviewer-safe log excerpts are included here instead of access-dependent report URLs.

## Local verification

- `nix develop --quiet -c cabal test cardano-node-test` passes; all five top-level SIGTERM properties pass 100 tests.
- `nix build --no-link .#hydraJobs.x86_64-linux.windows.cardano-node` passes the same Windows cross-build that previously failed under `-Werror`.
- `nix develop --quiet -c cabal build cardano-testnet-test` passes after removing the timing-based property.
- `hlint` reports `No hints` for all touched Haskell files.

# Checklist

- [x] New tests are added if needed and existing tests are updated
- [x] No additional changelog entry is needed; the existing Next-version entry already notes clean exit on `SIGTERM`
- [x] No version bounds need updating; there are no dependency changes
- [ ] CI passes. The following CI checks are required:
  - [x] Code is linted with `hlint`
  - [x] Code is formatted with `stylish-haskell`
  - [ ] Code builds on Linux, macOS and Windows for `ghc-9.6` and `ghc-9.12`
- [x] Self-reviewed the diff

# Note on CI

This PR is from a fork, so the necessary CI jobs will need to be triggered by someone with write privileges.

### DIFF
--- a/cardano-node/cardano-node.cabal
+++ b/cardano-node/cardano-node.cabal
@@ -249,6 +249,7 @@ test-suite cardano-node-test
                         Test.Cardano.Node.Gen
                         Test.Cardano.Node.Json
                         Test.Cardano.Node.POM
+                        Test.Cardano.Node.TopLevel
                         Test.Cardano.Tracing.NewTracing.Consistency
 
   ghc-options:          -threaded -rtsopts "-with-rtsopts=-N -T"
--- a/cardano-node/src/Cardano/Node/Handlers/TopLevel.hs
+++ b/cardano-node/src/Cardano/Node/Handlers/TopLevel.hs
@@ -1,8 +1,14 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE TypeApplications #-}
+
+#if !defined(mingw32_HOST_OS)
+#define UNIX
+#endif
 
 module Cardano.Node.Handlers.TopLevel
   ( SigTermException (..)
   , SigTermPhase (..)
+  , installSigTermHandler
   , throwSigTerm
   , toplevelExceptionHandler
   ) where
@@ -53,12 +59,27 @@ import qualified Ouroboros.Network.Diffusion.Types as Network
 
 import           Prelude
 
-import           Control.Concurrent (ThreadId)
+import           Control.Concurrent
+                   ( ThreadId
+#ifdef UNIX
+                   , mkWeakThreadId
+                   , myThreadId
+#endif
+                   )
 import           Control.Exception
+#ifdef UNIX
+import           Control.Monad (forM_, void)
+#endif
 import           Control.Monad.Class.MonadAsync (ExceptionInLinkedThread (..))
+#ifdef UNIX
+import           GHC.Weak (deRefWeak)
+#endif
 import           System.Environment
 import           System.Exit
 import           System.IO
+#ifdef UNIX
+import qualified System.Posix.Signals as Signals
+#endif
 
 -- | Internal async exception used to route SIGTERM through the top-level
 -- handler without letting ordinary exception handlers catch it.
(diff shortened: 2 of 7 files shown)


Jimbo4350 (approved):
Nice fix. Walking through the root cause for posterity, since the "async" framing is subtle:

**The bug in one line:** the startup SIGTERM handler delivered `ExitSuccess` via `throwTo` — an *asynchronous delivery* of a value whose *type* (`ExitCode`) is **synchronously classified** (its `Exception` instance doesn't route through `SomeAsyncException`).

That mismatch bites because of *where* it lands. During startup the main thread is inside `Data.Yaml.decodeFileThrow`, which bottoms out in yaml's `mkHelper`:

```haskell
catches (parse …)
  [ Handler $ \pe  -> Left pe                              -- ParseException
  , Handler $ \ye  -> Left (InvalidYaml (Just ye))         -- YamlException
  , Handler $ \sae -> throwIO (sae :: SomeAsyncException)  -- async → rethrow untouched
  , Handler catchLeft ]                                    -- everything else → OtherParseException
```

`catches` matches purely on *type*, never on how the exception was delivered. `ExitSuccess` isn't in the `SomeAsyncException` box, so it skips handler #3 and falls into the catch-all #4 → `OtherParseException ExitSuccess`, which `decodeFileThrow` rethrows as a plain parse error → the top-level handler prints `Generic parse exception: ExitSuccess` and exits **1**. That's the Antithesis failure.

**The fix** keeps delivery identical and only changes the type classification — `SigTermException`'s `Exception` instance routes through `asyncExceptionToException`/`asyncExceptionFromException`, which is exactly what places its values in the `SomeAsyncException` box that handler #3 checks. So #3 now rethrows it untouched and `toplevelExceptionHandler` converts it to `ExitSuccess` → exit **0**. Runtime shutdown is unchanged (still `ExitSuccess`, which diffusion already treats as expected).

LGTM 👍

PR #6631, opened 2026-07-28
Migrate cardano-node-chairman, cardano-submit-api and cardano-testnet to herald

carbolymer:
# Description

Replaces `cardano-testnet`'s scriv changelog tooling with [herald](https://github.com/input-output-hk/cardano-dev/tree/main/herald), and extends changelog automation to `cardano-submit-api` and `cardano-node-chairman`.

Fragments live in per-package `.changes/` directories, so the `project:` field in each fragment can be inferred from the directory rather than written by hand. This needs herald 0.1.2.0 (per-package `changes-dir` support), which is why both new workflows pin `herald-ref: github:input-output-hk/cardano-dev/herald-0.1.2.0` explicitly - the released `herald-validate`/`herald-release` actions still default to herald 0.1.1.0.

The 27 pending scriv fragments in `cardano-testnet` were converted to herald YAML. Category mapping:

| scriv category | herald kind |
| --- | --- |
| Removed | breaking |
| Added | feature |
| Changed / Deprecated | compatible |
| Fixed / Security | bugfix |
| Tests | test |
| Documentation | documentation |
| Maintenance | maintenance |

Each fragment's `pr:` was resolved from the commit that added the original scriv file, so changelog entries keep linking to the PR that made the change. Original filename stems are preserved so ordering is unchanged.

`cardano-node-chairman` gains a `CHANGELOG.md` (it had none, and herald requires the file to exist before it can batch fragments into it) and is added to the root changelog index.

The stale `## vNext` heading is removed from `cardano-submit-api`'s changelog, because herald inserts each new release section immediately above the first `## ` heading.

scriv is removed: `.github/workflows/check-changelog.yml` and the `scriv` dev-shell entry in `nix/haskell.nix`.

## Reviewer notes

- The `--pr` check in `herald-validate` reports mismatches for the converted fragments, because they intentionally carry their *original* PR numbers (the PR that made the change they describe) rather than this PR's number.
- Both new checks (`herald-validate`, and the release workflow) are intentionally **not** required, so they do not block CI.
- scriv's `no-changelog-needed` label escape hatch has no herald equivalent: once the validate check is made required, every PR touching `cardano-node-chairman/`, `cardano-submit-api/` or `cardano-testnet/` will need a changelog fragment.
- herald attributes a changed file to a project by the directory of that project's `changelog:` path, so PRs touching other packages (`cardano-node`, `cardano-tracer`, `trace-forward`, `cardano-node-capi`) are unaffected. The `cardano-node` / `cardano-node-chairman` name-prefix overlap does not cross-attribute, since attribution is by directory, not name.

## Verification

- `herald validate` (0.1.2.0) reports `All changelog fragments valid.`
- `herald next cardano-testnet` yields `11.1.0`.
- `actionlint` passes on both new workflows.

# Checklist

- [ ] New tests are added if needed and existing tests are updated. These may include:
  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] CI passes. See note on CI. The following CI checks are required:
  - [ ] Code is linted with `hlint`. See `.github/workflows/check-hlint.yml` to get the `hlint` version
  - [ ] Code is formatted with `stylish-haskell`. See `.github/workflows/stylish-haskell.yml` to get the `stylish-haskell` version
- [x] Self-reviewed the diff

# Note on CI
You will need to get someone with write privileges. Please contact IOG node developers to do this
for you.


.github/workflows/herald-validate.yml
```diff
@@ -0,0 +1,45 @@
...
+    # PR number context is unavailable.  The trigger is kept so the check
+    # reports "skipped" (passes) when this is a required status check.
+    if: ${{ github.event_name != 'merge_group' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          # Fetch full history so herald can determine the fork point
+          # for --diff and --pr checks
+          fetch-depth: 0
```
palas, line 28:
```suggestion
          fetch-depth: 0
          persist-credentials: false
```
It is not a big deal, but for consistency...

.github/workflows/herald-release.yml
```diff
@@ -0,0 +1,61 @@
...
+#
+# What it does:
+#   1. Batches changelog fragments into a new changelog section
+#   2. Bumps the version (cabal file or version.txt, auto-computed or explicit)
+#   3. Creates a release commit (tag is deferred to the manual signing step)
+#   4. Adds a release changelog fragment for the next cycle
+#   5. Pushes to a release/PACKAGE-VERSION branch
+#   6. Opens a PR with signing instructions (and optionally CHaP submission commands)
+
+name: Herald release
```
palas, line 13: I am not familiar on how this action works, but currently this repo only releases `cardano-node`, so I am not sure how releases of other packages are going to go. I enquired in the past, and I was told it would be best to release everything at the same time as node, so I don't know how that plays with herald. Worth keeping in mind

cardano-testnet/.changes/20260219_005200_palas_scriv_testnet.yml
```diff
@@ -0,0 +1,7 @@
+pr: 6446
+kind:
+  - maintenance
+description: |
+  Added `scriv` to nix development shell
+  Set-up `scriv` for tracking changelogs under `cardano-testnet`
+  Added CI action that checks that changelog entries are added when `cardano-testnet` is modified
```
palas: I suspect these are going to be rendered as a single line in markdown, and because there isn't even punctuation at the end of the line (my bad), it is going to be a bit uninteligible. Maybe would be best to add the punctuation and/or add a line in between.

palas (approved):
It looks all correct. Just a couple of really minor nits

Jimbo4350 (approved):

.github/workflows/herald-release.yml
carbolymer: This action just collects changelogs, bumps and tags the version.

PR #6639, opened 2026-08-05
CDFs for EB announcements

### LINKED ISSUE #1026: Measure diffusion distributions
## Why

We need to know how well the node is performing in a given network in terms of block diffusion. This is essential for the protocol security argument. It also gives us the visibility whether we need to improve things. Consistently fast diffusion will give us high confidence in the implementation (for a given network)

## What

- For EB announcements
- For EB bodies
- For EB closures
- Like the blocks diffused within 1s and 3s CDFs
- CDF over multiple entities read from many nodes in a given time window, directly from the traces.
- Use existing tracers or add new ones, and derive prometheus metrics in the node for them 

Image

coot:
# Description

- **CDF counters - refactorisation**
- **CDFs for EB announcements**

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/cardano-node/src/Cardano/Node/Tracing/Cdf.hs
+++ b/cardano-node/src/Cardano/Node/Tracing/Cdf.hs
@@ -1,0 +1,104 @@
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+-- | CDFs
+--
+-- This module should be imported qualified.
+--
+module Cardano.Node.Tracing.Cdf
+  ( Counter (..)
+  , Config (..)
+  , State
+  , size
+  , null
+  , empty
+  , minPriority
+  , defaultConfig
+  , processDataPoint
+  ) where
+
+import           Prelude hiding (null)
+import           Data.Int (Int64)
+import           Data.IntPSQ (IntPSQ)
+import qualified Data.IntPSQ as Pq
+import           Data.Time (NominalDiffTime)
+
+data Counter = Counter {
+    limit   :: !Double
+  , counter :: !Int64
+}
+
+decCdf :: Double -> Counter -> Counter
+decCdf v cdf@Counter{..}
+  | v < limit = cdf {counter = counter - 1}
+  | otherwise = cdf
+
+incCdf :: Double -> Counter -> Counter
+incCdf v cdf@Counter{..}
+  | v < limit = cdf {counter = counter + 1}
+  | otherwise = cdf
+
+
+-- | We keep the results in a priority queue, to be able to evict the oldest
+-- entry when the data sets become larger than `numOfDataPoints`.
+--
+newtype State p = State { cdfState :: IntPSQ p NominalDiffTime }
+
+empty :: State p
+empty = State Pq.empty
+
+null :: State p -> Bool
+null = Pq.null . cdfState
+
+size :: State p -> Int
+size = Pq.size  . cdfState
+
+minPriority :: Ord p => State p -> Maybe p
+minPriority State { cdfState } = case Pq.minView cdfState of
+    Nothing -> Nothing
+    Just (_, p, _, _) -> Just p
+
+newtype Config = Config { numOfDataPoints :: Int }
+
+-- | Default `Config` keeps `k/2` data points.
+defaultConfig :: Config
+defaultConfig = Config { numOfDataPoints = 1080 }
+
+
+processDataPoint
+  :: forall f p.
+     ( Ord p
+     , Functor f
+     )
+  => Config
+  -> (Int, p, NominalDiffTime)
+  -- ^ index, priority, value
+  -> State p
+  -> f Counter
+  -> Maybe (f Counter, State p)
+processDataPoint Config { numOfDataPoints } (idx, p, delay) (State m) cdfs
+    | idx `Pq.member` m
+    = Nothing
+
+    | otherwise
+    = if Pq.size m' > numOfDataPoints
+      then
+        case Pq.minView m' of
+          Nothing -> Nothing
+          Just (_, minVal, minDelay, m'')
+            | minVal == p
+            -> Nothing
+
+            | otherwise
+            -> Just (adjustCdf (realToFrac minDelay) <$> cdfs, State m'')
+      else
+        Just (updateCdf <$> cdfs, State m')
+  where
+    m' = Pq.insert idx p delay m
+
+    updateCdf :: Counter -> Counter
+    updateCdf = incCdf (realToFrac delay)
+
+    adjustCdf :: Double -> Counter -> Counter
+    adjustCdf d = updateCdf . decCdf d
--- a/cardano-node/src/Cardano/Node/Tracing/Tracers.hs
+++ b/cardano-node/src/Cardano/Node/Tracing/Tracers.hs
@@ -58,6 +58,8 @@ import qualified Ouroboros.Network.BlockFetch.ClientState as BlockFetch
 import           Ouroboros.Network.ConnectionId (ConnectionId)
 import qualified Ouroboros.Network.Diffusion as Diffusion
 
+import           LeiosDemoTypes (TraceLeiosKernel (..))
+
 import           Codec.CBOR.Read (DeserialiseFailure)
 import           Control.Monad (unless)
 import           "contra-tracer" Control.Tracer (mkTracer)
@@ -360,6 +362,15 @@ mkConsensusTracers configReflection trBase trForward mbTrEKG _trDataPoint trConf
                 ["Consensus", "LeiosKernel"]
     configureTracers configReflection trConfig [leiosKernelTr]
 
+    !leiosMetricsTr <- do
+      tr1 <- foldTraceM (\cm lc -> pure . calculateLeiosMetrics cm lc) initialLeiosMetrics
+                (metricsFormatter
+                  (mkMetricsTracer mbTrEKG))
+      pure $ filterTrace (\(_, msg) -> case msg of 
+                         TraceLeiosAnnouncementAccepted{} -> True
+                         _ -> False)
+               tr1
+
     !leiosPeerTr <- mkCardanoTracer
                 trBase trForward mbTrEKG
                 ["Consensus", "LeiosPeer"]
@@ -421,6 +432,7 @@ mkConsensusTracers configReflection trBase trForward mbTrEKG _trDataPoint trConf
           traceWith txCountersTracer
       , Consensus.leiosKernelTracer = mkTracer $
           traceWith leiosKernelTr
+          <> traceWith leiosMetricsTr
       , Consensus.leiosPeerTracer = mkTracer $
           traceWith leiosPeerTr
       }
--- a/cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
+++ b/cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
@@ -15,11 +15,16 @@
 {-# OPTIONS_GHC -Wno-orphans  #-}
 
 module Cardano.Node.Tracing.Tracers.Consensus
-  (
+  ( -- * ClientMetrics
     initialClientMetrics
   , calculateBlockFetchClientMetrics
   , servedBlockLatest
   , ClientMetrics
+    -- * Leios Metrics
+  , initialLeiosMetrics
+  , calculateLeiosMetrics
+  , LeiosMetrics
+    -- * Tx-Submission / Mempool
   , txsMempoolTimeoutSoftCounterName
   , txsSyncDurationTotalCounterName
   , impliesMempoolTimeoutSoft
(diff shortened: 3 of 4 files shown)


cardano-node/src/Cardano/Node/Tracing/Cdf.hs
```diff
@@ -0,0 +1,106 @@
...
+size = Pq.size  . cdfState
+
+minPriority :: Ord p => State p -> Maybe p
+minPriority State { cdfState } = case Pq.minView cdfState of
+    Nothing -> Nothing
+    Just (_, p, _, _) -> Just p
+
+data Config = Config { numOfDataPoints :: Int }
+
+-- | Default `Config` keeps `k/2` data points.
```
nfrisby, line 64: The proportionality to `k` is easier to justify for some CDFs than for others, but it's probably fine for most?

cardano-node/src/Cardano/Node/Tracing/Cdf.hs
```diff
@@ -0,0 +1,106 @@
...
+  :: forall f p.
+     ( Ord p
+     , Functor f
+     )
+  => Config
+  -> (Int, p, NominalDiffTime)
+  -- ^ index, priority, value
+  -> State p
+  -> f Counter
+  -> (f Counter, State p, Bool)
```
nfrisby: A couple questions that a comment should answer:

- What does the `Bool` mean?
- Why do we need to maintain the `Counter` and the `State` as separate things?

cardano-node/src/Cardano/Node/Tracing/Cdf.hs
```diff
@@ -0,0 +1,106 @@
...
+processDataPoint Config { numOfDataPoints } (idx, p, delay) s@(State m) cdfs
+    | idx `Pq.member` m
+    = nothingToDo
+
+    | otherwise
+    = if Pq.size m' > numOfDataPoints
+      then
+        case Pq.minView m' of
+          Nothing -> (cdfs, s, False)
```
nfrisby: use `nothingToDo` here too?

cardano-node/src/Cardano/Node/Tracing/Cdf.hs
```diff
@@ -0,0 +1,106 @@
...
+            -> nothingToDo
+
+            | otherwise
+            -> (adjustCdf (realToFrac minDelay) <$> cdfs, State m'', True)
+      else
+        (updateCdf <$> cdfs, State m', True)
+  where
+    nothingToDo = (cdfs, s, False)
+
+    delay_ :: Double
```
nfrisby: Just inline this as `(realToFrac delay :: Double)`?

cardano-node/src/Cardano/Node/Tracing/Cdf.hs
```diff
@@ -0,0 +1,106 @@
...
+  | v < limit = cdf {counter = counter - 1}
+  | otherwise = cdf
+
+incCdf :: Double -> Counter -> Counter
+incCdf v cdf@Counter{..}
+  | v < limit = cdf {counter = counter + 1}
+  | otherwise = cdf
+
+
+newtype State p = State { cdfState :: IntPSQ p NominalDiffTime }
```
nfrisby, line 46: Please add a comment explaining why a Cdf needs a notion of priority?

cardano-node/src/Cardano/Node/Tracing/Tracers.hs
```diff
@@ -360,6 +362,15 @@ mkConsensusTracers configReflection trBase trForward mbTrEKG _trDataPoint trConf
                 ["Consensus", "LeiosKernel"]
     configureTracers configReflection trConfig [leiosKernelTr]
 
+    !leiosMetricsTr <- do
```
nfrisby, line 365: :+1: Matches the extant pattern, eg:

https://github.com/IntersectMBO/cardano-node/blob/ee7ea74fc8e3990357844de678f26ed99732d2bf/cardano-node/src/Cardano/Node/Tracing/Tracers.hs#L258-L266

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
```diff
@@ -76,18 +83,17 @@ import           Control.Monad (guard)
...
 import           Data.Word (Word32, Word64)
+import           Data.IntPSQ (IntPSQ)
+import qualified Data.IntPSQ as Pq
 import           Network.TypedProtocol.Core
 
-import           LeiosDemoTypes (TraceLeiosKernel (..), TraceLeiosPeer (..),
+import           LeiosDemoTypes (AnnouncementFields (..), TraceLeiosKernel (..), TraceLeiosPeer (..),
                    traceLeiosKernelToObject, traceLeiosPeerToObject)
 import           LeiosUtils.CallTrace (SomeJsonCallTrace (..), callTraceToObject)
+import LeiosDemoLogic.Announcements.ElBimap (ElId(..))
```
nfrisby: inconsistent alignment and not in alphabetical order

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
```diff
@@ -2331,6 +2300,145 @@ instance MetaTrace KESAgentClientTrace where
 -- Leios
 --------------------------------------------------------------------------------
 
+-- | Leios metrics which construct CDFs for EB announcements.
+data LeiosMetrics' c = LeiosMetrics {
+    lmCdfState     :: Cdf.State SlotNo
+    -- ^ internal state which collects all data points
+  , lmSlotMapLate  :: IntPSQ SlotNo ()
```
nfrisby: Why isn't this just another `c`?

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
```diff
@@ -2331,6 +2300,145 @@ instance MetaTrace KESAgentClientTrace where
...
+             , cdfMetric "leios.eb.announcement.delay.cdf1400ms" lmCdf1400msVar
+             , cdfMetric "leios.eb.announcement.delay.cdf1600ms" lmCdf1600msVar
+             , cdfMetric "leios.eb.announcement.delay.cdf1800ms" lmCdf1800msVar
+             , cdfMetric "leios.eb.announcement.delay.cdf2000ms" lmCdf2000msVar
+             ]
+        else []
+    where
+      size                   = Cdf.size lmCdfState
+      cdfMetric name var     = DoubleM name (fromIntegral (Cdf.counter var) / fromIntegral size)
+      lateAnnoucementsAllMetric = [ CounterM "leios.eb.announcement.late.all" Nothing | lmDelay > 2 ]
```
nfrisby: Typos:

- `lateAnnoucementsAllMetric` -> `lateAnnoumcementsAllMetric`
- `lateAnnoucementsPerSlotMetric` -> `lateAnnoucementsPerSlotMetric`

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
nfrisby: And maybe `(Maybe SlotNo, Int)` would be a simpler data structure to understand?

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
```diff
@@ -2331,6 +2300,145 @@ instance MetaTrace KESAgentClientTrace where
...
+                          if annDelay > 2
+                          then Pq.insert (fromIntegral slotNo_) slotNo () lmSlotMapLate'
+                          else lmSlotMapLate'
+                    }
+  where
+    nothingToDo = lm {lmTraceIt = False}
+
+calculateLeiosMetrics lm _lc _ = lm
+
+instance LogFormatting LeiosMetrics where
```
nfrisby, line 2378: :+1: matches https://github.com/IntersectMBO/cardano-node/blob/ee7ea74fc8e3990357844de678f26ed99732d2bf/cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs#L535-L552

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
```diff
@@ -2331,6 +2300,145 @@ instance MetaTrace KESAgentClientTrace where
...
+      , ("leios.eb.announcement.delay.cdf400ms", "probability for EB announcemement to complete within 400ms")
+      , ("leios.eb.announcement.delay.cdf600ms", "probability for EB announcemement to complete within 600ms")
+      , ("leios.eb.announcement.delay.cdf800ms", "probability for EB announcemement to complete within 800ms")
+      , ("leios.eb.announcement.delay.cdf1000ms", "probability for EB announcemement to complete within 1000ms")
+      , ("leios.eb.announcement.delay.cdf1200ms", "probability for EB announcemement to complete within 1200ms")
+      , ("leios.eb.announcement.delay.cdf1400ms", "probability for EB announcemement to complete within 1400ms")
+      , ("leios.eb.announcement.delay.cdf1600ms", "probability for EB announcemement to complete within 1600ms")
+      , ("leios.eb.announcement.delay.cdf1800ms", "probability for EB announcemement to complete within 1800ms")
+      , ("leios.eb.announcement.delay.cdf2000ms", "probability for EB announcemement to complete within 2000ms")
+      , ("leios.e.b.announcement.late.all", "number of late EB announcements that took longer than 2s: all peers are included")
```
nfrisby: I wasn't expecting "all peers are included" here. These metrics are driven by `TraceLeiosAnnouncementAccepted`, which is deduped---it only fires the _first_ time some announcement arrives.

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
```diff
@@ -2331,6 +2300,145 @@ instance MetaTrace KESAgentClientTrace where
...
+      , ("leios.e.b.announcement.late.bySlot", "number of slots for which the earliest EB announcements took longer than 2s")
```
nfrisby: It looks to me like `lmSlotMapLate` is counting announcements rather than counting slots. Am I confused?

nfrisby: Overall, I'm not clear on what `lmSlotMapLate` is actually doing... partly because I'm not sure what `IntM` is, but if I recall correctly (I can't find its definition right now), it's a _gauge_, so it's not accumulating anything.

nfrisby (changes requested):

cardano-node/src/Cardano/Node/Tracing/Cdf.hs
nfrisby: But one other option is to just force each callers to choose a number

cardano-node/src/Cardano/Node/Tracing/Cdf.hs
nfrisby: There's only two call sites to this function and they both ignore the return value when the returned `Bool` is `False`.

So return `Maybe (f Counter, State p)` instead?

cardano-node/src/Cardano/Node/Tracing/Cdf.hs
coot: That's how it's done, we just happen to provide a `defaultConfig` which is used by all applications (block-fetch and leios announcements).  Do you have a better value for leios announcment?

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
coot: `IntM` indeed is a gauge, ~I wanted to count late slots in a fixed interval of time - but this is not how late blocks are counted.  I'll implement this differently to match block-fetch late block counter.~

coot: I wanted to count how many announcements where late in a given slot, unlike the `late.all` which counts all late announcements since node has started.  I think this is not needed, one can inter that from the `late.all` graph (by looking at the slope of the curve).

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
nfrisby: > all peers are included

What is this supposed to mean? I still don't understand

nfrisby: Also, there's a typo in the name: `.e.b.`

nfrisby (approved):
There are some "typo"-level minor issues in this thread https://github.com/IntersectMBO/cardano-node/pull/6639/changes#r3722072409 but otherwise looks good to me. Thanks for addressing my comments 👍

cardano-node/src/Cardano/Node/Tracing/Tracers/Consensus.hs
coot: Fixed.

PR #6641, opened 2026-08-05
Fix haddock generation in CI by using GHC 9.12.4

palas:
# Description

The GitHub Pages haddock generation workflow is currently broken. This PR tries to address it by adding a nix development shell with `ghc9124`, to the flake and using it when building the haddock documentation in `.github/workflows/github-page.yml`.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [x] Self-reviewed the diff

### DIFF
--- a/.github/workflows/github-page.yml
+++ b/.github/workflows/github-page.yml
@@ -26,8 +26,9 @@ jobs:
         extra_nix_config: |
           experimental-features = nix-command flakes
           allow-import-from-derivation = true
-          extra-substituters = https://cache.iog.io/
-          extra-trusted-public-keys = hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ=
+          # The substituters (cache.iog.io) come from the flake's nixConfig:
+          accept-flake-config = true
+          access-tokens = github.com=${{ secrets.GITHUB_TOKEN }}
 
     - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
       with:
@@ -35,20 +36,18 @@ jobs:
 
     # The `.#haddock` shell (see flake.nix) provides a GHC whose haddock does not
     # panic on `type data` declarations (GHC issue #25739): GHC 9.6.7's does.
-    - name: Fetch nix cache and update cabal indices
-      run: |
-        nix develop .#haddock --command \
-          cabal update
+    - uses: rrbutani/use-nix-shell-action@59a52b2b9bbfe3cc0e7deb8f9059abe37a439edf # v1
+      with:
+        devShell: .#haddock
+
+    - name: Update cabal indices
+      run: cabal update
 
     - name: Build whole project
-      run: |
-        nix develop .#haddock --command \
-          cabal build all
+      run: cabal build all
 
     - name: Build documentation
-      run: |
-        nix develop .#haddock --command \
-          cabal haddock-project --local --output=./haddocks --internal --foreign-libraries
+      run: cabal haddock-project --local --output=./haddocks --internal --foreign-libraries
 
     - name: Compress haddocks
       run: |
@@ -61,6 +60,16 @@ jobs:
       with:
         name: haddocks
         path: ./haddocks.tgz
+
+    # cabal only prints a summary line when building docs for a dependency
+    # fails; the actual haddock error is hidden in these logs.
+    - name: Upload cabal logs
+      uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
+      if: ${{ failure() }}
+      continue-on-error: true
+      with:
+        name: cabal-haddock-logs
+        path: ~/.cache/cabal/logs/
 
     - name: Deploy documentation to gh-pages 🚀
       if: github.ref == 'refs/heads/master'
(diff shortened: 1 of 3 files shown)


.github/workflows/github-page.yml
carbolymer, line 1: Could you pull the workflow simplifications from https://github.com/IntersectMBO/cardano-node/pull/6581? Concretely:

1. `rrbutani/use-nix-shell-action@v1` with `devShell: .#ghc9124` - enters the shell once, so each of the three `cabal` steps stops paying its own `nix develop` startup and shrinks to a plain `cabal ...` one-liner. Worth SHA-pinning like the other actions here.
2. `accept-flake-config = true` and `access-tokens = github.com=${{ secrets.GITHUB_TOKEN }}` in `extra_nix_config`, in place of the hand-copied substituter and key lines - the flake's `nixConfig` already declares cache.iog.io, so the workflow doesn't need its own copy that can drift.
3. The upload-cabal-logs-on-failure step (`~/.cache/cabal/logs/`, `if: failure()`) - next time haddock breaks, those logs are the first thing anyone will ask for.

The action version bumps from 6581 are already covered on master, and its `pull_request` trigger was a testing leftover - your `workflow_dispatch` does that job properly.

flake.nix
```diff
@@ -187,12 +190,24 @@
...
+        {
+          inherit (shell) devops workbench-shell;
+          default = shell.dev;
+          cluster = shell;
+          profiled = project.profiled.shell;
+        }
+        # Shell used by the "Haddock documentation" workflow (github-page.yml)
+        // optionalAttrs (hostPlatform.system == "x86_64-linux") {
+          ghc9124 =
+            (project.appendModule {
```
carbolymer: What about defining this next to `profiled` and `asserted` at the bottom of `nix/haskell.nix`?

```nix
haddocked = final.appendModule {
  compiler-nix-name = "ghc9124";
  shell.withHoogle = final.pkgs.lib.mkForce false;
  shell.tools = final.pkgs.lib.mkForce {};
};
```

Then the devShell line here mirrors the profiled one: `ghc9124 = project.haddocked.shell;`. It's the same species as those two - an alternative compilation config of the project - and it puts the compiler choice in the same file as the `doHaddock` version gate it exists to get past. I realise `windowsCompilerNixName` points the other way, but that one is cross-compilation wiring for ciJobs rather than a project variant, so `profiled`/`asserted` feel like the closer precedent. The x86_64-linux gating would stay here either way.

flake.nix
```diff
@@ -187,12 +190,24 @@
...
+      in
+        {
+          inherit (shell) devops workbench-shell;
+          default = shell.dev;
+          cluster = shell;
+          profiled = project.profiled.shell;
+        }
+        # Shell used by the "Haddock documentation" workflow (github-page.yml)
+        // optionalAttrs (hostPlatform.system == "x86_64-linux") {
+          ghc9124 =
```
carbolymer: Naming nit: the attr name bakes in the compiler version, so the next haddock-compiler bump has to touch this flake *and* three lines in `github-page.yml`. Named after its job instead:

```nix
haddock = (project.appendModule { compiler-nix-name = haddockCompilerNixName; ... }).shell;
```

the workflow becomes `nix develop .#haddock` and never changes again. You already made `haddockCompilerNixName` the single knob - the attr name is the only place still leaking the version.

nix/haskell.nix
```diff
@@ -137,8 +137,14 @@ let
...
-
-            # GHC 9.6.7 haddock panics on TopTx type family (tyConStupidTheta)
+          })
+          # The haddock of GHC < 9.12.3 panics on `type data` declarations, e.g. the
+          # TopTx type family or AllLedgerPeers ("tyConStupidTheta" panic, GHC issue
+          # #25739, fixed in GHC 9.12.3). Disable haddock of the affected dependencies
+          # for those compilers only, so that shells with a fixed compiler (e.g. the
+          # `.#ghc9124` one used by github-page.yml) still provide dependencies with
+          # documentation.
+          ({ config, lib, ... }: lib.mkIf (builtins.compareVersions config.compiler.version "9.12.3" < 0) {
```
carbolymer: Nit: `lib.versionOlder config.compiler.version "9.12.3"` says the same thing and reads at a glance.

carbolymer (approved):

Jimbo4350 (approved):

flake.nix
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/pull/6641/changes/b78bf67f99f5c589820eb2f89371c36b053b0d6a

flake.nix
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/pull/6641/changes/b78bf67f99f5c589820eb2f89371c36b053b0d6a

nix/haskell.nix
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/pull/6641/changes/b78bf67f99f5c589820eb2f89371c36b053b0d6a

.github/workflows/github-page.yml
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/pull/6641/changes/90fa6bcd71ed8bb38276da52ea04ec612d683b96

PR #6643, opened 2026-08-06
wb | fschain-6912k profiles for 6.75MiB-block chain creation (+ fix broken blocksize768k overlay)

Russoul:
## Description

Adds `fschain-6912k` / `fschain-6912k-xs` — the next scale step after `fschain-768k` for creating benchmark chain fragments: 7077888-byte (6.75 MiB) blocks holding **~18.8k value txs each (~9x the 2086 of the 768k chain)**, at the same 0.015 block density and Oct2021 dataset (4M utxo, 1M delegators).

### Regression fix (first commit)

When the epoch-timeline format changed, overlay files started being parsed as an `Epoch` object with protocol parameters under an `epoch_params` key. `blocksize64k.json` was migrated, but `blocksize768k.json` was not — so it silently decoded as an *empty* overlay, and `fschain-768k*` profiles generated chains with the default 88 KiB block size instead of 768 KiB. Fixed by adding the missing wrapper.

### New profiles (second commit)

- New `blocksize6912k` pparams overlay (6.75 MiB) and primitive.
- `durationChain`/`durationChainM` parametrized by tx count; the 768k profiles keep their 90000.
- `valueXXLBlock`: 900 tps submission pressure (block consumption at saturation is ~282 tx/s; the margin mirrors `fschain-768k`'s 100 tps vs ~31 tx/s).
- `V.fundsQuadruple` (40T) and `baseNoFunds`: the tx-generator's split tree needs ~21.3M lovelace per submitted tx, so 800k txs exceed the 10T default (`FS.base` semantics unchanged for other profiles).

### Verification

`fschain-6912k-xs` run on the supervisor backend: after the generator's split phase (~slot 780), all value-tx blocks are saturated at 7039682–7039688 bytes (99.5% of the cap, no room for one more tx) with ~18.8k txs per block; the mempool stays pinned at its 2-blocks capacity (37.6k txs). A full 3000-slot `fschain-6912k` run produced 49 blocks, of which **35 are saturated value-tx blocks** (7039682–7039689 bytes, ~18.8k txs each) starting at slot ~565 — comfortable margin over the ~19 viable blocks needed for a chain fragment even under an unlucky leader schedule (an earlier 2400-slot attempt drew 1.9σ low and yielded only 14).


fmaste (approved):
LGTM

PR #6647, opened 2026-08-10
Remove constraint on crypton-x509-system

erikd:
# Description

This constraint was a blocker on upgrading to crypton >= 1.1 and the migration from the memory package to the ram package.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -22,7 +22,9 @@ active-repositories:
 
 constraints:
   -- haskell.nix patch does not work for 1.6.8
-  , any.crypton-x509-system < 1.6.8
+  -- grapesy currently depends on crypton-x509-system (>=1.6 && <1.7)
+  , any.crypton-x509-system < 1.6.8 || >= 1.7
+
   -- haskell.nix tries to use 0.7.1.5 for some reason
   , any.proto-lens >= 0.7.1.7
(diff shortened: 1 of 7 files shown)


cabal.project
```diff
@@ -13,16 +13,18 @@ repository cardano-haskell-packages
...
 
 active-repositories:
   , :rest
   , cardano-haskell-packages:override
 
 constraints:
   -- haskell.nix patch does not work for 1.6.8
-  , any.crypton-x509-system < 1.6.8
+  -- grapesy currently depends on crypton-x509-system (>=1.6 && <1.7)
+  , any.crypton-x509-system < 1.6.8 || >= 1.7
```
carbolymer, line 26: So this makes the lower bound here useless, no?

carbolymer (approved):

cabal.project
erikd: It excludes the version that causes a failure in `hakell.nix`.

The `crypton-x509-system < 1.6.8` is needed for `grapesy` . Version `1.6.8` causes a `hackage.nix` failure. Version `>= 1.7` is needed for when `grapesy` gets updated (which I am chasing).

PR #6668, opened 2026-08-28
gRPC: Enable HTTP endpoint with TLS support

carbolymer:
# Description

Adds configuration for the node's gRPC server to listen over HTTP/2 instead of only a unix socket, in plaintext (h2c) or with TLS.

- New CLI flags `--grpc-listen-address`/`--grpc-listen-port` (config keys `RpcListenAddress`/`RpcListenPort`) switch the server from the unix socket to HTTP/2.
  The listen address accepts an IPv4 or IPv6 address and defaults to `127.0.0.1`.
- New CLI flags `--grpc-tls-certificate`/`--grpc-tls-private-key`/`--grpc-tls-chain-certificate` (config keys `RpcTlsCertificateFile`/`RpcTlsPrivateKeyFile`/`RpcTlsChainCertificateFiles`) switch the HTTP/2 endpoint to TLS.
- Conflicting or incomplete combinations are rejected at configuration parse time.
  For example, a unix socket path together with a listen port, or a TLS private key without a certificate.
- A new `RPC.ServerListening` trace announces the endpoint the gRPC server is listening on at startup.

Also fixes `--port` and the tracer socket CLI options.
They previously parsed port numbers with a plain `Int`/`Word16` `read` and silently wrapped out-of-range values, for example `70000` became `4464`.
They now reject out-of-range and non-decimal input.

This PR also carries two changes that cannot ship separately from the endpoint work:

- The bump to cardano-api 11.7 and cardano-rpc 11.3 (both released to CHaP today), together with the CHaP index and flake pin update.
  cardano-rpc 11.3 replaced the socket path in `RpcConfigF` with the `RpcEndpoint` type, so the bump only compiles with the endpoint changes here, and the endpoint changes only compile against 11.3.
- The UTxO RPC test update for the latest v1beta spec, previously #6666.
  The spec made the `FetchBlock` request reference and response block fields repeated, so the old test does not compile against cardano-rpc 11.3 and the updated one does not compile against 11.2.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/cardano-node/src/Cardano/Node/Configuration/POM.hs
+++ b/cardano-node/src/Cardano/Node/Configuration/POM.hs
@@ -34,12 +34,13 @@ import           Cardano.Network.ConsensusMode (ConsensusMode (..), defaultConse
 import qualified Cardano.Network.Diffusion.Configuration as Cardano
 import           Cardano.Network.PeerSelection (NumberOfBigLedgerPeers (..))
 import           Cardano.Node.Configuration.LedgerDB
+import           Cardano.Node.Configuration.NodeAddress (NodeHostIPAddress (..))
 import           Cardano.Node.Configuration.Socket (SocketConfig (..))
 import           Cardano.Node.Handlers.Shutdown
 import           Cardano.Node.Protocol.Types (Protocol (..))
 import           Cardano.Node.Types
 import           Cardano.Rpc.Server.Config (PartialRpcConfig, RpcConfig, RpcConfigF (..),
-                   RpcEndpoint (..), defaultRpcListenAddress, makeRpcConfig)
+                   RpcEndpoint (..), RpcTlsFiles (..), defaultRpcListenAddress, makeRpcConfig)
 import           Ouroboros.Consensus.Ledger.SupportsMempool
 import           Ouroboros.Consensus.Mempool (MempoolCapacityBytesOverride (..))
 import           Ouroboros.Consensus.Node (NodeDatabasePaths (..))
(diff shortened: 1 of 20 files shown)


Jimbo4350 (approved):

cardano-node/src/Cardano/Node/Configuration/POM.hs
```diff
@@ -722,6 +720,46 @@ instance FromJSON PartialNodeConfiguration where
...
+          (Just socketPath, Nothing, Nothing, Nothing) -> pure . Just $ RpcEndpointUnixSocket socketPath
+          (Nothing, Just _, Nothing, _) -> fail "RpcListenAddress requires RpcListenPort to be set"
+          (Nothing, _, Nothing, Just _) -> fail "TLS configuration requires RpcListenPort to be set"
+          (Nothing, _, Just listenPort, Just tlsFiles) ->
+            pure . Just $
+              RpcEndpointHttps (fromMaybe defaultRpcListenAddress mListenAddress) (fromIntegral listenPort) tlsFiles
+          (Nothing, _, Just listenPort, Nothing) ->
+            pure . Just $
+              RpcEndpointHttp (fromMaybe defaultRpcListenAddress mListenAddress) (fromIntegral listenPort)
+          (Nothing, Nothing, Nothing, Nothing) -> pure Nothing
```
palas, line 742:
```suggestion
        rpcEndpoint <- case mSocketPath of
          Just socketPath -> do
            when (isJust mListenAddress) $ fail "RpcSocketPath and RpcListenAddress are mutually exclusive"
            when (isJust mListenPort) $ fail "RpcSocketPath and RpcListenPort are mutually exclusive"
            when (isJust mTlsFiles) $ fail "RpcSocketPath and TLS configuration are mutually exclusive"
            pure . Just $ RpcEndpointUnixSocket socketPath
          Nothing ->
            case mListenPort of
              Nothing -> do
                when (isJust mListenAddress) $ fail "RpcListenAddress requires RpcListenPort to be set"
                when (isJust mTlsFiles) $ fail "TLS configuration requires RpcListenPort to be set"
                pure Nothing 
              Just listenPort ->
                case mTlsFiles of
                  Just tlsFiles ->
                    pure . Just $
                      RpcEndpointHttps (fromMaybe defaultRpcListenAddress mListenAddress) (fromIntegral listenPort) tlsFiles
                  Nothing ->
                    pure . Just $
                      RpcEndpointHttp (fromMaybe defaultRpcListenAddress mListenAddress) (fromIntegral listenPort)
```
I think it would be better to split this case for clarity, even if it results on nested cases. Because otherwise it is too many combinations to reason about. The suggestion goes a bit further and uses `when` for the errors too, instead of underscores.

palas (commented):
I still haven't had time to look at it all, but here is a suggestion

cardano-node/src/Cardano/Node/Configuration/POM.hs
carbolymer: I'm not super happy about how the `case` looks like. I think your snippet makes it a bit harder to reason about the cases depending on the value of two or three fields. I'll stick to my solution. I'll think if there's a better way to simplify this.

carbolymer:
hydra stuck for days, force merging

PR #6669, opened 2026-08-29
Align BlockFetch.Decision and ClientMetrics namespaces with what is emitted

### LINKED ISSUE #6667: [BUG] - `BlockFetch.Decision` documents namespaces that can never fire, and `BlockFetch.Client.ClientMetrics` is unknown to the config check
Two tracers have namespaces that disagree between what is emitted, what is documented, and what a trace configuration is allowed to name. Both come from the same root cause: a tracer's type is written out by hand in `Documentation.hs` and `Consistency.hs`, and neither is type-checked against the tracer actually wired in `Tracers.hs`, so the three can drift apart silently.

Verified on `39d586694` against released `ouroboros-consensus-4.1.0.0`.

## 1. `BlockFetch.Decision` — three phantom namespaces, two missing ones

`Consensus.Tracers.blockFetchDecisionTracer` has carried `TraceDecisionEvent` since `ouroboros-consensus-4.1.0.0`:

```haskell
-- ouroboros-consensus-4.1.0.0, Ouroboros/Consensus/Node/Tracers.hs:73
, blockFetchDecisionTracer :: f (TraceDecisionEvent remotePeer (Header blk))
```

`Tracers.hs` wires that field, so what the node emits under `BlockFetch.Decision` is `TraceDecisionEvent`, whose namespaces are `PeersFetch` and `PeerStarvedUs` (`Tracers/Consensus.hs:671`).

But both declaration sites still name the pre-`TraceDecisionEvent` type:

```haskell
-- Tracing/Documentation.hs:322
blockFetchDecisionTrDoc <- documentTracer (blockFetchDecisionTr ::
   Logging.Trace IO [BlockFetch.TraceLabelPeer remotePeer
                                  (FetchDecision [Point (Header blk)])])

-- Tracing/Consistency.hs:174
blockFetchDecisionNS = map (nsGetTuple . nsReplacePrefix ["BlockFetch", "Decision"])
                (allNamespaces :: [Namespace [BlockFetch.TraceLabelPeer remotePeer
                                              (FetchDecision [Point (Header blk)])]])
```

whose namespaces are `EmptyPeersFetch`, `Accept` and `Decline`.

### Effect

- `bench/trace-schemas/newNamespaces.txt:10-12` lists `BlockFetch.Decision.Accept`, `.Decline` and `.EmptyPeersFetch`. None of them can ever be emitted.
- `BlockFetch.Decision.PeersFetch` and `.PeerStarvedUs` are emitted but appear in no documentation, and `checkTraceConfiguration` rejects them as unknown namespaces — so the two real messages cannot be re-levelled or silenced from a config file.

## 2. `BlockFetch.Client.ClientMetrics` — documented and emitted, but rejected by the config check

`ClientMetrics` is folded from blockfetch events and composed onto the blockfetch client tracer in `Tracers.hs`, and documented at the `["BlockFetch","Client"]` prefix:

```haskell
-- Tracing/Documentation.hs:341
blockFetchClientMetricsDoc <- documentTracer (blockFetchClientMetricsTr ::
    Logging.Trace IO ClientMetrics)
```

so `BlockFetch.Client.ClientMetrics` is in `newNamespaces.txt:3`. But `Consistency.hs` has **no** `ClientMetrics` entry — `getAllNamespaces` never mentions it.

### Effect

The namespace is emitted and documented, yet naming it in a trace configuration produces a `Config namespace error` from `checkTraceConfiguration`.

## Suggested fix

1. In `Documentation.hs` and `Consistency.hs`, annotate the blockfetch decision tracer as `TraceDecisionEvent remotePeer (Header blk)`, matching the record field.
2. Add a `ClientMetrics` entry to `getAllNamespaces` at prefix `["BlockFetch","Client"]`, alongside the existing `blockFetchClientNS`.
3. Regenerate `bench/trace-schemas/newNamespaces.txt` and the trace schemas.

Worth confirming whether the now-unused `MetaTrace`/`LogFormatting` instances for `FetchDecision [Point header]` and `[TraceLabelPeer peer (FetchDecision [Point header])]` still have a consumer; if not, they can go with (1).

## Root cause, and preventing recurrence

Each tracer's type is stated three times — once by the `Consensus.Tracers` field, once in `Documentation.hs`, once in `Consistency.hs` — and only the first is checked by the compiler. The other two build their own tracer values purely to enumerate namespaces, so a wrong annotation there compiles cleanly and surfaces only as a wrong name in the generated docs or a rejected config entry.

Anything that derives all three from one place would close this off. Failing that, a test asserting that the namespace set from `getAllNamespaces` matches the set `docTracersFirstPhase` produces would have caught both of these, since they disagree in opposite directions.

pucedoteth:
# Description

Fixes #6667, taking the localized approach @mgmeier suggested there — restoring consistency across usage, documentation and configuration inside the existing pattern, rather than deriving all three from one place.

**`BlockFetch.Decision`.** `Consensus.blockFetchDecisionTracer` has carried `TraceDecisionEvent` since `ouroboros-consensus-4.1.0.0`, and `Tracers.hs:385` wires that field, so the node emits `PeersFetch` and `PeerStarvedUs`. Both declaration sites still spelled the pre-`TraceDecisionEvent` type by hand:

```haskell
Logging.Trace IO [BlockFetch.TraceLabelPeer remotePeer (FetchDecision [Point (Header blk)])]
```

whose namespaces are `EmptyPeersFetch`, `Accept` and `Decline`. So the three documented namespaces can never fire, and the two that do fire are unknown to `checkTraceConfiguration` — they cannot be re-levelled or silenced from a config file.

**`BlockFetch.Client.ClientMetrics`** had the mirror-image problem: composed onto the blockfetch client tracer in `Tracers.hs:389` and documented at the `["BlockFetch","Client"]` prefix, but absent from `getAllNamespaces`, so naming it in a configuration produced a `Config namespace error`.

Both sites are now annotated with the record's own type, `ClientMetrics` is added at the `BlockFetch.Client` prefix, and `newNamespaces.txt` is updated accordingly (`Accept`/`Decline`/`EmptyPeersFetch` out, `PeerStarvedUs`/`PeersFetch` in; the file stays C-locale sorted).

The `Ouroboros.Network.BlockFetch.Decision` import is dropped from both modules, since `FetchDecision` was its only use in either and `cabal.project` sets `-Werror`.

# Notes for the reviewer

**I could not build this.** I have no GHC/cabal toolchain available, and a cardano-node build needs libsodium/secp256k1/blst besides. Rather than guess, here is what I did check statically — please let CI be the judge:

- `ClientMetrics` is exported from `Cardano.Node.Tracing.Tracers.Consensus` (export list line 21), so the new import in `Consistency.hs` resolves.
- No import cycle: `Tracers.Consensus` does not import `Consistency`, and it was already reachable via `Consistency -> Documentation -> Tracers.Consensus`.
- Dropping the `Decision` import is safe in both modules: `FetchDecision` was its only symbol in use, while `Point` (from `Ouroboros.Network.Block`) and the qualified `BlockFetch.` (from `BlockFetch.ClientState`) are still used 12x/7x and 26x respectively, so those imports stay live.
- `TraceDecisionEvent` is spelled exactly as `Tracers/Consensus.hs:79` imports it, and applied at the arity used by the `MetaTrace` instance at line 657 and the consensus record field.
- Import lines follow the surrounding stylish-haskell alignment.

I have not run `hlint` or `stylish-haskell` either, for the same reason.

# Checklist

- [ ] New tests are added if needed and existing tests are updated
  - The namespace consistency check is itself the test here: `checkTraceConfiguration` compares `getAllNamespaces` against the configuration, and `newNamespaces.txt` is the golden list this PR updates.
- [ ] The version bounds in `.cabal` files are updated — not applicable, no dependency change
- [ ] CI passes. See note above: I could not run the build or the linters locally.

### DIFF
--- a/bench/trace-schemas/newNamespaces.txt
+++ b/bench/trace-schemas/newNamespaces.txt
@@ -7,9 +7,8 @@ BlockFetch.Client.CompletedFetchBatch
 BlockFetch.Client.RejectedFetchBatch
 BlockFetch.Client.SendFetchRequest
 BlockFetch.Client.StartedFetchBatch
-BlockFetch.Decision.Accept
-BlockFetch.Decision.Decline
-BlockFetch.Decision.EmptyPeersFetch
+BlockFetch.Decision.PeerStarvedUs
+BlockFetch.Decision.PeersFetch
 BlockFetch.Remote.Receive.BatchDone
 BlockFetch.Remote.Receive.Block
 BlockFetch.Remote.Receive.ClientDone
--- a/cardano-node/ChangeLog.md
+++ b/cardano-node/ChangeLog.md
@@ -2,6 +2,15 @@
 
 ## Next version
 
+- **Bugfix:** `BlockFetch.Decision` now documents the namespaces it actually emits,
+  `PeersFetch` and `PeerStarvedUs`, instead of `Accept`, `Decline` and `EmptyPeersFetch`,
+  which no longer correspond to any message. The two real namespaces can now be named in a
+  trace configuration; previously `checkTraceConfiguration` rejected them as unknown.
+
+- **Bugfix:** `BlockFetch.Client.ClientMetrics` is recognised by `checkTraceConfiguration`.
+  It was emitted and documented, but naming it in a trace configuration produced a
+  `Config namespace error`.
+
 ## 11.1.0 -- August 2026
 
 - **Behaviour change:** snapshot options set directly under `LedgerDB` alongside a
--- a/cardano-node/src/Cardano/Node/Tracing/Consistency.hs
+++ b/cardano-node/src/Cardano/Node/Tracing/Consistency.hs
@@ -32,6 +32,7 @@ import           Cardano.Node.Tracing.Documentation (docTracersFirstPhase)
 import           Cardano.Node.Tracing.Formatting ()
 import qualified Cardano.Node.Tracing.StateRep as SR
 import           Cardano.Node.Tracing.Tracers.BlockReplayProgress
+import           Cardano.Node.Tracing.Tracers.Consensus (ClientMetrics)
 import           Cardano.Node.Tracing.Tracers.ConsensusStartupException
 import           Cardano.Node.Tracing.Tracers.KESInfo ()
 import           Cardano.Node.Tracing.Tracers.LedgerMetrics (LedgerMetrics)
@@ -68,7 +69,7 @@ import           Ouroboros.Consensus.Protocol.Praos.AgentClient (KESAgentClientT
 import qualified Ouroboros.Consensus.Storage.ChainDB as ChainDB
 import           Ouroboros.Network.Block (Point (..), SlotNo, Tip)
 import qualified Ouroboros.Network.BlockFetch.ClientState as BlockFetch
-import           Ouroboros.Network.BlockFetch.Decision
+import           Ouroboros.Network.BlockFetch.Decision.Trace (TraceDecisionEvent)
 import           Ouroboros.Network.ConnectionHandler (ConnectionHandlerTrace (..))
 import           Ouroboros.Network.ConnectionId (ConnectionId)
 import qualified Ouroboros.Network.ConnectionManager.Core as ConnectionManager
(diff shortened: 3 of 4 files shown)


pucedoteth:
Superseded by the fix for #6667 that is now on `master`, so I have left the conflict alone: resolving it would produce almost nothing.

Checked file by file against `master`:

- `bench/trace-schemas/newNamespaces.txt` — identical to this branch already.
- `cardano-node/src/Cardano/Node/Tracing/Documentation.hs` — identical already.
- `Consistency.hs` — `master` has the same `TraceDecisionEvent remotePeer (Header blk)` annotation, differing only in line wrapping, and goes further with `checkNodeTraceConfigurationWith` and a shared `DocTracer` pass.
- `ChangeLog.md` — `master`'s entry already describes this fix, including `BlockFetch.Client.ClientMetrics` now being accepted by the consistency check, plus a regression test comparing documented namespaces against the check's inventory, which this PR did not have.

Happy to close it, or feel free to close it directly.

pucedoteth:
Closing as superseded by the #6667 fix on `master`, which covers this and adds the regression test.

PR #6670, opened 2026-08-31
Update o-c and add tracer for LeiosTally

johnalotoski:
# Description

Node-side companion to the consensus change that adds `TraceLeiosTally`, making the
Leios certification margin observable.  https://github.com/IntersectMBO/ouroboros-consensus/pull/2256

* `cabal.project` bumps the `ouroboros-consensus` source-repository-package
* `Cardano/Node/Tracing/Tracers/Consensus.hs` registers the new constructor at the three
sites the tracing framework requires: a `forHuman` clause, a `namespaceFor` mapping to
`Namespace [] ["Tally"]`, and the corresponding entry in `allNamespaces`. `forMachine`
needs nothing, since it delegates to consensus `traceLeiosKernelToObject`.

Severity falls through to `Info`, matching `Voted` and `VoteAcquired`. That makes the
trace inert under stock configs, which filter at `Notice`.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -92,12 +92,12 @@ allow-older:
 -- Do NOT add more source-repository-package stanzas here unless they are strictly
 -- temporary! Please read the section in CONTRIBUTING about updating dependencies.
 
--- Points to ouroboros-consensus/leios-prototype
+-- Points to ouroboros-consensus/jl/leios-prototype-w35-patched
 source-repository-package
   type: git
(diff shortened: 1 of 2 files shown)


ch1bo:
Cherry picked, but then changed to re-use the `VoteAcquired` trace in https://github.com/IntersectMBO/ouroboros-consensus/pull/2271. Now released as `prototype-2026w35` on ouroboros-leios

PR #6675, opened 2026-09-02
Fixes #6667.

### LINKED ISSUE #6667: [BUG] - `BlockFetch.Decision` documents namespaces that can never fire, and `BlockFetch.Client.ClientMetrics` is unknown to the config check

jutaro:
The new tracing system states each tracer's message type in three places — the runtime wiring in `Tracers.hs`, the documentation generator in `Documentation.hs`, and the configuration consistency check in `Consistency.hs` — and only the first is compiler-checked. The other two
had drifted:

- `Documentation.hs` and `Consistency.hs` still annotated the blockfetch decision tracer with
  the pre-`TraceDecisionEvent` type `[TraceLabelPeer peer (FetchDecision [Point (Header blk)])]`.
  The generated docs therefore listed three namespaces that can never be emitted
  (`BlockFetch.Decision.{Accept,Decline,EmptyPeersFetch}`), while the two namespaces the node
  actually emits — `BlockFetch.Decision.PeersFetch` and `BlockFetch.Decision.PeerStarvedUs` —
  were undocumented and rejected by `checkTraceConfiguration`, so they could not be re-levelled
  or silenced from a config file.
- `BlockFetch.Client.ClientMetrics` was documented and emitted, but had no entry in
  `getAllNamespaces`, so the consistency check rejected it as unknown.

## Changes

- Retype the tracer annotations in `Documentation.hs` and `Consistency.hs` to
  `TraceDecisionEvent remotePeer (Header blk)`, matching the
  `Consensus.Tracers.blockFetchDecisionTracer` field; add a `ClientMetrics` entry to
  `getAllNamespaces` at prefix `BlockFetch.Client`.
- Remove the now-unconsumed `MetaTrace`/`LogFormatting` instances for the old list type (their
  `connectedPeers` metric documentation was never backed by an emitter;
  `LogFormatting (FetchDecision …)` stays — it is still used when formatting `PeersFetch`),
  and the schema generator's dead `EmptyPeersFetch` fallback.
- Regenerate `bench/trace-schemas/newNamespaces.txt` (only the `BlockFetch.Decision` hunk; the
  file has accumulated unrelated staleness on `master`, which is left for a follow-up so this
  diff stays reviewable).
- `goodConfig.yaml` now names `BlockFetch.Decision.PeerStarvedUs` and
  `BlockFetch.Client.ClientMetrics`, so the existing golden test proves both previously
  rejected namespaces are accepted.

## New test: `prop_namespaceInventory`

Guards the root cause: the tracer inventory is written out by hand twice, and nothing checked
the two copies against each other. The property runs one documentation pass
(`docTracersFirstPhase`, now shared with the golden tests via
`checkNodeTraceConfigurationWith` instead of three separate runs) and compares its namespace
set against `getAllNamespaces` using the new `namespaceInventoryDiff`. Each direction fails
with its own actionable message: *documented but unknown to the consistency check* (a config
naming it would be rejected) vs. *consistency-checked but never documented*. Eleven
pre-existing discrepancies (`Net.Handshake.*`, `Net.Mux.*`, `Net.DNSResolver`,
`Forge.ThreadStats`, `Reflection`, `NodeInfo`, `NodeStartupInfo`) are allowlisted by prefix
pending a follow-up issue; a stale-entry check fails once an allowlisted prefix no longer
matches any discrepancy, so the allowlist can only shrink. The test catches every future
divergence between the two files, though not a type going stale identically in both.

### DIFF
--- a/bench/trace-schemas/newNamespaces.txt
+++ b/bench/trace-schemas/newNamespaces.txt
@@ -7,9 +7,8 @@ BlockFetch.Client.CompletedFetchBatch
 BlockFetch.Client.RejectedFetchBatch
 BlockFetch.Client.SendFetchRequest
 BlockFetch.Client.StartedFetchBatch
-BlockFetch.Decision.Accept
-BlockFetch.Decision.Decline
-BlockFetch.Decision.EmptyPeersFetch
+BlockFetch.Decision.PeerStarvedUs
+BlockFetch.Decision.PeersFetch
 BlockFetch.Remote.Receive.BatchDone
 BlockFetch.Remote.Receive.Block
 BlockFetch.Remote.Receive.ClientDone
--- a/bench/trace-schemas/scripts/schema-gen/GhciSchemaGen.hs
+++ b/bench/trace-schemas/scripts/schema-gen/GhciSchemaGen.hs
@@ -1459,9 +1459,6 @@ updateSchemaForNamespaceWithWarning emitWarning config msgOutDir typeOutDir nsMa
 fallbackCtorForNamespace :: [String] -> Maybe ConstructorName
 fallbackCtorForNamespace parts
   | ["Net", "Handshake"] `isListPrefixOf` parts = Just diffusionHandshakeCtor
-  | ["BlockFetch", "Decision"] `isListPrefixOf` parts
-  , lastMay parts == Just "EmptyPeersFetch" =
-      Just listEmptyCtor
   | ["Net", "ConnectionManager"] `isListPrefixOf` parts
   , lastMay parts == Just "UnexpectedlyFalseAssertion" =
       Just connectionManagerUnexpectedlyFalseAssertionCtor
--- a/cardano-node/ChangeLog.md
+++ b/cardano-node/ChangeLog.md
@@ -2,6 +2,19 @@
 
 ## Next version
 
+- Fix `BlockFetch.Decision` trace namespace drift: documentation and the
+  configuration consistency check now use the runtime `TraceDecisionEvent`
+  type, so the documented message namespaces are
+  `BlockFetch.Decision.PeersFetch` and `BlockFetch.Decision.PeerStarvedUs`
+  instead of the stale `BlockFetch.Decision.{Accept,Decline,EmptyPeersFetch}`,
+  and both can now be configured. `BlockFetch.Client.ClientMetrics` is now
+  accepted by the consistency check. Removed the obsolete
+  `[TraceLabelPeer peer (FetchDecision [Point header])]` tracer instances,
+  including the never-emitted `connectedPeers` metric documentation. Added a
+  regression test comparing the documented namespaces against the consistency
+  check's namespace inventory.
+  ([#6667](https://github.com/IntersectMBO/cardano-node/issues/6667))
+
 ## 11.1.0 -- August 2026
 
 - **Behaviour change:** snapshot options set directly under `LedgerDB` alongside a
--- a/cardano-node/cardano-node.cabal
+++ b/cardano-node/cardano-node.cabal
@@ -240,6 +240,7 @@ test-suite cardano-node-test
                       , cardano-diffusion:{api, cardano-diffusion, orphan-instances}
                       , cardano-node
                       , cardano-slotting
+                      , containers
                       , contra-tracer
                       , directory
                       , filepath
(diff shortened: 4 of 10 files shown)


mgmeier (approved):
LGTM, thank you @jutaro

mgmeier (approved):

PR #6685, opened 2026-09-09
cardano-testnet: support gRPC over HTTP (h2c) via --enable-grpc-http

carbolymer:
## Description

`cardano-testnet cardano` can now serve the node gRPC endpoint over raw HTTP/2 without TLS (h2c), next to the existing unix socket toggle. When gRPC is enabled, every node's endpoint is printed in the startup log.

```
cardano-testnet cardano
│
├── (no gRPC flag)                       gRPC disabled (default)
│
├── --enable-grpc                        gRPC over a unix socket
│                                        (rpc.sock next to each node's N2C socket)
│
└── --enable-grpc-http                   gRPC over HTTP/2 without TLS (h2c)
    │                                    default: 127.0.0.1, random free port per node
    │
    ├── --grpc-listen-address IP         listen IP for all nodes, IPv4 or IPv6
    │                                    optional, default 127.0.0.1
    │
    └── --grpc-listen-port-base PORT     base port: node i listens on PORT+i-1
                                         e.g. 7301 gives 7301,7302,7303 for the
                                         default 3 nodes
                                         optional, random free ports when omitted
```

Parser rules:
- `--enable-grpc` and `--enable-grpc-http` are mutually exclusive (parse error when both are given).
- `--grpc-listen-address` and `--grpc-listen-port-base` require `--enable-grpc-http` (`Missing: --enable-grpc-http` otherwise).
- The base port must be a plain decimal in the range 1-65535; hex and out-of-range values are rejected at parse time.
- A base port whose derived range would exceed port 65535 for the node count aborts startup before any node spawns.

Runtime behaviour:
- Startup logs one line per node, e.g. `gRPC endpoint of node1: http://127.0.0.1:7301` (IPv6 hosts are bracketed), or `gRPC endpoint of node1: unix socket <path>` with `--enable-grpc`.
- Fixed ports are derived from the base, so duplicate ports are impossible by construction. Random ports are drawn up front on the configured listen address and deduplicated against each other and the nodes' P2P ports.
- cardano-node swallows a gRPC HTTP bind failure silently (its RPC server async is not linked), so cardano-testnet probes each HTTP endpoint after node start and fails loudly on timeout instead of handing tests a dead endpoint.
- IPv6 listen addresses are supported end to end: the parser accepts bare IPv6 literals, the node binds the right address family, and the client side brackets the host both in the logged URL and in the HTTP/2 `:authority` header.
- `TestnetNode` now carries the node's gRPC endpoint (`nodeRpcEndpoint`), and `nodeGrpcServer` (exported from `Cardano.Testnet`) maps it to a grapesy client `Server` for tests. The `RPC ReadGenesis` E2E now runs over h2c; the remaining RPC tests keep covering the unix socket path.

Testing:
- Golden help files regenerated; the full `cardano-testnet-golden` suite is green.
- `RPC ReadGenesis` (h2c, random-port path) and `RPC FollowTip` (unix socket) pass with retries disabled.
- The `--grpc-listen-port-base` path and IPv6 listen addresses are covered by parse-time and startup validation plus source-level verification; no automated test drives them end to end yet.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff

# Note on CI
for you.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cardano-testnet/src/Parsers/Cardano.hs
+++ b/cardano-testnet/src/Parsers/Cardano.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
 module Parsers.Cardano
@@ -6,20 +7,24 @@ module Parsers.Cardano
   , parseNodeSpecs
   ) where
 
-import           Cardano.Api (AnyShelleyBasedEra (..))
+import           Cardano.Api (AnyShelleyBasedEra (..), (?!))
 
 import           Cardano.CLI.EraBased.Common.Option hiding (pNetworkId)
 import           Cardano.Prelude (readMaybe)
+import           Cardano.Rpc.Server.Config (defaultRpcListenAddress)
 
 import           Prelude
 
 import           Control.Applicative (optional, (<|>))
-import           Control.Monad (unless)
+import           Control.Monad (unless, when)
+import           Data.Char (isDigit)
 import           Data.Default.Class (def)
+import           Data.IP (IP)
 import qualified Data.List as L
 import           Data.List.NonEmpty (NonEmpty ((:|)))
 import           Data.Maybe
 import           Data.Word (Word64)
+import           Network.Socket (PortNumber)
 import           Options.Applicative (CommandFields, Mod, Parser)
 import qualified Options.Applicative as OA
 import           Options.Applicative.Types (readerAsk)
@@ -99,11 +104,51 @@ pEnableNewEpochStateLogging = OA.switch
   )
 
 pEnableRpc :: Parser RpcSupport
-pEnableRpc = OA.flag RpcDisabled RpcEnabled
-  (   OA.long "enable-grpc"
-  <>  OA.help "[EXPERIMENTAL] Enable gRPC endpoint on all of testnet nodes. The listening socket file will be the same directory as node's N2C socket."
-  <>  OA.showDefault
-  )
+pEnableRpc =
+      OA.flag' RpcEnabledUnixSocket
+        (   OA.long "enable-grpc"
+        <>  OA.help "[EXPERIMENTAL] Enable gRPC endpoint on all of testnet nodes. The listening socket file will be the same directory as node's N2C socket. Listens on a unix socket."
+        )
+  <|> RpcEnabledHttp <$>
+        (   OA.flag' ()
+              (   OA.long "enable-grpc-http"
+              <>  OA.help "[EXPERIMENTAL] Enable gRPC endpoint over HTTP without TLS (h2c) on all of testnet nodes. Listens on 127.0.0.1 on a random port per node unless --grpc-listen-address/--grpc-listen-port-base are given."
+              )
+        *>  (   RpcHttpOptions
+            <$> OA.option ipReader
+                  (   OA.long "grpc-listen-address"
+                  <>  OA.metavar "IP"
+                  <>  OA.value defaultRpcListenAddress
+                  <>  OA.showDefault
+                  <>  OA.help "IP address the gRPC HTTP endpoints of all nodes listen on. Requires --enable-grpc-http."
+                  )
+            <*> OA.optional
+                  (   OA.option portReader
+                        (   OA.long "grpc-listen-port-base"
+                        <>  OA.metavar "PORT"
+                        <>  OA.help "Base port for gRPC HTTP listeners: testnet node i listens on PORT+i-1. Random free ports are used when omitted. Requires --enable-grpc-http."
+                        )
+                  )
+            )
+        )
+  <|> pure RpcDisabled
+
+-- | Parse an IP address, mirroring cardano-node's own @--grpc-listen-address@ parser
+ipReader :: OA.ReadM IP
+ipReader = OA.eitherReader $ \raw ->
+  maybe (Left $ "Failed to parse IP address: " <> raw) Right (readMaybe raw)
+
+-- | Parse a port number. Decimal digits only (rejecting the sign/hex/whitespace a plain
+-- 'Integer' read would otherwise accept) and in range before narrowing to 'PortNumber',
+-- since a derived 'Read' on 'PortNumber' silently wraps out-of-range values.
+portReader :: OA.ReadM PortNumber
+portReader = OA.eitherReader $ \token -> do
+  when (null token || not (all isDigit token)) $
+    Left $ "Not a valid port number: " <> token
+  port :: Integer <- readMaybe token ?! ("Not a valid port number: " <> token)
+  when (1 > port || port > 65_535) $
+    Left $ "Port number out of range (1 - 65535): " <> show port
+  pure $ fromIntegral port
 
 pKesSource :: Parser PraosCredentialsSource
 pKesSource = OA.flag UseKesKeyFile UseKesSocket
--- a/cardano-testnet/src/Testnet/Ping.hs
+++ b/cardano-testnet/src/Testnet/Ping.hs
@@ -6,6 +6,9 @@ module Testnet.Ping
   , checkSprocket
   , waitForSprocket
   , waitForPortClosed
+  , checkTcpPort
+  , waitForTcpPort
+  , randomFreePort
   , TestnetMagic
   , CNP.PingClientException
   ) where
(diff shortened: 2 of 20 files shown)


cardano-testnet/src/Parsers/Cardano.hs
```diff
@@ -99,11 +103,61 @@ pEnableNewEpochStateLogging = OA.switch
...
+                  <>  OA.showDefault
+                  <>  OA.help "IP address the gRPC HTTP endpoints of all nodes listen on. Requires --enable-grpc-http."
+                  )
+            <*> OA.optional
+                  (   OA.option portsReader
+                        (   OA.long "grpc-listen-ports"
+                        <>  OA.metavar "PORTS"
+                        <>  OA.help "Comma-separated list of gRPC HTTP listener ports, one per testnet node. Random free ports are used when omitted. Requires --enable-grpc-http."
+                        )
+                  )
```
palas, line 131: Hm, I don't like there is a list of nodes specific for `grpc-listen-ports`. The elements of a list of nodes should, in my opinion, contain **all** the configuration specific to a node, cause otherwise we end up we several different lists that all must have the same elements and in the same order, and then inconsistencies arise.

I would ask you to put it in `NodeWithOptions`, but that is for creation and not for runtime, so it doesn't really fit there either.

I think we will eventually have to have something like `NodeWithOptions` for runtime options. But since this is the first list for runtime as far as I know, I guess we can leave it as is for now? But break it later?

But you can already see that it would probably make sense to have some nodes with grpc enabled and others not, and that would be an extra list with this approach, and that is probably a reason why you didn't do it that way, because a list of `--enable-grpc yes,no,no,no...` together with `--grpc-listen-ports 100,101,102,103...` would have been weird.

Maybe we just need a TUI/GUI for `cardano-testnet`...

Jimbo4350 (approved):

cardano-testnet/src/Parsers/Cardano.hs
```diff
@@ -99,11 +103,61 @@ pEnableNewEpochStateLogging = OA.switch
...
+      | otherwise = case readMaybe token :: Maybe Integer of
+          Just port
+            | 1 <= port && port <= 65535 -> Right $ fromIntegral port
+            | otherwise -> Left $ "Port number out of range (1 - 65535): " <> show port
+          Nothing -> Left $ "Not a valid port number: " <> token
+
+    splitOnCommas :: String -> [String]
+    splitOnCommas s = case break (== ',') s of
+      (token, "") -> [token]
+      (token, _:rest) -> token : splitOnCommas rest
```
palas: We need to make sure the ports are different. Apparently if you write the same port more than once, one of the nodes dies, but the startup checks still pass, because they connect twice to the same node

cardano-testnet/src/Testnet/Ping.hs
```diff
@@ -81,13 +73,54 @@ checkSprocket sprocket = liftIOAnnotated $ do
...
       pure (Left e)
 
+-- | Repeat the check until it succeeds or the timeout expires; returns the last result.
+waitFor :: MonadIO m
+        => MT.DiffTime -- ^ timeout
+        -> MT.DiffTime -- ^ interval
+        -> IO (Either IOException ())
+        -> m (Either IOException ())
+waitFor timeout interval check = liftIOAnnotated $ do
+  lastResult <- newIORef (Right ())
```
palas: This returns `Right ()` by default on timeout, but it should probably return an error on timeout

cardano-testnet/src/Testnet/Start/Cardano.hs
```diff
@@ -375,8 +394,30 @@ cardanoTestnet
...
+        NodeRpcUnixSocket{} -> pure () -- readiness covered by the sprocket wait
+        NodeRpcHttp ip rpcHttpPort ->
+          Ping.waitForTcpPort 120 0.2 ip rpcHttpPort >>=
+            either
+              (const . throwString $ mconcat
+                [ "gRPC HTTP endpoint of ", nodeName
+                , " did not come up on ", show ip, ":", show rpcHttpPort
+                , " - port collision?"
+                ])
+              pure
```
palas, line 423: It would be nice to do a more sophisticated check on the node. Even better if we can actually identify the node and ensure it is the one it is supposed to be in that port. That would prevent the issue with the repeated ports too

cardano-testnet/src/Testnet/Start/Cardano.hs
```diff
@@ -366,6 +374,17 @@ cardanoTestnet
           keys@SpoNodeKeys{poolNodeKeysVrf} = mkTestnetNodeKeyPaths i
       pure (Just keys, kesSourceCliArg <> shelleyCliArgs <> byronCliArgs)
 
+    (mRpcHttpEndpoint, grpcArgs) <- case cardanoEnableRpc of
+      RpcDisabled -> pure (Nothing, [])
+      RpcEnabledUnixSocket -> pure (Nothing, ["--grpc-enable"])
+      RpcEnabledHttp RpcHttpOptions{rpcHttpListenAddress} -> do
+        rpcPort <- maybe (H.randomPort testnetDefaultIpv4Address) pure $ Map.lookup i rpcPortsMap
```
palas: I wonder whether this can also generate the same port more than once. Also, `randomPort` is using `testnetDefaultIpv4Address`,  but it should be using `rpcHttpListenAddress`.

palas (commented):
There are a couple of minor issues that could cause glitches. And then there is a design decision which I think won't work well in the long term, but my suggestion may be premature optimisation. Let me know what you think

cardano-testnet/src/Testnet/Start/Cardano.hs
carbolymer: The node staying up after a gRPC bind failure is deliberate: the RPC server async is intentionally not linked, so RPC is best-effort and never takes the node down. Identifying which node answers on a port is not really solvable from the outside either, since testnet nodes are clones of each other (same genesis, same chain), so there is nothing node-specific a probe could check without adding an identity endpoint to the node just for this.

What I can do is remove the ways a wrong listener can appear on the probed port: fixed ports are now derived from a single `--grpc-listen-port-base` (node i listens on base+i-1), so duplicates are impossible by construction, and randomly allocated ports are drawn up front and deduplicated against each other and the P2P ports (in the next push). After that, the only wrong-listener case left is an unrelated process squatting on the port. If we want to catch that too, the probe could do a real gRPC call (e.g. ReadTip) instead of a bare TCP connect; I would treat that as a follow-up.

cardano-testnet/src/Parsers/Cardano.hs
carbolymer: Agreed on the parallel-lists problem, so I'm dropping the list entirely: `--grpc-listen-ports` becomes `--grpc-listen-port-base`, and node i listens on base+i-1 (random free ports stay the default). That removes the ordering coupling and makes duplicate ports impossible by construction. When a second genuinely per-node runtime option shows up, I think the natural home already exists: the `--nodes` spec grammar (`spo:node-bin=...`) can grow attributes like `grpc=off|unix|http` and `grpc-port=N`, with runtime attributes routed internally. But I'd rather not open that up for a single derived value.

cardano-testnet/src/Parsers/Cardano.hs
carbolymer: Good catch, confirmed: the losing node's gRPC bind fails silently and the TCP probe passes against the winner's listener. Rather than validating the list, I'm replacing it with `--grpc-listen-port-base` (node i on base+i-1), so duplicates become impossible by construction. Coming in the next push.

cardano-testnet/src/Testnet/Ping.hs
carbolymer: Fixed in the next push: the result now starts as a failure, so a timeout that cancels the first in-flight check reports an error instead of success. A timeout after completed-but-failing checks already returned the last real error; the untouched success seed was the only hole.

cardano-testnet/src/Testnet/Start/Cardano.hs
carbolymer: You're right on both points. I measured it: the kernel guarantees distinct ports only for simultaneously bound sockets; once the probe socket closes, the port returns to a pool that bind(0) samples roughly uniformly (about 14k odd ports on Linux), so a default run drawing 3 P2P + 3 RPC ports collides roughly once in 900 runs. Fix in the next push: random RPC ports are drawn up front and deduplicated against each other and the P2P ports, and the port picking and readiness probe use the configured `--grpc-listen-address` instead of the hardcoded 127.0.0.1.

palas (approved):
Looks good 👍

Issue #6687, opened 2026-09-11
[FR] - cardano-testnet: machine-readable manifest and status command


palas:
**Internal/External**
*Internal* 

**Area**
*cardano-testnet*

### Problem

When `cardano-testnet cardano` starts a network it knows every socket path, port, and key location. But it only communicates them as human-oriented log text and an undocumented directory layout.

Programs that start `cardano-testnet` must therefore scrape the output directory and poll with their own retry loops. Consumers usually build this layer independently: `cardano-transaction-lib` parses the run directory ([their integration doc](https://github.com/Plutonomicon/cardano-transaction-lib/blob/develop/doc/cardano-testnet-testing.md)), the YACD Kubernetes operator wraps the binary, and Tweag's `cardano-peras` repo wraps it again.

The request is old and still open: [#5848](https://github.com/IntersectMBO/cardano-node/issues/5848) (2024) asks for "more structured output (e.g. JSON) so that we can parse the TMP dir to get the node socket path, etc.".

### Proposal

1. **Manifest.** On successful startup, write `<output-dir>/manifest.json`: network magic, era, and per node: name, role, socket path, port, config path; plus the funded key files. Document it as a stable interface, so the rest of the directory layout can keep changing freely.
2. **Status command.** `cardano-testnet status --output-dir DIR [--json]`: reports ready/synced/tip, exit code 0 only when the chain is producing blocks.

### References

[#5848](https://github.com/IntersectMBO/cardano-node/issues/5848) · [#6405](https://github.com/IntersectMBO/cardano-node/pull/6405) (keep-running behavior)

carbolymer:
Makes sense to me :thumbsup:
ad 1) One small caveat: if the user does not specify `--output-dir` , the output dir is non-deterministic and has to be obtained from the command output, which is not structured atm. Additionally, it would be nice to have schema defined somewhere (e.g jsonschema) for the manifest lying somewhere in the repo.
ad 2) We should spec out status command. What should be the behaviour when one node isn't healthy in multi-node network? 

I think this issue should be split into two, because manifest and status command are separate, although one is depending on the other I presume.

palas:
@carbolymer 

> ad 1) One small caveat: if the user does not specify --output-dir , the output dir is non-deterministic and has to be obtained from the command output, which is not structured atm. Additionally, it would be nice to have schema defined somewhere (e.g jsonschema) for the manifest lying somewhere in the repo.

[...]

PR #6693, opened 2026-09-14
cardano-testnet: write a machine-readable manifest.json on startup

### LINKED ISSUE #6690: [FR] cardano-testnet: write a machine-readable manifest.json on startup
Part of #6687 (subtask 1 of 2). Delivers the structured output asked for in #5848.

## Problem

When the network starts, cardano-testnet knows every socket, port and key path, but only
prints them as log text. Consumers re-discover them by scanning the directory layout:
cardano-transaction-lib parses `socket/node1/sock`, `node-data/nodeN/port` and
`utxo-keys/utxoN/utxo.skey`; our own bench/tx-generator does the same in
`TestnetDiscovery.hs`, which even carries a comment asking for a metadata file.

## Proposal

After the existing readiness checks pass (block no. >= 1 on every node, SPOs visible in the
ledger), and right before the "Testnet started" log line, write `<output-dir>/manifest.json`:

- `schemaVersion` (starts at 1), `createdAt`, `cardanoTestnetVersion`
- `network`: `magic`, `era`, `systemStart`
- `paths`: node config file, the five genesis files
- `nodes[]`: `name`, `role` ("spo"/"relay"), `host`, `port`, `socketPath`, `grpc`
  (null when disabled; otherwise transport + endpoint), `pid` (nullable), `pidFile`,
  `topologyFile`, `stdoutFile`, `stderrFile`
- `wallets[]`: the funded utxo keys — `name`, bech32 `address`, `signingKeyFile`,
  `verificationKeyFile`

Rules:

1. All paths are relative to the manifest's own directory. Exception: a value starting with
   `\\.\pipe\` is a Windows named pipe, used as-is. (Relative paths survive moved/mounted
   dirs and avoid the ~108-char unix socket path limit.)
2. Written atomically (temp file + rename). Readers never see a partial file.
3. Written once, never modified. The file appearing IS the ready signal; health questions
   are the status command's job (follow-up issue).
4. The run command deletes any pre-existing manifest.json before starting nodes, so a
   waiting script can never read a stale one.
5. Stability: the manifest is the supported interface; the rest of the directory layout
   stays free to change. Additive changes only within a schemaVersion; removals/renames
   bump it. A JSON Schema (draft 2020-12, like bench/trace-schemas/) lives in the repo.
   A golden test pins the writer's output, and the golden sample is validated against the
   schema with check-jsonschema (the validator bench/trace-schemas already uses), so the
   writer, the sample and the schema cannot drift apart.

palas:
# Description

Part of #6687 (first of its two subtasks). Addresses https://github.com/IntersectMBO/cardano-node/issues/6690.

When `cardano-testnet cardano` starts a network, it knows every socket path, port and key
location, but until now it only communicated them as log text and an undocumented directory
layout. Consumers rebuild discovery by hand: cardano-transaction-lib parses
`socket/node1/sock`, `node-data/nodeN/port` and `utxo-keys/`, and our own
`bench/tx-generator` does the same in `TestnetDiscovery.hs` — which carries a comment asking
for exactly this metadata file.

This PR writes `<output-dir>/manifest.json` once the existing readiness checks pass (every
node at block number >= 1, stake pools visible in the ledger), right before the
`Testnet started` log line. The file appearing doubles as the ready signal, so
`until [ -f testnet/manifest.json ]; do sleep 1; done` is a correct wait loop.

The file lists: schema version, creation time, tool version, network (magic, era, system
start), the node config and genesis file paths, per node (name, role spo/relay, host, port,
socket path, gRPC endpoint or null, pid, pid/topology/stdout/stderr files), and the three
funded wallets (bech32 address + key file paths).

Rules, documented in the README:
  
- All paths are relative to the manifest's own directory (Windows named pipes are the one
  exception, used as-is). Relative paths survive moved/mounted directories and avoid the
  ~108-char unix socket path limit.
- Written atomically (temp file + rename, default file permissions); never modified after.
- A pre-existing manifest is deleted before anything else on a new run in the same
  directory — even one that fails during environment creation — so a waiting script can
  never read stale data.
- The manifest is the supported interface; the rest of the directory layout stays free to
  change. A strict JSON Schema (draft 2020-12, `additionalProperties: false`, same style
  and validator as `bench/trace-schemas/`) lives at
  `cardano-testnet/schemas/manifest.schema.json`. `make testnet-manifest-validate`
  validates the golden sample (and any `MANIFEST=path`) with `check-jsonschema`.

Testing: a golden test pins the exact JSON shape (plus a FromJSON round-trip); an
integration test starts a default network and decodes the real file; live verification
additionally checked PIDs against the pid files, socket connectivity via the relative
paths, wallet address against `utxo.addr`, persistence after shutdown, stale-manifest
deletion on a failed rerun, and schema validation of a real run's manifest.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff


cardano-testnet/src/Testnet/Manifest.hs
```diff
@@ -0,0 +1,379 @@
...
+    }
+
+buildNode :: FilePath -> TestnetNode -> IO ManifestNode
+buildNode outputDir node = do
+  mPid <- Process.getPid (nodeProcessHandle node)
+  pure ManifestNode
+    { mnodeName         = nodeName node
+    , mnodeRole         = if isTestnetNodeSpo node then "spo" else "relay"
+    , mnodeHost         = showIpv4Address (nodeIpv4 node)
+    , mnodePort         = fromIntegral (nodePort node)
```
carbolymer: Why convert from better-typed types, to more generic strings and ints?

carbolymer (changes requested):
Nice work. At first glance looks ok. I need more time for a closer look. I think we should make sure that we do it right at the first attempt.

cardano-testnet/test/cardano-testnet-test/cardano-testnet-test.hs
```diff
@@ -146,6 +147,7 @@ tests = do
           , ignoreOnMacAndWindows "Supports dumping/loading config files" Cardano.Testnet.Test.DumpConfig.hprop_dump_config
           , ignoreOnMacAndWindows "Can have its start time modified" Cardano.Testnet.Test.UpdateTimeStamps.hprop_update_time_stamps
           , ignoreOnMacAndWindows "Can get on-chain parameters from blockfrost files" Cardano.Testnet.Test.MainnetParams.hprop_mainnet_params
+          , ignoreOnWindows "Writes a machine-readable manifest on startup" Cardano.Testnet.Test.Manifest.hprop_manifest
```
carbolymer, line 150: shouldn't we test this on windows as well?

Makefile
```diff
@@ -37,6 +37,11 @@ trace-schemas-overrides-coverage: ## Fail when generated schema files change wit
 trace-schemas-validate: ## Validate trace message schemas against meta.schema.json
 	nix run .#validate-trace-schemas
 
+testnet-manifest-validate: ## Validate manifest golden file (and optionally MANIFEST=path) against its schema
```
carbolymer, line 40: This is not executed by any github workflow afaik. We should definitely run this on merges.

cardano-testnet/src/Testnet/Start/Cardano.hs
```diff
@@ -486,6 +490,20 @@ cardanoTestnet
...
+  finalShelleyGenesis <- case eitherDecode systemStartBytes of
+    Right sg -> return sg
+    Left err -> throwString $ "Could not read shelley genesis for manifest: " <> err
+  let systemStartTime = sgSystemStart finalShelleyGenesis
+
+  -- Build and atomically write the manifest.  The file appearing is the
+  -- ready signal — it is written only after all readiness checks pass.
+  manifest <- liftIOAnnotated $
+    buildManifest tmpAbsPath runtime (eraToString Defaults.defaultEra) systemStartTime
+  liftIOAnnotated $ writeManifest tmpAbsPath manifest
```
carbolymer, line 515: The README says the file appearing is the ready signal, but this isn't the last thing that can fail: `startLedgerNewEpochStateLogging` below (on by default) runs after it and does IO that can throw. A script that starts using the cluster as soon as the file shows up can race a `cardanoTestnet` that's about to die. We should move this write after that block, so it's the last step before `pure runtime`.

cardano-testnet/src/Testnet/Start/Cardano.hs
```diff
@@ -486,6 +490,20 @@ cardanoTestnet
 
   assertExpectedSposInLedgerState stakePoolsFp nPools execConfig
 
+  -- Re-read the shelley genesis for the (potentially updated) system start time.
+  -- The in-memory 'shelleyGenesis' may be stale when timestamps were updated above.
+  systemStartBytes <- liftIOAnnotated (LBS.readFile shelleyGenesisFile)
+  finalShelleyGenesis <- case eitherDecode systemStartBytes of
```
carbolymer: Instead of reading and parsing the genesis file a second time, the `UpdateTimestamps` block above could return the updated genesis and we'd use that value here. That also fixes `chainForecastHorizon`, which still uses the stale in-memory genesis when timestamps were updated.

cardano-testnet/src/Parsers/Run.hs
```diff
@@ -104,6 +105,10 @@ runCardanoOptions = \case
     conf <- mkConfigAbs dirName
     runSimpleApp . runResourceT $ do
       logInfo $ "Creating environment: " <> display (tempAbsPath conf)
+      -- Fresh-run rule: remove any manifest left by a previous run in this
+      -- directory before anything else, so a script waiting for the file can
+      -- never pick up stale data — even if environment creation fails below.
+      liftIO $ removeStaleManifest (unTmpAbsPath (tempAbsPath conf))
```
carbolymer: This only protects the `cardano-testnet cardano` path. `create-testnet-env` and `createAndRunTestnet` don't remove the stale manifest before environment creation, so in a reused directory a manifest from an old run keeps signalling a ready cluster while the new run is still creating the env, or after it failed. The same call should go into those two entry points.

cardano-testnet/src/Testnet/Manifest.hs
```diff
@@ -0,0 +1,379 @@
...
+    , mnodeSocketPath   = relPath (sprocketSystemName (nodeSprocket node))
+    , mnodeGrpc         = buildGrpc (nodeRpcEndpoint node)
+    , mnodePid          = fmap fromIntegral mPid
+    , mnodePidFile      = "logs" </> nodeName node </> "node.pid"
```
carbolymer: These two path shapes are copies: `Testnet/Runtime.hs` builds the same pid-file path and `Start/Cardano.hs` builds the same topology path, and nothing ties them together at compile time. That's exactly what `Cardano.Node.Testnet.Paths` is for, and this module already uses it for the config and genesis files. Please add helpers there and use them from both sides.

cardano-testnet/src/Testnet/Manifest.hs
```diff
@@ -0,0 +1,379 @@
...
+    { manifestSchemaVersion        = 1
+    , manifestCreatedAt            = now
+    , manifestCardanoTestnetVersion = cardanoTestnetVersionString
+    , manifestNetwork              = ManifestNetwork
+        { mnMagic       = testnetMagic
+        , mnEra         = era
+        , mnSystemStart = systemStart
+        }
+    , manifestPaths                = ManifestPaths
+        { mpNodeConfigFile = defaultConfigFile
```
carbolymer: These two are recomputed from constants, but `TestnetRuntime` already carries the real paths in `configurationFile` and `shelleyGenesisFile`, set at startup to whatever the node was actually given. Relativise those instead (`makeManifestRelPath outputDir (unFile configurationFile)`), so the manifest cannot drift from the running configuration. The other four genesis files have no runtime field, so the constants are fine for them.

cardano-testnet/src/Testnet/Manifest.hs
```diff
@@ -0,0 +1,379 @@
...
+  , buildManifest
+  ) where
+
+import           Cardano.Api (CardanoEra (..), File (..))
+
+import           Cardano.Node.Testnet.Paths (defaultConfigFile, defaultGenesisFilepath)
+
+import           Prelude
+
+import           Control.Exception (onException, try)
```
carbolymer: We should use `safe-exceptions` instead

cardano-testnet/src/Testnet/Manifest.hs
```diff
@@ -0,0 +1,379 @@
...
+        , mwalletSigningKeyFile      = relPath (signingKeyFp paymentKeyInfoPair)
+        , mwalletVerificationKeyFile = relPath (verificationKeyFp paymentKeyInfoPair)
+        }
+    relPath = makeManifestRelPath outputDir
+
+-- | Make a path relative to the output directory, normalised (no leading @./@).
+-- Windows named pipes (starting with @\\.\pipe\@) are kept as-is.
+makeManifestRelPath :: FilePath -> FilePath -> FilePath
+makeManifestRelPath outputDir path
+  | "\\\\.\\pipe\\" `isPrefixOf` path = path
```
carbolymer: Related to my comment about Windows testing: this function is pure, so both branches (pipe prefix and normal relativisation) can be unit-tested on every platform without starting a testnet. Right now the pipe branch has no coverage anywhere, since `hprop_manifest` is `ignoreOnWindows`.

cardano-testnet/src/Testnet/Manifest.hs
```diff
@@ -0,0 +1,379 @@
...
+-- ---------------------------------------------------------------------------
+
+-- | The manifest file name, always written into the output directory root.
+manifestFileName :: FilePath
+manifestFileName = "manifest.json"
+
+-- | The cardano-testnet package version as a string (e.g. @"11.1.1"@),
+-- matching the output of @cardano-testnet version@.
+cardanoTestnetVersionString :: String
+cardanoTestnetVersionString = showVersion version
```
carbolymer: `Parsers/Version.hs` inlines the same `showVersion version` expression. Nothing enforces they stay the same if one of them changes. Version.hs could just use this constant.

cardano-testnet/schemas/manifest.schema.json
```diff
@@ -0,0 +1,187 @@
...
+    "required": [
+        "schemaVersion",
+        "createdAt",
+        "cardanoTestnetVersion",
+        "network",
+        "paths",
+        "nodes",
+        "wallets"
+    ],
+    "additionalProperties": false,
```
carbolymer, line 16: This contradicts the policy on line 21: with `additionalProperties: false` on every object, adding a field without bumping `schemaVersion` makes every consumer that pinned this schema reject valid manifests. The closed `role` enum and `transport` consts have the same problem with new values. Either drop `additionalProperties: false` from objects that are allowed to grow, or keep this strict schema for our own golden check and publish an open one as the contract. Needs a decision either way, right now the promise and the schema can't both be true. The README's "Stability promise" paragraph states both halves back to back ("unknown fields are an error" vs "fields may be added within a schema version"), so whichever way the decision goes, that paragraph needs the same fix.

cardano-testnet/schemas/manifest.schema.json
```diff
@@ -0,0 +1,187 @@
...
+            "type": "string",
+            "description": "The cardano-testnet package version string."
+        },
+        "network": {
+            "type": "object",
+            "required": ["magic", "era", "systemStart"],
+            "additionalProperties": false,
+            "properties": {
+                "magic": {
+                    "type": "integer",
```
carbolymer, line 38: magic is a Word32, so this should get `"minimum": 0, "maximum": 4294967295`.

cardano-testnet/schemas/manifest.schema.json
```diff
@@ -0,0 +1,187 @@
...
+                                "additionalProperties": false,
+                                "properties": {
+                                    "transport":  { "type": "string", "const": "unix-socket" },
+                                    "socketPath": { "type": "string" }
+                                }
+                            }
+                        ]
+                    },
+                    "pid": {
+                        "description": "OS process ID at start time, or null if the process handle had no PID.",
```
carbolymer: The pid is captured when the manifest gets built, after the readiness checks pass, not at node start time. The values agree today only because nothing restarts a node in between. This should say "at the time the manifest was written"; the start-time pid is what's in the pid file.

Makefile
```diff
@@ -37,6 +37,11 @@ trace-schemas-overrides-coverage: ## Fail when generated schema files change wit
...
+	nix run nixpkgs#check-jsonschema -- --schemafile cardano-testnet/schemas/manifest.schema.json \
```
carbolymer: The targets above use the repo flake (`nix run .#validate-trace-schemas`), pinned by flake.lock. `nixpkgs#check-jsonschema` resolves through the local flake registry, so everyone potentially validates with a different version. We should expose check-jsonschema from our flake and use `.#` here too.

cardano-testnet/test/cardano-testnet-golden/Cardano/Testnet/Test/Golden/Manifest.hs
```diff
@@ -0,0 +1,100 @@
...
+import qualified Hedgehog.Extras.Test.Process as H
+
+-- | A fixed sample manifest for golden testing.
+-- All values are deterministic (fixed ports, PIDs, times) so the output
+-- is stable across runs.
+sampleManifest :: Manifest
+sampleManifest = Manifest
+  { manifestSchemaVersion        = 1
+  , manifestCreatedAt            = UTCTime (fromGregorian 2026 1 1) 0
+  , manifestCardanoTestnetVersion = cardanoTestnetVersionString
```
carbolymer: This couples the golden file to the live package version, so the test breaks on every cardano-testnet version bump, starting with the one this PR's own changelog fragment triggers at release. Same trap as golden_VersionCmd. A fixed literal here would make the golden change only when the manifest shape changes.

cardano-testnet/test/cardano-testnet-golden/Cardano/Testnet/Test/Golden/Manifest.hs
```diff
@@ -0,0 +1,100 @@
...
+          , mgfAlonzo   = "alonzo-genesis.json"
+          , mgfConway   = "conway-genesis.json"
+          , mgfDijkstra = "dijkstra-genesis.json"
+          }
+      }
+  , manifestNodes =
+      [ ManifestNode
+          { mnodeName = "node1", mnodeRole = "spo", mnodeHost = "127.0.0.1"
+          , mnodePort = 30001, mnodeSocketPath = "socket/node1/sock"
+          , mnodeGrpc = Nothing, mnodePid = Just 12345
```
carbolymer: All three nodes use `mnodeGrpc = Nothing` and a `Just` pid, so the JSON field names of both `ManifestGrpc` constructors and the `pid: null` encoding are pinned by nothing, a typo in those instances would slip through. Varying the fixture would close it: one node with `ManifestGrpcHttp`, one with `ManifestGrpcUnixSocket`, one with `mnodePid = Nothing`.

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Manifest.hs
```diff
@@ -0,0 +1,47 @@
...
+import qualified Hedgehog.Extras as H
+
+-- | Integration test: start a default testnet, then verify that
+-- @manifest.json@ exists and can be decoded into the 'Manifest' type,
+-- and that basic counts match expectations (3 nodes, 3 wallets, magic 42).
+--
+-- Execute with:
+-- @DISABLE_RETRIES=1 cabal test cardano-testnet-test --test-options '-p "/Manifest/"'@
+hprop_manifest :: Property
+hprop_manifest = integrationRetryWorkspace 0 "manifest" $ \tmpDir -> H.runWithDefaultWatchdog_ $ do
```
carbolymer: why 0? every other full-spinup test in this suite uses `integrationRetryWorkspace 2`, and cluster startup is the flakiest thing we do. This would be the only such test with no retries.

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Manifest.hs
```diff
@@ -0,0 +1,47 @@
...
+  conf <- mkConf tmpDir
+  _runtime <- createAndRunTestnet def def conf
+
+  let manifestPath = tmpDir </> manifestFileName
+  H.assertFileExists manifestPath
+
+  bs <- liftIO $ LBS.readFile manifestPath
+  manifest :: Manifest <- evalEither $ A.eitherDecode' bs
```
carbolymer: We use `H.leftFail` for this in this suite (frozen call stack, failure points at the right line). Same for the `evalEither` in the golden test.

cardano-testnet/src/Testnet/Manifest.hs
carbolymer: To expand on this: the conversion happens at construction instead of serialisation, so `Manifest` ends up as a second data model kept in sync by hand rather than domain types plus a JSON mapping. `fromIntegral (nodePort ...)` and `show ip` here, plus `buildGrpc` absorbing the `Maybe` instead of being total on `NodeRpcEndpoint` with `fmap` at the call site. Keeping `PortNumber`/`IP` in the record and converting inside `ToJSON` removes the drift risk.

cardano-testnet/src/Testnet/Manifest.hs
carbolymer: This guard is dead code: `\\.\` parses as a drive (via `splitDrive`'s share fallback), so `makeRelative` already returns pipe paths unchanged, and `normalise` keeps the prefix intact. Checked on filepath 1.4.200.1 (ghc 9.6 boot lib) and 1.5.5.0 (ghc 9.12), both the Windows and Posix flavours. The whole function can be `normalise . makeRelative outputDir`, with the pipe behaviour kept as a haddock note.

carbolymer (changes requested):

cardano-testnet/test/cardano-testnet-test/cardano-testnet-test.hs
palas: We could do a separate test, but the thing is pipes in Windows work differently than file sockets (different type of path), so the golden file won't be the same

cardano-testnet/src/Testnet/Manifest.hs
palas: It was a bit like a serialization artefact. But you are right, why not keep the types till the last moment.
Addressed in: https://github.com/IntersectMBO/cardano-node/commit/e7d550de75e258f07dc2d5c04519edbc863cea9c

Makefile
palas: Added it to `check-mainnet-config` here: https://github.com/IntersectMBO/cardano-node/pull/6693/changes/cf8c1d74ca33acf5cf0f0f6899f49331637ab2b7

cardano-testnet/src/Testnet/Start/Cardano.hs
palas: Moved it to later here: https://github.com/IntersectMBO/cardano-node/commit/5b6c2582ad85979971840dec229acbd062b9cc88

cardano-testnet/src/Testnet/Start/Cardano.hs
palas: Good idea, addressed here: https://github.com/IntersectMBO/cardano-node/commit/5b6c2582ad85979971840dec229acbd062b9cc88

cardano-testnet/src/Parsers/Run.hs
palas: You are right. Now we delete both at the start of creating an env and at the start of using it, and create it at the very end. Addresed here: https://github.com/IntersectMBO/cardano-node/commit/5b6c2582ad85979971840dec229acbd062b9cc88

cardano-testnet/src/Testnet/Manifest.hs
palas: 100%. Addressed here: https://github.com/IntersectMBO/cardano-node/commit/e7d550de75e258f07dc2d5c04519edbc863cea9

cardano-testnet/src/Testnet/Manifest.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-node/commit/e7d550de75e258f07dc2d5c04519edbc863cea9c

cardano-testnet/src/Testnet/Manifest.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-node/commit/e7d550de75e258f07dc2d5c04519edbc863cea9c

cardano-testnet/schemas/manifest.schema.json
palas: You are right. I think the README is wrong. We should keep the schema strict so that it breaks when something changes, and then we are forced to keep it up to date. We could have a version for users, like you say, but then we would need to keep them both synchronised, and we'll be bumping the version number anyway. Addressed in: https://github.com/IntersectMBO/cardano-node/commit/1f16031e630272f2e8d3a3ab0d7e1aaa2424dca2

cardano-testnet/schemas/manifest.schema.json
palas: Addressed here: https://github.com/IntersectMBO/cardano-node/commit/1f16031e630272f2e8d3a3ab0d7e1aaa2424dca2

cardano-testnet/schemas/manifest.schema.json
palas: Addressed here: https://github.com/IntersectMBO/cardano-node/commit/1f16031e630272f2e8d3a3ab0d7e1aaa2424dca2

Makefile
palas: Addressed here: https://github.com/IntersectMBO/cardano-node/commit/21f7a4e6ef60ec144f81286242b95fbbca62042c

cardano-testnet/test/cardano-testnet-golden/Cardano/Testnet/Test/Golden/Manifest.hs
palas: Good catch, now it uses the literal "0.0.0.0", so the golden file only changes if the shape changes. Addressed here: https://github.com/IntersectMBO/cardano-node/commit/565707cdb782c09cd03b817b9e94040910a4a28d

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Manifest.hs
palas: Right, I didn't see that. Addressed here: https://github.com/IntersectMBO/cardano-node/commit/565707cdb782c09cd03b817b9e94040910a4a28d

cardano-testnet/test/cardano-testnet-golden/Cardano/Testnet/Test/Golden/Manifest.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/565707cdb782c09cd03b817b9e94040910a4a28d
This really should be done with property-based testing, pity there is no JSON schema validation in Haskell

cardano-testnet/src/Testnet/Manifest.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/e7d550de75e258f07dc2d5c04519edbc863cea9c

cardano-testnet/src/Testnet/Manifest.hs
```diff
@@ -0,0 +1,439 @@
...
+
+buildWallets :: FilePath -> [PaymentKeyInfo] -> IO [ManifestWallet]
+buildWallets outputDir = zipWithM build [(1::Int)..]
+  where
+    build i PaymentKeyInfo{paymentKeyInfoPair, paymentKeyInfoAddr} = do
+      -- The runtime carries the address as the raw file contents; parsing it
+      -- here means the manifest can only ever hold a valid address.
+      addr <- case deserialiseAddress AsAddressAny paymentKeyInfoAddr of
+        Just a  -> pure a
+        Nothing -> fail $ "buildManifest: invalid wallet address: " <> Text.unpack paymentKeyInfoAddr
```
carbolymer, line 423:
```suggestion
        Nothing -> throwString $ "buildManifest: invalid wallet address: " <> Text.unpack paymentKeyInfoAddr
```
We're dropping callstack this way

cardano-testnet/src/Testnet/Manifest.hs
```diff
@@ -0,0 +1,439 @@
...
+writeManifest :: FilePath -> Manifest -> IO ()
+writeManifest outputDir manifest = do
+  let manifestPath = outputDir </> defaultManifestFile
+  -- Default permissions (not openTempFile's owner-only 0600): the manifest
+  -- must be as readable as the rest of the output directory.
+  (tmpFile, tmpHandle) <- IO.openTempFileWithDefaultPermissions outputDir "manifest.json.tmp"
+  let cleanup = do
+        _ <- try (hClose tmpHandle) :: IO (Either IOError ())
+        _ <- try (removeFile tmpFile) :: IO (Either IOError ())
+        pure ()
```
carbolymer, line 328:
```suggestion
        void . try @IOError $ hClose tmpHandle
        void . try @IOError $ removeFile tmpFile
```
more readable I think

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Manifest.hs
```diff
@@ -0,0 +1,46 @@
...
+hprop_manifest :: Property
+hprop_manifest = integrationRetryWorkspace 2 "manifest" $ \tmpDir -> H.runWithDefaultWatchdog_ $ do
+  conf <- mkConf tmpDir
+  _runtime <- createAndRunTestnet def def conf
+
+  let manifestPath = tmpDir </> defaultManifestFile
+  H.assertFileExists manifestPath
+
+  bs <- liftIO $ LBS.readFile manifestPath
+  manifest :: Manifest <- evalEither $ A.eitherDecode' bs
```
carbolymer, line 39:
```suggestion
  manifest <- H.readJsonFileOk @Manifest $ tmpDir </> defaultManifestFile
```
this amounts to the same with better error reporting

cardano-testnet/src/Testnet/Manifest.hs
```diff
@@ -0,0 +1,439 @@
...
+    ]
+
+
+-- ---------------------------------------------------------------------------
+-- FromJSON instances
+-- ---------------------------------------------------------------------------
+
+instance FromJSON Manifest where
+  parseJSON = withObject "Manifest" $ \o -> Manifest
+    <$> o .: "schemaVersion"
```
carbolymer, line 214: `schemaVersion` decodes with a plain `.:`, so any integer is accepted, while the schema pins it to `const: 1`. Every other loose field got tightened in the last two commits, this one stayed behind.

It matters more than the others because it is the field whose job is to reject a manifest we cannot read. A future cardano-testnet writing `"schemaVersion": 2` with renamed fields would either fail somewhere confusing, or decode fine on the fields that happen to be unchanged and hand the caller a half-understood manifest. A small parser in the style of `parsePort`, failing with something like "unsupported schema version 2", turns that into a clear error.

Makefile
```diff
@@ -37,6 +37,11 @@ trace-schemas-overrides-coverage: ## Fail when generated schema files change wit
 trace-schemas-validate: ## Validate trace message schemas against meta.schema.json
 	nix run .#validate-trace-schemas
 
+testnet-manifest-validate: ## Validate manifest golden file (and optionally MANIFEST=path) against its schema
+	nix run .#check-jsonschema -- --schemafile cardano-testnet/schemas/manifest.schema.json \
+	  cardano-testnet/test/cardano-testnet-golden/files/golden/manifest.json
+	$(if $(MANIFEST),nix run .#check-jsonschema -- --schemafile cardano-testnet/schemas/manifest.schema.json $(MANIFEST),)
```
carbolymer, line 43: Nothing ever passes `MANIFEST=`, so the only file we validate is the golden sample, which we wrote by hand. The manifest the writer actually produces is never checked against the schema.

`hprop_manifest` only decodes it with aeson, and aeson cannot see what the schema is there to catch: an extra field under `additionalProperties: false`, or a field missing from a branch the golden fixture does not cover.

Validating from the test itself is awkward because `check-jsonschema` is not on PATH during `cabal test`. A CI step that starts a testnet with a fixed `--output-dir` and then runs `make testnet-manifest-validate MANIFEST=<dir>/manifest.json` would close it, and the override is already here for exactly that. See also my comment on the workflow step, the two fit together.

.github/workflows/check-mainnet-config.yml
```diff
@@ -57,3 +60,6 @@ jobs:
             exit 1
           fi
         done
+
+    - name: Validate cardano-testnet manifest golden file against its schema
+      run: make testnet-manifest-validate
```
carbolymer, line 65: This workflow only triggers on `push` and `merge_group`, so on a PR from a fork the manifest check first runs in the merge queue rather than during review. The other `check-*.yml` workflows all set `pull_request` for that reason.

A dedicated `check-testnet-manifest.yml` modelled on `check-cabal-files.yml` would fix the trigger and give a status name that says what broke, rather than a red "Check mainnet configuration" for a manifest schema error. The `permissions: contents: read` block you added here works the same in a standalone workflow, so the CodeQL complaint is not a reason to keep it folded in. The cost is losing the shared nix install step, roughly half a minute per run.

Also worth knowing either way: `check-mainnet-config` is not in master's required checks, so this does not gate merges today. If you do split it, that is the natural place to also validate a real manifest, see my comment on the Makefile.

cardano-testnet/src/Testnet/Manifest.hs
```diff
@@ -0,0 +1,439 @@
...
+    relPath = makeManifestRelPath outputDir
+
+-- | Make a path relative to the output directory, normalised (no leading @./@).
+--
+-- Windows named pipes (@\\.\pipe\...@) come through unchanged without a
+-- special case: 'System.FilePath.splitDrive' treats @\\.\@ as a drive, so
+-- 'makeRelative' returns pipe paths as-is and 'normalise' keeps the prefix
+-- (checked on filepath 1.4.301.0, both the Windows and Posix flavours).
+makeManifestRelPath :: FilePath -> FilePath -> FilePath
+makeManifestRelPath outputDir = normalise . makeRelative outputDir
```
carbolymer, line 439: Two small pure tests would be worth having here, and they cover the things no end-to-end test can reach.

First, this function through `System.FilePath.Windows`. The whole integration test is `ignoreOnWindows`, so on Windows there is no manifest coverage at all, and the named pipe path is the one thing that only exists on Windows. `System.FilePath.Windows` is importable on Linux, so a `propertyOnce` asserting a `\\.\pipe\...` path comes back unchanged pins the behaviour this function now relies on, on ordinary CI, with no Windows runner and no second golden file.

Second, `removeStaleManifest`'s delete branch. `integrationRetryWorkspace` mints a fresh directory per run and per retry, so the branch that deletes an existing manifest never executes in the integration test, however much we add to it. It is the fresh-run promise in the README, currently with no coverage, and it needs two runs in one directory to exercise.

Both run in milliseconds, so they are cheap next to the value.

cardano-testnet/schemas/manifest.schema.json
```diff
@@ -0,0 +1,201 @@
...
+            "description": "Schema version. Any change to the manifest shape, including new fields or values, bumps it."
+        },
+        "createdAt": {
+            "type": "string",
+            "format": "date-time",
+            "description": "UTC time the manifest was written (ISO-8601)."
+        },
+        "cardanoTestnetVersion": {
+            "type": "string",
+            "description": "The cardano-testnet package version string."
```
carbolymer, line 30: The decoder is stricter than this says: it requires `Data.Version`'s grammar, so `11.1.1-dev` would pass the schema and then fail to decode. Worth saying so, since the schema is what a non-Haskell consumer reads.

```suggestion
            "description": "The cardano-testnet package version: digits separated by dots, for example \"11.1.1\"."
```

cardano-testnet/src/Testnet/Manifest.hs
```diff
@@ -0,0 +1,439 @@
...
+  } deriving (Eq, Show)
+
+data ManifestNetwork = ManifestNetwork
+  { mnMagic       :: !Word32
+  , mnEra         :: !AnyCardanoEra
+  , mnSystemStart :: !UTCTime
+  } deriving (Eq, Show)
+
+data ManifestPaths = ManifestPaths
+  { mpNodeConfigFile :: !FilePath
```
carbolymer, line 83: Optional, and I would not hold the PR for it: these path fields could use cardano-api's `File content dir` tags. Its derived JSON is byte-identical to a plain string, so there would be no wire change and no golden change, only compile-time tagging.

Two of them already have a tagged source upstream: `configurationFile` is a `NodeConfigFile In` on `TestnetRuntime`, and `Testnet.Types.nodeSocketPath` returns a `SocketPath` that `buildNode` currently bypasses by calling `sprocketSystemName` directly. Whether tags earn their keep in a record that exists to be serialised is a fair judgement call, so treat this as a question rather than a request.

cardano-testnet/test/cardano-testnet-test/cardano-testnet-test.hs
carbolymer: but this isn't a golden test, is it?

carbolymer (changes requested):
I think we should validate produced manifests against json schema - in addition to the golden test

cardano-testnet/src/Testnet/Manifest.hs
palas: Good idea. That way we protect the future status command from being used against the wrong version

Makefile
palas: No, that line is useless, but line 41 checks `cardano-testnet/test/cardano-testnet-golden/files/golden/manifest.json`. So if I change the encoding, then either me or the CI is going to run:
```
cabal test cardano-testnet
```

And that will fail with:
```
  golden_Manifest:      FAIL
      ✗ golden_Manifest failed at test/cardano-testnet-golden/Cardano/Testnet/Test/Golden/Manifest.hs:120:3
        after 1 test.
        shrink path: 1:
      
            ┏━━ test/cardano-testnet-golden/Cardano/Testnet/Test/Golden/Manifest.hs ━━━
        115 ┃ golden_Manifest :: Property
        116 ┃ golden_Manifest = propertyOnce $ do
        117 ┃   base <- H.getProjectBase
        118 ┃   let encoded = encodePretty sampleManifest
        119 ┃       actualJson = Text.unpack $ Text.decodeUtf8 $ LBS.toStrict encoded
        120 ┃   diffVsGoldenFile actualJson $
            ┃   │ Reading file: ./../cardano-testnet/test/cardano-testnet-golden/files/golden/manifest.json
            ┃   │ Golden test failed against the golden file.
            ┃   │ To recreate golden file, run with RECREATE_GOLDEN_FILES=1.
```

Then I'll be forced to re-run it with `RECREATE_GOLDEN_FILES=1`. And that will update `cardano-testnet/test/cardano-testnet-golden/files/golden/manifest.json`.

Then either me or the CI will run: `make testnet-manifest-validate`. And that will fail, and that will force me to update `cardano-testnet/schemas/manifest.schema.json`.

And that is how the `Aeson` and the schema are kept in sync (at least to the extent the golden file exercises it).

Granted, it would be better to do it with randomly generated manifests, but that would require the Haskell to depend on the schema validator, and I am not sure it is worth it

.github/workflows/check-mainnet-config.yml
palas: Oh. I was trying to save actions, but yes, let's just make a separate action for this then

cardano-testnet/test/cardano-testnet-test/cardano-testnet-test.hs
palas: Oh, you are right, I confused it with the other test

Makefile
palas: But you are right, why not validate what actually gets produced too? I'll add a test for that

cardano-testnet/src/Testnet/Manifest.hs
palas: Yes, for the second we can create a dummy `manifest.json` and just run `create-env`, by the end there should be no `manifest.json`

cardano-testnet/src/Testnet/Manifest.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/bd2862fc579c70f8bfdad8d5999347e4d170b20b

cardano-testnet/src/Testnet/Manifest.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/97d87cda0d428b565a462835bbe756b9f8f01626

cardano-testnet/src/Testnet/Manifest.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/aaa72e7380621be26a08f08540e2061c107480e8

cardano-testnet/schemas/manifest.schema.json
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/f6977472d9e9217a5636a6eecbe38f15a1c8f566

.github/workflows/check-mainnet-config.yml
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/0c7e140c13cd3efe9c37003c194ff2404e4d46cf

Makefile
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/a0ed79f95bad357a57f9cc503cee27b73614a2c4

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Manifest.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/f8540d550c6cd236c68069f202e6754e49ff38b9

cardano-testnet/test/cardano-testnet-test/cardano-testnet-test.hs
palas: I have tested the pipe-path rules in a pure way (https://github.com/IntersectMBO/cardano-node/commit/e0b8c5a869384798575fc46a500d9d58ed51a06c). I also tried un-gating the two integration tests, but Hydra's cross-Windows leg runs the suite under wine, where port allocation fails before anything starts (`Network.Socket.setSockOpt: WSAENOPROTOOPT`, https://ci.iog.io/build/14742303).

cardano-testnet/src/Testnet/Manifest.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/e0b8c5a869384798575fc46a500d9d58ed51a06c and https://github.com/IntersectMBO/cardano-node/commit/d02755190616d2b54cf4d1b0b57dc274e66822a5

cardano-testnet/src/Testnet/Manifest.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/commit/7433f6a61ac2981f0c3c79c3b4eb122da1e057d6

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Manifest.hs
```diff
@@ -0,0 +1,92 @@
...
+
+-- | Integration test: verify that 'createTestnetEnv' removes any
+-- pre-existing (stale) manifest from a previous run.  Only the
+-- environment is created (no cluster start); the manifest is written
+-- later by the run phase, so after 'createTestnetEnv' the file must
+-- be gone.
+--
+-- Execute with:
+-- @DISABLE_RETRIES=1 cabal test cardano-testnet-test --test-options '-p "/Manifest stale/"'@
+hprop_manifest_stale :: Property
```
carbolymer: This covers the `removeStaleManifest` call inside `createTestnetEnv`, but there is a second call at the top of `cardanoTestnet`, and that one is what protects the `StartFromEnv` path (reusing an existing env dir). Every existing test reaches it only through `createAndRunTestnet`, where `createTestnetEnv` has already deleted the file, so a regression at the second site would not show up anywhere.

Closing it needs a stale file planted and `cardanoTestnet` called directly, which costs a cluster start, so I would not insist. Flagging so the gap is a known one rather than an accidental one.

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Manifest.hs
```diff
@@ -0,0 +1,92 @@
...
+hprop_manifest_windows_pipe_path = H.propertyOnce $ do
+  let outputDir = "C:\\Users\\test\\testnet"
+      pipePath  = "\\\\.\\pipe\\cardanotestnet-node1"
+
+  -- (1) Under Windows-flavour filepath: pipe path passes through unchanged
+  FilePath.Windows.normalise (FilePath.Windows.makeRelative outputDir pipePath)
+    === pipePath
+
+  -- (2) Under the real (platform-native) makeManifestRelPath: pipe path unchanged
+  makeManifestRelPath outputDir pipePath === pipePath
```
carbolymer: Assertion (2) does not test what the comment says on Linux and macOS. There the ambient filepath is the Posix flavour, where a backslash is an ordinary character, so the string passes through `makeRelative` for a reason unrelated to `splitDrive` or drives. It only exercises the Windows mechanism on the windows-latest leg.

That is fine as coverage, assertion (1) carries the Windows semantics everywhere, but the comment should say so: something like "on Windows this is the same check as (1) through the real function; on Posix it passes because backslashes are not separators". Otherwise the next reader assumes (2) is a second pin of the drive behaviour.

.github/workflows/haskell.yml
```diff
@@ -128,6 +128,32 @@ jobs:
       run: |
         cabal test cardano-testnet cardano-node cardano-node-chairman cardano-submit-api
 
+    - name: Validate produced manifest files against schema
+      if: runner.os == 'Linux'
+      run: |
+        pipx install check-jsonschema==0.37.1
```
carbolymer: Minor, non-blocking: this pins check-jsonschema at 0.37.1 via pipx, while the Makefile target and check-testnet-manifest.yml use `nix run .#check-jsonschema`, which resolves to 0.33.0 from the locked nixpkgs. Two validator versions for one schema, and two places to bump. I get why, this job has no nix. A one-line comment saying so would stop the next person from "fixing" the mismatch in the wrong direction.

.github/workflows/haskell.yml
```diff
@@ -128,6 +128,32 @@ jobs:
       run: |
         cabal test cardano-testnet cardano-node cardano-node-chairman cardano-submit-api
 
+    - name: Validate produced manifest files against schema
+      if: runner.os == 'Linux'
```
carbolymer: This gate is only on the OS, so the cluster starts and gets torn down once per Linux GHC leg, three times per PR, for a manifest whose shape does not depend on the compiler. It is cheap today (about 40s a leg on the last run), so purely a tidy-up: adding a `matrix.ghc == '9.12'` style condition would run it once.

cardano-testnet/test/cardano-testnet-golden/Cardano/Testnet/Test/Golden/Manifest.hs
```diff
@@ -0,0 +1,132 @@
...
+  -- Round-trip check: decode the encoded JSON and re-encode; must match.
+  decodedManifest <- evalEither (eitherDecode encoded :: Either String Manifest)
+  let reEncoded = encodePretty decodedManifest
+  encoded === reEncoded
+
+  -- An unsupported schema version is rejected at decode time.
+  let v2 = encodePretty sampleManifest { manifestSchemaVersion = supportedSchemaVersion + 1 }
+  case eitherDecode v2 :: Either String Manifest of
+    Left _  -> pure ()
+    Right _ -> failure
```
carbolymer: Convention here is `assertWith`, and it also reports the wrongly decoded value on failure, where this prints only "failure" with no context.

```suggestion
  assertWith (eitherDecode v2 :: Either String Manifest) isLeft
```

Needs `isLeft` from `Data.Either` and `assertWith` added to the `Hedgehog.Extras.Test.Base` import.

.github/workflows/haskell.yml
```diff
@@ -128,6 +128,32 @@ jobs:
...
+        # the 108-character unix socket path limit ($RUNNER_TEMP is short).
+        dir="$RUNNER_TEMP/manifest-schema-check"
+        log="$RUNNER_TEMP/manifest-schema-check.log"
+        "$(cabal list-bin cardano-testnet)" cardano --output-dir "$dir" > "$log" 2>&1 &
+        testnet_pid=$!
+        # The manifest is written at readiness, so the file appearing is the ready signal.
+        for _ in $(seq 1 120); do
+          if [ -f "$dir/manifest.json" ]; then break; fi
+          sleep 5
+        done
```
carbolymer: The loop only watches for the file. If `cardano-testnet` dies early, nothing notices until the 120 iterations run out, so a crash costs the full ten minutes before the log gets dumped. Breaking out when the process is gone fails straight away instead:

```suggestion
        for _ in $(seq 1 120); do
          if [ -f "$dir/manifest.json" ]; then break; fi
          if ! kill -0 "$testnet_pid" 2>/dev/null; then break; fi
          sleep 5
        done
```

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Manifest.hs
```diff
@@ -0,0 +1,92 @@
...
+hprop_manifest_windows_pipe_path :: Property
+hprop_manifest_windows_pipe_path = H.propertyOnce $ do
+  let outputDir = "C:\\Users\\test\\testnet"
+      pipePath  = "\\\\.\\pipe\\cardanotestnet-node1"
+
+  -- (1) Under Windows-flavour filepath: pipe path passes through unchanged
+  FilePath.Windows.normalise (FilePath.Windows.makeRelative outputDir pipePath)
+    === pipePath
+
+  -- (2) Under the real (platform-native) makeManifestRelPath: pipe path unchanged
```
carbolymer: On the ubuntu and macos legs this passes for an unrelated reason, so the comment overstates what (2) pins:

```suggestion
  -- (2) Through the real function. On Windows this repeats (1); on Posix it
  --     passes only because backslashes are not separators there.
```

carbolymer (approved):
Nice work! LGTM!

.github/workflows/haskell.yml
palas: Yes, maybe it is overkill

.github/workflows/haskell.yml
palas: Hm, you are right. We should have this in mind for the status command too

.github/workflows/haskell.yml
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/pull/6693/changes/1594d675b515f430a90e639c9ff5561a86f07ece

.github/workflows/haskell.yml
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/pull/6693/changes/1594d675b515f430a90e639c9ff5561a86f07ece

cardano-testnet/test/cardano-testnet-golden/Cardano/Testnet/Test/Golden/Manifest.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/pull/6693/changes/1594d675b515f430a90e639c9ff5561a86f07ece

.github/workflows/haskell.yml
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/pull/6693/changes/1594d675b515f430a90e639c9ff5561a86f07ece

cardano-testnet/test/cardano-testnet-test/Cardano/Testnet/Test/Manifest.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-node/pull/6693/changes/1594d675b515f430a90e639c9ff5561a86f07ece

